PR #97, opened 2022-08-11
Handle produce error

akrambek:


jfallows:
@akrambek given that `WINDOW` `maximum` increases to accommodate pending `ack` without blocking `DATA` progress, we need to make sure this is honored by `http` binding to get the expected end-to-end behavior through `http`, `http-kafka` and `kafka` bindings without blocking progress while publishing large messages.

jfallows (approved):
Great work @akrambek !

PR #101, opened 2022-09-21
BugFix: process configURL if the protocol is http

Alfusainey:
`https` is repeated twice in the if condition. This means that if the protocol in the URL to the Zilla JSON config file is `http`, then the `java.net.http.HttpClient` client is not used to retrieve the JSON configuration. Instead, a `URLConnection` is used.

@jfallows I think this will still work with a `configURL.openConnection()`. If this was the intended behavior, then i think the other approach will be to simply the if condition

Signed-off-by: Alfusainey Jallow <alf.jallow@gmail.com>


jfallows (approved):
@Alfusainey good catch!

PR #105, opened 2022-09-27
Replace HttpRequest with buffer slot

akrambek:


runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
```diff
@@ -274,7 +276,7 @@ public MessageConsumer newStream(
                 binding.options != null && binding.options.overrides != null ? binding.options.overrides : EMPTY_OVERRIDES;
 
             final HttpClientPool clientPool = clientPools.computeIfAbsent(resolvedId, HttpClientPool::new);
-            newStream = clientPool.newStream(application, begin, overrides);
+            newStream = clientPool.newStream(buffer, index, length, application, overrides, binding.versions());
```
jfallows, line 279: Perhaps revert this back to keep the API simpler?

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
```diff
@@ -856,103 +859,148 @@ int decode(
...
+            final int extensionLength = begin.extension().sizeof();
+            if (extensionLength > maximumHeadersSize)
             {
-                final HttpExchange exchange = client.newExchange(sender, begin, overrides);
-                newStream = exchange::onApplication;
+                HttpBeginExFW beginEx = beginExRW.wrap(codecBuffer, 0, codecBuffer.capacity())
+                        .typeId(httpTypeId)
+                        .headersItem(h -> h.name(HEADER_STATUS).value(STATUS_431))
+                        .build();
+                newStream = rejectWithStatusCode(sender, begin, beginEx);
```
jfallows, line 901: Consider moving the construction of `beginEx` inside `rejectWithStatusCode` and pass only `STATUS_431` here instead of `beginEx`.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
```diff
@@ -856,103 +859,148 @@ int decode(
...
-                    dequeues.getAsLong();
+            headerSlotOffset = 0;
+            headerSlotLimit = 0;
+        }
 
-                    request.doFlushBegin(client);
-                }
+        private void cleanupHeaderSlotIfNecessary()
+        {
+            if (headerSlotOffset == headerSlotLimit && headerSlot != NO_SLOT)
```
jfallows, line 975: Logically we'd put the test for `headerSlot != NO_SLOT` first as that is gating for the other values to have meaning.

runtime/binding-http/src/test/java/io/aklivity/zilla/runtime/binding/http/internal/streams/rfc7230/client/ConnectionManagementPoolSize1IT.java
```diff
@@ -238,6 +238,7 @@ public void shouldResetRequestAndFreeConnectionWhenRequestWithContentLengthIsRes
         k3po.finish();
     }
 
+    @Ignore("There is no HTTP_MAXIMUM_QUEUED_REQUESTS_NAME")
```
jfallows, line 241: If this IT is still relevant, then let's remove `@Ignore` and make it pass, otherwise if not still relevant then we can remove the IT.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
```diff
@@ -1869,6 +1862,9 @@ private void cleanupEncodeSlotIfNecessary()
         private BudgetDebitor responseDeb;
         private long responseDebIndex = NO_DEBITOR_INDEX;
 
+        private long traceId;
+        private long authorization;
+
```
jfallows, line 1867: Are these needed just to retain their values while queuing `beginEx` in slot, or do we need them for the lifetime of the exchange?

If only needed to survive queuing, then we should store them along with the `streamId` and `beginEx` in the slot entry instead.

jfallows (changes requested):

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
akrambek: I did that but then in the case of 503 there is another header `retry-after` hence I moved the headers out of the method.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
akrambek: Yeah, good catch we only need them to survive queuing.

runtime/binding-http/src/test/java/io/aklivity/zilla/runtime/binding/http/internal/streams/rfc7230/client/ConnectionManagementPoolSize1IT.java
akrambek: I don't think it is relevant anymore.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
```diff
@@ -886,12 +884,11 @@ public MessageConsumer newStream(
             countRequests.getAsLong();
             this.versions = versions;
 
-            final BeginFW begin = beginRO.wrap(buffer, index, index + length);
             HttpClient client = supplyClient();
 
             MessageConsumer newStream = null;
 
-            final int extensionLength = begin.extension().sizeof();
+            final int extensionLength = begin.extension().sizeof() + 3 * Long.BYTES + Integer.BYTES;
```
jfallows, line 891: The terminology has shifted slightly, so we need to rename these variables for clarity.
This is no longer `extensionLength`, it is now `queuedRequestLength` and `maximumHeadersSize` is `maximumRequestQueueSize`, agree?

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
```diff
@@ -956,13 +953,16 @@ private void flushNext()
             {
                 final MutableDirectBuffer headerBuffer = bufferPool.buffer(headerSlot);
                 final long streamId = headerBuffer.getLong(headerSlotOffset);
-                final int length = headerBuffer.getInt(headerSlotOffset + Long.BYTES);
-                HttpExchange httpExchange = exchanges.get(streamId);
+                final long traceId = headerBuffer.getLong(headerSlotOffset + Long.BYTES);
+                final long authorization = headerBuffer.getLong(headerSlotOffset + 2 * Long.BYTES);
+                final int length = headerBuffer.getInt(headerSlotOffset +  3 * Long.BYTES);
+                final int metadataOffset = headerSlotOffset + 3 * Long.BYTES + Integer.BYTES;
```
jfallows, line 959: Let's use a flyweight for this, similar to `KafkaCacheEntry`, instead of manually calculating the offsets.

jfallows (changes requested):

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
akrambek: Yep

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
```diff
@@ -1033,7 +1036,7 @@ private void onUpgradedOrClosed(
 
         private int availableSlotSize()
         {
-            return bufferPool.slotCapacity() - headerSlotLimit - Long.BYTES - Integer.BYTES;
+            return bufferPool.slotCapacity() - httpQueueSlotLimit - Long.BYTES - Integer.BYTES;
```
jfallows, line 1039: Remind me why we subtract `Long.BYTES - Integer.BYTES` here. Is it still needed? If so, let's use the appropriate constant and if not, then let's remove it.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
```diff
@@ -1954,32 +1957,27 @@ private void onRequestBegin(
...
-                headerBuffer.putLong(headerSlotLimit, traceId);
-                headerSlotLimit += Long.BYTES;
-                headerBuffer.putLong(headerSlotLimit, authorization);
-                headerSlotLimit += Long.BYTES;
-                headerBuffer.putInt(headerSlotLimit, length);
-                headerSlotLimit += Integer.BYTES;
-                headerBuffer.putBytes(headerSlotLimit, beginEx.buffer(), beginEx.offset(), length);
-                client.pool.headerSlotLimit += headerSlotLimit + length;
+                final MutableDirectBuffer httpQueueBuffer = bufferPool.buffer(client.pool.httpQueueSlot);
+                int headerSlotLimit = client.pool.httpQueueSlotLimit;
```
jfallows, line 1962: `final` ?

jfallows (changes requested):

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
akrambek: Yeah it is a bug, seems like have a wrong mental model. Removing.

jfallows (approved):

Issue #68, opened 2022-06-01
Simplify configuration for JWT identity provider signing keys


jfallows:
Zilla supports JWT authorization and requires the public signing keys from the JWT identity provider to verify token integrity.
```
    "guards":
    {
        "jwt0":
        {
           "type":"jwt",
           "options":
           {
              "issuer": "https://auth.example.com/",
              "audience": "https://api.example.com/",
              "keys":
              [
                  ...
              ]
           }
        }
    }
```
This requires copying the contents of the identity provider supplied JSON Web Key Set (JWKS).

However, identity providers can rotate their signing keys, requiring the configuration to be updated.

Instead, we can add support for specifying the URL where the JSON Web Key Set can be retrieved.

```
    "guards":
    {
        "jwt0":
        {
           "type":"jwt",
           "options":
           {
              "issuer": "https://auth.example.com/",
              "audience": "https://api.example.com/",
              "keys": "https://auth.example.com/.well-known/jwks.json"
           }
        }
    }
```

In fact, we can derive a default value for `keys` URL by appending `.well-known/jwks.json` to the `issuer` URL, making it unnecessary to specify the `keys` in the configuration.

```
    "guards":
    {
        "jwt0":
        {
           "type":"jwt",
           "options":
           {
              "issuer": "https://auth.example.com/",
              "audience": "https://api.example.com/"
           }
        }
    }
```

The `keys` URL can be resolved when parsing configuration on startup.

Systems that prefer to configure the `keys` explicitly can still do so using the existing syntax.

Alfusainey:
@jfallows I want to wrap my head around the Zilla code base by working on this issue. Can I give it a try?

So the idea is to populate the `JwtOptionsConfig#keys` field upon startup when a URL (instead of a JSON array) is provided as the value for `keys`? So basically, the logic to do so should preferably be in `JwtOptionsConfigAdapter#adaptFromJson`:

```java
...
 List<JwtKeyConfig> keys = object.containsKey(KEYS_NAME)
                ? object.getJsonArray(KEYS_NAME)
                    .stream()
                    .map(JsonValue::asJsonObject)
                    .map(key::adaptFromJson)
                    .collect(toList())
                : KEYS_DEFAULT;
...
```
Or should this happen much earlier up?

Also, if the key is missing, an empty list of JwtKeyConfig is currently returned. But a missing key can now mean that the JWKS should be retrieved using `https://auth.example.com/.well-known/jwks.json`. 

[...]

PR #106, opened 2022-09-28
[#68] Add schema for specifying an OpenID Provider discovery endpoint

Alfusainey:
Adds the schema changes required to provide the discovery endpoint for an OpenID Provider to use for retrieving its JWKS metadata.

Signed-off-by: Alfusainey Jallow <alf.jallow@gmail.com>


specs/guard-jwt.spec/src/main/scripts/io/aklivity/zilla/specs/guard/jwt/config/guard-issue-68.json
```diff
@@ -0,0 +1,17 @@
...
+            "options":
+            {
+                "issuer": "https://auth.example.com",
+                "audience": "https://api.example.com",
+                "keys": "https://auth.example.com/.well_known/jwks.json",
+                "challenge": 30
+            }
+        }
+    }
+}
```
jfallows, line 17: Please name this file as if it had been present before the issue was filed, rather than tying it back to the issue number.

For example, `guard-keys-dynamic.json` rather than `guard-issue-68.json`.

specs/guard-jwt.spec/src/test/java/io/aklivity/zilla/specs/guard/jwt/config/SchemaTest.java
```diff
@@ -38,5 +43,19 @@ public void shouldValidateGuard()
         JsonObject config = schema.validate("guard.json");
 
         assertThat(config, not(nullValue()));
+        JsonValue value = config.getValue("/guards/jwt0/options/keys");
+        assertThat(value, is(instanceOf(JsonArray.class)));
     }
+
+    @Test
+    public void shouldValidateGuardWithUrlAsKey()
```
jfallows, line 51: Please rename to `shouldValidateGuardWithDynamicKeys()`

jfallows (changes requested):
To handle the case where `keys` is omitted, and derived from `issuer` then we will need to differentiate another config scenario, `guard-keys-implicit.json` which should be supported by schema and test.

Alfusainey:
> To handle the case where `keys` is omitted, and derived from `issuer` then we will need to differentiate another config scenario, `guard-keys-implicit.json` which should be supported by schema and test.

I think the current schema already provides support for this since `keys` is (currently) not a required config property. So adding a guard config with missing `keys` successfully validates against the current schema.

The newly added commit contains a test case for implicit keys

jfallows (dismissed):

jfallows:
@Alfusainey the config and schema tests look good, this is a great checkpoint. 👍 

Heads up that we'll not be merging this PR until after the implementation is complete to ensure the `develop` branch is always ready for release (with no partial features).

Suggest converting this PR to a draft and updating the title and description to match the broader intent of the PR.

Note that you need to put `Fixes #68` in the main PR description above to associate this PR as the resolution for issue #68, then when the PR is merged, issue #68 will be closed automatically.  Recommend then also removing `[#68]` from the PR title prefix.

Let me know if you have any questions about the implementation approach, otherwise feel free to forge ahead. 😄

Alfusainey:
@jfallows The changes are added in commit fc4e34e3. I look forward to your feedback

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -128,31 +120,11 @@ public Void call() throws Exception
...
-            configText = response.body();
-
-        }
         else
         {
-            URLConnection connection = configURL.openConnection();
-            try (InputStream input = connection.getInputStream())
+            try
             {
-                configText = new String(input.readAllBytes(), UTF_8);
```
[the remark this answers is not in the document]
Alfusainey: nice thinking 👍 ok, will update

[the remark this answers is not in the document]
Alfusainey: @jfallows am trying to think about how this would really work after a `NamespaceConfig` is already resolved from the JSON string.

My understanding is that the `keys` URL should be resolve during the time when the `configText` is been adapted to a `NamespaceConfig`. In other words, the `keys` should be resolved during the following call:

```java
NamespaceConfig namespace = jsonb.fromJson(configText, NamespaceConfig.class);
```

and not afterwards. Because if its done after the JSON string is already adapted to a `NamespaceConfig`, then the keys would have already been resolved. So I do not see how the `readURL` callback would be helpful in that case. Can you please clarify?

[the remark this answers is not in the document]
jfallows: > Then in `JwtGuardHandler` constructor, we can use `readURL` to resolve the keys.

@Alfusainey ultimately the `JwtGuardHandler` needs the resolved keys, but it is not a requirement to resolve them during JSON parsing.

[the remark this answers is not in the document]
Alfusainey: ok, i see 👍

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/GuardConfig.java
```diff
@@ -17,13 +17,20 @@
...
+
 public class GuardConfig
 {
     public transient long id;
 
     public final String name;
     public final String type;
     public final OptionsConfig options;
+    public Function<URL, String> readURL;
+    public Jsonb jsonb;
```
jfallows, line 33: Please mark these as `transient` and move up alongside other `transient` fields.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/NamespaceConfig.java
```diff
@@ -26,6 +28,7 @@
 {
     public transient int id;
     public transient ToLongFunction<String> resolveId;
+    public Function<URL, String> readURL;
```
jfallows, line 31: Please mark this field as `transient` too.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -128,31 +129,11 @@ public Void call() throws Exception
...
+                configText = readURL(configURL);
             }
             catch (IOException ex)
```
jfallows, line 138: Okay to let exception escape, will be reported by `zilla` engine.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -220,12 +201,27 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
...
 
+            namespace.readURL = configURL ->
+            {
+                try
+                {
+                    return this.readURL(configURL);
+                }
+                catch (IOException | URISyntaxException | InterruptedException e)
+                {
+                    rethrowUnchecked(e);
```
jfallows, line 212: Please move `rethrowUnchecked(...)` behavior inside `readURL` and use `ex` for exception instead of `e` (by convention).

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -220,12 +201,27 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
...
+                }
+                return null;
```
jfallows, line 214: Avoid multiple returns, instead use variable outside `try` initialized to `null`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -300,4 +296,35 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
 
         return null;
     }
+
+
+    private String readURL(URL url) throws IOException, URISyntaxException, InterruptedException
```
jfallows, line 301: Please remove `throws` clause and use `rethrowUnchecked` in this method body instead.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -54,39 +63,43 @@ public class JwtGuardHandler implements GuardHandler
     private final Long2ObjectHashMap<JwtSessionStore> sessionStoresByContextId;
 
     public JwtGuardHandler(
-        JwtOptionsConfig options,
-        LongSupplier supplyAuthorizedId)
+            GuardConfig guard,
+            LongSupplier supplyAuthorizedId)
```
jfallows, line 67: Recommend reverting this change in parameter types, and adding `readURL` as an additional parameter instead.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -54,39 +63,43 @@ public class JwtGuardHandler implements GuardHandler
...
-        Map<String, JsonWebKey> keys = new HashMap<>();
+        Map<String, JsonWebKey> tmpKeys = new HashMap<>();
         if (options.keys != null)
         {
-            for (JwtKeyConfig key : options.keys)
+            tmpKeys = getKeys(options.keys);
+        }
+        else
+        {
+            String keysURL = options.keysURL.orElse(getJwksURL(issuer));
```
jfallows, line 82: It might be better to default this in the parser rather than in the handler.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -54,39 +63,43 @@ public class JwtGuardHandler implements GuardHandler
...
+            LongSupplier supplyAuthorizedId)
     {
+
+        JwtOptionsConfig options = (JwtOptionsConfig) guard.options;
         this.issuer = options.issuer;
         this.audience = options.audience;
         this.challenge = options.challenge.orElse(null);
 
-        Map<String, JsonWebKey> keys = new HashMap<>();
+        Map<String, JsonWebKey> tmpKeys = new HashMap<>();
```
jfallows, line 75: Recommend avoiding variable names like `tmpKeys`, prefer more descriptive names like `resolvedKeys` instead.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -54,39 +63,43 @@ public class JwtGuardHandler implements GuardHandler
...
+                    JsonArray keysArray = guard.jsonb.fromJson(keysText, JsonArray.class);
+                    List<JwtKeyConfig> keysConfig = keysArray
+                            .stream()
+                            .map(JsonValue::asJsonObject)
+                            .map(new JwtKeyConfigAdapter()::adaptFromJson)
+                            .collect(toList());
+                    tmpKeys = getKeys(keysConfig);
                 }
-                catch (JoseException ex)
+                catch (MalformedURLException | JsonbException ignored)
```
jfallows, line 96: Better to let this exception escape so we can tell what's wrong instead of hiding it.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -244,6 +257,43 @@ private JwtSessionStore supplySessionStore(
         return sessionStoresByContextId.computeIfAbsent(contextId, JwtSessionStore::new);
     }
 
+    private Map<String, JsonWebKey> getKeys(List<JwtKeyConfig> keysConfig)
```
jfallows, line 260: Please check all method definitions and put parameters on separate newlines below method name.

specs/guard-jwt.spec/src/main/scripts/io/aklivity/zilla/specs/guard/jwt/config/guard-keys-dynamic.json
```diff
@@ -0,0 +1,17 @@
...
+    "guards":
+    {
+        "jwt0":
+        {
+            "type": "jwt",
+            "options":
+            {
+                "issuer": "https://aklivity.us.auth0.com/",
+                "audience": "http://localhost:8080/tasks",
+                "keys": "https://aklivity.us.auth0.com/.well-known/jwks.json",
```
jfallows, line 12: Perhaps use a different URL from the default when specified explicitly, otherwise we cannot tell that it is being used instead of the default.

specs/guard-jwt.spec/src/main/scripts/io/aklivity/zilla/specs/guard/jwt/schema/jwt.schema.patch.json
```diff
@@ -27,31 +27,40 @@
...
                                 {
-                                    "kty":
-                                    {
-                                        "title": "Key Type",
-                                        "type": "string"
-                                    },
-                                    "kid":
+                                    "type": "string",
+                                    "format": "uri",
+                                    "pattern": "^(https?|wss?|ftp)://"
```
jfallows, line 34: Let's restrict to `http` or `https` only for now.

jfallows (changes requested):

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
Alfusainey: I actually use the `GuardConfig` parameter to access both the `readURL` and `jsonb` transient fields. Should I also add `jsonb` as a parameter here?

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
Alfusainey: what do you mean by "check all method definitions"? Check if the `keysConfig` parameter is null?

jfallows: Sorry, I just meant formatting.

For each method, make sure it is of the form:
```
... <method-name>(
  <param-type> <param-name>,
  <param-type> <param-name>)
```

rather than on a single line. 😄

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
jfallows: Yes, that works.

specs/guard-jwt.spec/src/main/scripts/io/aklivity/zilla/specs/guard/jwt/config/guard-keys-dynamic.json
Alfusainey: ok, i reverted this change to the original one

Alfusainey:
@jfallows PR updated

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -310,4 +288,44 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
 
         return null;
     }
+
+
+    private String readURL(URL url)
```
jfallows, line 293: Please move parameter to next line, and rename `url` to `location`.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -244,6 +265,34 @@ private JwtSessionStore supplySessionStore(
         return sessionStoresByContextId.computeIfAbsent(contextId, JwtSessionStore::new);
     }
 
+    private Map<String, JsonWebKey> getKeys(
+            List<JwtKeyConfig> keysConfig)
```
jfallows, line 269: Just one block indentation for method params.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -54,39 +66,48 @@ public class JwtGuardHandler implements GuardHandler
...
-                }
+                String keysText = readURL.apply(URI.create(options.keysURL.get()).toURL());
+                JsonObject keysObject = jsonb.fromJson(keysText, JsonObject.class);
+                List<JwtKeyConfig> keysConfig = keysObject
+                        .getJsonArray(KEYS_NAME)
+                        .stream()
+                        .map(JsonValue::asJsonObject)
+                        .map(new JwtKeyConfigAdapter()::adaptFromJson)
+                        .collect(toList());
+                resolvedKeys = getKeys(keysConfig);
```
jfallows, line 102: We actually want to be able to call something like the following:
```
JwtKeySetConfig jwks = jsonb.fromJson(keysText, JwtKeySetConfig.class);
```
Where `JwtKeySetConfig` represents the syntax of the response from the keysURL, and has a field called `keys` of type `List<JwtKeyConfig>`.
This means we'd need to create a local `jsonb` properly configured with the corresponding (new) class `JwtKeySetConfigAdapter` that encapsulates the parsing logic above, making use of `JwtKeyConfigAdapter` internally to parse each key.

Then we don't need to pass the global `jsonb` here any more either.

Initialization of the local `jsonb` would follow something like:
```
            JsonbConfig config = new JsonbConfig()
                    .withAdapters(new JwtKeySetConfigAdapter());
                    .withConfig(config)
                    .build();
```

jfallows (changes requested):

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
Alfusainey: this is a neat insight!

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -56,37 +67,45 @@ public class JwtGuardHandler implements GuardHandler
     public JwtGuardHandler(
         JwtOptionsConfig options,
         LongSupplier supplyAuthorizedId)
+    {
+        this(options, supplyAuthorizedId, url -> "{}");
+    }
```
jfallows, line 72: This constructor seems to be used only by unit tests, let's remove it and supply the last parameter from the unit test instead.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -56,37 +67,45 @@ public class JwtGuardHandler implements GuardHandler
...
-                }
-                catch (JoseException ex)
-                {
-                    rethrowUnchecked(ex);
-                }
+                rethrowUnchecked(ex);
             }
         }
-        this.keys = keys;
+        this.keys = resolvedKeys;
```
jfallows, line 108: Seems like the flow control structure above can be simplified slightly.

```
List<JwtKeyConfig> keysConfig = options.keys;
if (keysConfig == null && options.keysURL.isPresent())
{
    try
    {
        ...
        keysConfig = jwks.keys;
    }
    catch (MalformedURLException | JsonbException ex)
    {
        rethrowUnchecked(ex);
    }
}

Map<String, JsonWebKey> resolvedKeys = new HashMap<>();
if (keysConfig != null)
{
    // inline getKeys method here
}

```

jfallows (changes requested):

Alfusainey:
@jfallows thanks much! PR updated

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -97,14 +86,20 @@ else if (options.keysURL.isPresent())
...
             catch (MalformedURLException | JsonbException ex)
             {
                 rethrowUnchecked(ex);
             }
         }
+
+        Map<String, JsonWebKey> resolvedKeys = new HashMap<>();
+        if (keysConfig != null)
+        {
+            resolvedKeys = getKeys(keysConfig);
```
jfallows, line 100: I was requesting you inline `getKeys` method body here, no need for separate method any more.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/config/JwtOptionsConfigAdapter.java
```diff
@@ -130,15 +130,16 @@ public OptionsConfig adaptFromJson(
         {
             if (issuer != null)
             {
-                keysUrl = (issuer.endsWith("/")) ?
-                        String.format("%s.well-known/jwks.json", issuer) : String.format("%s/.well-known/jwks.json", issuer);
+                keysURL = (issuer.endsWith("/"))
```
jfallows, line 133: Please remove redundant parens around `issuer.endsWith("/")`.

runtime/guard-jwt/src/test/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandlerTest.java
```diff
@@ -39,13 +41,15 @@
 
 public class JwtGuardHandlerTest
 {
+    private Function<URL, String> keysURL = url -> "{}";
```
jfallows, line 44: Please make this `static` `final` and rename to `READ_KEYS_URL`.

jfallows (changes requested):

runtime/guard-jwt/src/test/java/io/aklivity/zilla/runtime/guard/jwt/internal/config/AbstractJwtConfigAdapterTest.java
```diff
@@ -0,0 +1,32 @@
...
+{
+    protected Jsonb jsonb;
+
+    protected void initJson(JsonbAdapter<?, ?> adapter)
+    {
+        JsonbConfig config = new JsonbConfig()
+                .withAdapters(adapter);
+        jsonb = JsonbBuilder.create(config);
+    }
+}
```
jfallows, line 32: Please remove this file and duplicate the code in each subclass instead.
The design for tests (linearity) is different from the design for implementation (abstraction, reuse).

jfallows (changes requested):

runtime/guard-jwt/src/test/java/io/aklivity/zilla/runtime/guard/jwt/internal/config/AbstractJwtConfigAdapterTest.java
Alfusainey: ok, thanks. will do

jfallows (approved):

jfallows (approved):

PR #121, opened 2022-10-24
Implement zilla dump command similar to tcpdump

### LINKED ISSUE #114: Implement `zilla dump` command similar to `tcpdump`
The Zilla runtime has a a pluggable command structure, currently used for `start` and `stop`.

```
./zilla
usage: zilla <command> [ <args> ]

Commands are:
    help    Display help information
    start   Start engine
    stop    Stop engine

See 'zilla help <command>' for more information on a specific command.
```

Add a new `zilla dump` command to produce a virtual packet capture of Zilla shared memory streams that can be dissected in Wireshark as `tcp`, `tls`, `http`, `mqtt`, `amqp`, etc.

Capturing the raw byte stream for each binding in `zilla.json` in [`.pcap format`][pcap-format] is sufficient to let Wireshark dissectors present the structured form of the protocol frames.

[pcap-format]: https://datatracker.ietf.org/doc/html/draft-gharris-opsawg-pcap-02

bmaidics:
Fix #114


bmaidics:
What's still needed here:

- Turned out Pcap4j is not working with java 17. I was using java 11 during the development. Another drawback is that it is a wrapper around a native lib. Based on these we should drop the lib, and come up with something
- Couldn't get rid of some Wireshark warnings: previous segment not arrived/dup ack etc.. We need to figure out a way to smartly assign sequence number to packet and the segment types as well.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/DumpHandlers.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+    @Override
+    public void onBegin(BeginFW begin)
+    {
+        final long routeId = begin.routeId();
+        final long streamId = begin.streamId();
+        final int bindingId = (int) (routeId >> 0) & 0xffff_ffff;
+        Inet6Address address = createAddress(bindingId, streamId);
+        TcpPort port = getPort(streamId);
+        writePacketsToPcap(createPseudoEthernetPacket(createPseudoTcpPacketBuilder(address, port).syn(true),
```
jfallows, line 72: This will need to execute at very high performance, racing to keep up with the Zilla engine itself.

Therefore, while this approach is functionally correct, please try _not_ to create any objects during processing, as they would be created at a high rate, and trigger corresponding garbage collection effort.

Switching over to using generated `pcap` flyweights may help address this point as well as other issues using the `pcap4j` library.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/DumpStreamsCommand.java
```diff
@@ -0,0 +1,93 @@
...
+        RingBufferSpy.SpyPosition position,
+        String pcapLocation)
+    {
+        super(config, hasFrameType, verbose, continuous, affinity, position);
+        try
+        {
+            phb = Pcaps.openDead(DataLinkType.EN10MB, 65536);
+            String timeStamp = new SimpleDateFormat("yyyy_MM_dd_HH_mm_ss").format(System.currentTimeMillis());
+            String fileName = "zilla_dump" + timeStamp + ".pcap";
+            dumper = phb.dumpOpen(Paths.get(pcapLocation, fileName).toString());
```
jfallows, line 60: `tcpdump` specifies the exact file for output, let's take the same approach here.

The timestamp effect can be achieved on the command line using:
```
./zilla dump --output "zilla_dump`date +%Y_%m_%d_%H_%M_%`.pcap"
```

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
```diff
@@ -0,0 +1,115 @@
...
+import com.github.rvesse.airline.annotations.Option;
+
+import io.aklivity.zilla.runtime.command.ZillaCommand;
+import io.aklivity.zilla.runtime.command.dump.internal.airline.spy.RingBufferSpy;
+import io.aklivity.zilla.runtime.engine.Configuration;
+import io.aklivity.zilla.runtime.engine.EngineConfiguration;
+
+@Command(name = "dump", description = "Dump stream content")
+public final class ZillaDumpCommand extends ZillaCommand
+{
```
jfallows, line 39: Suggest adding an option to specify which bindings should be recorded in the `.pcap` output.

incubator/command-log/pom.xml
```diff
@@ -85,22 +90,6 @@
...
-          <packageName>io.aklivity.zilla.runtime.command.log.internal.types</packageName>
-        </configuration>
-        <executions>
-          <execution>
-            <goals>
-              <goal>generate</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
```
jfallows, line 103: Please keep the generated flyweights local to the module using them.

Java Virtual Machine escape analysis can then better determine only one usage of them inside the same Java module and can then apply better optimizations for these high performance codepaths as desired.

incubator/command-dump/src/test/java/io/aklivity/zilla/runtime/command/dump/internal/TestDumpCommand.java
```diff
@@ -0,0 +1,89 @@
...
+        properties.put("zilla.engine.directory", "src/test/resources/zilla_test_stream");
+        final EngineConfiguration config = new EngineConfiguration(new Configuration(), properties);
+        Predicate<String> hasFrameTypes = t -> true;
+        boolean verbose = true;
+        boolean continuous = false;
+        long affinity = -1;
+        RingBufferSpy.SpyPosition position = RingBufferSpy.SpyPosition.ZERO;
+        Runnable command = new DumpStreamsCommand(config, hasFrameTypes, verbose, continuous, affinity, position,
+            tempDir.getAbsolutePath());
+        command.run();
```
jfallows, line 62: This bootstrap part of the test should be using `ZillaDumpCommand` and specifying public command line arguments in the command fields rather than exercising the internal implementation directly.

Then the internal implementation can be refactored at will without destabilizing the test as it relies only on what amounts to the public API of the command line syntax.

incubator/command-dump/src/test/java/io/aklivity/zilla/runtime/command/dump/internal/TestDumpCommand.java
```diff
@@ -0,0 +1,89 @@
...
+                Assertions.assertEquals(expected, actual);
+                expected = expectedPhb.getNextPacketEx().getPacket();
+            }
+            catch (EOFException e)
+            {
+                break;
+            }
+        }
+        expectedPhb.close();
+        actualPhb.close();
```
jfallows, line 87: Assuming we drop the `pcap4j` dependency, this part of the test will need to change.

Given that we can visualize the `.pcap` output (both expected and actual) via `wireshark`, perhaps a simple binary comparison is sufficient here.

incubator/command-dump/src/test/java/io/aklivity/zilla/runtime/command/dump/internal/TestDumpCommand.java
```diff
@@ -0,0 +1,89 @@
...
+{
+
+    @TempDir
+    File tempDir;
+
+    @Test
+    public void testDump() throws PcapNativeException, NotOpenException, IOException, TimeoutException
+    {
+        Properties properties = new Properties();
+        properties.put("zilla.engine.directory", "src/test/resources/zilla_test_stream");
```
jfallows, line 53: Please move all resources used by the test to the parallel package path for this test class, but (deeper) under `src/test/resources/...`.

Suggest using `src/test/resources/io/aklivity/zilla/runtime/command/dump/internal/engine` and `src/test/resources/io/aklivity/zilla/runtime/command/dump/internal/expected.pcap` or similar.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/DumpCommand.java
```diff
@@ -0,0 +1,128 @@
+/*
```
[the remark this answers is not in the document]
bmaidics: Yes, I kept it, because it was easier to debug

incubator/command-dump/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
jfallows, line 1: Needs to be updated to Aklivity license.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
jfallows: You would need to specify binding names, such as `http0` or `tls0`, not binding types.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
```diff
@@ -0,0 +1,222 @@
...
+
+    @Option(name = {"--version"})
+    public boolean version = false;
+
+    @Option(name = {"--continuous"})
+    public boolean continuous = true;
+
+    @Option(name = {"-t", "--bindingTypes"},
+        description = "Dump specific bindings types only, e.g http")
+    public List<String> bindingTypes = new ArrayList<>();
```
jfallows, line 64: This one can be removed.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
```diff
@@ -0,0 +1,222 @@
...
+{
+    @Option(name = {"-v", "--verbose"},
+        description = "Show verbose output")
+    public boolean verbose;
+
+    @Option(name = {"--version"})
+    public boolean version = false;
+
+    @Option(name = {"--continuous"})
+    public boolean continuous = true;
```
jfallows, line 60: Always continuous, no need for option.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/LoggableStream.java
```diff
@@ -0,0 +1,120 @@
+/*
```
jfallows, line 1: No longer used? If so please remove.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/Logger.java
```diff
@@ -0,0 +1,21 @@
+/*
```
jfallows, line 1: No longer used? If so please remove.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/Handlers.java
```diff
@@ -0,0 +1,38 @@
+/*
```
jfallows, line 1: No longer used? If so please remove.

jfallows (changes requested):

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
bmaidics: I only created this for the testing. I'll delete this option, and create a protected field for this. 

I still don't like it (creating a field that will only be used in test), but I can't think of any other quick solution, so we can terminate the process in the test.

bmaidics: But I'm open for suggestion, if you have some idea

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/LoggableStream.java
bmaidics: It's still used. I'd keep it, since this class wraps objects/methods which are needed to read from a stream, and the code is more readable this way.
But if you think, it's not well aligned with the other commands, I can move these out to the ZillaDumpCommand class.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
bmaidics: How can you identify if e.g a BeginFW belongs to http0 or tls0? I can only see the extension type, that is available there. Or did I misunderstood something?

jfallows (changes requested):

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/DumpHandlers.java
```diff
@@ -104,7 +104,9 @@ public void onBegin(BeginFW begin)
         final ExtensionFW extension = begin.extension().get(extensionRO::tryWrap);
```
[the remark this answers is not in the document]
bmaidics: Oh, yes. I forgot to delete it, it was not even used anymore. Done.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/DumpHandlers.java
```diff
@@ -122,7 +124,8 @@ public void onData(DataFW data)
         final ExtensionFW extension = data.extension().get(extensionRO::tryWrap);
         if (extension != null)
         {
-            if (hasExtensionType.test(labelManager.lookupLabel(extension.typeId())))
+            int bindingId = (int) data.routeId();
+            if (hasExtensionType.test(labelManager.lookupLabel(bindingId)))
```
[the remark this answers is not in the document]
bmaidics: Ohh, I see. I didn't understand the first line on the labels file in the test. But that is the namespace....
Thanks

jfallows (changes requested):
I built this PR locally, and modified `zilla-examples` `tcp.echo` example `stack.yml` to use `develop-SNAPSHOT` instead of `latest` image, then followed the README to launch the stack.

I then executed the following to get a shell inside the `zilla` docker service:
```
docker exec -it $(docker ps -q -f name=example_zilla) sh
```

When I then run `./zilla dump -v` or `./zilla dump -v -o zilla.pcap` the executable exits without producing any output nor error which is unexpected.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
```diff
@@ -146,6 +147,17 @@ private static byte[] hexStringToByteArray(String s)
...
+        {
+            try
+            {
+                directory = new URI(System.getProperty("user.dir"));
+            }
+            catch (URISyntaxException e)
+            {
+                System.out.println("Cannot parse working directory");
+            }
+        }
```
jfallows, line 160: Lets change the type of `directory` to `Path`, default it to `Paths.get(".zilla", "engine")` and remove this if block.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
```diff
@@ -0,0 +1,580 @@
...
+@Command(name = "dump", description = "Dump stream content")
+public final class ZillaDumpCommand extends ZillaCommand
+{
+    @Option(name = {"-v", "--verbose"},
+        description = "Show verbose output")
+    public boolean verbose;
+
+    @Option(name = {"-b", "--bindingNames"},
+        description = "Dump specific namespaces, bindings only, e.g example.http0,example.kafka0")
+    public List<String> bindingNames = new ArrayList<>();
```
jfallows, line 75: Even though this contains "binding names" we should just call it `bindings` as there is no other interpretation for the string command line argument and keeps it simpler plus easier to type.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
```diff
@@ -0,0 +1,580 @@
...
+            byte[] bytes = new byte[data.offset() + data.limit()];
+
+            if (data.payload() != null)
+            {
+                DirectBuffer buffer = data.payload().buffer();
+                buffer.getBytes(0, bytes, data.offset(), data.limit());
+
+                TcpHeaderFW tcpHeader = createTcpHeader(streamId, Flag.PSH);
+                PcapPacketHeaderFW pcapHeader = createPcapPacketHeader(PSEUDO_ETHERNET_FRAME.length +
+                    tcpHeader.sizeof() + bytes.length, data.timestamp());
```
jfallows, line 344: Use `data.sizeof()` instead of `bytes.length`.

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
```diff
@@ -0,0 +1,580 @@
...
+    }
+
+    private void onBegin(
+        BeginFW begin)
+    {
+        if (allowedBindings.isEmpty() || allowedBindings.contains(begin.routeId()))
+        {
+            final long streamId = begin.streamId();
+            TcpHeaderFW tcpHeader = createTcpHeader(streamId, Flag.SYN);
+            PcapPacketHeaderFW pcapHeader = createPcapPacketHeader(PSEUDO_ETHERNET_FRAME.length + tcpHeader.sizeof(),
```
jfallows, line 321: Is the IPv6 packet structure included in the `PSEUDO_ETHERNET_FRAME` bytes?
If so, how is the length conveyed to the IPv6 packet in the generated output?

jfallows (changes requested):

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
bmaidics: Logic changed, so it's not relevant anymore. IPv6 is modelled as flyweight.

jfallows (approved):

PR #122, opened 2022-10-25
Remove merged from kafka binding configuration

### LINKED ISSUE #108: Remove merged from kafka binding configuration
Cleanup `merged` in configs, remove from schema and online docs.

ankit-kumar-25:
Fix #108

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
@@ -155,10 +155,10 @@ public final class KafkaMergedFactory implements BindingHandler
     private final MergedBudgetCreditor creditor;
 
     public KafkaMergedFactory(
-        KafkaConfiguration config,
-        EngineContext context,
-        LongFunction<KafkaBindingConfig> supplyBinding,
-        MergedBudgetCreditor creditor)
+            KafkaConfiguration config,
+            EngineContext context,
+            LongFunction<KafkaBindingConfig> supplyBinding,
+            MergedBudgetCreditor creditor)
     {
         this.kafkaTypeId = context.supplyTypeId(KafkaBinding.NAME);
         this.writeBuffer = new UnsafeBuffer(new byte[context.writeBuffer().capacity()]);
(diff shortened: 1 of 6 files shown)


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
```diff
@@ -204,42 +204,37 @@ public MessageConsumer newStream(
...
-                    authorization,
-                    topicName,
-                    resolvedId,
-                    capabilities,
-                    initialOffsetsById,
-                    defaultOffset,
-                    isolation,
-                    deltaType,
-                    ackMode)::onMergedMessage;
-        }
```
jfallows, line 242: This change loses the `binding != null` condition too.

jfallows (changes requested):
Please also consider schema validation and the potential downstream impact on https://github.com/aklivity/zilla-examples/ and https://docs.aklivity.io/zilla/get-started configurations.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
ankit-kumar-25: `binding` is not getting used at any other place, so I wasn’t sure if a null check is required.

ankit-kumar-25:
For schema validation and the potential downstream impact, I was think of introducing another constructor for KafkaOptionsConfig with merged attritbute and make it deprecated. So that we have backward compatibility.
@jfallows Can you please share insights on this?

jfallows:
> For schema validation and the potential downstream impact, I was think of introducing another constructor for KafkaOptionsConfig with merged attritbute and make it deprecated. So that we have backward compatibility. @jfallows Can you please share some insights on this?

@ankit-kumar-25 Yes, that's a good idea to deprecate the `merged` option initially, to minimize backwards incompatibility in the short term and schedule it for removal later.

Our contract with the users of zilla is the `zilla.json` configuration, not the internal code APIs, so for users to see the effect of such deprecation, it would have to be applied to JSON schema validation.

Take a look at https://github.com/aklivity/zilla/tree/develop/specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema and let me know your thoughts.

Suggest also removing all occurrences of `merged` `kafka` binding option from any `zilla-examples` currently specifying it explicitly. This will involve a separate PR on `zilla-examples` repository.

ankit-kumar-25:
@jfallows Update the Schema with deprecated flag for `merged` attribute. 

Also raised PR: https://github.com/aklivity/zilla-examples/pull/11 to remove all occurrences of merged kafka binding option in zilla-examples.

Can you please review this? Thanks!

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
ankit-kumar-25: done, reverted the `binding != null` condition.

jfallows (approved):

PR #126, opened 2022-11-01
Support SASL SCRAM authentication

ankit-kumar-25:


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientSaslHandshaker.java
```diff
@@ -430,8 +568,98 @@ protected final int decodeSaslAuthenticate(
                 client.decodableResponseBytes -= authenticateResponse.sizeof();
                 assert client.decodableResponseBytes >= 0;
 
-                client.onDecodeSaslResponse(traceId);
-                client.onDecodeSaslAuthenticateResponse(traceId, authorization, errorCode);
+                if (client.sasl != null && client.sasl.mechanism != null &&
```
jfallows, line 571: Let's also split `decodeSaslAuthenticate` into `decodeSaslPlainAuthenticate`, `decodeSaslScramAuthenticateFirst` and `decodeSaslScramAuthenticateFinal` to follow the state machine rather than needing to have if conditions inside a general purpose `decodeSaslAuthenticate`.

Note: as with encode, we may need a further `decodeSaslAuthenticate` field to delegate to one of `decodeSaslPlainAuthenticate`, `decodeSaslScramAuthenticateFirst` or `decodeSaslScramAuthenticateFinal` by setting that `decodeSaslAuthenticate` value accordingly while following the encoding and decoding state machine.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientSaslHandshaker.java
```diff
@@ -430,8 +568,98 @@ protected final int decodeSaslAuthenticate(
...
+
+                try
+                {
+                    byte[] serverKey = formatter.serverKey(client.saltedPassword);
+                    byte[] serverSignature = formatter.hmac(serverKey, client.authMessage);
+                    DirectBuffer serverFinalResponse = authenticateResponse.authBytes().value();
+                    String serverFinalMessage = serverFinalResponse.getStringWithoutLengthUtf8(0,
+                            serverFinalResponse.capacity());
+
+                    Matcher serverFinalResponseMatcher = SERVER_FINAL_MESSAGE.matcher(serverFinalMessage);
```
jfallows, line 641: Anywhere you need to use a `Matcher`, pre-create a non-static instance at the `KafkaClientSaslHandshaker` scope by calling `.matcher("")`, and then use `.reset(serverFinalMessage)` to reuse instead of creating a new one each time.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientSaslHandshaker.java
```diff
@@ -430,8 +568,98 @@ protected final int decodeSaslAuthenticate(
...
+                    {
+                        client.onDecodeSaslResponse(traceId);
+                        client.onDecodeSaslAuthenticateResponse(traceId, authorization, errorCode);
+                    }
+
+                }
+                catch (InvalidKeyException e)
+                {
+                    e.printStackTrace();
+                }
```
jfallows, line 662: Avoid these kinds of catch-all exception handlers as they will break the logic of the code when they occur.
Instead determine if the exception is really possible, by inspecting how you generated the inputs, say via `serverKey` and `saltedPassword` above, then remove the `throws` from `hmac(...)` method and use `LangUtil.rethrowUnchecked(ex)` to allow the exception out if it did occur, though you will have already proven to yourself that it is actually not possible.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientSaslHandshaker.java
```diff
@@ -430,8 +568,98 @@ protected final int decodeSaslAuthenticate(
...
+                    byte[] serverSignature = formatter.hmac(serverKey, client.authMessage);
+                    DirectBuffer serverFinalResponse = authenticateResponse.authBytes().value();
+                    String serverFinalMessage = serverFinalResponse.getStringWithoutLengthUtf8(0,
+                            serverFinalResponse.capacity());
+
+                    Matcher serverFinalResponseMatcher = SERVER_FINAL_MESSAGE.matcher(serverFinalMessage);
+                    while (serverFinalResponseMatcher.find())
+                    {
+                        serverFinalMessage = serverFinalResponseMatcher.group(1);
+                    }
```
jfallows, line 645: Is this really a `while` loop? or should it be an `if` with `.matches()` instead?
For example, is it legitimate to have multiple `v=...` in the same server final message?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientSaslHandshaker.java
```diff
@@ -430,8 +568,98 @@ protected final int decodeSaslAuthenticate(
...
+                    byte[] serverKey = formatter.serverKey(client.saltedPassword);
+                    byte[] serverSignature = formatter.hmac(serverKey, client.authMessage);
+                    DirectBuffer serverFinalResponse = authenticateResponse.authBytes().value();
+                    String serverFinalMessage = serverFinalResponse.getStringWithoutLengthUtf8(0,
+                            serverFinalResponse.capacity());
+
+                    Matcher serverFinalResponseMatcher = SERVER_FINAL_MESSAGE.matcher(serverFinalMessage);
+                    while (serverFinalResponseMatcher.find())
+                    {
+                        serverFinalMessage = serverFinalResponseMatcher.group(1);
```
jfallows, line 644: Consider using named matcher groups (instead of numerical) for improved readability.
In the SASL SCRAM spec this is called `v` for `verifier`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientSaslHandshaker.java
```diff
@@ -430,8 +568,98 @@ protected final int decodeSaslAuthenticate(
...
+                    }
+                    if (!Arrays.equals(Base64.getDecoder().decode(serverFinalMessage),
+                            serverSignature))
+                    {
+                        client.onDecodeSaslResponse(traceId);
+                        client.onDecodeSaslAuthenticateResponse(traceId, authorization, 58);
```
jfallows, line 650: Avoid hardcoded values like `58` and instead create a static constant for `ERROR_SASL_AUTHENTICATION_FAILED`.

Please make sure we have a negative test in the scripts that sends an invalid verifier value from the server to confirm we behave correctly.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/config/client.options.sasl.scram-256.json
```diff
@@ -0,0 +1,26 @@
...
+    {
+        "app0":
+        {
+            "type" : "kafka",
+            "kind": "client",
+            "options":
+            {
+                "sasl":
+                {
+                    "mechanism": "scram-sha-1",
```
jfallows, line 13: Configuration is called `scram-256` but mechanism is `scram-sha-1`, please make them consistent, and create a new one only if we need both.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -75,7 +75,7 @@
                                 {
                                     "title": "Mechanism",
                                     "type": "string",
-                                    "enum": [ "plain" ]
+                                    "enum": [ "plain", "scram-sha-256", "scram-sha-1" ]
```
jfallows, line 78: Do we need `scram-sha-512` as well?

jfallows (changes requested):
Excellent progress!

jfallows (approved):

Issue #138, opened 2022-12-21
Watch `zilla` configuration for changes and apply automatically


jfallows:
After `zilla` is configured and running successfully, support ability to change `zilla.yaml` and have `zilla` notice to apply the changes.

Typically this will involve tearing down an existing configuration `namespace`, and applying the new configuration `namespace`. This requires tearing down any existing streams for bindings in the old `namespace` before applying the new `namespace` configuration.

Note that the new `zilla.yaml` contents may be invalid syntax, and should then be ignored but with a suitable error message to indicate that the attempted configuration changes were unsuccessful, so that the failure can be detected in the logs.

bmaidics:
Edit config file using kubectl:
kubectl edit configmap zilla -n <kubernetes namespace>

PR #141, opened 2023-01-04
Dynamic config

bmaidics:
https://github.com/aklivity/zilla/issues/138


bmaidics:
Currently implemented the config watcher directly in ConfigTask.java to avoid code duplication. Let me know if you'd prefer it to be in its different class and I can refactor it.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/DispatchAgent.java
```diff
@@ -153,6 +154,12 @@ public class DispatchAgent implements EngineContext, Agent
     private final BufferPoolLayout bufferPoolLayout;
     private final RingBuffer streamsBuffer;
     private final MutableDirectBuffer writeBuffer;
+
+    //TODO: remove stream and throttle handlers AND tell them to mclean up after themselves
+    // (abort for streams, reset for throttles)
```
jfallows, line 159: Remove note comments

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/DispatchAgent.java
```diff
@@ -153,6 +154,12 @@ public class DispatchAgent implements EngineContext, Agent
...
+    private final LongConsumer detachStreams;
```
jfallows, line 160: Is this used locally? If not, no need for field.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/NamespaceRegistry.java
```diff
@@ -60,6 +62,7 @@ public NamespaceRegistry(
         this.bindingsById = new Int2ObjectHashMap<>();
         this.guardsById = new Int2ObjectHashMap<>();
         this.vaultsById = new Int2ObjectHashMap<>();
+        this.detachStreamsByBindingId = detachStreamsByBindingId;
```
jfallows, line 65: Rename to `detachBinding` ?

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureIT.java
```diff
@@ -0,0 +1,136 @@
...
+        .clean();
+
+    @Rule
+    public final TestRule chain = outerRule(engine).around(k3po).around(timeout);
+
+    @BeforeClass
+    public static void copyConfigToCorrectLocation() throws Exception
+    {
+        String source = new File(ReconfigureIT.class.getClassLoader()
+            .getResource("io/aklivity/zilla/runtime/engine/internal/zilla.reconfigure.original.json")
```
jfallows, line 70: Perhaps use `ReconfigureIT.class.getResource(...)` instead to resolve relative to class in resources directory.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -88,6 +88,7 @@ public void detach(
         BindingConfig binding)
     {
         router.remove(binding.id);
+        //
```
jfallows, line 91: Remove empty note comment

runtime/engine/src/test/resources/io/aklivity/zilla/runtime/engine/internal/zilla.reconfigure.json
```diff
@@ -0,0 +1,12 @@
...
+  "bindings":
+  {
+    "net0":
+    {
+      "type" : "test",
+      "kind": "proxy",
+      "exit": "app0"
+    }
+  }
+}
```
jfallows, line 12: You can rename these to `reconfigure-before.json` and `reconfigure-after.json`.

specs/binding-tcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tcp/config/server.reconfigure.json
```diff
@@ -0,0 +1,17 @@
...
+            "kind": "server",
+            "options":
+            {
+                "host": "0.0.0.0",
+                "port": 8080
+            },
+            "exit": "app0"
+        }
+    }
+}
```
jfallows, line 17: Remove, no longer needed.

specs/binding-tcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tcp/streams/application/rfc793/client.sent.data.reconfigure.delete_create/client.rpt
```diff
@@ -0,0 +1,28 @@
...
+    option zilla:transmission "duplex"
+
+connected
+
+write notify CONFIG_CHANGED
+
+connect "zilla://streams/app0"
+    option zilla:window 8192
+    option zilla:transmission "duplex"
+connected
```
jfallows, line 28: Only use `.` (not `_`) for separator in scenario names, it's a convention we've applied for the other specs.

specs/binding-tcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tcp/streams/application/rfc793/client.sent.data.reconfigure.delete_create/client.rpt
```diff
@@ -0,0 +1,28 @@
+#
```
jfallows, line 1: What is `delete_create` scenario?
I was expecting to see 3 scenarios; created, modified, deleted

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -158,8 +170,92 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
...
+        try
+        {
+            watchService = FileSystems.getDefault().newWatchService();
+            Paths.get(configURL.toURI()).getParent().register(watchService, StandardWatchEventKinds.ENTRY_MODIFY,
+                StandardWatchEventKinds.ENTRY_CREATE, StandardWatchEventKinds.ENTRY_DELETE);
+        }
+        catch (FileSystemNotFoundException e)
+        {
+            //Should only occur in tests
+            return;
```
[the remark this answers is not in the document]
bmaidics: I'm already modelling by schema. watchConfig is only called if we have file as a schema. FileSystemNotFoundException is thrown in one of the tests (I don't remember exactly which)

specs/binding-tcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tcp/streams/application/rfc793/client.sent.data.reconfigure.delete_create/client.rpt
bmaidics: I forgot to delete these from the TCP specs. Correcte.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/NamespaceRegistry.java
```diff
@@ -60,6 +62,7 @@ public NamespaceRegistry(
         this.bindingsById = new Int2ObjectHashMap<>();
         this.guardsById = new Int2ObjectHashMap<>();
         this.vaultsById = new Int2ObjectHashMap<>();
+        this.detachBinding = detachBinding;
```
jfallows, line 65: Move this assignment up below `this.supplyLoadEntry = supplyLoadEntry;`

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureIT.java
```diff
@@ -0,0 +1,139 @@
...
+    public void shouldConnectToZillaOnNewPortWhenConfigModified() throws Exception
+    {
+        k3po.start();
+
+        String source = new File(ReconfigureIT.class
+            .getResource("zilla.reconfigure.after.json")
+            .toURI()).getAbsolutePath();
+        String target = new File("target/test-classes/" +
+            "io/aklivity/zilla/runtime/engine/internal/zilla.reconfigure.json").getAbsolutePath();
+        FileUtils.copy(source, target);
```
jfallows, line 88:
```
import java.nio.file.Files;
```
Define some useful fields, designed to survive a refactoring.
```
private final String packageName = ReconfigureIT.class.getPackageName();
private Path configDir = Paths.get("target/test-classes", packageName.replace(".", "/"));
```
Use Files.copy to read from classpath and write to filesystem.
```
InputStream source = ReconfigureIT.class.getResourceAsStream("zilla.reconfigure.after.json");
Path target = configDir.resolve("zilla.reconfigure.json");
Files.copy(source, target);
```

Similar changes for the other methods.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureIT.java
```diff
@@ -0,0 +1,139 @@
...
+            "io/aklivity/zilla/runtime/engine/internal/zilla.reconfigure.missing.json").delete();
+    }
+
+    @Test
+    @Configuration("zilla.reconfigure.json")
+    @Specification({
+        "${app}/client.sent.data.reconfigure.modify/server",
+        "${net}/client.sent.data.reconfigure.modify/client"
+    })
+    public void shouldConnectToZillaOnNewPortWhenConfigModified() throws Exception
```
jfallows, line 79: Rename to `shouldReconfigureWhenModified()`.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureIT.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+
+    @Test
+    @Configuration("zilla.reconfigure.missing.json")
+    @Specification({
+        "${app}/client.sent.data.reconfigure.create/server",
+        "${net}/client.sent.data.reconfigure.create/client"
+    })
+    public void shouldConnectToZillaWhenConfigCreated() throws Exception
```
jfallows, line 103: Rename to `shouldReconfigureWhenCreated()`.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureIT.java
```diff
@@ -0,0 +1,139 @@
...
+        k3po.finish();
+    }
+
+    @Test
+    @Configuration("zilla.reconfigure.json")
+    @Specification({
+        "${app}/client.sent.data.reconfigure.delete/server",
+        "${net}/client.sent.data.reconfigure.delete/client"
+    })
+    public void shouldNotConnectToZillaWhenConfigDeleted() throws Exception
```
jfallows, line 127: Rename to `shouldReconfigureWhenDeleted()`.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -218,15 +218,21 @@ public Statement apply(
...
+                    if (configURL == null)
+                    {
+                        configURL = new URL(testClass.getClassLoader().getResource(configurationRoot).toString() +
+                            "/" + config.value());
+                    }
                 }
                 else
                 {
-                    String resourceName = String.format("%s-%s", testClass.getSimpleName(), config.value());
+                    resourceName = String.format("%s-%s", testClass.getSimpleName(), config.value());
```
jfallows, line 235: Are the changes in this class still necessary?

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureIT.java
```diff
@@ -0,0 +1,139 @@
...
+    {
+        k3po.start();
+        File configFile = new File("target/test-classes/" +
+            "io/aklivity/zilla/runtime/engine/internal/zilla.reconfigure.json");
+        configFile.delete();
+        Thread.sleep(2000);
```
jfallows, line 133: In general, it's better not to use `Thread.sleep(delay)` in the tests if it is avoidable as the sleep delays are magic values that work for some platforms but not others and end up being extra large to be conservative, slowing down the build.

Perhaps we can use the WatchService in the test to notice the event, and that is our indication that the engine has also been notified via WatchService? If this works, then the tests would be event-driven and always execute at full speed on all platforms.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/DispatchAgent.java
```diff
@@ -1311,9 +1349,16 @@ private MessageConsumer handleBeginInitial(
             if (newStream != null)
             {
                 final long replyId = supplyReplyId(initialId);
-                streams[streamIndex(initialId)].put(instanceId(initialId), newStream);
-                throttles[throttleIndex(replyId)].put(instanceId(replyId), newStream);
+                final int streamIndex = streamIndex(initialId);
+                final int throttleIndex = throttleIndex(replyId);
+                streams[streamIndex].put(instanceId(initialId), newStream);
+                throttles[throttleIndex].put(instanceId(replyId), newStream);
```
jfallows, line 1355: Look for calls to `streams[streamIndex].remove(...)` and `throttles[streamIndex].remove(...)` too, so we can cleanup binding streams and throttles properly too.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/DispatchAgent.java
```diff
@@ -867,7 +902,10 @@ private void onSystemSignal(
         switch (signalId)
         {
         case SIGNAL_TASK_QUEUED:
-            taskQueue.poll().run();
+            if (!taskQueue.isEmpty())
+            {
+                taskQueue.poll().run();
+            }
```
jfallows, line 908: What triggered the need for this change? Feels like we might be working around a bug.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/DispatchAgent.java
```diff
@@ -434,6 +438,37 @@ public void detachSender(
...
+                {
+                    throttles[throttleId].values().forEach(streamHandler ->
+                        doSyntheticAbort(throttleId, streamHandler)
+                    );
+                    throttles[throttleId] = new Int2ObjectHashMap<>();
+                }
+            );
+        }
+    }
+
```
jfallows, line 471: Please reevaluate this implementation after applying feedback for `handleBeginInitial` below.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -158,8 +172,86 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
...
+                        Thread.currentThread().interrupt();
+                    }
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            rethrowUnchecked(ex);
         }
+    }
```
jfallows, line 251: This method is fairly unwieldy and trying to handle a lot of different cases as if it was a linear flow.

I think we might need to restructure the code into 3 tasks.
 - watcher task
 - register task
 - unregister task

The watcher task is responsible for detecting a new configuration, either file based or http based, so initially `FileWatcherTask`, later `HttpWatcherTask`.

When the `FileWatcherTask` detects a file delete or file modify event on the config file it is watching, it submits an `UnregisterTask` to unregister the namespace, detaching any streams for bindings in the namespace.

When the `FileWatcherTask` detects a file create or file modify event on the config file it is watching, it submits a `RegisterTask` to register the new namespace.

I think this design will simplify the approach into more bite-sized chunks of functionality, and also lend it self to being extended for HttpWatcherTask via long-polling, while keeping the register and unregister behaviors unaffected.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/DispatchAgent.java
```diff
@@ -1311,9 +1349,16 @@ private MessageConsumer handleBeginInitial(
...
+                final int streamIndex = streamIndex(initialId);
+                final int throttleIndex = throttleIndex(replyId);
+                streams[streamIndex].put(instanceId(initialId), newStream);
+                throttles[throttleIndex].put(instanceId(replyId), newStream);
                 supplyLoadEntry.apply(routeId).initialOpened(1L);
+
+                streamIndicesByBinding.computeIfAbsent(NamespacedId.localId(routeId), k -> new IntHashSet())
+                    .add(streamIndex);
+                throttleIndicesByBinding.computeIfAbsent(NamespacedId.localId(routeId), k -> new IntHashSet())
+                    .add(throttleIndex);
```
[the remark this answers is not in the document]
bmaidics: Thanks. A lot of really useful informations!

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureIT.java
bmaidics: Forgot to rename after moving it from TCP binding. Done

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
bmaidics: This is needed for the 0-1 scenario, where we originally don't have the config file in place. This scenario should contain some configURL, if not, we can't watch the parent directory for changes. Since the current solution (testClass.getClassLoader().getResource(resourceName)) returns null if the file does not exists, we need to pass the configURL somehow. This is the workaround for that. Maybe there's a better solution, but this is the solution I could think of that requires the least amount of work.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
bmaidics: It made the code much simpler. I was also thinking of doing something similar, but wanted to get your opinion first (https://github.com/aklivity/zilla/pull/141#issuecomment-1375548091).

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/DispatchAgent.java
bmaidics: Bug found, and resolved. The WatcherTask didn't close properly as ForkJoinTask.cancel is not working as expected.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,169 @@
...
+                                    rootNamespace = commonPool().submit(
+                                        new RegisterTask(configURL, schemaTypes, guardsByType, supplyId, maxWorkers, tuning,
+                                            dispatchers, errorHandler, logger, context, config, extensions)
+                                    ).get();
+                                }
+                            }
+                        }
+                        key.reset();
+                    }
+                    catch (InterruptedException ex)
```
bmaidics, line 125: Just a note here: we discussed that maybe catching a ClosedWatchServiceException here would be enogh, so we don't need to store the current thread and do interrupt. However, we only get ClosedWatchServiceException, when we call take or poll on a closed WatchService. In our case, we might be already at the staet of unregistering/registering, when we got the close call, which closes the WatchService. In that case, we won't stop the detach. Tested it, and got the same NPE, when trying to close an already closed stream.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -153,9 +155,32 @@ public final class Engine implements AutoCloseable
...
             .collect(Collectors.toMap(g -> g.name(), g -> g));
 
-        final Callable<Void> configure =
-                new ConfigureTask(configURL, schemaTypes, guardsByType::get, labels::supplyLabelId, maxWorkers, tuning,
-                        dispatchers, errorHandler, logger, context, config, extensions);
+
+        registerTask = new RegisterTask(configURL, schemaTypes, guardsByType::get,
+            labels::supplyLabelId, maxWorkers, tuning, dispatchers, errorHandler, logger, context, config, extensions);
+
+        if (configURL == null || "file".equals(configURL.getProtocol()))
```
jfallows, line 162: Can `configURL` actually be null?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -153,9 +155,32 @@ public final class Engine implements AutoCloseable
...
+                public void setRootNamespace(NamespaceConfig rootNamespace)
+                {
+                }
+
+                @Override
+                public boolean run()
+                {
+                    return true;
+                }
+            };
```
jfallows, line 182: throw an exception here to indicate the protocol is not supported instead

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,169 @@
...
+        return true;
+    }
+
+    @Override
+    public void setRootNamespace(NamespaceConfig rootNamespace)
+    {
+        this.rootNamespace = rootNamespace;
+    }
+
+    private byte[] getConfigHash()
```
jfallows, line 146: Rename method to `computeHash`

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -197,24 +221,25 @@ public long counter(
...
             AgentRunner.startOnThread(runner, Thread::new);
         }
-
-        this.configureRef = commonPool().submit(configure);
-
-        return configureRef;
+        Future<NamespaceConfig> register = commonPool().submit(registerTask);
+        NamespaceConfig rootNamespace = register.get();
+        watcherTask.setRootNamespace(rootNamespace);
+        commonPool().submit(watcherTask);
```
jfallows, line 233: I think we need to simplify the flow here a bit more.
The `watcherTask` should be in charge of noticing a new configuration, even at startup.
When it detects a change, the reaction should be established here, with an unregister of the old config (possible no-op) and a register of the new config (possible no-op).
Use a listener callback approach on `watcherTask` so we can easily setup the same reactions for http vs file or any other watcher implementation.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,169 @@
...
+        this.supplyId = supplyId;
+        this.maxWorkers = maxWorkers;
+        this.tuning = tuning;
+        this.dispatchers = dispatchers;
+        this.errorHandler = errorHandler;
+        this.logger = logger;
+        this.context = context;
+        this.config = config;
+        this.extensions = extensions;
+    }
```
jfallows, line 83: This constructor simplifies quite a bit after we move the config changed reactions to outside this class, as mentioned above.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/stream/Target.java
```diff
@@ -231,6 +235,11 @@ private boolean handleWriteInitial(
                 handled = streamsBuffer.test(msgTypeId, buffer, index, length);
                 streams[streamIndex(streamId)].remove(instanceId(streamId));
                 supplyLoadEntry.apply(routeId).initialClosed(1L).initialErrored(1L);
+                LongHashSet streamIdSet = streamSets.get(routeId);
+                if (streamIdSet != null)
+                {
+                    streamIdSet.remove(streamId);
+                }
```
jfallows, line 242: Can `streamIdSet` actually be `null`?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/stream/Target.java
```diff
@@ -299,6 +308,11 @@ private boolean handleWriteReply(
                 handled = streamsBuffer.test(msgTypeId, buffer, index, length);
                 streams[streamIndex(streamId)].remove(instanceId(streamId));
                 correlations.remove(streamId);
+                LongHashSet streamIdSet = streamSets.get(routeId);
+                if (streamIdSet != null)
+                {
+                    streamIdSet.remove(streamId);
+                }
```
jfallows, line 315: Can `streamIdSet` actually be `null`?

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
bmaidics: Originally, it could be null in ITs. But I changed so it can't be by setting it to the expected URL, so we can watch config file change for the 0-1 (create) scenario.

bmaidics: So you're right. I can delete this check

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/stream/Target.java
bmaidics: Yes, it's needed, because when detaching a stream during reconfiguration, the DispatchAgent already removes the streamIdSet before triggering doSyntheticAbort/reset (which triggers these calls). This removal is only needed for the other syntheticAbort/reset calls, when closing Zilla regularly.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
bmaidics: I actually started a refactor similar to this, when did the rebase on the HTTP watcher. You're right, the code is mutch cleaner this way.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -197,24 +218,23 @@ public long counter(
...
     {
         for (AgentRunner runner : runners)
         {
             AgentRunner.startOnThread(runner, Thread::new);
         }
-
-        this.configureRef = commonPool().submit(configure);
-
-        return configureRef;
+        reconfigure();
```
bmaidics, line 227: I think we should call the initial configuration here instead of the WatcherTask. Tried it that way, and since we won't wait for the WatcherTask to finish (long running task), it can happen (and happened as well) that the Zilla starts the stopping process before the configuration is actually finished. A thing that we can do here instead of this, is to wait for the register CountDownLatch. Let me know if you prefer that approach.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,132 @@
...
+    public boolean run()
+    {
+        try
+        {
+            doInitialConfiguration();
+
+            Path configPath = Paths.get(new File(configURL.getPath()).getAbsolutePath());
+
+            WatchService watchService = FileSystems.getDefault().newWatchService();
+            configPath.getParent().register(watchService, ENTRY_MODIFY, ENTRY_CREATE, ENTRY_DELETE);
```
jfallows, line 45: This call to `register` returns a `WatchKey`.

Suggest retaining a mapping of watched configURL paths to (parent) `WatchKey`. This will help with the test below when a `WatchKey` is returned from `watchService.take()`. If the `event.context()` path is not present then the event can be ignored.

We'd also need to retain a mapping of watched configURL paths to computed hash value.

The change listener callback signature will need to change to `(URL configURL, String oldValue, String newValue)`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/RegisterTask.java
```diff
@@ -305,9 +258,11 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
...
-
-        // TODO: repeat to detect and apply changes
-
         return null;
     }
+
+    public void setConfigText(String configText)
+    {
+        this.configText = configText;
+    }
```
jfallows, line 267: As discussed, in the Engine, we should track `configURL` to parsed `NamespaceConfig` which may be initially `null` before the first parse of a new `configURL`.

When the engine detects a config change from the watcher task, it can trigger the unregister of the old namespace and a parse plus register of the the new config.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -152,9 +160,25 @@ public final class Engine implements AutoCloseable
         final Map<String, Guard> guardsByType = guards.stream()
             .collect(Collectors.toMap(g -> g.name(), g -> g));
 
-        final Callable<Void> configure =
-                new ConfigureTask(config.configURL(), schemaTypes, guardsByType::get, labels::supplyLabelId, maxWorkers, tuning,
-                        dispatchers, errorHandler, logger, context, config, extensions);
+        this.rootConfigURL = config.configURL();
```
jfallows, line 163: Given that we now default the configuration URL, i don't think this can be `null` any more, so we don't need code to treat it specially.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -152,9 +160,25 @@ public final class Engine implements AutoCloseable
...
+            this.watcherTask = new FileWatcherTask(this::reconfigure);
+        }
+        else if ("http".equals(this.rootConfigURL.getProtocol()) || "https".equals(this.rootConfigURL.getProtocol()))
+        {
+            throw new UnsupportedOperationException();
+        }
+        else
+        {
+            this.watcherTask = new FileWatcherTask(this::reconfigure);
+        }
```
jfallows, line 175: Instead of assuming we only support http, https and file, then testing for the ones we don't support, better to test for the one we do support (file) and fail if not that, then continue without `else` as throw will escape out of the entire code block anyway.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -213,15 +213,21 @@ public Statement apply(
             Configuration config = description.getAnnotation(Configuration.class);
             if (config != null)
             {
+                String resourceName;
```
jfallows, line 216: Please undo this change if not relied upon elsewhere.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+
+public class FileWatcherTask extends WatcherTask
+{
+    private final Map<Path, byte[]> configHashes;
+    private WatchService watchService;
+
+    public FileWatcherTask(BiConsumer<URL, String> configChangeListener)
```
jfallows, line 33: Please format parameters on separate lines. (all methods)

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,147 @@
...
+        configChangeListener.accept(configURL, configText);
+    }
+
+    private String readConfigText(URL configURL)
+    {
+        String configText;
+        if (configURL == null)
+        {
+            configText = CONFIG_TEXT_DEFAULT;
+        }
```
jfallows, line 114: I don't think we want to be adding semantics about config defaulting into the file watcher, that seems more appropriate at the engine level or similar, so that file vs http watcher doesn't have to replicate defaulting logic.

The watcher should really be dealing with both `null` (not found) and non-`null` values for config.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,147 @@
...
+        {
+            rethrowUnchecked(ex);
+        }
+        return true;
+    }
+
+    @Override
+    protected void doInitialConfiguration(URL configURL)
+    {
+        Path configPath = Paths.get(new File(configURL.getPath()).getAbsolutePath());
```
jfallows, line 92: Simplifies to `Path configPath = Paths.get(configURL.toURI()).getAbsolutePath();` ?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,147 @@
...
+            configPath.getParent().register(watchService, ENTRY_MODIFY, ENTRY_CREATE, ENTRY_DELETE);
+        }
+        catch (IOException ignored)
+        {
+        }
+
+        configChangeListener.accept(configURL, configText);
+    }
+
+    private String readConfigText(URL configURL)
```
jfallows, line 108: Simpler to pass `Path` here instead of `URL`, as we are dealing only with `file` scheme URLs.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+        return configText;
+    }
+
+    private byte[] computeHash(String configText)
+    {
+        byte[] hash = new byte[0];
+        try
+        {
+            MessageDigest md5Digest = MessageDigest.getInstance("MD5");
```
jfallows, line 138: Call this `md5` and hoist to constructor, instead of reinstantiating for each call to `computeHash`.

This might require creating a private static method to handle the exception, calling from constructor to initialize and return new message digest instance.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,147 @@
...
+            hash = md5Digest.digest(configText.getBytes(UTF_8));
+        }
+        catch (NoSuchAlgorithmException ex)
+        {
+            return hash;
```
jfallows, line 143: Use `LangUtil.rethrowUnchecked(ex);` here.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,147 @@
...
+            {
+                configText = CONFIG_TEXT_DEFAULT;
+            }
+        }
+        return configText;
+    }
+
+    private byte[] computeHash(String configText)
+    {
+        byte[] hash = new byte[0];
```
jfallows, line 135: This will no longer be needed once the feedback below is applied.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,147 @@
...
+                                configHashes.put(changed, newConfigHash);
+                                configChangeListener.accept(configURLs.get(changed), configText);
+                            }
+                        }
+                    }
+                    key.reset();
+                }
+                catch (InterruptedException ex)
+                {
+                    watchService.close();
```
jfallows, line 77: Instead of using an explicit `Thread.interrupt()` to escape the `take()` method and react by closing the watch service, can we have a `close()` method on the watcher task that calls `watcherService.close()` and triggers `ClosedWatchServiceException` to escape the `take()` method?

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
bmaidics: Done. Just a note here: we need to watch for "jar" as protocol, since in the tests, the configs are inside a jar.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,147 @@
...
+        {
+            while (true)
+            {
+                try
+                {
+                    final WatchKey key = watchService.take();
+
+                    for (WatchEvent<?> event : key.pollEvents())
+                    {
+                        final Path changed = ((Path) key.watchable()).resolve((Path) event.context());
```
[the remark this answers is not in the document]
bmaidics: Yes, it's relative. I was thinking the same, but after test failures + debug I realized it's the only way.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
bmaidics: We already discussed this: https://github.com/aklivity/zilla/pull/141/commits/d9d197ad2ffb2dbb87770810be015d5a49e5f2f1#r1082477238
Short answer: it's not enough

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -244,6 +266,22 @@ public void close() throws Exception
...
 
+    private void reconfigure(URL configURL, String configText)
+    {
+        try
+        {
+            NamespaceConfig namespace = namespaces.get(configURL);
+            unregisterTask.setRootNamespace(namespace);
+            commonPool().submit(unregisterTask).get();
+            registerTask.setConfigText(configText);
+            namespaces.put(configURL, commonPool().submit(registerTask).get());
```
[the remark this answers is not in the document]
bmaidics: I actually like having seperate tasks. If we'd put these parts in the engine, it'd have lots of fields that are only necessary for configuration. I think it's also beneficial that now the engine is (almost) only responsible for lifecycle management of Zilla (start, close). Other than those, it creates the tasks, and let them handle the configuration part. But maybe merging register + unregister is worth considering, since they're always called together now. 

Also if you're not convinced, I'm open to change these to a simple function :)

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
bmaidics: If we get IOException during read, we now let the RegisterTask to detect that and set the the config to default. The watcherTask still registers the WatchService for the file

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
bmaidics: It's not enoguh, since we have configfiles inside jars. If we're using only Paths, that will be a FileSystemNotFoundException.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -213,15 +213,21 @@ public Statement apply(
...
                 if (configurationRoot != null)
                 {
-                    String resourceName = String.format("%s/%s", configurationRoot, config.value());
+                    resourceName = String.format("%s/%s", configurationRoot, config.value());
                     URL configURL = testClass.getClassLoader().getResource(resourceName);
+                    if (configURL == null)
+                    {
+                        configURL = new URL(testClass.getClassLoader().getResource(configurationRoot).toString() +
+                            "/" + config.value());
+                    }
```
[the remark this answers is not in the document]
bmaidics: I already replied to this. Please find it here: https://github.com/aklivity/zilla/pull/141/files#r1068938356
If you have other suggestion, how to do this, please let me know.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
jfallows: Yeah, I recall that discussion but was under the impression it revolved around Thread.interrupt() not working on commonPool task submissions.

I was asking if you don't rely on Thread.interrupt(), using WatcherService.close() instead, then is that sufficient for commonPool too.

No worries if it is really insufficient, just making sure. 😄 

For example, the following unit test passes for me.

```
public class WatchServiceTest
{
    @Test
    public void shouldClose() throws Exception
    {
        final WatchService service = FileSystems.getDefault().newWatchService();

        Runnable task = () ->
        {
            watch:
            while (true)
            {
                try
                {
                    service.take();
                }
                catch (InterruptedException | ClosedWatchServiceException ex)
                {
                    break watch;
                }
            }
        };

        final Future<?> future = ForkJoinPool.commonPool().submit(task);

        service.close();

        future.get();
    }
}
```

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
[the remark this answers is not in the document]
jfallows: It might be helpful to combine register + unregister and it's not really as task any more, no need to submit on another thread, perhaps just a helper class to manage the relevant state, single instance, simple method call to trigger behavior when needed.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -196,24 +220,23 @@ public long counter(
...
+        return CompletableFuture.completedFuture(null);
     }
 
     @Override
     public void close() throws Exception
     {
         final List<Throwable> errors = new ArrayList<>();
 
-        configureRef.cancel(true);
+        watcherTask.interrupt();
```
jfallows, line 239: See reply comment in PR conversation about considering `watcherTask.close()` approach here.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -196,24 +220,23 @@ public long counter(
...
         {
             AgentRunner.startOnThread(runner, Thread::new);
         }
-
-        this.configureRef = commonPool().submit(configure);
-
-        return configureRef;
+        commonPool().submit(watcherTask);
+        watcherTask.onURLDiscovered(rootConfigURL);
+        return CompletableFuture.completedFuture(null);
```
jfallows, line 231: If `start()` always returns a completed future, then perhaps it should return `void` instead.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
jfallows: Is there any issue with `jar:` protocol config URLs?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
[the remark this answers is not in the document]
jfallows: Okay, then let's hoist `keyPath` outside the loop of poll events.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,140 @@
...
+                        if (configHashes.containsKey(changed))
+                        {
+                            String newConfigText = readConfigText(configURLs.get(changed));
+                            byte[] oldConfigHash = configHashes.get(changed);
+                            byte[] newConfigHash = computeHash(newConfigText);
+                            if (!Arrays.equals(oldConfigHash, newConfigHash))
+                            {
+                                configHashes.put(changed, newConfigHash);
+                                configChangeListener.accept(configURLs.get(changed), newConfigText);
```
jfallows, line 75: Define `changedURL` as `configURLs.get(changed)` to improve readability.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,140 @@
...
+{
+    private final Map<Path, byte[]> configHashes;
+    private final Map<Path, URL> configURLs;
+    private MessageDigest md5;
+    private WatchService watchService;
+
+    public FileWatcherTask(
+        BiConsumer<URL, String> configChangeListener)
+    {
+        super(configChangeListener);
```
jfallows, line 38: Suggest renaming `configChangeListener` to `changeListener` since this is all about config already.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
jfallows: Marked resolved but not applied?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+        catch (IOException ex)
+        {
+            rethrowUnchecked(ex);
+        }
+        return true;
+    }
+
+    @Override
+    protected void doInitialConfiguration(
```
jfallows, line 96: There's no such thing as initial configuration now, right?
Let's rename this, and also determine if it is still needed on the superclass.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/RegisterTask.java
```diff
@@ -307,9 +270,12 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
...
-        // TODO: repeat to detect and apply changes
-
         return null;
     }
+
+    public void setConfigText(
+        String configText)
+    {
+        this.configText = configText;
+    }
```
jfallows, line 280: This state field seems necessary only to temporarily hold a parameter for RegisterTask prior to submitting on a different thread.

This can be removed after changing to a direct method call on RegisterTask, passing as a parameter instead.

Note: RegisterTask should probably be renamed as it's not really a task.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/RegisterTask.java
```diff
@@ -296,7 +258,8 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
             }
             future.join();
 
-            extensions.forEach(e -> e.onConfigured(context));
+            extensions.forEach(e -> e.onRegistered(context));
+            return namespace;
```
jfallows, line 262: Define `NamespaceConfig namespace = null;` at top of method, and use single return at end of method.

At the moment, it looks like you might be skipping over the error reporting after the try/catch.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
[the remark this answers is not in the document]
jfallows: Sorry, the linked comment is not showing up, please summarize what breaks when omitted.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
bmaidics: No problem with them. Just need to use FIleWatcher for them, so in the if statement, we need it as well.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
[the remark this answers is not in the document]
bmaidics: This is needed for the 0-1 scenario, where we originally don't have the config file in place. This scenario should contain some configURL, if not, we can't watch the parent directory for changes. Since the current solution (testClass.getClassLoader().getResource(resourceName)) returns null if the file does not exists, we need to pass the configURL somehow. This is the workaround for that. Maybe there's a better solution, but this is the solution I could think of that requires the least amount of work.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
bmaidics: Ohh, yes. I don't know why I marked this as resolved. When we use config files in a jar, we get: 'java.nio.file.FileSystemNotFoundException' exception if trying to get the Path from the URI

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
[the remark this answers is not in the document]
jfallows: Yeah, I'd prefer we revert the changes in this code and find a different solution if possible, perhaps specific to the 0-1 test case.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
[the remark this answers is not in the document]
bmaidics: I like that idea. I created a ConfigManager class for this.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
bmaidics: That's right, we can actually remove it. But, we'll need a new way for Zilla to wait for the initial configuration. I have ideas. Let's discuss this.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
bmaidics: Actually, this looks like it's enoguh. I also tried this method, but what I think I missed is the future.get call. And it really makes sense that without that, it's not working .

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/WatchedConfig.java
```diff
@@ -0,0 +1,179 @@
...
+            {
+                Path targetPath = Files.readSymbolicLink(configPath);
+                if (!targetPath.isAbsolute())
+                {
+                    targetPath = configPath.getParent().resolve(targetPath);
+                }
+                // This is needed so if the symlink contains a symlink chain, those are watched as well
+                observablePaths.addLast(targetPath);
+                // We need to watch for the actual file changes as well
+                watchedPaths.add(toRealPath(targetPath));
```
jfallows, line 75: Why do we need to add this here?
Isn't it covered by `if (watchedPaths.add(observablePath))` at line `81` below?
What breaks if we don't use `toRealPath` ?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/WatcherTask.java
```diff
@@ -0,0 +1,22 @@
...
+    protected final BiConsumer<URL, String> changeListener;
+
+    protected WatcherTask(
+        BiConsumer<URL, String> changeListener)
+    {
+        this.changeListener = changeListener;
+    }
+
+    public abstract void onURLDiscovered(
+        URL configURL);
```
jfallows, line 20: Rename to `public abstract void watch(URL configURL)`.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -246,15 +247,16 @@ public void evaluate() throws Throwable
                 final ErrorHandler errorHandler = ex ->
                 {
                     errors.add(ex);
+                    ex.printStackTrace();
```
jfallows, line 250: Not needed.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigManager.java
```diff
@@ -224,7 +180,7 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
             namespace.id = supplyId.applyAsInt(namespace.name);
 
             // TODO: consider qualified name "namespace::name"
-            namespace.resolveId = name -> name != null ? NamespacedId.id(namespace.id, supplyId.applyAsInt(name)) : 0L;
+            namespace.resolveId = idResolver(namespace);
```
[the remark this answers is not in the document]
bmaidics: I only created this, because the namespace variable has to be effectively final, as we declare it at the beginning of the method now. I just resolved this by making a final copy of namespace.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -196,24 +216,23 @@ public long counter(
...
         for (AgentRunner runner : runners)
         {
             AgentRunner.startOnThread(runner, Thread::new);
         }
-
-        this.configureRef = commonPool().submit(configure);
-
-        return configureRef;
+        watcherTaskRef = commonPool().submit(watcherTask);
+        watcherTask.onURLDiscovered(rootConfigURL);
```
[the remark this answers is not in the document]
bmaidics: I just renamed it to handleConfigURL.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/WatchedConfig.java
bmaidics: Suppose we have this, as in the test: 
set config: zilla.reconfigure.modify.complex.chain.json -> data/configs/zilla.reconfigure.modify.complex.chain.json
data -> symlink
configs -> realconfigs
Real config is at: symlink/realconfigs/zilla.reconfigure.modify.complex.chain.json
If we remove this line, we don't watch for the file actual config file change, only the symlinks. In our case: 
......../symlink/realconfigs/zilla.reconfigure.modify.complex.chain.json will not be added to watchedPaths

jfallows: It seems this is a nuance to make sure we register on the right path, for all watched paths, so it seems odd to handle it as a special case here, but not on other calls to `watchedPaths.add(...)`.

Perhaps it would be more consistent to call `toRealPath` just before registering with the watch service?

bmaidics: That's not good I think, since then we're not watching the symlink's parent itself, rather we'll watch the real path's parent.

jfallows: Yep, this is at the core of my question.

We have different treatment for the actual `configPath` file versus each intermediate step of observable paths. Only the `configPath` is being processed with `toRealPath`.

Does that leave some edge cases open where we are not properly watching the real path of the other observable paths? Perhaps directory-based symlinks are implicitly followed during `Path` registration, making the call `toRealPath` unnecessary for directories?

It seems like the logic in `if (Files.isSymbolicLink(configPath))` block should be applied inside the `while (!observablePaths.isEmpty())` loop to handle some edge cases, like perhaps `config/zilla.yaml => ../current/zilla.yaml`, `current/zilla.yaml => v1/zilla.yaml`, where there are different intermediate real paths of files to watch in different parent directories, not just one for the real path of `config/zilla.yaml`?

Btw, I noticed the `toRealPath` call is currently commented out in the PR.

bmaidics: I'm not sure if I got this. One configPath can only have 1 realPath. So in the given example, the realpath of config/zilla.yaml will be (if symlinks created in the root folder): /root/current/v1/zilla.yaml. 
By intermediate real path, what do you mean?

jfallows: In the example of the previous comment, the real path of config/zilla.yaml would resolve through both symlinks, to give v1/zilla.yaml, right?

In that case we would not be watching the intermediate symlink parent for changes, v1 directory, it seems - please double check.

bmaidics: In this example, this will be the content of the watchedPaths:

- /Users/barnabas/test/config/zilla.yaml, 
- /Users/barnabas/test/config/../current/zilla.yaml, 
- /Users/barnabas/test/current/v1/zilla.yaml
This means, that we'll watch both symlink's parent dir for changes: 

- for the 1st symlink:  /Users/barnabas/test/config
- for the 2nd symlink: /Users/barnabas/test/config/../current (this will actually be /Users/barnabas/test/current I think, but just to make sure, we cann call .realPath on it)

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigurationManager.java
```diff
@@ -292,28 +244,49 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
...
         }
+        return namespace;
+    }
 
-        // TODO: repeat to detect and apply changes
+    private void logError(
+        String message)
+    {
+        logger.accept("Configuration parsing error: " + message);
+    }
```
jfallows, line 264: Please move this private method to the bottom of the class, after the public methods.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,156 @@
...
+        WatchedConfig watchedConfig = new WatchedConfig(configURL, watchService);
+        watchedConfig.register();
+        watchedConfig.keys().forEach(k -> watchedConfigs.put(k, watchedConfig));
+        String configText = readConfigText(configURL);
+        watchedConfig.setConfigHash(computeHash(configText));
+        return changeListener.apply(configURL, configText);
+    }
+
+    @Override
+    public void doInitialConfiguration(
```
jfallows, line 116: Is this separate concept of `doInitialConfiguration` still needed?
Perhaps the caller can take this responsibility instead?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -0,0 +1,156 @@
...
+        {
+            URLConnection connection = configURL.openConnection();
+            try (InputStream input = connection.getInputStream())
+            {
+                configText = new String(input.readAllBytes(), UTF_8);
+            }
+        }
+        catch (IOException ex)
+        {
+            return "";
```
jfallows, line 146: Should this be the default config text instead of empty string?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/WatcherTask.java
```diff
@@ -0,0 +1,27 @@
...
+    {
+        this.changeListener = changeListener;
+    }
+
+    public abstract NamespaceConfig watch(
+        URL configURL);
+
+    public abstract void doInitialConfiguration(
+        URL configURL) throws Exception;
+
```
jfallows, line 26: See earlier comment on `FileWatcherTask` about potentially removing this method.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -223,7 +223,11 @@ public void start() throws Exception
             AgentRunner.startOnThread(runner, Thread::new);
         }
         watcherTaskRef = commonPool().submit(watcherTask);
-        watcherTask.doInitialConfiguration(rootConfigURL);
+        NamespaceConfig initialConfig = watcherTask.watch(rootConfigURL).get();
+        if (initialConfig == null)
+        {
+            throw new Exception("Parsing of the initial configuration failed.");
+        }
```
jfallows, line 230: This is good, however I was thinking the exception would be set on the future internally by the watcher task, for example via `CompletableFuture.failedFuture(ex)`, rather than returning null and converting it at the caller, then you just need to call `.get()` and the exception will be thrown for you if unsuccessful.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigurationManager.java
```diff
@@ -208,10 +181,11 @@ public Void call() throws Exception
 
             namespace.id = supplyId.applyAsInt(namespace.name);
 
-            namespace.readURL = this::readURL;
+            namespace.readURL = this.readURL;
 
             // TODO: consider qualified name "namespace::name"
-            namespace.resolveId = name -> name != null ? NamespacedId.id(namespace.id, supplyId.applyAsInt(name)) : 0L;
+            final NamespaceConfig finalNamespace = namespace;
+            namespace.resolveId = name -> name != null ? NamespacedId.id(finalNamespace.id, supplyId.applyAsInt(name)) : 0L;
```
jfallows, line 188: Is the rationale for adding `finalNamespace` here to make sure that `namespace` is not changing before lambda is executed?

If so, then please use `namespace0` instead of `finalNamespace` as we use that convention elsewhere in the codebase.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigurationManager.java
```diff
@@ -276,69 +250,49 @@ public Void call() throws Exception
...
-                        .GET()
-                        .uri(location.toURI())
-                        .build();
-
-                HttpResponse<String> response = client.send(
-                        request,
-                        HttpResponse.BodyHandlers.ofString());
+            future = CompletableFuture.allOf(future, dispatcher.attach(namespace));
+        }
+        future.join();
```
jfallows, line 274: I think this can be further simplified using streams, something like:
```
dispatchers.stream()
  .map(d -> d.attach(namespace))
  .reduce(CompletableFuture::allOf)
  .ifPresent(CompletableFuture::join);
```

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigurationManager.java
bmaidics: Yes, it's necessary, otherwise it's not building: "Variable used in lambda expression should be final or effectively final". Renamed it to namespace0.

jfallows (approved):

PR #142, opened 2023-01-05
Modify the layout of zilla files on the docker image

### LINKED ISSUE #140: Follow standard layout for zilla files in docker image
In support of Kubernetes config maps which requires mapping an entire directory for configuration, we need to move `zilla.json` to its own directory. At the same time, we should move the other parts to their standard locations as well.

`/etc/zilla/zilla.json`
`/opt/zilla/zilla`
`/opt/zilla/zilla/.zilla`
`/var/run/zilla`
`/var/run/zilla/.engine`

attilakreiner:
fixes #140 
required for https://github.com/aklivity/zilla-examples/pull/17

### DIFF
--- a/runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
+++ b/runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
@@ -15,6 +15,7 @@
  */
 package io.aklivity.zilla.runtime.command.start.internal.airline;
 
+import static io.aklivity.zilla.runtime.engine.EngineConfiguration.ENGINE_CONFIG_URL;
 import static io.aklivity.zilla.runtime.engine.EngineConfiguration.ENGINE_DIRECTORY;
 import static io.aklivity.zilla.runtime.engine.EngineConfiguration.ENGINE_VERBOSE;
 import static io.aklivity.zilla.runtime.engine.EngineConfiguration.ENGINE_WORKERS;
@@ -23,6 +24,7 @@ import static org.agrona.LangUtil.rethrowUnchecked;
 
 import java.io.IOException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -51,7 +53,7 @@ public final class ZillaStartCommand extends ZillaCommand
     @Option(name = { "-c", "--config" },
             description = "Configuration location",
             hidden = true)
-    public URI configURL = Paths.get("zilla.json").toUri();
+    public URI cliConfigURL;
 
     @Option(name = { "-v", "--verbose" },
             description = "Show verbose output")
@@ -76,6 +78,9 @@ public final class ZillaStartCommand extends ZillaCommand
     {
         Runtime runtime = getRuntime();
         Properties props = new Properties();
+        // default config file is zilla.json in the current directory if it's not specified neither in the properties
+        // file nor in the command line
+        URI configURL = cliConfigURL == null ? Paths.get("zilla.json").toUri() : cliConfigURL;
         props.setProperty(ENGINE_DIRECTORY.name(), ".zilla/engine");
 
         Path path = Paths.get(properties != null ? properties : OPTION_PROPERTIES_DEFAULT);
@@ -84,8 +89,13 @@ public final class ZillaStartCommand extends ZillaCommand
             try
             {
                 props.load(Files.newInputStream(path));
+                // if config file is specified in the command line, override the value specified in the properties file
+                if (cliConfigURL == null && props.getProperty(ENGINE_CONFIG_URL.name()) != null)
+                {
+                    configURL = new URI(props.getProperty(ENGINE_CONFIG_URL.name()));
+                }
             }
-            catch (IOException ex)
+            catch (IOException | URISyntaxException ex)
             {
                 System.out.println("Failed to load properties: " + path);
                 rethrowUnchecked(ex);
--- a/runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
+++ b/runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
@@ -85,17 +85,17 @@ public final class ZillaStartCommand extends ZillaCommand
             try
             {
                 props.load(Files.newInputStream(path));
-                // if config URL is specified in the command line, let's override the value specified in the properties file
-                if (configURL != null)
-                {
-                    props.setProperty(ENGINE_CONFIG_URL.name(), configURL.toString());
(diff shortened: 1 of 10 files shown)


jfallows (commented):
Use `.zilla/zilla.properties` to default config location and engine directory.

Specifying standard `/etc/zilla/zilla.json` config location in the image will let it work by default, and avoids adding unnecessary complexity to the examples.

Specifying the `/var/run/zilla` engine directory moves the writable runtime files to the appropriate location in the image, per the requirements in the issue description.

runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
```diff
@@ -51,7 +51,7 @@ public final class ZillaStartCommand extends ZillaCommand
     @Option(name = { "-c", "--config" },
             description = "Configuration location",
             hidden = true)
-    public URI configURL = Paths.get("zilla.json").toUri();
+    public String configFile;
```
jfallows, line 54: Why change from URL to file while adding the indirection through `zilla.properties`?
The configuration is located at a URL, for example we'll be supporting `http(s)` URLs soon too.

jfallows (changes requested):

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -221,14 +213,14 @@ public Statement apply(
...
+                    String configURL = requireNonNull(testClass.getClassLoader().getResource(resourceName)).toString();
+                    configure(ENGINE_CONFIG_URL.name(), configURL);
                 }
                 else
                 {
                     String resourceName = String.format("%s-%s", testClass.getSimpleName(), config.value());
-
-                    configURL = testClass.getResource(resourceName);
+                    String configURL = requireNonNull(testClass.getClassLoader().getResource(resourceName)).toString();
+                    configure(ENGINE_CONFIG_URL.name(), configURL);
```
jfallows, line 223: This one should be using `testClass.getResource(...)` instead of `testClass.getClassLoader().getResource(...)`.

Note: there are quite a few calls to `configURL.toString()` in the codebase, consider changing the type of the `ENGINE_CONFIG_URL` property to `URL` instead of `String` to eliminate them.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
```diff
@@ -70,6 +76,19 @@ public class EngineConfiguration extends Configuration
...
+            {
+                return URI.create(i).toURL();
+            }
+            catch (MalformedURLException ex)
+            {
+                LangUtil.rethrowUnchecked(ex);
+                return null;
+            }
+        };
+        ENGINE_CONFIG_URL = config.property(URL.class, "config.url", decodeFunction, "file:zilla.json");
```
jfallows, line 91: Add a private static method called `configURL` taking `Configuration` and `String` parameters, with implementation same as anonymous `decodeFunction` above, then pass `EngineConfiguration::configURL` instead of `decodeFunction`.

See `KafkaConfiguration.cacheDirectory(Configuration, String)` as an example.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -221,14 +213,14 @@ public Statement apply(
...
                 }
                 else
                 {
                     String resourceName = String.format("%s-%s", testClass.getSimpleName(), config.value());
-
-                    configURL = testClass.getResource(resourceName);
+                    String configURL = requireNonNull(testClass.getResource(resourceName)).toString();
+                    configure(ENGINE_CONFIG_URL.name(), configURL);
```
jfallows, line 223: There is an overload of `configure` that takes parameter types `PropertyDef<T>, T`.

```
URL configURL = testClass.getResource(resourceName));
configure(ENGINE_CONFIG_URL, configURL);
```

Please move `requireNotNull` checks to `configure(..., value)` overloads instead of making the call site more complex.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -221,14 +213,14 @@ public Statement apply(
                 if (configurationRoot != null)
                 {
                     String resourceName = String.format("%s/%s", configurationRoot, config.value());
-
-                    configURL = testClass.getClassLoader().getResource(resourceName);
+                    String configURL = requireNonNull(testClass.getClassLoader().getResource(resourceName)).toString();
+                    configure(ENGINE_CONFIG_URL.name(), configURL);
```
jfallows, line 217: There is an overload of `configure` that takes parameter types `PropertyDef<T>, T`.

```
URL configURL = testClass.getClassLoader().getResource(resourceName));
configure(ENGINE_CONFIG_URL, configURL);
```

Please move `requireNotNull` checks to `configure(..., value)` overloads instead of making the call site more complex.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
```diff
@@ -284,6 +294,19 @@ private static int defaultBufferPoolCapacity(
         return ENGINE_BUFFER_SLOT_CAPACITY.get(config) * 64;
     }
 
+    private static URL configURL(Configuration config, String url)
+    {
```
jfallows, line 298: Please fix parameter indentation.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
```diff
@@ -284,6 +294,19 @@ private static int defaultBufferPoolCapacity(
...
+        try
+        {
+            return URI.create(url).toURL();
+        }
+        catch (MalformedURLException ex)
+        {
+            LangUtil.rethrowUnchecked(ex);
+            return null;
+        }
```
jfallows, line 307: Please use single `return` for these `rethrowUnchecked` blocks.

```
String configURL = null;

try
{
    configURL = URI.create(url).toURL();
}
catch (MalformedURLException ex)
{
    LangUtil.rethrowUnchecked(ex);
}

return configURL;
```

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
attilakreiner: Done.

jfallows (approved):

PR #143, opened 2023-01-09
Adding support for Expression Resolver

### LINKED ISSUE #91: Support `{{ mustache }}` syntax in zilla.json
Support [{{ mustache }}](https://mustache.github.io/) syntax in zilla.json.

Consider making the root level objects pluggable, with initial support for `env` to provide access to environment variables from the executing shell.

This avoids the need to put any potentially sensitive information directly into the configuration.

For example:
```
...
    "options":
    {
        "sasl":
        {
            "mechanism": "plain",
            "username": "{{env.SASL_USERNAME}}",
            "password": "{{env.SASL_PASSWORD}}"
        }
    }
...
```

Note: initial support for `env` is already implemented, but disabled by default for now.

aDaemonThread:
fix #91


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
```diff
@@ -99,7 +99,7 @@ public class EngineConfiguration extends Configuration
         ENGINE_CREDITOR_CHILD_CLEANUP_LINGER_MILLIS = config.property("child.cleanup.linger", SECONDS.toMillis(5L));
         ENGINE_VERBOSE = config.property("verbose", false);
         ENGINE_WORKERS = config.property("workers", Runtime.getRuntime().availableProcessors());
-        ENGINE_CONFIG_SYNTAX_MUSTACHE = config.property("config.syntax.mustache", false);
+        ENGINE_CONFIG_SYNTAX_MUSTACHE = config.property("config.syntax.mustache", true);
```
jfallows, line 102: We should also rename this to something like `config.resolve.expressions` and change the constant name and method name accordingly.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -162,11 +162,14 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
 
         if (config.configSyntaxMustache())
         {
-            configText = Mustache.resolve(configText, System::getenv);
+            ExpressionResolver expressionResolver = ExpressionResolver.instantiate();
+            for (String context: expressionResolver.names())
+            {
+                configText = expressionResolver.resolve(context, configText);
+            }
```
jfallows, line 169: We can do this in a single pass instead of a loop, no need to pass each context separately.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/stream/resolver/EnvironmentResolverSpi.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Optional;
+import java.util.function.UnaryOperator;
+import java.util.regex.Pattern;
+
+import io.aklivity.zilla.runtime.engine.resolver.ExpressionResolverSpi;
+
+public class EnvironmentResolverSpi implements ExpressionResolverSpi
+{
+
+    private static final Pattern MUSTACHE_PATTERN = Pattern.compile("\\$\\{\\{\\s*env\\.([^\\s\\}]*)\\s*\\}\\}");
```
jfallows, line 12: Rename to `EXPRESSION_PATTERN` instead.

Also, this can be moved to the `ExpressionResolver` itself, match the pattern, extract the context, then use the appropriate `ExpressionResolverSpi` to resolve the value.

Each `ExpressionResolverSpi` implementation should be as simple as possible to avoid duplication when we extend to support more contexts.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/stream/resolver/EnvironmentResolverSpi.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    public String resolve(String config)
+    {
+        UnaryOperator<String> env = System::getenv;
+        return MUSTACHE_PATTERN
+                .matcher(config)
+                .replaceAll(r -> Optional.ofNullable(env.apply(r.group(1))).orElse(""));
+    }
```
jfallows, line 27: This implementation can simplify to `return System.getenv(var);` where the `String` parameter represents the variable name, such as `SASL_USERNAME` in `env.SASL_USERNAME`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/resolver/ExpressionResolver.java
```diff
@@ -0,0 +1,49 @@
...
+
+        return new ExpressionResolver(unmodifiableMap(resolverSpisByName));
+    }
+
+    public Iterable<String> names()
+    {
+        return resolverSpis.keySet();
+    }
+
+    public ExpressionResolver(Map<String, ExpressionResolverSpi> resolverSpis)
```
jfallows, line 44: This should be a `private` constructor so that only `instantiate` is used to create a new instance via service loader.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/resolver/ExpressionResolver.java
```diff
@@ -0,0 +1,49 @@
...
+        Map<String, ExpressionResolverSpi> resolverSpisByName = new HashMap<>();
+        resolvers.forEach(resolverSpi -> resolverSpisByName.put(resolverSpi.name(), resolverSpi));
+
+        return new ExpressionResolver(unmodifiableMap(resolverSpisByName));
+    }
+
+    public Iterable<String> names()
+    {
+        return resolverSpis.keySet();
+    }
```
jfallows, line 42: We probably don't need this one exposed externally as lookup is happening inside `resolve(...)`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/resolver/ExpressionResolver.java
```diff
@@ -0,0 +1,49 @@
...
+    private final Map<String, ExpressionResolverSpi> resolverSpis;
+
+    public static ExpressionResolver instantiate()
+    {
+        return instantiate(load(ExpressionResolverSpi.class));
+    }
+
+    public String resolve(
+            String name,
+            String config)
```
jfallows, line 22: Just a single argument, `config` needed here.

Applying the pattern to the `config` allows us to find each occurrence of `${{ context.var }}` and then lookup the resolver SPI for `context` to resolve `var`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/resolver/ExpressionResolver.java
```diff
@@ -0,0 +1,49 @@
...
+    {
+        requireNonNull(name, "name");
+
+        ExpressionResolverSpi resolverSpi = requireNonNull(resolverSpis.get(name), () -> "Unrecognized resolver name: " + name);
+
+        return resolverSpi.resolve(config);
+    }
+
+    private static ExpressionResolver instantiate(ServiceLoader<ExpressionResolverSpi> resolvers)
```
jfallows, line 31: Please check your indentation of method parameters (all methods), single indent, next line per parameter.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/resolver/ExpressionResolverSpi.java
```diff
@@ -0,0 +1,9 @@
+package io.aklivity.zilla.runtime.engine.resolver;
+
+public interface ExpressionResolverSpi
+{
+    String name();
+
+    String resolve(String config);
```
jfallows, line 7: Rename `config` to `var`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/resolver/ExpressionResolverSpi.java
```diff
@@ -0,0 +1,9 @@
+package io.aklivity.zilla.runtime.engine.resolver;
```
jfallows, line 1: Please refactor package name to `io.aklivity.zilla.runtime.engine.expression` instead.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/expression/ExpressionResolver.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public static ExpressionResolver instantiate()
+    {
+        return instantiate(load(ExpressionResolverSpi.class));
+    }
+
+    public String resolve(
+        String config)
+    {
+        return EXPRESSION_PATTERN.matcher(config)
```
jfallows, line 27: Instead of recreating the `Matcher` on every call to `resolve(config)`, suggest instead creating the matcher via `EXPRESSION_MATCHER.matcher("")` in constructor, then `matcher.reset(config).replaceAll(...)` in `resolve` method.

Also suggest changing parameter names from `config` to `template`, since config is just one possible usage of this expression resolver.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/expression/EnvironmentResolverSpi.java
```diff
@@ -0,0 +1,24 @@
...
+    {
+        return "env";
+    }
+
+    @Override
+    public String resolve(
+        String var)
+    {
+        UnaryOperator<String> env = System::getenv;
+        return Optional.ofNullable(env.apply(var)).orElse("");
```
jfallows, line 22: Avoid object allocation of `Optional` on each call to `resolve`.
Return `null` if not found, default to `""` is happening at caller `ExpressionResolver`.
Simplify to `return System.getenv(var);`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/ConfigureTask.java
```diff
@@ -160,16 +160,14 @@ else if ("http".equals(configURL.getProtocol()) || "https".equals(configURL.getP
...
 
-        if (config.configSyntaxMustache())
+        if (config.configResolveExpressions())
         {
-            ExpressionResolver expressionResolver = ExpressionResolver.instantiate();
-            for (String context: expressionResolver.names())
-            {
-                configText = expressionResolver.resolve(context, configText);
-            }
+            ExpressionResolver resolver = ExpressionResolver.instantiate();
```
jfallows, line 165: Let's move this instantiation to the constructor, so it is done once, even when detecting changed configuration (later).

Suggest calling it `expressions` so the code reads as `expressions.resolve(...)`.

jfallows (commented):
We need a replacement test for what used to be `MustacheTest` now `ExpressionResolverTest` to verify behavior.

See `BindingFactoryTest` as an example.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/expression/TestExpressionResolverSpi.java
```diff
@@ -0,0 +1,20 @@
...
+    public String name()
+    {
+        return "test";
+    }
+
+    @Override
+    public String resolve(
+        String var)
+    {
+        return "PASSWORD".equals(var) ? "ACTUALPASSWORD" : "";
```
jfallows, line 18: Even though it's just a test implementation, let's return null instead of "" as the defaulting will happen at the ExpressionResolver instead.

jfallows (changes requested):

jfallows (approved):

PR #150, opened 2023-01-20
Adding yaml support for zilla config

### LINKED ISSUE #144: Support YAML syntax for Zilla configuration
Today we use JSON syntax for Zilla configuration, and we can support YAML too (instead?) for better alignment with Kubernetes native YAML syntax.

aDaemonThread:
fix #144


runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
```diff
@@ -109,6 +109,22 @@ public void run()
...
+                        Paths.get(config.configURL().getPath()).getParent() + "/zilla.json" : "zilla.json";
+                props.setProperty(ENGINE_CONFIG_URL.name(), "file:" + jsonConfig);
+                config = new EngineConfiguration(props);
+                System.out.println("zilla.yaml file not found, loading zilla.json instead");
+            }
+            if (config.configURL().toString().endsWith(".json"))
+            {
+                System.out.println("warning: json syntax is deprecated, migrate to yaml");
+            }
+        }
```
jfallows, line 127: This block of code calls `config.configURL()` 6 times, seems like we should hoist it into a variable, no? 😄 

First, rename field named `configURL` of type `URI` to `configURI` instead.

```
EngineConfiguration config = new EngineConfiguration(props);
URL configURL = config.configURL();
if ("file".equals(configURL.getProtocol()))
{
    Path configPath = Paths.get(configURL.getPath());
    if (configPath.endsWith("zilla.yaml") && Files.notExists(configPath))
    {
        String configJSON = String.format("file:%s", configPath.resolveSibling("zilla.json"));
        props.setProperty(ENGINE_CONFIG_URL.name(), configJSON);
        config = new EngineConfiguration(props);
        System.out.println("zilla.yaml file not found, loading zilla.json instead");
    }
    if (configPath.getFilename().getName(0).endsWith(".json"))
    {
        System.out.println("warning: json syntax is deprecated, migrate to yaml");
    }
}
```

jfallows (changes requested):

runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
jfallows: Also, is reinstantiation of `config = new EngineConfiguration(props);` actually necessary, or is it sufficient to just update properties?

runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
```diff
@@ -109,18 +111,18 @@ public void run()
...
-                        Paths.get(config.configURL().getPath()).getParent() + "/zilla.json" : "zilla.json";
-                props.setProperty(ENGINE_CONFIG_URL.name(), "file:" + jsonConfig);
-                config = new EngineConfiguration(props);
+                String configJSON = String.format("file:%s", configPath.resolveSibling("zilla.json"));
+                props.setProperty(ENGINE_CONFIG_URL.name(), configJSON);
+                configPath = Paths.get(config.configURL().getPath());
                 System.out.println("zilla.yaml file not found, loading zilla.json instead");
             }
-            if (config.configURL().toString().endsWith(".json"))
+            if (configPath.getFileName().endsWith(".json"))
```
jfallows, line 125: Does this code actually cause the deprecation warning to be printed?
Javadoc for [java.nio.file.Path](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#endsWith-java.lang.String-) says that `endsWith(String)` is checking the path segment as a whole.

Hence we need the path segment name, to test with `String.endsWith(String)` instead.
This was in the previous snippet as `configPath.getFileName().getName(0).endsWith(".json")` but didn't make it into the PR.

Could you please double check?

jfallows (commented):
Please double check that deprecation warning is actually emitted when using `zilla.json` instead of `zilla.yaml`.

jfallows (approved):

jfallows (approved):

PR #156, opened 2023-01-27
Http dynamic configuration

bmaidics:


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
```diff
@@ -0,0 +1,212 @@
...
+{
+    private static final String INTIAL_ETAG = "INIT";
+    private final Map<URI, String> etags;
+    private final Map<URI, byte[]> configHashes;
+    private final Map<URI, CompletableFuture<Void>> futures;
+    private final Queue<URI> configWatcherQueue;
+
+    //If server does not support long-polling use this interval
+    private final int pollIntervalSeconds;
+    private volatile boolean closed = false;
```
jfallows, line 36: No need to explicitly initialize to default value.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BiFunction;
+
+import io.aklivity.zilla.runtime.engine.config.NamespaceConfig;
+
+public class HttpWatcherTask extends WatcherTask
+{
+    private static final String INTIAL_ETAG = "INIT";
```
jfallows, line 28: Leave blank line after statics before non-statics for readability.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
```diff
@@ -0,0 +1,212 @@
...
+        futures.put(configURI, future);
+    }
+
+    private Void handleException(
+        Throwable throwable,
+        URI configURI)
+    {
+        try
+        {
+            TimeUnit.SECONDS.sleep(pollIntervalSeconds);
```
jfallows, line 127: In general, I think we need to schedule the next async send rather than sleeping synchronously.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
```diff
@@ -0,0 +1,212 @@
...
+            rethrowUnchecked(ex);
+        }
+        catch (InterruptedException ex)
+        {
+            Thread.currentThread().interrupt();
+        }
+        return config;
+    }
+
+    private URI getUri(
```
jfallows, line 198: Rename to `toURI(configURL)`, matching the name of `configURL.toURI()`.

jfallows (changes requested):
Reviewed `HttpWatcherTask`.

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/streams/application/reconfigure.create.via.http/client.rpt
```diff
@@ -0,0 +1,45 @@
...
+connected
+
+write http:method "GET"
+write http:header "If-None-Match"
+write http:header "Prefer" "wait=86400"
+write close
+
+read http:status "200" "OK"
+read http:header "Etag" "AAAAAAA"
+read '{"name": "test","bindings":{"net0":{"type" : "test","kind": "proxy","exit": "app0"}}}'
```
jfallows, line 38: If you want to make this more readable, you can format it over multiple lines such as this.
```
read '{'
       '"name": "test".'
       '"bindings":`
       '{'
         '"net0":'
         '{'
           '"type": "test",'
           '"kind": "proxy",'
           '"exit": "app0"'
         '}'
       '}'
     '}'
```

Note: these should also be `yaml` syntax now. 😉

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -27,44 +28,45 @@
...
 
-        this.md5 = md5;
         this.watchService = watchService;
 
     }
 
+    @Override
+    public Future<Void> submit()
+    {
+        return commonPool().submit(this);
```
jfallows, line 67: Is this already fixed in the other PR, if not then we need to address it here to avoid using `commonPool()`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
```diff
@@ -0,0 +1,260 @@
...
+    private final Map<URI, String> etags;
+    private final Map<URI, byte[]> configHashes;
+    private final Map<URI, CompletableFuture<Void>> futures;
+    private final BlockingQueue<URI> configWatcherQueue;
+    private final ScheduledExecutorService executor;
+    //If server does not support long-polling use this interval
+    private final int pollIntervalSeconds;
+
+    public HttpWatcherTask(
+        BiFunction<URL, String, NamespaceConfig> changeListener,
```
jfallows, line 43: Feels like this `URL` type in `changeListener` is causing more exception handling than desired, like `MalformedURLException`.

Perhaps it would be better to use `URI` here instead to simplify the code and remove unnecessary try/catch blocks?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
```diff
@@ -0,0 +1,260 @@
...
+
+public class HttpWatcherTask extends WatcherTask
+{
+    private static final String INTIAL_ETAG = "INIT";
+    private static final URI CLOSE_REQUESTED = URI.create("http://localhost:12345");
+
+    private final Map<URI, String> etags;
+    private final Map<URI, byte[]> configHashes;
+    private final Map<URI, CompletableFuture<Void>> futures;
+    private final BlockingQueue<URI> configWatcherQueue;
```
jfallows, line 37: Please rename to `configQueue`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
```diff
@@ -0,0 +1,260 @@
...
+    private final ScheduledExecutorService executor;
+    //If server does not support long-polling use this interval
+    private final int pollIntervalSeconds;
```
jfallows, line 40: Please rename to `pollSeconds`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
```diff
@@ -0,0 +1,260 @@
...
+{
+    private static final String INTIAL_ETAG = "INIT";
+    private static final URI CLOSE_REQUESTED = URI.create("http://localhost:12345");
+
+    private final Map<URI, String> etags;
+    private final Map<URI, byte[]> configHashes;
+    private final Map<URI, CompletableFuture<Void>> futures;
+    private final BlockingQueue<URI> configWatcherQueue;
+    private final ScheduledExecutorService executor;
+    //If server does not support long-polling use this interval
```
jfallows, line 39: This comment can be removed.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
```diff
@@ -0,0 +1,260 @@
...
+        URI configURI,
+        String etag)
+    {
+        HttpClient client = HttpClient.newBuilder()
+            .version(HTTP_2)
+            .followRedirects(NORMAL)
+            .build();
+        HttpRequest request = HttpRequest.newBuilder()
+            .GET()
+            .headers("If-None-Match", etag, "Prefer", "wait=86400")
```
jfallows, line 159: Do we actually send `if-none-match` with value `""` or is that header omitted from the request when the etag is empty string?

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureHttpIT.java
```diff
@@ -0,0 +1,149 @@
...
+    @Rule
+    public final TestRule chain = outerRule(k3po).around(engine).around(timeout);
+
+    @Before
+    public void setupRegisterLatch() throws Exception
+    {
+        //Make sure that the initial configuration has finished
+        EngineTest.TestEngineExt.registerLatch.await();
+        //Register new CountDownLatch
+        EngineTest.TestEngineExt.registerLatch = new CountDownLatch(1);
```
jfallows, line 70: These 2 comments can be removed.
If you determine that more context is needed, then feel free to rename the method describing what is being setup.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureHttpIT.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Test
+    @Configure(name = ENGINE_CONFIG_POLL_INTERVAL_SECONDS, value = "1")
+    @Configuration("http://localhost:8080/")
+    @Specification({
+        "${app}/reconfigure.modify.via.http/server",
+        "${net}/reconfigure.modify.via.http/client"
+    })
+    public void shouldReconfigureWhenModifiedHTTP() throws Exception
```
jfallows, line 80: Please avoid UPPERCASE words in method names, so `Http` instead of `HTTP`.

Kindly verify all other method names in this class with the same feedback.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
bmaidics: If we change to URI here, I think we also should to change in a lot of places, e.g: in the guard to be consistent. If we only change it here, we're just delegating the exception handling to the readURL part, where we'd transform the URI to URL. But the readURL is dependent on the URL.openConnection method so I think we have to stick with URL there. If you think it's worth delegating the exception handling there, I'm up for it.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
bmaidics: As far as I understood, the etag that we get back from the zilla config server should never be empty. Even if the file is empty, we'll get the hash of the empty byte[]. So no, we should never send it with empty string. At the beginning we're sending with "INIT" constant.

jfallows: `etag` is not a required HTTP response field, so you cannot assume it is always present.

jfallows: At the beginning, you should not be sending any value for `if-match`, rather than `INIT`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/FileWatcherTask.java
```diff
@@ -27,44 +27,44 @@
...
-        this.md5 = md5;
         this.watchService = watchService;
 
     }
 
+    @Override
+    public Future<Void> submit()
+    {
+        return executor.submit(this);
+    }
```
jfallows, line 67: Missing blank line between methods.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/HttpWatcherTask.java
jfallows: Yeah, let's discuss the impact of changing it to `URI` instead of `URL`.

jfallows (approved):

PR #166, opened 2023-02-14
Provide http(s) configuration server for zilla.yaml

### LINKED ISSUE #139: Support dynamic `zilla` configuration via `https`
Retrieve initial `zilla.yaml` configuration via `https`, then detect a new configuration available at the same URL.

This can be done via infrequent `GET if-none-match` polling, or ideally using `prefer: wait=86400` header instead, allowing the response to unwind only when the `1 day` timeout expires or immediately when the content changes.

The reaction to the changed configuration will be the same as for #138 as only the `https` retrieval mechanism differs.

bmaidics:
fix #139


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/FileSystemWatcher.java
```diff
@@ -0,0 +1,274 @@
+package io.aklivity.zilla.runtime.engine;
```
bmaidics, line 1: Few things we need to decide with this:

- Where should it belong? Create a separate dir for this?
- It has some common parts with the ConfigWatcher, so after both is merged (or before this is merged), we can do a little cleanup. Maybe create a FileWatcherHelper?
- Currently, we're always creating the FileSystemWatcher, even if filesystem binding is never used. We can implement it in a way, that it's only created once if we need to watch for a file. Maybe when we're adding a FileSystemBinding, we can create it?

runtime/binding-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/filesystem/internal/stream/FileSystemServerFactory.java
```diff
@@ -107,8 +120,20 @@ public FileSystemServerFactory(
...
+        try
+        {
+            md5 = MessageDigest.getInstance("MD5");
+        }
+        catch (NoSuchAlgorithmException ex)
+        {
+            rethrowUnchecked(ex);
+        }
+
+        this.md5 = md5;
```
jfallows, line 136: Move this to `private` `static` `initMessageDigest` method.

runtime/binding-http-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/http/filesystem/internal/config/HttpFileSystemWithResolver.java
```diff
@@ -29,17 +29,21 @@
...
     private static final String8FW HEADER_METHOD_NAME = new String8FW(":method");
+    private static final String8FW HEADER_IF_NONE_MATCH_NAME = new String8FW("if-none-match");
+    private static final String8FW HEADER_PREFER_NAME = new String8FW("prefer");
     private static final String16FW HEADER_METHOD_VALUE_GET = new String16FW("GET");
     private static final String16FW HEADER_METHOD_VALUE_HEAD = new String16FW("HEAD");
 
+    private final String16FW etagRO = new String16FW();
     private final HttpFileSystemWithConfig with;
     private final Matcher paramsMatcher;
+    private final Matcher timoutMatcher;
```
jfallows, line 46: Typo `timeoutMatcher`.

Actually, let's rename this to `preferWaitMatcher`.

runtime/binding-http-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/http/filesystem/internal/config/HttpFileSystemWithResolver.java
```diff
@@ -68,20 +73,36 @@ public HttpFileSystemWithResult resolve(
...
+        }
+        HttpHeaderFW ifNotMatched = httpBeginEx.headers().matchFirst(h -> HEADER_IF_NONE_MATCH_NAME.equals(h.name()));
+        String16FW etag = new String16FW("");
+        if (ifNotMatched != null)
+        {
+            String16FW value = ifNotMatched.value();
+            etag = etagRO.wrap(value.buffer(), value.offset(), value.limit());
+        }
+        HttpHeaderFW prefer = httpBeginEx.headers().matchFirst(h -> HEADER_PREFER_NAME.equals(h.name()));
+        int timeout = 0;
```
jfallows, line 97: Rename to `wait`.

runtime/binding-http-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/http/filesystem/internal/config/HttpFileSystemWithResolver.java
```diff
@@ -68,20 +73,36 @@ public HttpFileSystemWithResult resolve(
...
+        if (prefer != null)
+        {
+            Matcher timeoutMatcher = timoutMatcher.reset(prefer.value().asString());
+            if (timeoutMatcher.find())
+            {
+                timeout = Integer.parseInt(timeoutMatcher.group(1));
             }
         }
-        return result;
+        return new HttpFileSystemWithResult(path, capabilities, etag, timeout);
```
jfallows, line 106: Pass last argument as `TimeUnit.SECONDS.toMillis(wait)` so we have the timeout in millis for use with the signaler.

runtime/binding-http-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/http/filesystem/internal/stream/HttpFileSystemProxyFactory.java
```diff
@@ -617,18 +620,41 @@ private void onFileSystemBegin(
...
+                {
+                    httpBeginExBuilder.headersItem(h -> h.name(HEADER_ETAG_NAME).value(tag));
+                }
+                httpBeginEx = httpBeginExBuilder.build();
+            }
 
             delegate.doHttpBegin(traceId, authorization, affinity, httpBeginEx);
         }
 
+        private String16FW getStatus(FileSystemBeginExFW fsBeginEx)
```
jfallows, line 643: Method parameter formatting.

runtime/binding-http-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/http/filesystem/internal/stream/HttpFileSystemProxyFactory.java
```diff
@@ -617,18 +620,41 @@ private void onFileSystemBegin(
...
+                final HttpBeginExFW.Builder httpBeginExBuilder =
+                    httpBeginExRW.wrap(extBuffer, 0, extBuffer.capacity())
                         .typeId(httpTypeId)
-                        .headersItem(h -> h.name(HEADER_STATUS_NAME).value(HEADER_STATUS_VALUE_200))
+                        .headersItem(h -> h.name(HEADER_STATUS_NAME).value(getStatus(fsBeginEx)))
                         .headersItem(h -> h.name(HEADER_CONTENT_TYPE_NAME).value(type))
-                        .headersItem(h -> h.name(HEADER_CONTENT_LENGTH_NAME).value(length))
-                        .build();
+                        .headersItem(h -> h.name(HEADER_CONTENT_LENGTH_NAME).value(length));
+                if (tag != null && tag.asString() != null)
```
jfallows, line 633: Can we use `tag.length() != -1` instead of `tag.asString() != null` here?

runtime/binding-http-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/http/filesystem/internal/stream/HttpFileSystemProxyFactory.java
```diff
@@ -617,18 +620,41 @@ private void onFileSystemBegin(
...
+                }
+                httpBeginEx = httpBeginExBuilder.build();
+            }
 
             delegate.doHttpBegin(traceId, authorization, affinity, httpBeginEx);
         }
 
+        private String16FW getStatus(FileSystemBeginExFW fsBeginEx)
+        {
+            if (fsBeginEx.tag().asString() == null)
```
jfallows, line 645: Try to avoid calling `asString()` if possible.

runtime/binding-http-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/http/filesystem/internal/stream/HttpFileSystemProxyFactory.java
```diff
@@ -934,6 +960,8 @@ private MessageConsumer newFileSystemStream(
                 .typeId(fsTypeId)
                 .capabilities(resolved.capabilities())
                 .path(resolved.path())
+                .tag(resolved.tag())
+                .timeout(resolved.timeout() * 1000)
```
jfallows, line 964: After applying feedback above, remove `* 1000` here.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/FileSystemWatcher.java
jfallows: Suggest keeping this `FileSystemWatcher` in `binding-filesystem` until we have a design that could support reuse by both engine config watcher and filesystem binding.

The main issue is that moving it up to `DispatchAgent` should really make `DispatchAgent` responsible for calling `.take()`, because if multiple bindings were to do that it would get very confusing as they encountered each others watch events. Then if `DispatchAgent` is calling `.take()` (or non-blocking equivalent) then it needs to dispatch those watch events via some clean abstractions.

We do have something similar for network I/O via `Poller` and `PollerKey`, which is included in the `doWork` loop of `DispatchAgent`.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -218,6 +218,8 @@ public Statement apply(
                 {
                     String resourceName = String.format("%s/%s", configurationRoot, config.value());
                     URL configURL = testClass.getClassLoader().getResource(resourceName);
+                    //   io/aklivity/zilla/specs/binding/filesystem/config/server.yaml
+                    //   io/aklivity/zilla/specs/binding/filesystem/config/server.json
```
jfallows, line 222: Remove comments.

jfallows (changes requested):

runtime/binding-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/filesystem/internal/stream/FileSystemStreamFactory.java
```diff
@@ -14,13 +14,15 @@
...
+import io.aklivity.zilla.runtime.binding.filesystem.internal.FileSystemWatcher;
 import io.aklivity.zilla.runtime.engine.binding.BindingHandler;
 import io.aklivity.zilla.runtime.engine.config.BindingConfig;
 
 public interface FileSystemStreamFactory extends BindingHandler
 {
     void attach(
-        BindingConfig binding);
+        BindingConfig binding,
+        FileSystemWatcher fileSystemWatcher);
```
jfallows, line 25: Feels like it shouldn't be necessary to add a separate parameter to `attach` since you have control over the construction of the implementation of this interface.

jfallows (changes requested):

runtime/binding-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/filesystem/internal/stream/FileSystemServerFactory.java
```diff
@@ -283,25 +330,113 @@ private void onAppBegin(
...
                 .capabilities(capabilities)
-                .path(path)
+                .path(relativePath)
                 .type(type)
                 .payloadSize(size)
-                .modifiedTime(modifiedTime)
+                .tag(newTag)
                 .build();
+            return replyBeginEx;
+        }
```
jfallows, line 376: We typically don't code building of flyweights like this, as return values for a method, because they must be only used on stack just before the call to write them out before they are available for reuse.

Recommend moving this code into `doAppBegin(tag)` instead, perhaps with an overload of `doAppBegin(tag, capabilities)`.

runtime/binding-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/filesystem/internal/FileSystemBindingContext.java
```diff
@@ -64,11 +79,21 @@ public void detach(
...
     @Override
     public String toString()
     {
         return String.format("%s %s", getClass().getSimpleName(), factories);
     }
+
+    private FileSystemWatcher supplyFileSystemWatcher()
+    {
+        return fileSystemWatcher;
+    }
```
jfallows, line 97: Suggest refactoring this and renaming the field as:
```
public FileSystemWatcher supplyWatcher()
{
    return watcher;
}
```

runtime/binding-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/filesystem/internal/FileSystemBindingContext.java
```diff
@@ -64,11 +79,21 @@ public void detach(
         {
             factory.detach(binding.id);
         }
+        if (--bindings == 0)
```
jfallows, line 82: Please put a blank line between these `if` statements to avoid a misread as `else-if`.

jfallows (changes requested):

runtime/binding-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/filesystem/internal/FileSystemBindingContext.java
```diff
@@ -91,9 +90,8 @@ public String toString()
         return String.format("%s %s", getClass().getSimpleName(), factories);
     }
 
-    private FileSystemWatcher supplyFileSystemWatcher()
+    public FileSystemWatcher supplyWatcher()
```
jfallows, line 93: Keep `private`.

jfallows (changes requested):

jfallows (approved):

PR #178, opened 2023-03-06
Add log + rollback on reconfigure errors

bmaidics:


runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
```diff
@@ -133,6 +133,7 @@ public void run()
         try (Engine engine = Engine.builder()
             .config(config)
             .errorHandler(this::onError)
+            .report(report)
```
bmaidics, line 136: I was thinking wether it's the best to propagate the consumer here. Maybe we just log everytime? Or since the reconfigure method will be called at initial configuration as well the try-catch here can be even removed. I think that'd be the best option.

jfallows: Can we use the existing error handler instead of introducing `report`?

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -272,8 +272,17 @@ private NamespaceConfig reconfigure(
...
-            configurationManager.register(newNamespace);
-            namespaces.put(configURL, newNamespace);
+            try
+            {
+                configurationManager.register(newNamespace);
+                namespaces.put(configURL, newNamespace);
+            }
+            catch (Exception ex)
+            {
+                context.onError(ex);
```
jfallows, line 282: What is an example of an exception that can be caught here?

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
bmaidics: Basically almost any exception that we can get at startup. What I faced: changed the configfile, so the keystore location is pointing to a non-existing file. In this case, we get FileNotFoundException.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/DispatchAgent.java
```diff
@@ -603,6 +603,7 @@ public int doWork() throws Exception
         {
             ex.addSuppressed(new Exception(String.format("[%s]\t[0x%016x] %s",
                                                          agentName, lastReadStreamId, streamsLayout)));
+            ex.printStackTrace();
```
jfallows, line 606: Please remove this.

jfallows (changes requested):

jfallows (approved):

PR #181, opened 2023-03-20
Enhance kafka binding to notify transition from historical to live messages

### LINKED ISSUE #113: Enhance idempotency support for HTTP-Kafka binding

### LINKED ISSUE #172: Enhance kafka binding to notify transition from historical to live messages
When reading messages from `kafka` binding cache, the latest message in the Kafka partition might be higher than the most recently received message in the cache, for example if the remaining messages are still in flight.

Therefore, we know that reading historical messages that have already occurred will be completed when we receive that initially observed latest message in the cache.

However, due to filtering, a `kafka` binding cache client might not receive any messages at all, and be waiting for a new message to arrive.

In this case, it is useful for the `kafka` binding cache to detect it is up-to-date with all historical messages, but awaiting newer live messages.

This can be communicated per partition by sending a `FLUSH` reply frame with a `KafkaFetchFlushEx` indicating the updated progress offset at the initial latest offset.

For `kafka` binding merged clients, we can collect the `FLUSH` reply frames for each partition, and when all partitions have been accounted for, then we can send a `FLUSH` frame with `KafkaMergedFlushEx` indicating the progress offsets for all partitions, matching the initial latest offsets for all partitions received with the `BEGIN` reply frame and associated `KafkaMergedBeginEx`.

Receiving this `FLUSH` frame before any `DATA` frames will notify the `kafka` binding merged client of the transition from historical to live messages, while continuing to receive future messages.

aDaemonThread:
fix #172 
fix #113


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -913,6 +915,7 @@ private void doClientFanoutReplyWindow(
             this.deltaType = deltaType;
             this.isolation = isolation;
             this.isolatedOffset = isolation == READ_COMMITTED ? () -> group.stableOffset : () -> group.latestOffset;
+            this.targetOffset = isolation == READ_COMMITTED ? () -> initialGroupStableOffset : () -> initialGroupLatestOffset;
```
jfallows, line 918: Please rename `targetOffset` to `initialGroupIsolatedOffset` as this is the same condition as `isolatedOffset` but specific to initial group value.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -1052,6 +1055,8 @@ private void doClientReplyBegin(
 
             this.initialGroupPartitionOffset = group.partitionOffset;
             this.initialGroupLatestOffset = group.latestOffset;
+            this.initialGroupStableOffset = group.stableOffset;
+            group.dataFrameReceived = false;
```
jfallows, line 1059: We are attempting to detect whether or not a specific client has delivered a `DATA` frame, not the group.

If we track at the group level, then it will not behave differently for each client, which may have different filters, requiring different behavior.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -1102,6 +1107,15 @@ private void doClientReplyDataIfNecessary(
             {
                 final KafkaCacheEntryFW nextEntry = cursor.next(entryRO);
 
+                if (nextEntry == null &&
+                        group.partitionOffset >= targetOffset.getAsLong() &&
+                        !group.dataFrameReceived &&
+                        cursor.offset == 0)
```
jfallows, line 1113: `cursor.offset` might never be `0` if Kafka has cleaned up old messages, then earliest offset in the partition would be `> 0`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -1440,6 +1455,35 @@ private void doClientReplyFlush(
...
+                                    .build()
+                                    .sizeof()));
+
+            replySeq += reserved;
+
+            assert replyAck <= replySeq;
+
+            cursor.advance(partitionOffset + 1);
+        }
+
```
jfallows, line 1486: This should be named `doClientReplyFlush` for consistency with other `doClientReply*` methods.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -1164,6 +1178,7 @@ private void doClientReplyData(
         {
             assert nextEntry != null;
 
+            group.dataFrameReceived = true;
```
jfallows, line 1181: Consider the case with 2 client streams, each one would need to receive a `FLUSH` if there are no `DATA` frames sent before catching up to latest / stable.

Therefore, we cannot track this at the group level as the first client would get the right behavior, but all future clients would not get the `FLUSH` as they would count the first client's `DATA` frame as their own.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -1071,6 +1071,7 @@ else if (length != 0)
                         client.replyId, client.topic, client.partitionId, client.decodableRecordSetBytes);
                 }
 
+
```
jfallows, line 1074: Please remove whitespace change.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -2066,6 +2070,10 @@ private final class KafkaFetchClient extends KafkaSaslClient
             private long nextOffset;
             private long stableOffset;
             private long latestOffset;
+            private long initialLatestOffset;
+            private long initialStableOffset;
+            private boolean dataFrameSent = false;
```
jfallows, line 2075: These should be tracked on the `KafkaFetchStream`, not the `KafkaFetchClient`.
The `KafkaFetchStream` would then determine if anything needs to be done when the `FetchResponse` is decoded.

Also, rather than a boolean flag consider tracking `flushOrDataFramesSent` and incrementing in both `doApplicationData` and `doApplicationFlush`, then adding `doApplicationFlushIfNecessary` that checks if `flushOrDataFramesSent` is `0`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -2066,6 +2070,10 @@ private final class KafkaFetchClient extends KafkaSaslClient
...
+            private int fetchRequestCount = 0;
```
jfallows, line 2076: Not sure why we need to track this one.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -3064,6 +3073,12 @@ private void onDecodeFetchResponse(
 
                 if (topicPartitions.get(partitionId) == leaderId)
                 {
+                    fetchRequestCount++;
+                    if (fetchRequestCount == 2 && !dataFrameSent)
+                    {
+                        client.triggerFlushIfRequired(traceId);
+                    }
+                    dataFrameSent = false;
```
jfallows, line 3081: Let `client.triggerFlushIfRequired` implement the logic to determine what to do, and keep this part simple.

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/CacheMergedIT.java
```diff
@@ -185,6 +185,17 @@ public void shouldFetchMergedMessagesWithNoFilter() throws Exception
...
+    @Test
+    @Configuration("cache.options.merged.yaml")
+    @Specification({
+        "${app}/merged.fetch.filter.sync/client",
+        "${app}/unmerged.fetch.filter.sync/server"})
+    public void shouldFetchMergedMessagesLatest() throws Exception
+    {
+        k3po.finish();
+    }
+
```
jfallows, line 198: Is this test non-deterministic on GitHub Actions?

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/fetch/cache.filter.sync/client.rpt
jfallows, line 1: We shouldn't need any cache-specific scripts. We don't have any for other cache scenarios, we reuse generic scripts for cache and non-cache scenarios.

specs/binding-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/kafka/streams/application/FetchIT.java
```diff
@@ -304,6 +304,24 @@ public void shouldReceiveMessagesWithNoFilter() throws Exception
...
+    }
+
+    @Test
+    @Specification({
+        "${app}/cache.filter.sync/client",
+        "${app}/cache.filter.sync/server"})
+    public void shouldReceiveCacheSync() throws Exception
+    {
+        k3po.finish();
+    }
```
jfallows, line 323: We shouldn't need any cache-specific scripts.

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/CacheFetchIT.java
```diff
@@ -453,6 +453,18 @@ public void shouldReceiveMessagesWithNoFilter() throws Exception
...
+    @Specification({
+        "${app}/cache.filter.sync/client",
+        "${app}/cache.filter.sync/server"})
+    @ScriptProperty("serverAddress \"zilla://streams/app1\"")
+    public void shouldFetchLatestMessage() throws Exception
+    {
+        partition.append(0L);
+        k3po.finish();
+    }
+
```
jfallows, line 467: We shouldn't need any cache-specific scripts, let's try to use a generic `filter.sync` script instead.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -1949,6 +1948,7 @@ private void doApplicationData(
             OctetsFW payload,
             Flyweight extension)
         {
+            flushOrDataFramesSent++;
```
jfallows, line 1951: Please move this to the end of the method, so it represents what has already happened.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -1963,6 +1963,7 @@ private void doApplicationFlush(
             int reserved,
             Flyweight extension)
         {
+            flushOrDataFramesSent++;
```
jfallows, line 1966: Please move this to the end of the method, so it represents what has already happened.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -1971,6 +1972,35 @@ private void doApplicationFlush(
...
+                        .fetch(f -> f
+                                .partition(p -> p
+                                        .partitionId(client.partitionId)
+                                        .partitionOffset(client.decodeRecordBatchLastOffset)
+                                        .stableOffset(client.initialStableOffset)
+                                        .latestOffset(client.initialLatestOffset)))
+                        .build();
+
+                doFlush(application, originId, routedId, replyId, replySeq, replyAck, replyMax,
+                        traceId, authorization, reserved, kafkaFlushEx);
```
jfallows, line 1996: Shouldn't `flushOrDataFramesSent` also be incremented here?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -3098,13 +3124,8 @@ private void onDecodeFetchResponse(
...
                 if (topicPartitions.get(partitionId) == leaderId)
                 {
-                    fetchRequestCount++;
-                    if (fetchRequestCount == 2 && !dataFrameSent)
-                    {
-                        client.triggerFlushIfRequired(traceId);
-                    }
-                    dataFrameSent = false;
-                    signaler.signalNow(routeId, initialId, SIGNAL_NEXT_REQUEST, 0);
+                    doApplicationFlushIfNecessary(traceId, authorization, 0);
```
jfallows, line 3127: If this is the only call site, and last argument is always `0`, then let's remove the last parameter from the method.

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/CacheMergedIT.java
```diff
@@ -185,13 +185,12 @@ public void shouldFetchMergedMessagesWithNoFilter() throws Exception
...
     }
 
-    @Ignore("GitHub Actions")
     @Test
     @Configuration("cache.options.merged.yaml")
     @Specification({
         "${app}/merged.fetch.filter.sync/client",
         "${app}/unmerged.fetch.filter.sync/server"})
-    public void shouldFetchMergedMessagesLatest() throws Exception
+    public void shouldFetchMergedLatestMessages() throws Exception
```
jfallows, line 193: Usually we name the test methods according to the spec scenarios.
In this case, i think that would be `shouldFetchMergedFilterSync`, works?

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/ClientFetchIT.java
```diff
@@ -514,7 +514,7 @@ public void shouldReceiveMessagesWithHeadersSkipManyFilter() throws Exception
     @Specification({
         "${app}/filter.sync/client",
         "${net}/filter.sync/server"})
-    public void shouldFetchMergedMessageLatest() throws Exception
+    public void shouldFetchLatestMessages() throws Exception
```
jfallows, line 517: Usually we name the test methods according to the spec scenarios.
In this case, i think that would be `shouldFetchFilterSync`, works?

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/ClientMergedIT.java
```diff
@@ -368,7 +368,7 @@ public void shouldFetchMergedMessagesWithHeadersSkipManyFilter() throws Exceptio
     @Specification({
         "${app}/merged.fetch.filter.sync/client",
         "${net}/unmerged.fetch.filter.sync/server"})
-    public void shouldFetchMergedMessageLatest() throws Exception
+    public void shouldFetchMergedLatestMessages() throws Exception
```
jfallows, line 371: Usually we name the test methods according to the spec scenarios.
In this case, i think that would be `shouldFetchMergedFilterSync`, works?

jfallows (changes requested):

jfallows (approved):

PR #200, opened 2023-04-05
Mqtt kafka specs

bmaidics:


specs/binding-kafka.spec/src/main/resources/META-INF/zilla/kafka.idl
```diff
@@ -223,6 +230,7 @@ scope kafka
             KafkaOffset[] progress;     // INIT only
             KafkaKey key;               // INIT only
             KafkaDelta delta;           // INIT + FIN
+            uint64 filters = 0;
```
jfallows, line 233: I think this is `INIT only` since it is saying what filter condition(s) matched to send the message, so it would have to be known at the start of the message.

Therefore, I think we can move this up with the other fixed width fields, after timestamp.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/publish.one.message/client.rpt
```diff
@@ -0,0 +1,50 @@
...
+                               .partition(-1, -1)
+                               .key("sensor/one")
+                               .header("zilla:topic", "sensor")
+                               .header("zilla:topic", "one")
+                               .header("zilla:local", "client")
+                               .headerInt("zilla:timeout-ms", 15000)
+                               .header("zilla:content-type", "text/plain")
+                               .header("zilla:format", "TEXT")
+                               .header("zilla:reply-to", "sensor/one")
+                               .header("zilla:correlation-id", "info")
```
jfallows, line 46: We can remove `zilla:reply-to` and `zilla:correlation-id` headers from scenarios not explicitly testing request-response pattern.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/mqtt/subscribe.no.local/client.rpt
```diff
@@ -0,0 +1,29 @@
...
+
+connect "zilla://streams/mqtt0"
+         option zilla:window 8192
+         option zilla:transmission "duplex"
+
+write zilla:begin.ext ${mqtt:beginEx()
+                              .typeId(zilla:id("mqtt"))
+                              .subscribe()
+                                .clientId("client")
+                                .filter("sensor/one", 1, "NO_LOCAL", "SEND_RETAINED")
```
jfallows, line 25: Please remove `SEND_RETAINED` flag from scenarios not explicitly testing `send_retained` behavior.

jfallows (changes requested):

PR #212, opened 2023-04-19
Support eager evaluation of all Kafka filters

### LINKED ISSUE #209: Support `eager` evaluation of all `kafka` filters and indicate which filters matched
Support `eager` evaluation of all `kafka` filters on `KafkaMergedBeginEx` or `KafkaFetchBeginEx`.

We currently evaluate each `KafkaFilter` lazily, stopping on first success, so we do not know which filter succeeded and which filters failed or were never executed after first successful filter.

In some cases, we need to understand the complete set of matching filters that could have contributed to deliver a message.

Therefore, we can add the concept of `evaluation` to the `filters`, indicating `LAZY` (backwards compatible) or `EAGER` behavior.

Regardless of execution strategy, we can include a filters mask in `KafkaMergedDataEx` or `KafkaFetchDataEx`, indicating which filters were evaluated to `true` when delivering the matching message.

Note: using `int64` to represent the `filters` mask limits the maximum number of filters for each stream to `64`, which is acceptable.

```
    enum KafkaEvaluation (uint8)
    {
        LAZY(0),
        EAGER(1)
    }
```
```
        struct KafkaMergedBeginEx
        {
            KafkaCapabilities capabilities = PRODUCE_AND_FETCH;
            string16 topic;
            KafkaOffset[] partitions;
            KafkaFilter[] filters; // ORed
            KafkaEvaluation evaluation = LAZY;
            KafkaIsolation isolation = READ_COMMITTED;
            KafkaDeltaType deltaType = NONE;
            KafkaAckMode ackMode = IN_SYNC_REPLICAS;
        }
```
```
        struct KafkaFetchBeginEx
        {
            string16 topic;
            KafkaOffset partition;
            KafkaFilter[] filters; // ORed
            KafkaEvaluation evaluation = LAZY;
            KafkaIsolation isolation = READ_UNCOMMITTED;
            KafkaDeltaType deltaType = NONE;
        }
```
```
        struct KafkaMergedDataEx
        {
            int32 deferred = 0;         // INIT only (TODO: move to DATA frame)
            int64 timestamp = 0;        // INIT only
            int64 filters;              // INIT only
            KafkaOffset partition;      // INIT only
            KafkaHeader[] headers;      // INIT + FIN (produce), INIT only (fetch)
        }
```
```
        struct KafkaFetchDataEx
        {
            int32 deferred = 0;         // INIT only (TODO: move to DATA frame)
            int64 timestamp = 0;        // INIT only
            int32 headersSizeMax = 4;   // INIT only
            int64 producerId = -1;      // INIT only
            int64 filters;              // INIT only
            KafkaOffset partition;      // INIT only
            KafkaHeader[] headers;      // FIN only
        }
```

aDaemonThread:
Fix #209


specs/binding-kafka.spec/src/main/java/io/aklivity/zilla/specs/binding/kafka/internal/KafkaFunctions.java
```diff
@@ -2446,6 +2499,12 @@ private boolean matchHeaders(
             {
                 return headersRW == null || headersRW.build().equals(mergedDataEx.headers());
             }
+
+            private boolean matchFilters(
+                final KafkaMergedDataExFW mergedDataEx)
+            {
+                return filters == null || filters == mergedDataEx.filters();
+            }
```
jfallows, line 2507: Need to call `matchFilters` from `match` above.

specs/binding-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/kafka/internal/KafkaFunctionsTest.java
```diff
@@ -903,6 +906,56 @@ public void shouldGenerateMergedFlushExtension()
...
+                .capabilities("PRODUCE_AND_FETCH")
+                .progress(0, 1L, 1L, 1L)
+                .filter()
+                .key("match")
+                .build()
+                .filter()
+                .header("name", "value")
+                .build()
+                .build()
+                .build();
```
jfallows, line 924: Use further indentation to show nested `.build()` objects.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -1001,7 +1090,14 @@ public KafkaFilterCondition asCondition(
...
             filters.forEach(f -> asConditions.add(asCondition(f)));
-            condition = asConditions.size() == 1 ? asConditions.get(0) : new KafkaFilterCondition.Or(asConditions);
+            if (KafkaEvaluation.EAGER == evaluation)
+            {
+                condition = asConditions.size() == 1 ? asConditions.get(0) : new KafkaFilterCondition.EagerOr(asConditions);
+            }
+            else
+            {
+                condition = asConditions.size() == 1 ? asConditions.get(0) : new KafkaFilterCondition.Or(asConditions);
+            }
```
jfallows, line 1100: Suggest changing this to `switch` on `evaluation`.

Note: for the optimization of single condition to still work, we would need to control the return value of `.test(...)` for the nested conditions, so they return the right bitmask when successful.

This approach may be helpful in general when implementing `EagerOr` and `LazyOr` as you can just bitwise-or the return values of `.test(...)` together to get the combined bitmask.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -962,6 +963,93 @@ public String toString()
             }
         }
 
+        private static final class EagerOr extends KafkaFilterCondition
```
jfallows, line 966: As discussed, suggest creating a superclass `Or` and subsclasses `LazyOr` and `EagerOr` for improved readability.

Only the `.test(...)` method implementation will differ between `LazyOr` and `EagerOr`, so everything else can be implemented in the `Or` superclass.

runtime/binding-kafka/src/main/zilla/internal.idl
```diff
@@ -36,6 +36,7 @@ scope internal
             kafka::KafkaHeader[] trailers;
             uint32 paddingLen;
             octets[paddingLen] padding;
+            int64 filters;
```
jfallows, line 39: This change needs to be reverted as `KafkaCacheEntry` is shared across multiple readers, whereas `filters` mask applies to each individual cache client stream.

specs/binding-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/kafka/internal/KafkaFunctionsTest.java
```diff
@@ -440,6 +441,7 @@ public void shouldGenerateMergedDataExtension()
                                      .typeId(0x01)
                                      .merged()
                                          .timestamp(12345678L)
+                                         .filters(0L)
```
jfallows, line 444: The example should probably show a non-zero filter mask, because filter mask zero means the message didn't match and therefore wouldn't be delivered.

specs/binding-kafka.spec/src/main/resources/META-INF/zilla/kafka.idl
```diff
@@ -219,6 +226,7 @@ scope kafka
         {
             int32 deferred = 0;         // INIT only (TODO: move to DATA frame)
             int64 timestamp = 0;        // INIT only
+            int64 filters = 0;          // INIT only
```
jfallows, line 229: Suggest we default to `-1` instead, as filters mask `0` means no match.

specs/binding-kafka.spec/src/main/resources/META-INF/zilla/kafka.idl
```diff
@@ -269,6 +278,7 @@ scope kafka
             int64 timestamp = 0;        // INIT only
             int32 headersSizeMax = 4;   // INIT only
             int64 producerId = -1;      // INIT only
+            int64 filters = 0;          // INIT only
```
jfallows, line 281: Suggest we default to `-1` instead, as filters mask `0` means no match.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -408,10 +413,10 @@ public int next(
             }
 
             @Override
-            public boolean test(
+            public long test(
                 KafkaCacheEntryFW cacheEntry)
             {
-                return cacheEntry != null;
+                return cacheEntry != null ? 1L : 0L;
```
jfallows, line 419: Does this need to be `-1` to generate the default, non-zero value for `None` filter?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -876,11 +881,11 @@ public String toString()
             }
         }
 
-        private static final class Or extends KafkaFilterCondition
+        private static class Or extends KafkaFilterCondition
```
jfallows, line 884: Let's mark this `abstract` too.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -943,11 +948,35 @@ public int next(
             }
 
             @Override
-            public boolean test(
-                KafkaCacheEntryFW cacheEntry)
+            public long test(KafkaCacheEntryFW cacheEntry)
+            {
+                return 0L;
+            }
```
jfallows, line 954: Remove method, now that class is abstract.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -943,11 +948,35 @@ public int next(
...
+            private LazyOr(
+                    List<KafkaFilterCondition> conditions)
+            {
+                super(conditions);
+                this.conditions = conditions;
+            }
+
+            @Override
+            public long test(
+                    KafkaCacheEntryFW cacheEntry)
```
jfallows, line 976: Remove extra indentation.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -962,6 +991,41 @@ public String toString()
...
+            {
+                long accept = 0L;
+                for (int i = 0; i < conditions.size(); i++)
+                {
+                    final KafkaFilterCondition condition = conditions.get(i);
+                    final long result = condition.test(cacheEntry);
+                    if (result != 0L)
+                    {
+                        accept |= i + 1;
+                    }
```
jfallows, line 1017: Let's rework this code so that it can simplify to `accept |= condition.test(cacheEntry);`.

That requires the condition to return the relevant bitmask for the filter index it is on, not just `0` or `1`.

Therefore, when each filter condition is created for non-Or conditions, it needs to know the filter index, so it can compute `1 << index` as bitmask to return when `.test(...)` is successful, instead of `1`.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -1127,35 +1145,37 @@ private KafkaFilterCondition asNotCondition(
...
     private KafkaFilterCondition asHeadersCondition(
+        long mask,
         KafkaHeadersFW headers)
     {
-        return new KafkaFilterCondition.HeaderSequence(checksum, valueMatchRO, headerRO, headers);
+        return new KafkaFilterCondition.HeaderSequence(mask, checksum, valueMatchRO, headerRO, headers);
     }
 
     private static KafkaFilterCondition.Key initNullKeyInfo(
+        long mask,
```
jfallows, line 1171: Before adding the `mask`, we used this as a shared instance for all `null` key conditions, but now we need the `mask` to be correct per client, so the shared part should shift to the `nullKeyRO` and we now need to create the mask-specific `Key` instance.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -197,11 +200,13 @@ public KafkaCacheEntryFW next(
 
                 // TODO: when doing reset, condition.reset(condition)
                 // TODO: remove nextOffset < offset from if condition
-                if (nextOffset < offset || !condition.test(nextEntry))
+                if (nextOffset < offset || condition.test(nextEntry) == 0L)
                 {
                     nextEntry = null;
                 }
 
+                filters =  nextEntry != null ? condition.test(nextEntry) : 0L;
```
jfallows, line 208: Please call `condition.test(nextEntry)` just once.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -876,11 +893,11 @@ public String toString()
             }
         }
 
-        private static final class Or extends KafkaFilterCondition
+        private abstract static class Or extends KafkaFilterCondition
```
jfallows, line 896: I thought the standard modifier order would be `static abstract`, no?

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
aDaemonThread: We generally follow `abstract static class` as a standard and also same is followed in zilla.

jfallows: Okay thanks, I must just have it reversed in my head. 👍

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -71,7 +73,7 @@ public KafkaCacheCursorFactory(
     {
         this.writeBuffer = writeBuffer;
         this.checksum = new CRC32C();
-        this.nullKeyInfo = initNullKeyInfo(checksum);
+        this.nullKeyRO = initNullKeyRO();
```
jfallows, line 76: We were previously using this approach to avoid copying the `null` key value for each client filter.
I don't see where we are using `nullKeyRO` after the changes.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -71,7 +72,6 @@ public KafkaCacheCursorFactory(
     {
         this.writeBuffer = writeBuffer;
         this.checksum = new CRC32C();
-        this.nullKeyInfo = initNullKeyInfo(checksum);
```
jfallows, line 74: I think you might have misinterpreted the feedback. 🙂 

I was saying that we want to preserve the original optimization of not copying `null` keys all the time, just the approach is slightly different now because we cannot pre-create the entire filter condition.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -1121,7 +1123,11 @@ private KafkaFilterCondition asKeyCondition(
         long mask,
         KafkaKeyFW key)
     {
-        return new KafkaFilterCondition.Key(mask, checksum, key);
+        final OctetsFW value = key.value();
+
+        return value == null ?
+            new KafkaFilterCondition.Key(mask, checksum, nullKeyRO) : new KafkaFilterCondition.Key(mask, checksum, key);
+
```
jfallows, line 1130: This is closer, but we still make the copy of `nullKeyRO` each time we construct a new `KafkaFilterCondition.Key`.

Follow the constructor hierarchy up through `super` to see the copy in `Equals`.

jfallows (changes requested):

jfallows (approved):

PR #224, opened 2023-04-27
Stream direction

attilakreiner:
## Description

Fixes # (issue)


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/NamespaceRegistry.java
```diff
@@ -153,13 +154,28 @@ private void setMetricHandlers(
...
                 {
-                    originMetricHandler = originMetricHandler.andThen(handler);
+                    if (direction == SENT || direction == BOTH)
+                    {
+                        sentOriginMetricHandler = sentOriginMetricHandler.andThen(handler);
+                    }
+                    if (direction == RECEIVED || direction == BOTH)
+                    {
+                        receivedOriginMetricHandler = receivedOriginMetricHandler.andThen(handler);
+                    }
```
jfallows, line 178: I feel like there should be a simpler way to handle this than nested `if-else`, but it works. 🙂

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/metrics/Metric.java
```diff
@@ -32,12 +34,24 @@
...
+        BOTH
+    }
+
     String name();
 
     Kind kind();
 
     Unit unit();
 
+    default StreamDirection streamDirection()
```
jfallows, line 50: Let's rename this method to `direction` too.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/metrics/MetricContext.java
```diff
@@ -25,6 +27,11 @@
 
     Metric.Kind kind();
 
+    default Metric.StreamDirection streamDirection()
+    {
+        return BOTH;
+    }
+
```
jfallows, line 34: Do we need this on both `Metric` and `MetricContext`?

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/NamespaceRegistry.java
attilakreiner: Yeah, unfortunately it looks like this, but I don't exactly see a better way. We could introduce a map, but we have two keys so it should be something like `Map<Direction, Map<MetricHandlerKind, MessageConsumer>>` so I am not sure it would be actually simpler. 🤔

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/metrics/Metric.java
attilakreiner: done

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/metrics/MetricContext.java
attilakreiner: No, we actually need it only in `MetricContext`. It made sense to add it to `Metric` semantically speaking as it feels like a core property of a metric (like `kind` or `unit`), but I can remove it from there and it would still work. Pls let me know if I should go ahead and remove it.

jfallows: Yes, please remove it from `Metric`.

Actually, this feels more like a property of the metric handler, so `MetricContext` is a better place for it.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/metrics/MetricContext.java
attilakreiner: done

incubator/metrics-http/src/main/java/io/aklivity/zilla/runtime/metrics/http/internal/HttpSizeHandler.java
```diff
@@ -0,0 +1,133 @@
...
+import io.aklivity.zilla.runtime.engine.binding.function.MessageConsumer;
+import io.aklivity.zilla.runtime.metrics.http.internal.types.HttpHeaderFW;
+import io.aklivity.zilla.runtime.metrics.http.internal.types.stream.AbortFW;
+import io.aklivity.zilla.runtime.metrics.http.internal.types.stream.BeginFW;
+import io.aklivity.zilla.runtime.metrics.http.internal.types.stream.DataFW;
+import io.aklivity.zilla.runtime.metrics.http.internal.types.stream.EndFW;
+import io.aklivity.zilla.runtime.metrics.http.internal.types.stream.FrameFW;
+import io.aklivity.zilla.runtime.metrics.http.internal.types.stream.ResetFW;
+
+public class HttpSizeHandler implements MessageConsumer
```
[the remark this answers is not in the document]
attilakreiner: done

PR #252, opened 2023-05-24
Resolve code duplication around accessing metrics

attilakreiner:
## Description

WIP


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/metrics/LayoutManager.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+import io.aklivity.zilla.runtime.engine.internal.layouts.metrics.CountersLayout;
+import io.aklivity.zilla.runtime.engine.internal.layouts.metrics.GaugesLayout;
+import io.aklivity.zilla.runtime.engine.internal.layouts.metrics.HistogramsLayout;
+import io.aklivity.zilla.runtime.engine.internal.layouts.metrics.MetricsLayout;
+
+public class LayoutManager
```
jfallows, line 33: Suggest we rename this to MetricsLayoutManager and move to `layouts` package.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/metrics/MetricsProcessor.java
jfallows, line 1: Suggest we keep the API limited to accessing specific (namespaced) binding metric, and leave all the decisions about how to output at the calling side, as this will different significantly for `zilla metrics`, various `exporters`, etc.

There is also a question of discovery, that we can add here as well, to find the list of all metrics recorded for each binding.

Then callers such as `zillla metrics`, `exporters`, etc can discover the full list of recorded metrics, request the accessors for each metric once, and implement their own logic for collecting and exporting the current metric values in the appropriate format.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/metrics/MetricsProcessor.java
attilakreiner: For now, I removed the `print` method, so the only public method is `findRecord` that was just added now. This resulted in some cleanup. This is not the final version, as you pointed out this class will definitely need some more changes when we reconcile all the use cases. However, I think it's good enough for now to get you unblocked. The next iteration where all the use cases (engine, command, exporter) will be using the same class should come in a next PR.

attilakreiner: I added the `getRecords` method to get the list of all metric records. I also made the string formatter functions injected in the constructor. This is a necessary step to generalize the `MetricsProcessor` so it can eventually be used by the command & exporter. This way I was able to add back all the unit tests I temporarily removed in a slightly modified form (the tests are checking the metric record values directly instead of the printed output).

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -387,5 +391,19 @@ public void onError(
...
+            String binding,
+            String metric)
+        {
+            if (metrics == null)
+            {
+                return () -> 0L;
+            }
+            MetricRecord record = metrics.findRecord(namespace, binding, metric);
+            return () -> record == null ? 0L : record.value();
+        }
```
jfallows, line 407: We need to do more of a review on `MetricRecord` etc before we consider pulling it up into `engine`. Please move it back into `command-metrics` for now.

In the meantime, let's simplify.
 - lookup `bindingId` and `metricId` based on provided `namespace`, `binding` and `metric` names from label manager
 - keep the dispatchers parameter to the constructor and retain in a field
 - walk the dispatchers to find a `LongSupplier` for each `counter`
 - in DispatcherAgent.supplyCounter(bindingId, metricId), return countersLayout.suppyReader(bindingId, metricId)
 - return an aggregating `LongSupplier` lambda that sums all the `LongSupplier`s to get the total counter value

jfallows (changes requested):

PR #261, opened 2023-05-26
Implement counter value reader in EngineExtContext

attilakreiner:
## Description

This change Implements a counter value reader method in EngineExtContext.


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -349,15 +351,18 @@ private final class ContextImpl implements EngineExtContext
...
         private final ErrorHandler errorHandler;
         private final Collection<DispatchAgent> dispatchers;
+        private final LabelManager labels;
 
         private ContextImpl(
             Configuration config,
             ErrorHandler errorHandler,
-            Collection<DispatchAgent> dispatchers)
+            Collection<DispatchAgent> dispatchers,
+            LabelManager labels)
```
jfallows, line 360: Let's pass in `supplyLabelId` as `ToIntFunction<String>` instead of the entire `LabelManager`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -81,7 +82,8 @@ public final class Engine implements AutoCloseable
     private final Collection<AgentRunner> runners;
     private final Tuning tuning;
     private final List<EngineExtSpi> extensions;
-    private final EngineExtContext context;
+    // visible for testing
+    final EngineExtContext context;
```
jfallows, line 86: Please keep this `private` and add package-private accessor method for test purposes instead.

jfallows (changes requested):

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/EngineContextTest.java
```diff
@@ -0,0 +1,103 @@
...
+        CountersLayout countersLayout1 = new CountersLayout.Builder()
+            .path(path1)
+            .capacity(8192)
+            .mode(CREATE_READ_WRITE)
+            .build();
+        CountersLayout countersLayout2 = new CountersLayout.Builder()
+            .path(path2)
+            .capacity(8192)
+            .mode(CREATE_READ_WRITE)
+            .build();
```
jfallows, line 88: The `labels` and `counters` layouts should already be created by the `engine`, right?

Perhaps we just need an `@Configuration` `zilla.yaml` that registers a `test` metric against a `test` binding and a method on `EngineRule` to supply the counter writer for a specific core# ?

jfallows (changes requested):

PR #272, opened 2023-06-01
Allow Kafka merged stream to change fetch filters dynamically

bmaidics:


specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/fetch/filter.change.none/client.rpt
```diff
@@ -0,0 +1,98 @@
+#
```
bmaidics, line 1: I don't like the name of this scenario, but I couldn't come up with the perfect one. Any suggestion?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
```diff
@@ -2565,6 +2583,23 @@ private void doFetchInitialBegin(
...
         }
 
+        private void doFetchInitialFlush(
+            long traceId,
+            long partitionOffset)
+        {
+            KafkaFlushExFW kafkaFlushEx = kafkaFlushExRW.wrap(extBuffer, 0, extBuffer.capacity())
+                .typeId(kafkaTypeId)
+                .fetch(f -> f
+                    .partition(p -> p.partitionId(partitionId)
```
bmaidics, line 2593: I believe this might not be 100% correct. While doing the end-to-end testing I discovered a weird indeterministic behaviour. Scenario:

- Mqtt flush comes in (adding a new subscription, with send retained true)
- send the kafka flush to the messages stream (this is what's triggering this fetch flush)
- open the retained stream with the undelivered retained filters
- deliver retained messages over the retained stream
- close the retained stream
- publish a new data, that arrives in the messages merged stream

After this the retained message that was already delivered over the retained stream gets delivered again (not always) in the MESSAGES stream. I feel like this might have to do with setting the partition offsets when doing the flush, but couldn't figure out yet.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -887,16 +888,16 @@ private void doClientFanoutReplyWindow(
...
         private final long authorization;
-        private final KafkaCacheCursor cursor;
         private final KafkaIsolation isolation;
         private final KafkaDeltaType deltaType;
         private final KafkaOffsetType maximumOffset;
         private final LongSupplier isolatedOffset;
         private final LongSupplier initialGroupIsolatedOffset;
 
+        private KafkaCacheCursor currentCursor;
+        private KafkaCacheCursor newCursor;
```
jfallows, line 898: Please revert the `currentCursor` rename back to `cursor`, and change `newCusor` to `nextCursor`.
Please also move `cursor` field back to where it was in the source.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -887,16 +888,16 @@ private void doClientFanoutReplyWindow(
...
         private int state;
         private int flushOrDataFramesSent;
-
```
jfallows, line 899: Please revert blank line remove, it is intentional for readability to separate debitor fields.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -962,6 +963,10 @@ private void onClientMessage(
                 final BeginFW begin = beginRO.wrap(buffer, index, index + length);
                 onClientInitialBegin(begin);
                 break;
+            case FlushFW.TYPE_ID:
+                final FlushFW flush = flushRO.wrap(buffer, index, index + length);
+                onClientInitialFlush(flush);
+                break;
```
jfallows, line 969: Looks good.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -1316,7 +1351,7 @@ private void doClientReplyData(
                 {
                     this.messageOffset = 0;
 
-                    cursor.advance(partitionOffset + 1);
+                    currentCursor.advance(partitionOffset + 1);
```
jfallows, line 1354: Before this line we want something like:
```
if (nextCursor != null)
{
    cursor = nextCursor;
    nextCursor = null;
}
```

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -1002,6 +1007,36 @@ private void onClientInitialBegin(
...
+            final KafkaFilterCondition condition = cursorFactory.asCondition(filters, evaluation);
+
+            newCursor = cursorFactory.newCursor(condition, deltaType);
+            Node segmentNode = group.partition.seekNotAfter(initialOffset);
+            if (segmentNode.sentinel())
+            {
+                segmentNode = segmentNode.next();
+            }
+
+            newCursor.init(segmentNode, currentCursor.offset, currentCursor.latestOffset);
```
jfallows, line 1033: Suggest instead adding a new `init` method that takes the `KafkaCacheCursor` and uses the fields necessary to initialize, such as `segmentNode`, `offset` and `latestOffset`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -83,15 +83,15 @@ public KafkaCacheCursor newCursor(
 
     public final class KafkaCacheCursor implements AutoCloseable
     {
-        private final KafkaFilterCondition condition;
+        private KafkaFilterCondition condition;
```
jfallows, line 86: Please move this non-final field down below blank line, just above segmentNode.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCacheCursorFactory.java
```diff
@@ -83,15 +83,15 @@ public KafkaCacheCursor newCursor(
...
         private final KafkaDeltaType deltaType;
         private final LongHashSet deltaKeyOffsets; // TODO: bounded LongHashCache, evict -> discard
 
         private Node segmentNode;
         private KafkaCacheSegment segment;
         public long filters;
         public long offset;
-        private long latestOffset;
+        public long latestOffset;
```
jfallows, line 94: With the proposed new `init` method described later, this change can be reverted.

jfallows: Also please add a blank line before public fields to enhance readability.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
jfallows: We are not supposed to set the partition offsets during the `FLUSH` because we want the cursor offset position to remain unchanged, just pick up from where it is with an adjusted filter, agree?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -1002,6 +1007,36 @@ private void onClientInitialBegin(
...
+            if (messageOffset == 0)
+            {
+                currentCursor = newCursor;
```
jfallows, line 1036: Please add the following for alignment with suggested changes later below.
```
nextCursor = null;
```

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/fetch/filter.change.none/client.rpt
jfallows: This scenario name is okay.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
bmaidics: I moved the cursor because it's not final anymore. Let me know if you think we should move it back anyway.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
bmaidics: You're right. I did the change

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
jfallows: Good point, this positioning makes total sense then. 👍

jfallows (approved):

PR #273, opened 2023-06-06
update Helm logo & details and clean up README

vordimous:
## Description

The logo and other details listed on artifacthub.io come from the chart repo. 
The readme had some incorrect links and formatting.
the `<artifact-id>` needs to be updated based on the id from the [artifacthub control panel](https://artifacthub.io/docs/topics/repositories/#verified-publisher)

Fixes https://github.com/aklivity/zilla-docs/issues/63


cloud/helm-chart/src/main/helm/artifacthub-repo.yml
```diff
@@ -0,0 +1,6 @@
+# This file is used by ArtifactHub to verify the repository.
```
attilakreiner, line 1: I think @jfallows has already tried this approach and it did not work. The [documentation](https://artifacthub.io/docs/topics/repositories/#verified-publisher) says you have to put this file to the _repository_ root (not inside a helm chart). It mentions http and git repositories specifically, but there is no info on whether oci repositories are supported.

attilakreiner:
Looks good to me with one comment about artifacthub.

attilakreiner (dismissed):

cloud/helm-chart/src/main/helm/artifacthub-repo.yml
vordimous: Well that doesn't sound promising. Maybe adding the 'source' link in the chart will help? I'll do some more digging to double check.

vordimous: I found this repo that is verified and has the same setup: https://github.com/coder/coder/blob/main/helm/Chart.yaml https://artifacthub.io/packages/helm/coder-v2/coder
Hopefully that means it will work this time.

vordimous: I see so the deploy operation will need to either generate this file or deploy this file along with the chart? Here is how Chartmuseum adds this https://github.com/helm/chartmuseum#artifact-hub

attilakreiner: We decided to go with ghcr's oci based helm repository. I am not sure if you can just push arbitrary files (e.g. a yaml file) to the oci repo. Also there is no word in the artifacthub doc about how you can get an oci based repo verified. Chartmuseum implements a http (rest) based helm repo, so that's not an answer to our question.

vordimous: looks like we can push it seprately: https://artifacthub.io/docs/topics/repositories/helm-charts/#oci-support

cloud/helm-chart/src/main/helm/artifacthub-repo.yml
```diff
@@ -0,0 +1,6 @@
...
+# See: https://artifacthub.io/packages/helm/zilla/zilla
+repositoryID: <artifact-id>
```
jfallows, line 3:
```suggestion
repositoryID: 5c0f4454-3dbb-4aad-a73f-b760c0f0c6b5
```

jfallows (changes requested):

jfallows (approved):

PR #277, opened 2023-06-16
Create OpenTelemetry exporter

### LINKED ISSUE #112: Integrate OpenTelemetry collectors by exporting local metrics over OTLP
Note: requires design #109 and #111.

Support the `otlp` kind of telemetry `exporter`, contributed via the `telemetry-otlp` component.

```
    "exporters":
    {
      "otlp0":
      {
        "kind": "otlp",
        "endpoint": "otlp://myserver.local:55690"
        "endpoint": "https://example.com:4318"  <-- /v1/traces, /v1/metrics, /v1/logs (json or protobuf, grpc)
        "vault": ...
      }
    }
```

We need to determine the frequency of metrics reporting and potential aggregation across cores.

It may be simpler to export directly on each local metrics collection core individually until we determine whether or not we need more than one metrics collection core to fully keep up with the multiple data cores.

attilakreiner:
## Description

Create OpenTelemetry exporter to export collected and locally stored metrics to the OpenTelemetry Collector.

Fixes #112


incubator/pom.xml
```diff
@@ -28,6 +28,9 @@
     <module>command-log</module>
     <module>command-dump</module>
     <module>command-tune</module>
+
+    <module>exporter-otlp</module>
+    <module>exporter-otlp.spec</module>
```
jfallows, line 33: All specs go first, later they will all move to `specs` project.

incubator/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/schema/otlp.schema.patch.json
```diff
@@ -0,0 +1,66 @@
+[
```
jfallows, line 1: Is this configuration forwards compatible for the future use of gRPC instead of JSON for OTLP protocol?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/NamespaceRegistry.java
```diff
@@ -326,6 +327,14 @@ BindingRegistry findBinding(
         return bindingsById.get(bindingId);
     }
 
+    public KindConfig findBindingKind(
```
jfallows, line 330: Please rename to `resolveBindingKind`.

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/duplicated/MetricDescriptor.java
```diff
@@ -0,0 +1,27 @@
...
+ *   https://www.aklivity.io/aklivity-community-license/
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package io.aklivity.zilla.runtime.exporter.otlp.internal.duplicated;
+
+public interface MetricDescriptor
```
jfallows, line 17: I think we don't need to use `internalName` terminology here, as the names are the first class names of the metrics we have.

If an exporter needs to externalize the name, thats specific to the exporter, not generic to the engine concept of metrics.

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/duplicated/MetricRecord.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package io.aklivity.zilla.runtime.exporter.otlp.internal.duplicated;
+
+public interface MetricRecord
+{
+    String namespaceName();
+
+    String bindingName();
+
+    String metricName();
```
jfallows, line 23: We don't need to suffix `Name` on these methods, they all return strings so it is already pretty clear.

Perhaps these should be ids though?

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/duplicated/MetricsLayout.java
```diff
@@ -0,0 +1,163 @@
...
+    @Override
+    public void close()
+    {
+        unmap(buffer.byteBuffer());
+    }
+
+    public long[][] getIds()
+    {
+        Spliterator<long[]> spliterator = Spliterators.spliteratorUnknownSize(new MetricsIterator(), 0);
+        return StreamSupport.stream(spliterator, false).toArray(long[][]::new);
```
jfallows, line 54: This is _not_ called repetitively, right? (allocation)

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/OtlpExporterTask.java
```diff
@@ -0,0 +1,65 @@
...
+            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+            connection.setRequestMethod("POST");
+            connection.setRequestProperty("Content-Type", "application/json");
+            connection.setDoOutput(true);
+            OutputStream os = connection.getOutputStream();
+            os.write(json.getBytes());
+            os.flush();
+            os.close();
+            connection.getResponseCode();
+            connection.disconnect();
```
jfallows, line 58: Shouldn't we be using the new HttpClient API instead of HttpURLConnection?

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/exporter/ExporterContext.java
```diff
@@ -15,12 +15,19 @@
...
 import io.aklivity.zilla.runtime.engine.config.ExporterConfig;
+import io.aklivity.zilla.runtime.engine.config.KindConfig;
 
 public interface ExporterContext
 {
     ExporterHandler attach(
-        ExporterConfig config);
+        ExporterConfig config,
+        List<AttributeConfig> attributes,
+        Function<String, KindConfig> findBindingKind);
```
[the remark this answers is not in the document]
attilakreiner: done

incubator/pom.xml
attilakreiner: done

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/NamespaceRegistry.java
attilakreiner: renamed to `resolveKind` as per your previous comment

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/duplicated/MetricDescriptor.java
attilakreiner: I decided to remove this interface altogether (PR #279). It was about to be used by prometheus & otlp exporters, but it turned out they needed different methods after all, so creating this interface wasn't a good abstraction idea.

The actual implementations `OtlpMetricsDescriptor` and `PrometheusMetricDescriptor` indeed use the argument name `internalName`. The reason is that one of the things these classes do is translate the internal metric name to the external metric name, so using just `name` could be ambigious in this context.

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/duplicated/MetricRecord.java
attilakreiner: The implementations (`CounterGaugeRecord` and `HistogramRecord`) get the id's in the constructor and resolve them to names (strings) for the very reason of making the names available for the exporters' purposes.

To fix your other comment I have just added the method `bindingId()`, so I think it's just better to keep the method names like `bindingName()`, `bindingId()` etc. just to make sure there is no ambiguity. What do you think? 🤔

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/duplicated/MetricsLayout.java
attilakreiner: It is called only when we initiate the `MetricsProcessor` (renamed to `MetricsReader` in PR #279). We need to call it 3 times (for counters, gauges and histograms), because they are in their respective data files and we need to get the id's of all.

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/OtlpExporterTask.java
attilakreiner: done

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/duplicated/MetricDescriptor.java
jfallows: Right, I was making the point that it is not an internal name, it is just the name, and that anything being exported differently is an externalName or exportedName, so that should be the special case, not internalName.

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/duplicated/MetricsLayout.java
jfallows: Okay, understood it is only called during initialization.

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/duplicated/MetricRecord.java
attilakreiner: We put the binding name to Prometheus labels and Open Telemetry attributes, so we definitely need it.

jfallows: Right, but `EngineContext.supplyLocalName` already provides a lookup from ids to Strings that works for any `namespacedId` and `ExporterHandler` already has access to `EngineContext` so we only need the ids.

Lets have a design discussion on this to review the abstractions being introduced into the engine project for exporters.

PR #280, opened 2023-06-16
Fix dependencies to run TcpServerBM from command line

akrambek:
## Description

Fixes # (issue)


runtime/binding-tcp/src/main/moditect/module-info.java
```diff
@@ -25,4 +25,7 @@
 
     provides io.aklivity.zilla.runtime.engine.config.ConditionConfigAdapterSpi
         with io.aklivity.zilla.runtime.binding.tcp.internal.config.TcpConditionConfigAdapter;
+
+    provides jakarta.json.spi.JsonProvider
+        with org.leadpony.joy.yaml.YamlProvider;
```
jfallows, line 30: This should not be necessary because we don't use modules for the shaded test jar.

runtime/binding-tcp/src/main/resources/META-INF/services/jakarta.json.spi.JsonProvider
```diff
@@ -0,0 +1 @@
+org.leadpony.joy.yaml.YamlProvider
```
jfallows, line 1: Replicating this here doesn't seem correct, it should be coming from the engine JAR when shading instead.

runtime/binding-tcp/pom.xml
```diff
@@ -250,6 +249,14 @@
...
               <include>org.agrona:agrona</include>
               <include>io.aklivity.zilla:engine</include>
+              <include>jakarta.json:jakarta.json-api</include>
+              <include>jakarta.json.bind:jakarta.json.bind-api</include>
+              <include>org.eclipse:yasson</include>
+              <include>org.leadpony.justify</include>
+              <include>org.leadpony.joy:joy-yaml</include>
+              <include>org.leadpony.joy:joy-core</include>
+              <include>org.snakeyaml:snakeyaml-engine</include>
+              <include>com.ibm.icu:icu4j</include>
```
jfallows, line 259: Is there any shade plugin configuration to pull transitive dependencies implicitly?
Otherwise this will be brittle as we change the dependencies in engine over time.

runtime/binding-tcp/pom.xml
```diff
@@ -41,7 +41,6 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>engine</artifactId>
       <version>${project.version}</version>
-      <scope>provided</scope>
```
jfallows, line 44: This looks right, to let the shade plugin pick up the dependency.

runtime/binding-tcp/src/test/java/io/aklivity/zilla/runtime/binding/tcp/internal/bench/TcpServerBM.java
```diff
@@ -61,19 +66,34 @@
...
+                String configJSON = String.format("file:%s", configPath.resolveSibling("zilla.json"));
+                props.setProperty(ENGINE_CONFIG_URL.name(), configJSON);
+                configPath = Paths.get(config.configURL().getPath());
+                System.out.println("zilla.yaml file not found, loading zilla.json instead");
+            }
+            if (configPath.getFileName().toString().endsWith(".json"))
+            {
+                System.out.println("warning: json syntax is deprecated, migrate to yaml");
+            }
+        }
```
jfallows, line 89: We don't need to replicate this backwards compatibility code from `zilla start` in the benchmark, please remove it.

The integration with OpenJMH will add support for picking up a configuration similar to ITs, so it won't be coming from command line or be file based.

runtime/binding-tcp/src/test/java/io/aklivity/zilla/runtime/binding/tcp/internal/bench/TcpServerBM.java
```diff
@@ -61,19 +66,34 @@
     //private long routedId;
 
     {
-        Properties properties = new Properties();
-        properties.setProperty(ENGINE_DIRECTORY.name(), "target/zilla-benchmarks");
-        properties.setProperty(ENGINE_STREAMS_BUFFER_CAPACITY.name(), Long.toString(1024L * 1024L * 16L));
+        Properties props = new Properties();
```
jfallows, line 69: Please revert this rename back to `properties`.

jfallows (changes requested):

runtime/binding-tcp/pom.xml
```diff
@@ -250,13 +249,22 @@
...
+              <include>org.snakeyaml:snakeyaml-engine</include>
+              <include>com.ibm.icu:icu4j</include>
               <include>org.openjdk.jmh:jmh-core</include>
               <include>net.sf.jopt-simple:jopt-simple</include>
               <include>org.apache.commons:commons-math3</include>
               <include>commons-cli:commons-cli</include>
               <include>com.github.biboudis:jmh-profilers</include>
             </includes>
           </artifactSet>
+          <includeDependencies>true</includeDependencies>
```
jfallows, line 267: If this picks up dependencies transitively, can we trim the list above to no longer need to include any engine dependencies explicitly?

That would be better to implicitly keep the engine dependencies in sync.

runtime/binding-tcp/src/test/java/io/aklivity/zilla/runtime/binding/tcp/internal/bench/TcpServerBM.java
```diff
@@ -65,6 +65,9 @@
         properties.setProperty(ENGINE_DIRECTORY.name(), "target/zilla-benchmarks");
         properties.setProperty(ENGINE_STREAMS_BUFFER_CAPACITY.name(), Long.toString(1024L * 1024L * 16L));
 
+        properties.setProperty(ENGINE_DIRECTORY.name(), "target/zilla-benchmarks");
+        properties.setProperty(ENGINE_STREAMS_BUFFER_CAPACITY.name(), Long.toString(1024L * 1024L * 16L));
+
```
jfallows, line 70: These lines can now be removed as they duplicate the lines above, right?

runtime/engine/src/main/resources/META-INF/services/jakarta.json.spi.JsonProvider
```diff
@@ -0,0 +1 @@
+org.leadpony.joy.yaml.YamlProvider
```
jfallows, line 1: This file is already present in `joy-yaml` JAR, and not required by `engine` project otherwise.

If dependencies are being picked up correctly by `shade` plugin, shouldn't `joy-yaml` dependency cause this to be included in the shaded JAR without having to explicitly put it into our sources?

runtime/binding-tcp/pom.xml
jfallows: Also, suggest moving this configuration to top level `pom.xml` instead, with the rest of the default configuration for `shade` plugin as each project will need the same configuration.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #290, opened 2023-06-28
Mqtt retained feature

bmaidics:


incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -194,7 +217,18 @@ private final class MqttSubscribeProxy
...
+
+        private final Queue<List<Subscription>> deferredRetainedFiltersQueue = new LinkedList<>();
+        private final List<Subscription> retainedFilters = new ArrayList<>();
+        private final IntArrayList messagesSubscriptionIds = new IntArrayList();
+        private final IntArrayList retainedSubscriptionIds = new IntArrayList();
+        private final Long2ObjectHashMap<Boolean> retainAsPublished = new Long2ObjectHashMap<>();
+
+        private String clientId;
+        private Array32FW<MqttTopicFilterFW> currentFilters;
+        private Array32FW<MqttTopicFilterFW> currentRetainedFilters;
```
jfallows, line 231: Please add a blank line between fields and constructor.

jfallows: Also please remove "current" prefix for filters and retained filters.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -268,7 +303,36 @@ private void onMqttBegin(
...
 
-            delegate.doKafkaBegin(traceId, authorization, affinity);
+            final OctetsFW extension = begin.extension();
+            final MqttBeginExFW mqttBeginEx = extension.get(mqttBeginExRO::tryWrap);
+
+            assert mqttBeginEx.kind() == MqttBeginExFW.KIND_SUBSCRIBE;
+            final MqttSubscribeBeginExFW mqttSubscribeBeginEx = mqttBeginEx.subscribe();
+
+            clientId = mqttSubscribeBeginEx.clientId().asString();
+            currentFilters = mqttSubscribeBeginEx.filters();
```
jfallows, line 313: Why do we need to keep these filters around as fields? Can they live on stack instead?

Note that the reference refers to flyweights backed by message from streams ring buffer, so becomes invalid when stack unwinds.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -282,52 +346,95 @@ private void onMqttFlush(
...
 
             assert mqttFlushEx.kind() == MqttFlushExFW.KIND_SUBSCRIBE;
             final MqttSubscribeFlushExFW mqttSubscribeFlushEx = mqttFlushEx.subscribe();
 
-            filters = mqttSubscribeFlushEx.filters();
-            subscriptionIds.clear();
+            currentFilters = mqttSubscribeFlushEx.filters();
+            messagesSubscriptionIds.clear();
+
+            final KafkaFlushExFW kafkaFlushEx = createKafkaFlushEx(currentFilters);
```
jfallows, line 365: We aren't really "creating" the extension as it's a flyweight, prefer to do this on-stack please instead of hoisted to a separate method.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -282,52 +346,95 @@ private void onMqttFlush(
...
 
+            messages.doKafkaFlush(traceId, authorization, budgetId, reserved, kafkaFlushEx);
 
-            final KafkaFlushExFW kafkaFlushEx =
-                kafkaFlushExRW.wrap(writeBuffer, FlushFW.FIELD_OFFSET_EXTENSION, writeBuffer.capacity())
-                    .typeId(kafkaTypeId)
-                    .merged(m ->
+            if (retainAvailable)
+            {
+                retainedFilters.removeIf(rf -> !currentFilters.anyMatch(f -> f.pattern().asString().equals(rf.filter)));
```
jfallows, line 371: Can we avoid the `.asString()` call here?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -282,52 +346,95 @@ private void onMqttFlush(
...
+                final List<Subscription> newRetainedFilters = new ArrayList<>();
+                currentFilters.forEach(filter ->
+                {
+                    final boolean sendRetained = (filter.flags() & SEND_RETAIN_FLAG) != 0;
+                    final boolean alreadySentRetained =
+                        retainedFilters.stream().anyMatch(rf -> rf.filter.equals(filter.pattern().asString()));
```
jfallows, line 377: also here

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -282,52 +346,95 @@ private void onMqttFlush(
...
+                    if (sendRetained && !alreadySentRetained)
+                    {
+                        newRetainedFilters.add(new Subscription(
+                            (int) filter.subscriptionId(), filter.pattern().asString(), filter.qos(), filter.flags()));
```
jfallows, line 381: here too 😄

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -282,52 +346,95 @@ private void onMqttFlush(
...
+                    }
+                    else
                     {
-                        m.capabilities(c -> c.set(KafkaCapabilities.FETCH_ONLY));
-                        filters.forEach(filter ->
+                        retained.state = 0;
+                        retained.replySeq = 0;
+                        retained.replyAck = 0;
+                        retained.replyMax = 0;
+                        retained.doKafkaBegin(traceId, authorization, 0, newRetainedFilters);
```
jfallows, line 399: It might be clearer to reinitialize the retained state inside `doKafkaBegin` of `retained` instead of modifying it outside.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -207,7 +241,8 @@ private MqttSubscribeProxy(
             this.routedId = routedId;
             this.initialId = initialId;
             this.replyId = supplyReplyId.applyAsLong(initialId);
-            this.delegate = new KafkaProxy(originId, resolvedId, this);
+            this.messages = new KafkaMessagesProxy(originId, resolvedId, this);
+            this.retained = new KafkaRetainedProxy(originId, resolvedId, this);
```
jfallows, line 245: The `resolvedId` could be different for `messages` vs `retained`, based on routing, right?

Suggest using the information from the `KafkaBeginEx` in the factory newStream method to determine if `retained` is needed, passing `0` or `non-zero` for `retainedId` (resolved retained id).

Then retained would be initialized only for non-zero `retainedId`, and subsequent methods would test for `retained != null` instead of `retainAvailable`, which can then be removed from `MqttKafkaConfiguration`.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -346,7 +453,14 @@ private void onMqttData(
             assert initialAck <= initialSeq;
 
             doMqttReset(traceId);
-            delegate.doKafkaAbort(traceId, authorization);
+            if (MqttKafkaState.initialOpened(messages.state))
+            {
+                messages.doKafkaAbort(traceId, authorization);
+            }
```
jfallows, line 459: Why is this conditional at the caller based on internal state of `messages` stream?

Usually we check internal state inside the method and behave accordingly.
Please take a look at other similar method calls in the onMqtt* method below.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -431,20 +566,33 @@ private void onMqttWindow(
             replyMax = maximum;
             replyPad = padding;
             state = MqttKafkaState.openReply(state);
+            this.budgetId = window.budgetId();
```
jfallows, line 569: Field name convention for reply budget id is `replyBud`, please rename.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -530,16 +678,99 @@ private void doMqttReset(
...
+        {
+            return (int)(replySeq - replyAck);
+        }
+
+        private int replyWindow()
+        {
+            return replyMax - replyPendingAck();
+        }
+
+        private MqttDataExFW createSubscribeDataEx(
```
jfallows, line 692: Try to avoid this pattern of "creating" flyweights before use.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -569,13 +800,25 @@ private void doKafkaBegin(
...
+                initialMax = mqtt.initialMax;
+                state = MqttKafkaState.openingInitial(state);
+                mqtt.currentFilters.forEach(filter ->
+                {
+                    int subscriptionId = (int) filter.subscriptionId();
+                    if (!mqtt.messagesSubscriptionIds.contains(subscriptionId))
+                    {
+                        mqtt.messagesSubscriptionIds.add(subscriptionId);
+                    }
+                });
```
jfallows, line 816: Try to avoid manipulating internal state of the other side if possible as it breaks the encapsulation.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -698,93 +942,109 @@ private void onKafkaData(
             assert acknowledge <= sequence;
             assert sequence >= replySeq;
 
-            replySeq = sequence + reserved;
+            replySeq += reserved;
```
jfallows, line 945: Is this change correct?
Note: it seems to assume `sequence` doesn't advance, but the other side can move `sequence` forward if it wants to as well.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -903,15 +1163,395 @@ private void doKafkaWindow(
             int padding,
             int capabilities)
         {
-            replyAck = delegate.replyAck;
-            replyMax = delegate.replyMax;
-            replyPad = delegate.replyPad;
+            //
```
jfallows, line 1166: Remove?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -194,7 +217,18 @@ private final class MqttSubscribeProxy
...
         private int replyPad;
-
+        private long budgetId;
+        private int mqttSharedBudget;
+
+        private final Queue<List<Subscription>> deferredRetainedFiltersQueue = new LinkedList<>();
+        private final List<Subscription> retainedFilters = new ArrayList<>();
+        private final IntArrayList messagesSubscriptionIds = new IntArrayList();
+        private final IntArrayList retainedSubscriptionIds = new IntArrayList();
+        private final Long2ObjectHashMap<Boolean> retainAsPublished = new Long2ObjectHashMap<>();
```
jfallows, line 227: Fields that are not non-final constants should be initialized in the constructor please.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -282,52 +346,95 @@ private void onMqttFlush(
...
+                        final boolean rap = (filter.flags() & RETAIN_AS_PUBLISHED_FLAG) != 0;
+                        retainAsPublished.put((int) filter.subscriptionId(), rap);
+                    }
+                });
+
+                if (!newRetainedFilters.isEmpty())
+                {
+                    if (MqttKafkaState.initialOpened(retained.state) && !MqttKafkaState.initialClosed(retained.state))
+                    {
+                        deferredRetainedFiltersQueue.add(newRetainedFilters);
```
jfallows, line 391: What problem are we trying to solve by queuing the retained filters?
If we can solve it without introducing buffering state, that would be better. 😄

jfallows (changes requested):

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1201,7 +1201,7 @@ private final class MqttServer
         private final Int2ObjectHashMap<MqttPublishStream> publishStreams;
         private final Int2ObjectHashMap<MqttSubscribeStream> subscribeStreams;
         private final Int2ObjectHashMap<String> topicAliases;
-        private final Map<Integer, Integer> subscribePacketIds;
+        private final Map<String, Integer> subscribePacketIds;
```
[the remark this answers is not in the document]
bmaidics: I found Object2IntHashMap, which uses primitives. Let me know it this is good enough or not.

incubator/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/network/disconnect.close.subscribe.and.publish/client.rpt
```diff
@@ -0,0 +1,58 @@
...
+write [0x30 0x14]                   # PUBLISH
+      [0x00 0x0a] "sensor/two"      # topic name
+      [0x00]                        # properties
+      "message"                     # payload
+
+write [0xe0 0x02]                   # DISCONNECT
+      [0x00]                        # normal disconnect
+      [0x00]                        # properties = none
+
+write close
```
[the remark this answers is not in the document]
bmaidics: No, it's not necessary. We close the streams at disconnect as well.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/MqttKafkaConfiguration.java
```diff
@@ -21,11 +21,15 @@ public class MqttKafkaConfiguration extends Configuration
...
     public static final PropertyDef<String> KAFKA_MESSAGES_TOPIC;
+    public static final PropertyDef<String> KAFKA_RETAINED_MESSAGES_TOPIC;
+    public static final BooleanPropertyDef RETAIN_AVAILABLE;
 
     static
     {
         final ConfigurationDef config = new ConfigurationDef("zilla.binding.mqtt.kafka");
         KAFKA_MESSAGES_TOPIC = config.property("messages.topic", "mqtt_messages");
+        KAFKA_RETAINED_MESSAGES_TOPIC = config.property("retained.messages.topic", "mqtt_retained");
+        RETAIN_AVAILABLE = config.property("retain.available", true);
```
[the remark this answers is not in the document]
bmaidics: I added this so that we won't see retain streams in non-retain related scenarios in the scripts: like for sessions in the mqtt binding. But in reality, this is always true (if we don't specify it from the test). 
Since it's always true in prod, in reality we're responding to the MqttPublish and MqttSubscribe streams flags.
Let me know if you think we should delete it and adapt the scenarios for the change.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: I think this was necessary because of the buffered messages, as we don't increase the sequence at onKafkaData, but later.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: This was a valid concern of yours:
"we might not need to buffer if we initiate the new retained stream for any new filters and defer sending reply window to let it send messages until after the previous retained stream is caught up to live and closed
so there would potentially be more than one retained stream established, but only one allowed to send messages at a time"

So we don't allow more than 1 retained stream to be open. But multiple new retained filter can come in a short amount of time.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: I'm not sure if we can. As we can defer retained filters, we need to remember to retained filters that we have already opened the stream for. So let's say we receive these filters with two mqtt flush: 

1. [<sensor/one>, <sensor/two, retained>]
2. [<sensor/one>, <sensor/two, retained>, <sensor/three, retained>]

In this case for the first mqtt flush, we'll open the sensor/two retained stream. BUT the client subscribed for another filter. We get all the filters again, but we can't open sensor/two stream again. If we use the FW object, we're not "remembering" the ones that were opened already, as it changes constantly.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: This is somewhat similar to the deferred retained filters. If we can come up with a different solution for that, this might be solved differently as well.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: "The resolvedId could be different for messages vs retained, based on routing, right?"
I think we didn't talk about this yet. Do we want to allow the user to control the mqtt_retained details? Let's talk about this.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: It can happen that e.g we get an mqtt abort, but the messages stream is not even open yet, as we still delivering the retained messages.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: What do you mean by "before use"? On both places, we're using it right after it wast "created". Do you mean we should not use a method for this?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/MqttKafkaConfiguration.java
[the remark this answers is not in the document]
jfallows: I think we should be responding to the flags in the beginEx only and not have separate configuration. 

The scenarios should be kept simple (without retained as input) and therefore not send a retained stream.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
jfallows: I understand that we might need to hold a reference, just asking if it has to be utf-8 decoded each time or if we can do a more direct comparison, say using StringFW.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
jfallows: That's a comment about why we need the logic. My feedback was not that the logic is unnecessary, it was about where the logic lives.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
jfallows: There is a way to pass a function to the builder so it receives a nested builder without us needing to model is separately and pass by value.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/MqttKafkaConfiguration.java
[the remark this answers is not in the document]
bmaidics: Sounds good for subscribe retained, but I think we discussed that the publish retained stream is always open, and if the client did not send any retained message, we just keep that stream in sync

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
jfallows: But if the reply sender advances to he sequence, this will break, right?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: Of course, I have the Subscription object for that

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
jfallows: Ah, by buffer I had meant deferred reaction, not changing internal representation. I wasn't expecting an object data structure.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/MqttKafkaConfiguration.java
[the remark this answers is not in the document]
jfallows: Understood. Is there configuration in `mqtt` binding to disable `retained` publishing?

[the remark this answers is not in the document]
bmaidics: For mqtt publish beginEx add a flag: retain available or not. If not we can assume that we won't get publish with retain flag and can reject them if we do. Honor the retain flag in the subscribe stream beginEx

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: Use the already opened retained stream and change the filters (check if it introduces any problems)

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -3638,7 +3608,7 @@ private void onPublishWindow(
                 if (MqttState.initialClosing(state) &&
                     !MqttState.initialClosed(state))
                 {
-                    doPublishNetEnd(traceId, authorization, EMPTY_OCTETS);
+                    doPublishAppEnd(traceId, authorization);
```
jfallows, line 3611: `doPublishAppEnd` should already have a check for `if (!MqttState.initialClosed(state))` and we can remove it from this `if` condition then.

Note also that if it was closed, we wouldn't get the `WINDOW` frame because the `throttle` handler would already be removed by the engine.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -3761,8 +3731,7 @@ private void doPublishAppEnd(
             {
                 if (!MqttState.replyClosed(state))
                 {
-                    setPublishAppClosed();
-                    doEnd(application, originId, routedId, replyId, replySeq, replyAck, replyMax,
+                    doEnd(application, originId, routedId, initialId, initialSeq, initialAck, initialMax,
```
jfallows, line 3734: This check should `if (!MqttState.initialClosed(state))`, agree?

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -3811,7 +3780,6 @@ private void cleanupEnd(
                 long traceId,
                 long authorization)
             {
-                doPublishNetEnd(traceId, authorization, EMPTY_OCTETS);
```
jfallows, line 3814: Suggest that the publish expiration should be cleaned up implicitly when the stream is closed, i.e. called from `doPublishAppEnd` and `doPublishAppAbort`, agree?

Then `cleanupEnd` reduces to `doPublishAppEnd` and calls to `cleanupEnd` can be replaced by `doPublishAppEnd` and `cleanupEnd` can be removed.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -4240,7 +4194,6 @@ private void cleanupEnd(
                 long traceId,
                 long authorization)
             {
-                doSubscribeNetEnd(traceId, authorization);
```
jfallows, line 4243: Now that `cleanupEnd` has reduced to `doSubscribeAppEnd`, we can replace calls to `cleanupEnd` with `doSubscribeAppEnd` and remove `cleanupEnd`.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -986,18 +1624,21 @@ private void doEnd(
...
-            .originId(originId)
-            .routedId(routedId)
-            .streamId(streamId)
-            .sequence(sequence)
-            .acknowledge(acknowledge)
-            .maximum(maximum)
-            .traceId(traceId)
-            .authorization(authorization)
-            .build();
+        if (receiver != null)
```
jfallows, line 1627: This defensive coding seems like it is hiding a different bug.
Why would `receiver` be `null` here?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -108,12 +110,10 @@ public class MqttKafkaPublishFactory implements BindingHandler
...
-    private KafkaKeyFW key;
-
-    private OctetsFW[] topicNameHeaders;
-    private OctetsFW clientIdOctets;
-    private String16FW binaryFormat;
-    private String16FW textFormat;
+    private final String16FW binaryFormat;
+    private final String16FW textFormat;
+    private final String kafkaTopic;
+    private final String kafkaRetainedTopic;
```
jfallows, line 116: These topic names can be stored as `StringFW` to `utf-8` encode once, rather than storing as `String` and `utf-8` encoding per usage.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -280,7 +266,36 @@ private void onMqttBegin(
...
+                topicNameHeaders[i] = new OctetsFW().wrap(topicHeader.value(), 0, topicHeader.length());
+            }
+            clientIdOctets = new OctetsFW()
+                .wrap(mqttPublishBeginEx.clientId().value(), 0, mqttPublishBeginEx.clientId().length());
+            final DirectBuffer topicNameBuffer = mqttPublishBeginEx.topic().value();
+            key = new KafkaKeyFW.Builder()
+                .wrap(keyBuffer, 0, keyBuffer.capacity())
+                .length(topicNameBuffer.capacity())
+                .value(topicNameBuffer, 0, topicNameBuffer.capacity())
+                .build();
```
jfallows, line 291: The storage for this `key`, i.e. the `keyBuffer`, is shared and will be overwritten by subsequent publish streams for different mqtt topics on the same core.

I think we have this backwards. The flyweight and builder can be shared, as they are simply strongly-typed access to the underlying key buffer, but the key buffer itself cannot be shared as it would be overwritten.

The same feedback applies to the client id and the topic headers.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -1209,4 +1853,20 @@ private void doReset(
...
     }
+
+    private static final class Subscription
+    {
+        private final int id;
+        private final String filter;
+        private final int qos;
+        private final int flags;
+
+        Subscription(int id, String filter, int qos, int flags)
```
jfallows, line 1864: Parameter formatting.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -1066,25 +1707,28 @@ private MessageConsumer newKafkaStream(
         int maximum,
         long traceId,
         long authorization,
-        long affinity)
+        long affinity,
+        String clientId,
+        String topic,
```
jfallows, line 1712: Lets pass these as `StringFW` instead - likely this change is implicit when other feedback above has been applied..

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -903,15 +1150,406 @@ private void doKafkaWindow(
...
+
+        private void doKafkaWindow(
+            long traceId,
+            long authorization,
+            long budgetId,
+            int padding,
+            int capabilities)
+        {
+            if (MqttKafkaState.initialOpening(state) &&
+                !MqttKafkaState.initialClosing(state))
```
jfallows, line 1529: This is reply window, so we should be checking reply state, agree?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -903,15 +1150,406 @@ private void doKafkaWindow(
...
+
+            assert acknowledge <= sequence;
+            assert sequence >= replySeq;
+
+            replySeq = sequence;
+            state = MqttKafkaState.closeReply(state);
+
+            assert replyAck <= replySeq;
+
+            if (!mqtt.deferredRetainedFiltersQueue.isEmpty())
```
jfallows, line 1414: Do we still need a queue here? I thought we would either immediately begin or immediately flush with updated retained filters, no?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -903,15 +1150,406 @@ private void doKafkaWindow(
...
+                final OctetsFW key = kafkaMergedDataEx != null ? kafkaMergedDataEx.key().value() : null;
+                final long filters = kafkaMergedDataEx != null ? kafkaMergedDataEx.filters() : 0;
+
+                if (key != null)
+                {
+                    String topicName = kafkaMergedDataEx.key().value()
+                        .get((b, o, m) -> b.getStringWithoutLengthUtf8(o, m - o));
+                    helper.visit(kafkaMergedDataEx);
+                    final Flyweight mqttSubscribeDataEx = mqtt.createSubscribeDataEx(kafkaMergedDataEx, filters, topicName,
+                        true, mqtt.retainedSubscriptionIds);
```
jfallows, line 1389: This is still using the "createXX" approach we discussed removing.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
jfallows: this change doesn't seem to have been applied to the code

jfallows (changes requested):

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: Yep, I did not push the changes yet. Once I address a comment locally, I immediately mark it as resolved, so I have a clean overview what still needs to be done.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: TBH, I'm not sure why it was needed. I just deleted it, and everything still seems to work fine

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: We can't use the same solution as we did for the deferred retained filters, as we decided to solve the issue using Kafka flush. Since we first need to server the retained messages, we need to store the filters. I'm using flyweights for that, but using a buffer per client.

incubator/binding-mqtt-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/KafkaIT.java
```diff
@@ -107,6 +143,15 @@ public void shouldSendOneMessage() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${kafka}/publish.retained/client",
+        "${kafka}/publish.retained/server"})
+    public void shouldSendRetainedMessage() throws Exception
```
jfallows, line 150: `shouldPublishRetainedMessage`

incubator/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishProxyIT.java
```diff
@@ -110,6 +110,46 @@ public void shouldAbortWhenServerSentData() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Configuration("proxy.yaml")
+    @Specification({
+        "${mqtt}/publish.retained.server.sent.abort/client",
+        "${kafka}/publish.retained.server.sent.abort/server"})
+    public void shouldReceiveServerSentRetainedAbort() throws Exception
```
jfallows, line 118: Please review these test method names vs script scenario names.
For example, this one seems like it should be `shouldPublishRetainedThenReceiveServerSentAbort`

incubator/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/application/disconnect.after.subscribe.and.publish/client.rpt
```diff
@@ -0,0 +1,39 @@
...
+
+read zilla:data.ext ${mqtt:matchDataEx()
+                              .typeId(zilla:id("mqtt"))
+                              .subscribe()
+                                 .topic("sensor/one")
+                                 .subscriptionId(1)
+                                 .format("TEXT")
+                                 .build()
+                              .build()}
+read "message"
```
jfallows, line 39: Missing the `publish` stream to make this symmetrical with `server.rpt`?

incubator/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/network/disconnect.after.subscribe.and.publish/server.rpt
```diff
@@ -0,0 +1,58 @@
...
+      [0x0b 0x01]                   # subscription id = 1
+      [0x00 0x0a] "sensor/one"      # topic filter
+      [0x20]                        # options = at-most-once
+
+write [0x90 0x04]                   # SUBACK
+      [0x00 0x01]                   # packet id = 1
+      [0x00]                        # properties = none
+      [0x00]                        # reason code
+
+read [0x30 0x14]                   # PUBLISH
```
jfallows, line 50: Comment alignment is off by one 😉

incubator/binding-mqtt.spec/src/test/java/io/aklivity/zilla/specs/binding/mqtt/streams/network/ConnectionIT.java
```diff
@@ -373,4 +373,13 @@ public void shouldConnectAndSubscribeUnfragmented() throws Exception
     {
         k3po.finish();
     }
+
+    @Test
+    @Specification({
+        "${net}/disconnect.after.subscribe.and.publish/client",
+        "${net}/disconnect.after.subscribe.and.publish/server"})
+    public void shouldCloseSubscribeAndPublish() throws Exception
```
jfallows, line 381: `shouldDisconnectAfterSubscribeAndPublish`?

incubator/binding-mqtt/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/ConnectionIT.java
```diff
@@ -119,6 +119,22 @@ public void shouldConnectThenDisconnect() throws Exception
...
+    @Configuration("server.yaml")
+    @Specification({
+        "${net}/disconnect.after.subscribe.and.publish/client",
+        "${app}/disconnect.after.subscribe.and.publish/server"})
+    @Configure(name = SESSION_AVAILABLE_NAME, value = "false")
+    @Configure(name = WILDCARD_SUBSCRIPTION_AVAILABLE_NAME, value = "true")
+    @Configure(name = SHARED_SUBSCRIPTION_AVAILABLE_NAME, value = "true")
+    @Configure(name = MAXIMUM_QOS_NAME, value = "2")
+    @Configure(name = SESSION_EXPIRY_INTERVAL_NAME, value = "0")
+    public void shouldCloseSubscribeAndPublish() throws Exception
```
jfallows, line 132: `shouldDisconnectAfterSubscribeAndPublish`?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -223,7 +209,11 @@ private MqttPublishProxy(
             this.routedId = routedId;
             this.initialId = initialId;
             this.replyId = supplyReplyId.applyAsLong(initialId);
-            this.delegate = new KafkaProxy(originId, resolvedId, this);
+            this.keyBuffer = new UnsafeBuffer(new byte[bufferCapacity]);
+            this.headerIntValueBuffer = new UnsafeBuffer(new byte[bufferCapacity]);
+            this.kafkaHeadersBuffer = new UnsafeBuffer(new byte[bufferCapacity]);
```
jfallows, line 214: This is a lot of buffer allocation to use per stream, let's discuss.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -207,7 +231,13 @@ private MqttSubscribeProxy(
             this.routedId = routedId;
             this.initialId = initialId;
             this.replyId = supplyReplyId.applyAsLong(initialId);
-            this.delegate = new KafkaProxy(originId, resolvedId, this);
+            this.messagesSubscriptionIds = new IntArrayList();
+            this.retainedSubscriptionIds = new IntArrayList();
+            this.retainedSubscriptions = new ArrayList<>();
+            this.retainAsPublished = new Long2ObjectHashMap<>();
+            this.filtersBuffer = new UnsafeBuffer(new byte[bufferCapacity]);
```
jfallows, line 238: Let's review this per-stream buffer too.

jfallows (changes requested):

jfallows (approved):

PR #301, opened 2023-07-14
Mqtt kafka options

bmaidics:


incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaOptionsConfigAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+        OptionsConfig options)
+    {
+        MqttKafkaOptionsConfig mqttKafkaOptions = (MqttKafkaOptionsConfig) options;
+
+        JsonObjectBuilder object = Json.createObjectBuilder();
+
+        MqttKafkaTopicsConfig topics = mqttKafkaOptions.topics;
+
+        if (topics != null &&
+            !(TOPICS_DEFAULT.equals(topics)))
```
jfallows, line 68: Surprised this wasn't caught by checkstyle for redundant parens after `!`, please remove them. Also 3 more occurrences below.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -177,8 +177,9 @@ public MessageConsumer newStream(
 
         final MqttKafkaBindingConfig binding = supplyBinding.apply(routedId);
 
-        final MqttKafkaRouteConfig resolved = binding != null ? binding.resolve(authorization) : null;
 
+        final MqttKafkaRouteConfig resolved = binding != null ? binding.resolve(authorization) : null;
+        kafkaMessagesTopic = binding.messagesTopic();
```
jfallows, line 182: No need to store this in a field, as it would be overwritten by a different binding stream anyway, agree?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -161,8 +160,10 @@ public MessageConsumer newStream(
 
         final MqttKafkaBindingConfig binding = supplyBinding.apply(routedId);
 
-        final MqttKafkaRouteConfig resolved = binding != null ? binding.resolve(authorization) : null;
+        final MqttKafkaRouteConfig resolved = binding != null ?
+            binding.resolve(authorization) : null;
 
+        kafkaMessagesTopic = binding.messagesTopic();
```
jfallows, line 166: Same feedback here about storage of topic name.

jfallows (changes requested):

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
bmaidics: Yes, instead we should store it in the subscribeProxy. I think storing somehow is necessary, as the sending of the kafka begin happens in response to onMqttBegin and not at stream creation, when we can access the binding config.

PR #304, opened 2023-07-17
Mqtt kafka options

bmaidics:


incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaBindingConfig.java
```diff
@@ -45,4 +50,19 @@ public MqttKafkaRouteConfig resolve(
...
+
+    public String sessionsTopic()
+    {
+        return options.topics.sessions.asString();
+    }
+
+    public String retainedTopic()
+    {
+        return options.topics.retained.asString();
+    }
```
jfallows, line 67: Please return `StringFW` and avoid the call to `.asString()` here.

If we keep the return `String`, then we need to re-convert `String` back to `StringFW` for each usage, meaning that we are UTF-8 encoding the same string repeatedly, instead of just once during configuration parsing.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaOptionsConfigAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final String TOPICS_NAME = "topics";
+    private static final String SESSIONS_NAME = "sessions";
+    private static final String MESSAGES_NAME = "messages";
+    private static final String RETAINED_NAME = "retained";
+
+    private static final String8FW SESSIONS_DEFAULT = new String8FW("mqtt_sessions");
+    private static final String8FW MESSAGES_DEFAULT = new String8FW("mqtt_messages");
+    private static final String8FW RETAINED_DEFAULT = new String8FW("mqtt_retained");
+
+    public static final MqttKafkaTopicsConfig TOPICS_DEFAULT =
```
jfallows, line 39: If we don't reference this anywhere outside the class, then please make it `private` instead of `public`.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaTopicsConfig.java
```diff
@@ -0,0 +1,36 @@
...
+public class MqttKafkaTopicsConfig
+{
+    public final String8FW sessions;
+    public final String8FW messages;
+    public final String8FW retained;
+
+    public MqttKafkaTopicsConfig(
+        String8FW sessions,
+        String8FW messages,
+        String8FW retained)
```
jfallows, line 29: I think these should be `String16FW` instead of `String8FW` so that we can use them directly in the `kafka` extension metadata.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -206,6 +202,8 @@ private MqttPublishProxy(
             this.replyId = supplyReplyId.applyAsLong(initialId);
             this.messages = new KafkaMessagesProxy(originId, resolvedId, this);
             this.retained = new KafkaRetainedProxy(originId, resolvedId, this);
+            this.kafkaMessagesTopic = kafkaMessagesTopic;
+            this.kafkaRetainedTopic = kafkaRetainedTopic;
```
jfallows, line 206: Both `kafkaMessagesTopic` and `kafkaRetainedTopic` should probably be `String16FW` type instead, per comments above.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -235,6 +236,8 @@ private MqttSubscribeProxy(
             this.retainAsPublished = new Long2ObjectHashMap<>();
             this.messages = new KafkaMessagesProxy(originId, resolvedId, this);
             this.retained = new KafkaRetainedProxy(originId, resolvedId, this);
+            this.kafkaMessagesTopic = kafkaMessagesTopic;
+            this.kafkaRetainedTopic = kafkaRetainedTopic;
```
jfallows, line 240: Perhaps we should pass these topic names into the `KafkaMessagesProxy` and `KafkaRetainProxy`, then just call the field `topic` and avoid passing it to methods on the `KafkaXxxProxy` classes?

jfallows (changes requested):

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -173,15 +169,16 @@ public MessageConsumer newStream(
...
+            binding.resolve(authorization) : null;
 
         MessageConsumer newStream = null;
 
         if (resolved != null)
         {
             final long resolvedId = resolved.id;
-            newStream = new MqttSubscribeProxy(mqtt, originId, routedId, initialId, resolvedId)::onMqttMessage;
+            newStream = new MqttSubscribeProxy(mqtt, originId, routedId, initialId, resolvedId,
+                binding.messagesTopic(), binding.retainedTopic())::onMqttMessage;
```
jfallows, line 181: Please hoist these into local variables for readability.

incubator/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaOptionsConfigAdapterTest.java
```diff
@@ -0,0 +1,87 @@
...
+                "}";
+
+        MqttKafkaOptionsConfig options = jsonb.fromJson(text, MqttKafkaOptionsConfig.class);
+
+        assertThat(options, not(nullValue()));
+        assertThat(options.topics, not(nullValue()));
+        assertThat(options.topics.sessions.asString(), equalTo("mqtt_sessions"));
+        assertThat(options.topics.messages.asString(), equalTo("mqtt_messages"));
+        assertThat(options.topics.retained.asString(), equalTo("mqtt_retained"));
+    }
```
jfallows, line 63: Suggest having two tests, both `shouldReadOptionsWithDefaults` and `shouldReadOptions`, where `shouldReadOptionsWithDefaults()` parses empty options and has default values, then `shouldReadOptions()` parses explicit (non-default) values for each property and verifies those non-default values are returned.

jfallows (changes requested):

jfallows (approved):

PR #307, opened 2023-07-19
MQTT guard implementation

bmaidics:


incubator/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/config/server.authorization.password.credentials.yaml
```diff
@@ -0,0 +1,39 @@
...
+    options:
+      authorization:
+        test0:
+          credentials:
+            connect:
+              password: Bearer {credentials}
+    routes:
+      - exit: app0
+        guarded:
+          test0: []
```
jfallows, line 39: Lets rename this file to `server.credentials.password.yaml`.

incubator/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/config/server.authorization.username.credentials.yaml
```diff
@@ -0,0 +1,39 @@
...
+    options:
+      authorization:
+        test0:
+          credentials:
+            connect:
+              username: Bearer {credentials}
+    routes:
+      - exit: app0
+        guarded:
+          test0: []
```
jfallows, line 39: Lets rename this file to `server.credentials.username.yaml`.

incubator/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/schema/mqtt.schema.patch.json
```diff
@@ -32,7 +32,63 @@
...
+                        {
+                            "authorization":
+                            {
+                                "title": "Authorizations",
+                                "type": "object",
+                                "patternProperties":
+                                {
+                                    "^[a-zA-Z]+[a-zA-Z0-9\\._\\-]*$":
+                                    {
+                                        "title": "Authorization",
```
jfallows, line 47: `Guard` ?

incubator/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/schema/mqtt.schema.patch.json
```diff
@@ -32,7 +32,63 @@
...
+                                                {
+                                                    "connect":
+                                                    {
+                                                        "title": "Connect",
+                                                        "type": "object",
+                                                        "additionalProperties":
+                                                        {
+                                                            "type": "string",
+                                                            "pattern": ".*\\{credentials\\}.*"
+                                                        }
```
jfallows, line 65: This is not actually freeform, should be explicitly `username` or `password`, right?

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttOptionsConfigAdapter.java
```diff
@@ -0,0 +1,145 @@
...
+                JsonObject authorization = authorizations.getJsonObject(name);
+
+                MqttCredentialsConfig newCredentials = null;
+
+                JsonObject credentials = authorization.getJsonObject(AUTHORIZATION_CREDENTIALS_NAME);
+
+                if (credentials != null)
+                {
+                    List<MqttPatternConfig> newConnect =
+                        adaptPatternFromJson(credentials, AUTHORIZATION_CREDENTIALS_CONNECT_NAME);
```
jfallows, line 113: This needs to be locked down to `username` or `password` as well, agree?

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1670,9 +1677,35 @@ else if (length > MAXIMUM_CLIENT_ID_LENGTH)
...
                 keepAliveTimeout = Math.round(TimeUnit.SECONDS.toMillis(keepAlive) * 1.5);
                 doSignalKeepAliveTimeout();
+
+                long exchangeAuth = authorization;
+                if (guard != null)
+                {
+                    final String username = payload.username != null ? payload.username.asString() : null;
+                    final String password = payload.password != null ?
+                        payload.password.bytes().get((b, o, m) -> b.getStringWithoutLengthUtf8(o, m - o)) : null;
+                    final String credentialsMatch = credentials.apply(username, password);
```
jfallows, line 1687: We know it is just one or the other based on the options config, so seems like this should be directly targeted at the correct connect property, not checking to see if they are set or not, agree?

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1670,9 +1677,35 @@ else if (length > MAXIMUM_CLIENT_ID_LENGTH)
                 serverDefinedKeepAlive = keepAlive != connect.keepAlive();
                 keepAliveTimeout = Math.round(TimeUnit.SECONDS.toMillis(keepAlive) * 1.5);
                 doSignalKeepAliveTimeout();
+
+                long exchangeAuth = authorization;
```
jfallows, line 1681: The term `exchangeAuth` is an `http` term for the `http` `exchange`, i.e. the `http` `request-response`.
Let's rename this to `sessionAuth` or similar to refer to the `mqtt` concept instead.

jfallows (changes requested):

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttAuthorizationConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+        public MqttPatternConfig(
+            String name,
+            String pattern)
+        {
+            this.name = name;
+            this.pattern = pattern;
+        }
+    }
+}
```
jfallows, line 61: I think this will change after you lock down the schema to `username` or `password` explicitly.

Presumably, there would be an `enum` here indicating `username` or `password`, then a value for the pattern.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttAuthorizationConfig.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package io.aklivity.zilla.runtime.binding.mqtt.internal.config;
+
+import java.util.List;
+import java.util.function.Function;
+
+public final class MqttAuthorizationConfig
+{
+    public static final Function<String, String> DEFAULT_CREDENTIALS = x -> null;
```
jfallows, line 24: This constant belongs closer to the place it is used, potentially private too.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttOptionsConfigAdapter.java
```diff
@@ -0,0 +1,149 @@
...
+import io.aklivity.zilla.runtime.engine.config.OptionsConfig;
+import io.aklivity.zilla.runtime.engine.config.OptionsConfigAdapterSpi;
+
+public class MqttOptionsConfigAdapter implements OptionsConfigAdapterSpi, JsonbAdapter<OptionsConfig, JsonObject>
+{
+    private static final String AUTHORIZATION_NAME = "authorization";
+    private static final String AUTHORIZATION_CREDENTIALS_NAME = "credentials";
+    private static final String AUTHORIZATION_CREDENTIALS_CONNECT_NAME = "connect";
+    public static final String AUTHORIZATION_CREDENTIALS_USERNAME_NAME = "username";
+    public static final String AUTHORIZATION_CREDENTIALS_PASSWORD_NAME = "password";
```
jfallows, line 38: Please make these fields private.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1210,6 +1213,9 @@ private final class MqttServer
         private final Int2ObjectHashMap<String> topicAliases;
         private final Int2IntHashMap subscribePacketIds;
         private final Object2IntHashMap<String> unsubscribePacketIds;
+        private final GuardHandler guard;
+        private final Function<String, String> credentials;
+        private final String authFieldName;
```
jfallows, line 1218: This will become the same `enum` as mentioned above for `MqttCredentialsConfig`.

jfallows (changes requested):

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttAuthorizationConfig.java
```diff
@@ -0,0 +1,94 @@
...
+            case "username":
+            {
+                return USERNAME;
+            }
+            case "password":
+            {
+                return PASSWORD;
+            }
+            }
+            return null;
```
jfallows, line 76: Please eliminate the nested parens for each `case` and change this to single return.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttAuthorizationConfig.java
```diff
@@ -0,0 +1,94 @@
...
+        {
+            this.name = name;
+        }
+
+        public String getName()
+        {
+            return name;
+        }
+
+        public static MqttAuthField fromString(
```
jfallows, line 62: Please rename to `ofName`.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttAuthorizationConfig.java
```diff
@@ -0,0 +1,94 @@
...
+        public MqttCredentialsConfig(
+            List<MqttPatternConfig> connect)
+        {
+            this.connect = connect;
+        }
+    }
+
+    public enum MqttAuthField
+    {
+        USERNAME("username"), PASSWORD("password");
```
jfallows, line 47: Please put different enum constants on different lines.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttAuthorizationConfig.java
```diff
@@ -0,0 +1,94 @@
...
+
+        private final String name;
+
+        MqttAuthField(
+            String name)
+        {
+            this.name = name;
+        }
+
+        public String getName()
```
jfallows, line 57: Please rename to `name()`.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -468,6 +469,7 @@ public MessageConsumer newStream(
             final long budgetId = supplyBudgetId.getAsLong();
 
             newStream = new MqttServer(
+                binding,
```
jfallows, line 472: We don't typically pass the binding into the stream implementation, instead we keep these separate and extract what we need from the binding then pass to the constructor.

jfallows (changes requested):

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttAuthorizationConfig.java
bmaidics: We can't rename it to name(), as the enum already has a method, which returns the name of the enum (in our case USERNAME or PASSWORD with capital letters)

jfallows: Ah yes, of course. Any reason we cannot go with vanilla enum and handle the conversion to and from lowercase in the parser?

bmaidics: Of course we can resolve it in the parser. Why didn't I think of that? 😄

jfallows: Ah yes, of course. Any reason we cannot go with vanilla enum and handle the conversion to and from lowercase in the parser?

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttAuthorizationConfig.java
```diff
@@ -0,0 +1,94 @@
...
+                field = PASSWORD;
+                break;
+            }
+            return field;
+        }
+    }
+
+    public static final class MqttPatternConfig
+    {
+        public final MqttAuthField field;
```
jfallows, line 82: After renaming the enum class, suggest renaming `field` to `property`.

PR #311, opened 2023-07-20
API abstraction for SchemaRegistry in Zilla

### LINKED ISSUE #310: Design `zilla.yaml` configuration syntax for schema types
This abstraction will need to support; inline, karapace, confluent schema registry, aws glue, buf schema registry, etc

However, this task deals only with the generic schema types abstraction in the configuration, other implementation-specific details will be handled in later tasks.

aDaemonThread:
fixes #310


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/schema/SchemaHandler.java
```diff
@@ -0,0 +1,29 @@
...
+{
+    void configure();
+
+    void schemaRegistryClient();
+
+    int registerSchema(String schemaString, String subject, int version, int id);
+
+    <T> T getSchemasBySubjectAndId(String subject, int id);
+
+    <T> T getSchemasById(int id);
```
jfallows, line 28: I think we want to simplify this a bit.

`configure()` seems unnecessary, as we have `attach` and `detach` lifecycle on `SchemaContext` to help with any initialization or configuration based on parsed `options` or even implementation-specific `Configuration`.

`schemaRegistryClient()` seems unnecessary too.

The general mapping we discussed was `schemaId` => some kind of "schema description" that could be used to by caller to validate, encode or decode.

We also discussed a mapping from `schema name` (a.k.a `subject name`) plus optional `version` to `schemaId`.

`registerSchema` could probably be renamed to `register` (since the context of "schema" is already clear) and simplified; "schema / subject name" + "new schema description" => `schemaId`

Let me know if this is sufficient to handle all the use cases.

How do we want to represent the "schema description" and do we need more than one format?
How would a validator / encoder / decoder make use of such a "schema description"?

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/schema/TestSchemaFactorySpi.java
```diff
@@ -0,0 +1,36 @@
...
+    public String name()
+    {
+        return null;
+    }
+
+    @Override
+    public Schema create(
+        Configuration config)
+    {
+        return null;
```
jfallows, line 34: Does this need to return a `TestSchema`?

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/schema/TestSchemaHandler.java
```diff
@@ -0,0 +1,62 @@
...
+        return null;
+    }
+
+    @Override
+    public <T> T getSchemasById(
+        int id)
+    {
+        return null;
+    }
+}
```
jfallows, line 62: This needs some test behavior for us to try out test validators / encoders / decoders, say in the `kafka` binding for example.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/schema/config/TestSchemaOptionsConfigAdapter.java
```diff
@@ -0,0 +1,78 @@
...
+import jakarta.json.JsonObjectBuilder;
+
+import io.aklivity.zilla.runtime.engine.config.OptionsConfig;
+import io.aklivity.zilla.runtime.engine.config.OptionsConfigAdapterSpi;
+
+public class TestSchemaOptionsConfigAdapter implements OptionsConfigAdapterSpi
+{
+    private static final String HOST = "schema.com";
+    private static final String PORT = "8081";
+    private static final String CONTEXT = "default";
```
jfallows, line 29: Does the `test` `schema` need options?

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/engine.schema.json
```diff
@@ -70,6 +70,19 @@
             },
             "additionalProperties": false
         },
+        "schemas":
```
jfallows, line 73: I'm not sure if `schemas` is actually the right word here, since `schema` actually represents the types. not the system holding them.

The closest I have found for a collection of schemas is `catalog`.
https://stackoverflow.com/questions/7022755/whats-the-difference-between-a-catalog-and-a-schema-in-a-relational-database/7023159#7023159

This would give us `SchemaRegistryCatalog`, etc as we provide implementations.

What do you think?

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/DispatchAgent.java
```diff
@@ -765,6 +776,14 @@ public void onExporterDetached(
...
     }
 
+    @Override
+    public CatalogHandler supplyCatalog(
+        long catalogId)
+    {
+        CatalogRegistry catalog = configuration.resolveCatalog(catalogId);
+        return catalog != null ? catalog.handler() : null;
+    }
+
```
jfallows, line 786: Is this needed?

Note: we have a special case for some of the metrics related aspects to support `zilla metrics` command, which may not apply to catalogs.

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/engine.schema.json
```diff
@@ -70,15 +70,15 @@
             },
             "additionalProperties": false
         },
-        "schemas":
+        "catalog":
```
jfallows, line 73: Should this be `catalogs`? (plural)

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/engine.schema.json
```diff
@@ -70,15 +70,15 @@
...
         {
-            "title": "Schemas",
+            "title": "Catalog",
```
jfallows, line 75: If plural above then this title would be `Catalogs`.

jfallows (changes requested):

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/config/client.options.catalog.yaml
```diff
@@ -0,0 +1,38 @@
...
+    type: kafka
+    kind: client
+    options:
+      catalog:
+        test0:
+          topic: test
+          key:
+            strategy: topic
+            version: latest
+            id: 1
```
jfallows, line 30: suggest we want `strategy` + (optional) `version` OR `id` (not both), for `key` and `value`, agree?
the rationale is that `strategy` + `version` is used to resolve the `id`, or else we can specify the `id` explicitly

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/fetch/message.value.avro/client.rpt
```diff
@@ -0,0 +1,76 @@
...
+                                  .build()
+                              .build()}
+
+read zilla:data.ext ${kafka:matchDataEx()
+                             .typeId(zilla:id("kafka"))
+                             .fetch()
+                                 .partition(0, 10, 10)
+                                 .build()
+                             .build()}
+read [0x00 0x00 0x00 0x00 0x09 0x06 0x69 0x64 0x30 0x10 0x70 0x6f 0x73 0x69 0x74 0x69 0x76 0x65]
```
jfallows, line 76: Suggest the following to increase readability
```
read [0x00] ${kafka:varint(3)} "id0" ${kafka:varint(8)} "positive"
```
Please also make corresponding changes to `write ...` in server script.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/network/fetch.v5/message.value.avro/client.rpt
```diff
@@ -0,0 +1,134 @@
...
+     -1s
+     -1
+     1                                  # records
+     ${kafka:varint(18)}
+     [0x00]
+     ${kafka:varint(0)}
+     ${kafka:varint(0)}
+     ${kafka:varint(-1)}
+     ${kafka:varint(18)}
+     [0x00 0x00 0x00 0x00 0x09 0x06 0x69 0x64 0x30 0x10 0x70 0x6f 0x73 0x69 0x74 0x69 0x76 0x65]
```
jfallows, line 133: Update this too for readability.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/network/fetch.v5/message.value.avro/server.rpt
```diff
@@ -0,0 +1,128 @@
...
+      -1s
+      -1
+      1                                 # records
+      ${kafka:varint(18)}
+      [0x00]
+      ${kafka:varint(0)}
+      ${kafka:varint(0)}
+      ${kafka:varint(-1)}
+      ${kafka:varint(18)}
+      [0x00 0x00 0x00 0x00 0x09 0x06 0x69 0x64 0x30 0x10 0x70 0x6f 0x73 0x69 0x74 0x69 0x76 0x65]
```
jfallows, line 127: Update this too for readability.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -82,6 +82,52 @@
...
+                                    {
+                                        "topic":
+                                        {
+                                            "type": "string"
+                                        },
+                                        "key":
+                                        {
+                                            "strategy":
+                                            {
+                                                "type": "string"
```
jfallows, line 101: Please use an enum.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -82,6 +82,52 @@
...
+                                            },
+                                            "version":
+                                            {
+                                                "type": "string"
```
jfallows, line 105: Default to `latest`?

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -82,6 +82,52 @@
...
+                                            },
+                                            "id":
+                                            {
+                                                "type": "integer"
+                                            }
```
jfallows, line 110: Require `strategy` + (optional) `version` to resolve the `id` dynamically or fixed `id`.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -82,6 +82,52 @@
...
```
jfallows, line 125: Same as above.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/catalog/CatalogHandler.java
```diff
@@ -0,0 +1,27 @@
...
+
+public interface CatalogHandler
+{
+    <T> T register(String schemaDefinition, String subject, String dataFormat);
+
+    String getSchemasById(String schemaId);
+
+    String getSchemaBySubject(String subject);
+
+    String getVersionSchema(String subject, String version);
```
jfallows, line 26: `schemaId` should be an `int`, right?

The purpose of register is to create an official `id` for the new schema, so:
```
  int register(String name, String schema, String format);
```
Potentially don't need format if it can be discerned from the schema definition.

The get-style methods don't fit with the rest of our internal APIs, suggest we use `resolve` instead.
```
  String resolve(int schemaId);
  int resolve(String name, String version); // default version is "latest"
```

So we likely only need 3 methods here.

jfallows (changes requested):
Overall, it looks good - but we need to verify the abstraction with a Kafka-side implementation that makes use of the CatalogHandler API.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaTopicCatalogConfig.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.aklivity.zilla.runtime.binding.kafka.internal.config;
+
+public class KafkaTopicCatalogConfig
```
jfallows, line 18: Suggest `KafkaCatalogTopicConfig` instead, so all `KafkaCatalog...` types are aligned.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaTopicCatalogConfigHandler.java
```diff
@@ -0,0 +1,87 @@
...
+{
+    private static final String NAME_NAME = "name";
+    private static final String CATALOG_KEY = "key";
+    private static final String CATALOG_VALUE = "value";
+
+    private final KafkaSerDeConfigHandler schemaConfigHandler = new KafkaSerDeConfigHandler();
+
+    @Override
+    public JsonObject adaptToJson(
+        KafkaTopicCatalogConfig topicCatalogConfig)
```
jfallows, line 33: Just `kafkaTopic`

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -290,7 +292,8 @@ public MessageConsumer newStream(
                     leaderId,
                     initialOffset,
                     isolation,
-                    sasl)::onApplication;
+                    sasl,
+                    catalog)::onApplication;
```
jfallows, line 296: Unlikely that we would put the config so deep into the stream, instead it would probably be converted to behavior first, such as `Validator`.

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/ClientFetchIT.java
```diff
@@ -283,6 +283,16 @@ public void shouldReceiveMessageValueDistinct() throws Exception
...
+    /*@Test
+    @Configuration("client.options.catalog.yaml")
+    @Specification({
+        "${app}/message.value.avro/client",
+        "${net}/message.value.avro/server"})
+    public void shouldReceiveMessageValueAvro() throws Exception
+    {
+        k3po.finish();
+    }*/
+
```
jfallows, line 295: Uncomment?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/catalog/CatalogHandler.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.aklivity.zilla.runtime.engine.catalog;
+
+public interface CatalogHandler
+{
+    <T> T register(String schema, String subject);
```
jfallows, line 20: I think this `<T> T` doesn't quite work, we need to decide what the return type should be here.

It will probably become more obvious as we work through the next step, such as validators.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/StringValidator.java
```diff
@@ -0,0 +1,58 @@
...
+    {
+        this.encoding = config.encoding();
+        this.decoder = encoding.newDecoder();
+    }
+
+    @Override
+    public boolean validate(
+        OctetsFW payload)
+    {
+        boolean valid = false;
```
jfallows, line 42: Is `null` considered a valid `string`?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/StringValidator.java
```diff
@@ -0,0 +1,58 @@
...
+public class StringValidator implements Validator
+{
+    private Charset encoding;
+    private CharsetDecoder decoder;
+
+    public StringValidator(
+        ValidatorConfig config)
+    {
+        this.encoding = config.encoding();
+        this.decoder = encoding.newDecoder();
```
jfallows, line 35: Ideally, one instance of `CharsetDecoder` per `Charset` per core, so safe to share instance.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/StringValidator.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public boolean validate(
+        OctetsFW payload)
```
jfallows, line 40: Parameter should probably be called `data` or `value` instead of `payload`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/StringValidator.java
```diff
@@ -0,0 +1,58 @@
...
+    {
+        boolean valid = false;
+        if (payload != null)
+        {
+            byte[] payloadBytes = new byte[payload.sizeof()];
+            payload.value().getBytes(0, payloadBytes);
+            try
+            {
+                decoder.decode(ByteBuffer.wrap(payloadBytes));
```
jfallows, line 49: You can minimize the allocations by checking the following:
```
DirectBuffer buffer = payload.value();
ByteBuffer byteBuf = buffer.byteBuffer();
byte[] byteArray = buffer.byteArray();

ByteBuffer sliceBuf;
if (byteArray != null)
{
    int offset = buffer.addressOffset() - UnsafeBuffer.ARRAY_BASE_OFFSET;
    sliceBuf = ByteBuffer.wrap(byteArray, offset, buffer.capacity());
}
else
{
    int offset = buffer.addressOffset() - BufferUtil.address(byteBuf);
    sliceBuf = byteBuf.slice(offset, buffer.capacity());
}

try
{
     decoder.decode(sliceBuf);
}
catch (CharacterCodingException ex)
{
      valid = false;
}
```

Note: there is a more optimal approach for validation, say for UTF-8, where we check the bit mask of each byte for valid UTF-8 character sequences.

jfallows (changes requested):

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -78,6 +78,172 @@
...
+                                                        "json",
+                                                        "json_decoder",
+                                                        "json_schema",
+                                                        "list",
+                                                        "long",
+                                                        "short",
+                                                        "string",
+                                                        "uuid",
+                                                        "void",
+                                                        "wrapper_key"
```
jfallows, line 110: Some of these type only seem to be relevant for cases where application code would observe the decoded value via Java API for example. However, we are implementing this inside `zilla` so `byte_array` would seem indistinguishable from `bytes`, i.e. just binary data.

Please remove any of these types that do not have distinct meaning for the encoded representation.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -78,6 +78,172 @@
...
```
jfallows, line 193: Same feedback as above.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/config/cache.options.type.avro.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+---
+name: test
+catalogs:
+  test0:
+    type: test
+    options:
+      host: localhost
+      port: 8081
+      context: default
```
jfallows, line 25: These options for `test` catalog don't seem to be relevant.

Perhaps we should remove them.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogHandler.java
```diff
@@ -0,0 +1,52 @@
...
+        return null;
+    }
+
+    @Override
+    public String resolve(
+        String schemaId)
+    {
+        return "\"{\"fields\":[{\"name\":\"id\",\"type\":\"string\"}," +
+                "{\"name\":\"status\",\"type\":\"string\"}],\"name\":\"Event\"," +
+                "\"namespace\":\"io.aklivity.example\",\"type\":\"record\"}\"";
```
jfallows, line 50: Maybe we could let this be configured in yaml with options `subject`, `version`, `id` and `schema`?

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogHandler.java
```diff
@@ -0,0 +1,52 @@
...
+    public String resolve(
+        String subject,
+        String version)
+    {
+        return null;
+    }
+
+    @Override
+    public String resolve(
+        String schemaId)
```
jfallows, line 46: Shouldn't schemaId be `int`, also for the return value from `resolve(subject, version)` above?

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaOptionsConfigAdapterTest.java
```diff
@@ -82,14 +82,20 @@ public void shouldWriteOptions()
     {
         KafkaOptionsConfig options = new KafkaOptionsConfig(
                 singletonList("test"),
-                singletonList(new KafkaTopicConfig("test", LIVE, JSON_PATCH)),
+                singletonList(new KafkaTopicConfig("test", LIVE, JSON_PATCH, new KafkaTopicKeyValueConfig(
+               "string", "utf_8", singletonList(new KafkaCatalogConfig("test0", "topic", "latest", 0))),
+                    new KafkaTopicKeyValueConfig("avro", null, singletonList(new KafkaCatalogConfig(
+              "test0", "topic", "latest", 0))))),
```
jfallows, line 88: Indentation is slightly off.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/Validator.java
```diff
@@ -0,0 +1,24 @@
...
+ * under the License.
+ */
+package io.aklivity.zilla.runtime.binding.kafka.internal.validator;
+
+import io.aklivity.zilla.runtime.binding.kafka.internal.types.OctetsFW;
+
+public interface Validator
+{
+    boolean validate(
+        OctetsFW data);
```
jfallows, line 23: This might be better as `DirectBuffer`, `index` and `length`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaTopicValidators.java
```diff
@@ -0,0 +1,32 @@
...
+package io.aklivity.zilla.runtime.binding.kafka.internal.config;
+
+import io.aklivity.zilla.runtime.binding.kafka.internal.validator.Validator;
+
+public class KafkaTopicValidators
+{
+    public final Validator key;
+    public final Validator value;
+
+    public KafkaTopicValidators(
```
jfallows, line 25: Suggest rename to `KafkaTopicType`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/StringValidator.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    public boolean validate(
+        OctetsFW data)
+    {
+        boolean valid = true;
+        if (data != null)
+        {
+            byte[] payloadBytes = new byte[data.sizeof()];
+            data.value().getBytes(0, payloadBytes);
```
jfallows, line 39: I think we have proposed code in other feedback comments for getting the byteArray vs byteBuffer and using those directly instead of allocating a byte[] to make a copy, right?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaBindingConfig.java
```diff
@@ -73,4 +90,37 @@ public KafkaSaslConfig sasl()
...
     }
+
+    public KafkaDeltaType supplyDeltaType(
+        String topic,
+        KafkaDeltaType deltaType)
+    {
+        return topic != null &&
+                options != null &&
+                options.topics != null
+                    ? options.topics.stream().filter(t -> topic.equals(t.name)).findFirst().orElse(null).deltaType
```
jfallows, line 101: `.orElse(null).deltaType` will throw NPE, agree?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaBindingConfig.java
```diff
@@ -38,6 +48,13 @@ public KafkaBindingConfig(
         this.kind = binding.kind;
         this.options = KafkaOptionsConfig.class.cast(binding.options);
         this.routes = binding.routes.stream().map(KafkaRouteConfig::new).collect(toList());
+        this.validatorsMap = options != null &&
+                options.topics != null
+                    ? options.topics.stream()
+                    .collect(Collectors.toMap(t -> t.name, t -> new KafkaTopicValidators(
+                        t.key != null ? validator.create(t.key.type, supplyValidatorConfig(t.key.type, t.key)) : null,
+                        t.value != null ? validator.create(t.value.type, supplyValidatorConfig(t.value.type, t.value)) : null
+                        ))) : null;
```
jfallows, line 57: Indentation is slightly off.

jfallows (changes requested):

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -78,6 +78,166 @@
...
+                                            "properties":
+                                            {
+                                                "type":
+                                                {
+                                                    "title": "Type",
+                                                    "type": "string",
+                                                    "enum":
+                                                    [
+                                                        "byte_buffer",
+                                                        "bytes",
```
jfallows, line 95: What is the difference in validation between `byte_buffer` and `bytes`?
(also in duplicated enum below)

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -78,6 +78,166 @@
...
+                                                        "double",
+                                                        "float",
+                                                        "integer",
+                                                        "avro",
+                                                        "json",
+                                                        "json_decoder",
```
jfallows, line 101: What is the difference in validation between `json_decoder` and `json`?

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -78,6 +78,166 @@
...
+                                                        "list",
+                                                        "long",
+                                                        "short",
+                                                        "string",
+                                                        "uuid",
+                                                        "void"
```
jfallows, line 107: What does `void` validation mean?
(also in duplicated enum below)

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
jfallows: We need some validation here, since it only makes sense to specify `id` by itself, and not alongside `strategy` or `version`.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/fetch/message.value.avro.invalid/client.rpt
```diff
@@ -0,0 +1,76 @@
...
+                                  .build()
+                              .build()}
+
+read zilla:data.ext ${kafka:matchDataEx()
+                             .typeId(zilla:id("kafka"))
+                             .fetch()
+                                 .partition(0, 10, 10)
+                                 .build()
+                             .build()}
+read [0x00 0x00 0x00 0x00 0x09] ${kafka:varint(3)} "id0"
```
jfallows, line 76: Suggest changing this to better represent the logical meaning of the physical bytes.
```
read [0x00] 0x09 ${kafka:varint(3)} "id0"
```

Same for other scripts.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogHandler.java
```diff
@@ -0,0 +1,54 @@
...
+import io.aklivity.zilla.runtime.engine.test.internal.catalog.config.TestCatalogConfig;
+
+public class TestCatalogHandler implements CatalogHandler
+{
+    private final String schema;
+
+    public TestCatalogHandler(
+        CatalogConfig options)
+    {
+        this.schema = TestCatalogConfig.class.cast(options.options).schema;
```
jfallows, line 29: Please move this cast to the caller and change constructor parameter type to `TestCatalogConfig`.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogContext.java
```diff
@@ -0,0 +1,42 @@
...
+    public TestCatalogContext(
+        EngineContext context)
+    {
+    }
+
+    @Override
+    public CatalogHandler attach(
+        CatalogConfig catalog)
+    {
+        return new TestCatalogHandler(catalog);
```
jfallows, line 34: Please apply cast here instead of inside constructor.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/config/StringValidatorConfigAdapter.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public JsonObject adaptToJson(
+        ValidatorConfig config)
+    {
+        StringValidatorConfig validatorConfig = (StringValidatorConfig) config;
+        JsonObjectBuilder validator = Json.createObjectBuilder();
+
+        if (validatorConfig.encoding != null &&
+                !validatorConfig.encoding.isEmpty())
```
jfallows, line 35: Indentation

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/config/StringValidatorConfigAdapter.java
```diff
@@ -0,0 +1,59 @@
...
+package io.aklivity.zilla.runtime.binding.kafka.internal.validator.config;
+
+import jakarta.json.Json;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonObjectBuilder;
+import jakarta.json.bind.adapter.JsonbAdapter;
+
+public final class StringValidatorConfigAdapter implements ValidatorConfigAdapterSpi, JsonbAdapter<ValidatorConfig, JsonObject>
+{
+    private static final String ENCODING = "encoding";
```
jfallows, line 25: `ENCODING_NAME`

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/config/AvroValidatorConfig.java
```diff
@@ -0,0 +1,32 @@
...
+
+public final class AvroValidatorConfig extends ValidatorConfig
+{
+    public final List<KafkaCatalogConfig> catalogList;
+
+    public AvroValidatorConfig(
+        List<KafkaCatalogConfig> catalogList)
+    {
+        super("avro");
+        this.catalogList = catalogList;
```
jfallows, line 30: `catalogs`

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/StringValidator.java
```diff
@@ -0,0 +1,91 @@
...
+        int length)
+    {
+        byte[] payloadBytes = new byte[length];
+        data.getBytes(0, payloadBytes);
+        switch (encoding)
+        {
+        case "utf_8":
+            return isValidUTF8(payloadBytes);
+        }
+        return false;
```
jfallows, line 45: Use single return.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/StringValidator.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+package io.aklivity.zilla.runtime.binding.kafka.internal.validator;
+
+import org.agrona.DirectBuffer;
+
+import io.aklivity.zilla.runtime.binding.kafka.internal.validator.config.StringValidatorConfig;
+
+public final class StringValidator implements Validator
+{
+    private String encoding;
```
jfallows, line 24: `final`

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/StringValidator.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public StringValidator(
+        StringValidatorConfig config)
+    {
+        this.encoding = config.encoding;
```
jfallows, line 29: Perhaps change field type to `Predicate<byte[]>` and select `this::isValidUTF8` for `utf-8` with a default implementation for other charsets.

Then `validate` method would be something like
```
byte[] array = new byte[length];
data.getBytes(0, array);
return predicate.test(array);
```

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/LongValidator.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public boolean validate(
+        DirectBuffer data,
+        int index,
+        int length)
+    {
+        byte[] payloadBytes = new byte[length];
+        data.getBytes(0, payloadBytes);
+        return payloadBytes.length == 8 ? true : false;
```
jfallows, line 37: Use `length == 8` instead to remove need for `byte[]` allocation and copy.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/IntegerValidator.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public boolean validate(
+        DirectBuffer data,
+        int index,
+        int length)
+    {
+        byte[] payloadBytes = new byte[length];
+        data.getBytes(0, payloadBytes);
+        return payloadBytes.length == 4 ? true : false;
```
jfallows, line 36: Use `length == 4` instead to remove need for `byte[]` allocation and copy.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/AvroValidator.java
```diff
@@ -0,0 +1,92 @@
...
+    public AvroValidator(
+        AvroValidatorConfig config,
+        ToLongFunction<String> resolveId,
+        LongFunction<CatalogHandler> supplyCatalog)
+    {
+        this.handlersById = new Long2ObjectHashMap<>();
+        this.decoder = DecoderFactory.get();
+        this.catalogs = config.catalogList.stream().map(c ->
+        {
+            c.catalogId = resolveId.applyAsLong(c.name);
```
jfallows, line 53: I see now why we still pass `resolveId` down the stack.

I suspect this would only go away if we can move `validators` to a first class concept in the config, managed by the engine, so it can handle this for us.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/IntegerValidator.java
```diff
@@ -31,8 +31,6 @@ public boolean validate(
         int index,
         int length)
     {
-        byte[] payloadBytes = new byte[length];
-        data.getBytes(0, payloadBytes);
-        return payloadBytes.length == 4 ? true : false;
+        return length == 4 ? true : false;
```
jfallows, line 34: No need for ternary operator, since `length == 4` expression is already `boolean`.
So just need `return length == 4;`.
Same feedback for other similar validators.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/StringValidator.java
```diff
@@ -15,18 +15,20 @@
...
 public final class StringValidator implements Validator
 {
-    private String encoding;
+    private Predicate<byte[]> predicate;
 
     public StringValidator(
         StringValidatorConfig config)
     {
-        this.encoding = config.encoding;
+        this.predicate = config.encoding.equals("utf_8") ? this::isValidUTF8 : bytes -> false;
```
jfallows, line 31: If we only currently support utf8 charset, then please make sure the schema validation only accepts utf8 as a valid encoding, otherwise it will look like we support other encodings, but the implementation will always reject non-utf8.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogContext.java
```diff
@@ -31,7 +32,7 @@ public TestCatalogContext(
     public CatalogHandler attach(
         CatalogConfig catalog)
     {
-        return new TestCatalogHandler(catalog);
+        return new TestCatalogHandler(TestCatalogConfig.class.cast(catalog.options));
```
jfallows, line 35: Please rename `TestCatalogConfig` to `TestCatalogOptionsConfig`.
Feel free to introduce a `TestCatalogConfig` similar to `TestGuardConfig` if needed.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/AvroValidator.java
```diff
@@ -0,0 +1,92 @@
...
+
+        int schemaId = byteBuf.getInt();
+        int valLength = length - 1 - 4;
+        byte[] valBytes = new byte[valLength];
+        data.getBytes(length - valLength, valBytes);
+
+        try
+        {
+            reader = new GenericDatumReader(new Schema.Parser().parse(
+                handlersById.get(catalogs.get(0).id).resolve(schemaId)));
```
jfallows, line 83: Can some of this be hoisted to the constructor to avoid repeated instantiation per method call?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/AvroValidator.java
```diff
@@ -0,0 +1,92 @@
...
+            System.out.println("Unknown magic byte!");
+            return false;
+        }
+
+        int schemaId = byteBuf.getInt();
+        int valLength = length - 1 - 4;
+        byte[] valBytes = new byte[valLength];
+        data.getBytes(length - valLength, valBytes);
+
+        try
```
jfallows, line 80: Let's use single return for the method.
Suggest labeling the `try` block, then moving the code above inside, and `break <label>` instead of `return ...`, etc.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheServerFactory.java
```diff
@@ -47,6 +49,8 @@ public final class KafkaCacheServerFactory implements KafkaStreamFactory
     private final Int2ObjectHashMap<BindingHandler> factories;
     private final Long2ObjectHashMap<KafkaBindingConfig> bindings;
     private final KafkaCacheServerAddressFactory cacheAddressFactory;
+    private final LongFunction<CatalogHandler> supplyCatalog;
+    private final ValidatorFactory validator;
```
jfallows, line 53: This is a validator factory, not an individual validator, so it might be better to name it `validators` or `validatorFactory` instead.

Later this will move to the `engine` and provided via a method on `EngineContext` instead of instantiating the factory directly.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFactory.java
```diff
@@ -93,13 +97,15 @@ public KafkaCacheClientFactory(
...
+        this.supplyCatalog = context::supplyCatalog;
+        this.validator = ValidatorFactory.instantiate();
     }
 
     @Override
     public void attach(
         BindingConfig binding)
     {
-        KafkaBindingConfig kafkaBinding = new KafkaBindingConfig(binding);
+        KafkaBindingConfig kafkaBinding = new KafkaBindingConfig(binding, supplyCatalog, validator);
```
jfallows, line 108: `validator` variable rename, please see below.

jfallows (changes requested):

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/fetch/message.value.avro.invalid/client.rpt
aDaemonThread: I tried to use logical representation as mentioned below

`read [0x00] 9 ${kafka:varint(3)} "id0" ${kafka:varint(8)} "positive"`

But I’m encountering issues while decoding these bytes.

As the bytes coming from kafka looks like this 

`0 0 0 0 9`

But k3po, translate the logical representation to something like this

`0 9 0 0 0`

which is causing an issue while trying to extract schemaId.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
aDaemonThread: validate if value is null.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
aDaemonThread: json_decoder is required in case, we want to transform avro or proto event to json. But since we are planning to have transformation piece separately, it can be considered then.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
aDaemonThread: I don’t think we need these for validator. byte_buffer is just the bytes wrapped to give an instance of byte_buffer.

jfallows (approved):

PR #318, opened 2023-07-25
Mqtt kafka sessions

bmaidics:


incubator/binding-mqtt.spec/src/main/resources/META-INF/zilla/mqtt.idl
```diff
@@ -77,7 +77,8 @@ scope mqtt
 
     struct MqttSessionState
     {
-        string16 clientId;
+        uint8 version = 1;
+        string16 clientId = null;
```
jfallows, line 81: Given that `clientId` is removed from all the `${mqtt:session()...}` expressions in the scripts, do we now need to remove this unused field in session state?

jfallows (changes requested):

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1983,28 +1983,24 @@ private void onDecodeSubscribe(
...
                     );
 
                     final MqttSessionStateFW sessionState = sessionStateBuilder.build();
                     final int payloadSize = sessionState.sizeof();
 
-                    //TODO: is this correct? What is this?
-                    int reserved = payloadSize;
-
-                    sessionStream.doSessionData(traceId, reserved, sessionState);
+                    sessionStream.doSessionData(traceId, payloadSize, sessionState);
```
jfallows, line 2003: `reserved` is `payload.sizeof() + padding`, so we do need to calculate `reserved` instead of assuming that `padding` is `0`.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -4193,7 +4176,7 @@ private void onSubscribeWindow(
                             iterator.remove();
                         }
 
-                        sendNewSessionStateForUnsubscribe(traceId, authorization, ackedTopicFilters);
+                        sendNewSessionStateForUnsubscribe(traceId, ackedTopicFilters);
```
jfallows, line 4179: Does this method name need to be so specific, should it just be `doSendSessionState`?

jfallows (changes requested):

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -0,0 +1,1620 @@
...
+        {
+            this.mqtt = mqtt;
+            this.originId = originId;
+            this.routedId = routedId;
+            this.initialId = initialId;
+            this.replyId = supplyReplyId.applyAsLong(initialId);
+            this.session = new KafkaSessionProxy(originId, resolvedId, this);
+            this.group = new KafkaGroupProxy(originId, resolvedId, this);
+            this.sessionsTopicName = sessionsTopicName;
+            this.sessionIdentifier = new String16FW(sessionIds.get(bindingId));
```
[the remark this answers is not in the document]
bmaidics: You're right. One thing that we need to add here: if it arrives to the same zilla, and same core, session stealing won't work anyway, as we discussed that in this case the consumer group will reset the first client, and we expect it to disconnect. Do you think we still need some other key anyway to cover this case?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -0,0 +1,1620 @@
...
+            final MqttBeginExFW mqttBeginEx = extension.get(mqttBeginExRO::tryWrap);
+
+            assert mqttBeginEx.kind() == MqttBeginExFW.KIND_SESSION;
+            final MqttSessionBeginExFW mqttSessionBeginEx = mqttBeginEx.session();
+
+            final String16FW clientId = mqttSessionBeginEx.clientId();
+            this.clientId = new String16FW().wrap(clientId.buffer(), clientId.offset(), clientId.limit());
+            this.clientIdMigrate = new String16FW(clientId.asString() + MIGRATE_KEY_POSTFIX);
+
+            sessionExpiryMs = mqttSessionBeginEx.expiry() == 0 ? 30000 : mqttSessionBeginEx.expiry() * 1000;
```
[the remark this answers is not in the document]
bmaidics: This needs more discussion, as the kafka consumer group has some limitations of what the group timeout can be. The defaulting was also needed by the consumer group, as we can't provide 0 timeout. This was just a quick fix.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -0,0 +1,1620 @@
...
+            assert initialAck <= initialSeq;
+
+            final OctetsFW extension = begin.extension();
+            final MqttBeginExFW mqttBeginEx = extension.get(mqttBeginExRO::tryWrap);
+
+            assert mqttBeginEx.kind() == MqttBeginExFW.KIND_SESSION;
+            final MqttSessionBeginExFW mqttSessionBeginEx = mqttBeginEx.session();
+
+            final String16FW clientId = mqttSessionBeginEx.clientId();
+            this.clientId = new String16FW().wrap(clientId.buffer(), clientId.offset(), clientId.limit());
```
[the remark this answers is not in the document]
bmaidics: Ohhh, right. I think then we should store this as regular String, and create a flyweight only if needed. If we'd like to store this as flyweight, we'd need a per stream buffer, which is not good.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -0,0 +1,1620 @@
...
+        private long replyAck;
+        private int replyMax;
+        private int replyPad;
+
+        private String16FW clientId;
+        private String16FW clientIdMigrate;
+        private int sessionExpiryMs;
+
+        private boolean sessionLeader = false;
+        private boolean migrate = false;
```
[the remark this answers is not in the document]
bmaidics: Sure, I was already thinking of getting rid of these, but failed

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
bmaidics: Let's discuss this as well. We probably need replyPad as well for decoding network data, but I'd like to understand padding more, before doing this.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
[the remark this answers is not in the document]
bmaidics: Provide Integer.MAX_VALUE if session expiry is 0, pass anything else in ms, consumer group will restrict the value within the allowed range specified dynamically by the broker configuration.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.abort.reconnect.non.clean.start/client.rpt
```diff
@@ -0,0 +1,298 @@
...
+         option zilla:transmission "duplex"
+
+write zilla:begin.ext ${kafka:beginEx()
+                            .typeId(zilla:id("kafka"))
+                            .merged()
+                                .capabilities("PRODUCE_AND_FETCH")
+                                .topic("mqtt_sessions")
+                                .filter()
+                                    .key("client-1#migrate")
+                                    .headerNot("sender-id", "sender1")
```
jfallows, line 27: Sorry for nitpicking, but `sender1` should be `sender-1` in keeping with format of `client-1`, `member-1`, etc.
This applies to all `sender-id` header values in the scripts.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaProxyFactory.java
```diff
@@ -71,13 +71,19 @@ public void attach(
...
     }
 
     @Override
     public void detach(
         long bindingId)
     {
         bindings.remove(bindingId);
+
+        MqttKafkaSessionFactory sessionFactory = (MqttKafkaSessionFactory) factories.get(MqttBeginExFW.KIND_SESSION);
+        sessionFactory.onDetached(bindingId);
```
jfallows, line 86: Create an interface called `MqttKafkaStreamFactory` with `default` empty implementations of `onAttached` and `onDetached`, then make each factory implement the interface, overriding `onAttached` and `onDetached` on session factory.

Update `factories` map to have target type of `MqttKafkaStreamFactory` and eliminate cast here in `detach` and above in `attach` too.

Instead of special casing session factory, you could just call `onAttached` / `onDetached` on all the factories, some of them will not react as they just have the default implementation, but session factory will have some session-specific behavior.

```
factories.values().forEach(MqttStreamFactory::onAttached);
```
and
```
factories.values().forEach(MqttStreamFactory::onDetached);
```

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1983,26 +1946,23 @@ private void onDecodeSubscribe(
...
-                            {
-                                subscriptionBuilder.subscriptionId(subscription.id);
-                                subscriptionBuilder.flags(subscription.flags);
-                                subscriptionBuilder.pattern(subscription.filter);
-                            });
-                            sessionStream.unAckedSubscriptions.add(subscription);
-                        }
+                            subscriptionBuilder.subscriptionId(subscription.id);
+                            subscriptionBuilder.flags(subscription.flags);
+                            subscriptionBuilder.pattern(subscription.filter);
```
jfallows, line 1959: Use fluent API, chaining these into a single statement, eliminate need for `{...}` braces and reduce size of lambda variable, say `sub` and `state`.

Do we need to make a copy of the combined collection to build the session state?
Perhaps we could just iterate `sessionStream.subscriptions` and then `newSubscriptions`?

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1983,26 +1946,23 @@ private void onDecodeSubscribe(
...
+                        })
                     );
 
                     final MqttSessionStateFW sessionState = sessionStateBuilder.build();
                     final int payloadSize = sessionState.sizeof();
-
-                    //TODO: is this correct? What is this?
-                    int reserved = payloadSize;
+                    final int reserved = payloadSize + sessionStream.initialPad;
```
jfallows, line 1965: Since `initialPad` is private to `session` stream, better to handle it there, inside `doSessionData` instead of trying to calculate and pass to `doSessionData` from outside.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2117,9 +2070,8 @@ private void onDecodeUnsubscribe(
             }
         }
 
-        private void sendNewSessionStateForUnsubscribe(
+        private void doSendSessionState(
             long traceId,
-            long authorization,
             List<String> topicFilters)
         {
             List<Subscription> currentState = sessionStream.getSubscriptions();
```
jfallows, line 2077: We don't use `getXxx()` JavaBeans coding style, instead use `xxx()` record-style.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -3109,23 +3058,26 @@ private void onSessionReset(
...
-                    doEncodeConnack(traceId, authorization, reasonCode, assignedClientId, false, serverReference);
-                }
-                else
+
+                if (mqttResetEx != null)
                 {
-                    doEncodeDisconnect(traceId, authorization, reasonCode, serverReference);
-                }
+                    String16FW serverReference = mqttResetEx.serverReference();
+                    boolean serverReferenceExists = serverReference != null && serverReference.length() != 0;
```
jfallows, line 3066: Why do we have a special value check for empty server reference?
Shouldn't it just be `!= null` and use whatever is given, since we're not validating the value in any other way.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -0,0 +1,1644 @@
...
+            if (!MqttKafkaState.initialClosed(state))
+            {
+                state = MqttKafkaState.closeInitial(state);
+
+                doReset(mqtt, originId, routedId, initialId, initialSeq, initialAck, initialMax, traceId);
+            }
+        }
+    }
+
+    abstract class KafkaSessionProxy
```
jfallows, line 527: `private` too

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -0,0 +1,1644 @@
...
+            replyAck = delegate.replyAck;
+            replyMax = delegate.replyMax;
+            replyPad = delegate.replyPad;
+
+            doWindow(kafka, originId, routedId, replyId, replySeq, replyAck, replyMax,
+                traceId, authorization, budgetId, replyPad, 0, capabilities);
+        }
+    }
+
+    final class KafkaSessionSignalProxy extends KafkaSessionProxy
```
jfallows, line 851: `private` too

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -0,0 +1,1644 @@
...
+            assert initialAck <= initialSeq;
+
+            if (!wasOpen)
+            {
+                sendMigrateSignal(authorization, traceId);
+            }
+        }
+    }
+
+    final class KafkaSessionStateProxy extends KafkaSessionProxy
```
jfallows, line 911: `private` too

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -0,0 +1,1644 @@
...
+            replySeq = sequence;
+            state = MqttKafkaState.closeReply(state);
+
+            assert replyAck <= replySeq;
+
+            delegate.doMqttEnd(traceId, authorization);
+        }
+    }
+
+    final class KafkaGroupProxy
```
jfallows, line 1022: `private` too

jfallows (changes requested):

jfallows (approved):

PR #320, opened 2023-07-31
Fix compact partitionOffset bug

bmaidics:
## Description

Fixes # (issue)


jfallows:
We determined this is not a bug, should use topic bootstrap instead.

PR #324, opened 2023-08-03
Generate zilla.yaml from an OpenAPI definition

### LINKED ISSUE #254: Generate `zilla.yaml` from `OpenAPI` specification(s)
Generate `zilla.yaml` from a corresponding `OpenAPI` definition.

See the [OpenAPI specification 3.1.0](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md) and [OpenAPI examples](https://github.com/OAI/OpenAPI-Specification/blob/main/examples).

Note: we may need additional information beyond `OpenAPI` definition to be able to fully populate the generated `zilla.yaml`.

attilakreiner:
## Description

Generate zilla.yaml from a corresponding OpenAPI definition.

Fixes #254


incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,366 @@
...
+            ex.printStackTrace();
+            rethrowUnchecked(ex);
+        }
+        return openApi;
+    }
+
+    private NamespaceConfig createNamespaceConfig()
+    {
+        NamespaceConfigBuilder<NamespaceConfig> namespaceBuilder = NamespaceConfig.builder()
+            .name("example");
```
jfallows, line 98: In places where it feels necessary to break the flow to perform some dynamic behavior, it would be more convenient to inject the logic into the flow instead.

For example:
```
    return NamespaceConfig.builder()
        .name("example")
        .inject(this::injectJwtGuard)
        .binding()
            .type("tcp")
...
```

So I think we should add something like `R inject(Function<T, R> injector)` to `ConfigBuilder` subtypes.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/model/PathItem.java
```diff
@@ -0,0 +1,74 @@
...
+    public Operation patch;
+    public Operation trace;
+
+    public void initMethods()
+    {
+        methods = new LinkedHashMap<>();
+        if (get != null)
+        {
+            methods.put("GET", get);
+        }
```
jfallows, line 39: Suggest hoisting this to a method.
```
private static void putIfNotNull(
    Map<String, Operation> methods,
    String method,
    Operation operation)
{
    if (operation != null)
    {
        methods.put(method, operation);
    }
}
```
Then `initMethods` reads a good bit simpler.
```
public void initMethods()
{
    methods = new LinkedHashMap<>();
    putIfNotNull(methods, "GET", get);
    putIfNotNull(methods, "PUT", put);
    ...
}
```

It actually feels as if this logic is the beginning of the intermediate model we discussed, and therefore doesn't really belong directly on the OpenAPI model.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,366 @@
...
+                .name("tcp_server0")
+                .type("tcp")
+                .kind(SERVER)
+                .options(TcpOptionsConfig::builder)
+                    .host("0.0.0.0")
+                    .ports(resolvePortsForScheme("https"))
+                    .build()
+                .route()
+                    .exit("tls_server0")
+                    .build()
```
jfallows, line 133: Would you prefer to have a shorthand `.exit(String)` method directly on `binding` config builder?

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,366 @@
...
+                .build()
+            .binding()
+                .name("tcp_server1")
+                .type("tcp")
+                .kind(SERVER)
+                .options(TcpOptionsConfig::builder)
+                    .host("0.0.0.0")
+                    .ports(resolvePortsForScheme("http"))
```
jfallows, line 141: This feels as though we are pulling the OpenAPI specification into the zilla.yaml instead of generating zilla.yaml based on the contents of the OpenAPI specification.

For example, what if there are no `http` scheme URLs in the OpenAPI specification (e.g. they are all `https`)?

jfallows (changes requested):

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/airline/ConfigGenerator.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package io.aklivity.zilla.runtime.command.config.internal.airline;
+
+public interface ConfigGenerator
+{
+    String generateConfig();
```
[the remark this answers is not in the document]
attilakreiner: done

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/ConfigWriter.java
```diff
@@ -79,9 +104,15 @@ private void write0(
                 .withProvider(provider)
                 .withConfig(config)
                 .build();
-
             String jsonText = jsonb.toJson(namespace, NamespaceConfig.class);
-            JsonNode json = new ObjectMapper().readTree(jsonText);
+            JsonObject jsonObject = provider.createReader(new StringReader(jsonText)).readObject();
+            JsonObject patched = patch.apply(jsonObject);
+            StringWriter stringWriter = new StringWriter();
```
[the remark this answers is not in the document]
attilakreiner: done

incubator/command-config/src/test/resources/io/aklivity/zilla/runtime/command/config/internal/openapi/zilla.yaml
```diff
@@ -0,0 +1,125 @@
...
+      issuer: "${{env.JWT_ISSUER}}"
+      audience: "${{env.JWT_AUDIENCE}}"
+      keys:
+      - kty: "${{env.JWT_KTY}}"
+        "n": "${{env.JWT_N}}"
+        e: "${{env.JWT_E}}"
+        alg: "${{env.JWT_ALG}}"
+        crv: "${{env.JWT_CRV}}"
+        x: "${{env.JWT_X}}"
+        "y": "${{env.JWT_Y}}"
```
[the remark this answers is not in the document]
attilakreiner: I believe the reason why `YAMLMapper` needs to quote `n` and `y` and not the others is this:
https://yaml.org/type/bool.html
```
y|Y|yes|Yes|YES|n|N|no|No|NO
|true|True|TRUE|false|False|FALSE
|on|On|ON|off|Off|OFF
```
As per yaml spec `n` and `y` are boolean literals, so must be quoted if used as strings. There are some tricky things in yaml, I guess this is one of them 😅

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,366 @@
...
+        catch (Exception ex)
+        {
+            ex.printStackTrace();
+            rethrowUnchecked(ex);
+        }
+        return openApi;
+    }
+
+    private NamespaceConfig createNamespaceConfig()
+    {
```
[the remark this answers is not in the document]
attilakreiner: fixed

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,366 @@
...
+        NamespaceConfigBuilder<NamespaceConfig> namespaceBuilder = NamespaceConfig.builder()
```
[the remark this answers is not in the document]
attilakreiner: done

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,366 @@
...
+                            .build()
+                        .build()
+                    .build();
+
+        for (RouteConfig route : generateRoutes("http_client0", guardedRoutes))
+        {
+            httpServer0BindingBuilder.route(route);
+        }
+
+        NamespaceConfig namespace = httpServer0BindingBuilder
```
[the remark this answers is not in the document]
attilakreiner: done

incubator/command-config/src/main/moditect/module-info.java
```diff
@@ -0,0 +1,32 @@
...
+    requires io.aklivity.zilla.runtime.binding.http;
+    requires io.aklivity.zilla.runtime.binding.tcp;
+    requires io.aklivity.zilla.runtime.binding.tls;
+    requires io.aklivity.zilla.runtime.guard.jwt;
+    requires io.aklivity.zilla.runtime.vault.filesystem;
+
+    opens io.aklivity.zilla.runtime.command.config.internal.airline
+        to com.github.rvesse.airline;
+
+    opens io.aklivity.zilla.runtime.command.config.internal.openapi.model;
```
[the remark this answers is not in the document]
attilakreiner: Tried and failed; will get back to this and send you the details on slack.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/model/PathItem.java
attilakreiner: Yep, introducing an intermediate model would help to keep the OpenAPI model super simple/clean and let us deal with the required transformations/conversions in another class. I think this is a good idea, but I couldn't come up with a good name for it 😅. For now I created the package named `model2` and the class `PathItem2`.  Pls chk.

I will keep thinking about some better naming pattern, pls let me know if you have any suggestions.

I created `Server2` for the same reason, based on your review item about the `Server` class.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/model/Server.java
```diff
@@ -0,0 +1,39 @@
...
+    public String getUrl()
+    {
+        return url.toString();
+    }
+
+    public void setUrl(
+        String url) throws URISyntaxException
+    {
+        this.url = new URI(url);
+    }
```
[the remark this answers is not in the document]
attilakreiner: I need a type conversion here. I created an intermediate model to address this.

Pls the discussion about the intermediate model below, at your review item about `PathItem`.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/airline/ZillaConfigCommand.java
```diff
@@ -0,0 +1,68 @@
...
+import io.aklivity.zilla.runtime.command.config.internal.openapi.OpenApiHttpProxyConfigGenerator;
+
+@Command(name = "config", description = "Generate configuration file")
+public final class ZillaConfigCommand extends ZillaCommand
+{
+    @Option(name = {"-t", "--template"},
+        description = "Template name:\n" +
+            "- openapi.http.proxy\n")
+    @Required
+    public String template;
```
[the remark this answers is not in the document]
attilakreiner: The idea is to be able to extend the functionality of this command as we'll have more things to do here, but still keep the syntax simple e.g.
```
$ zilla config -t openapi.http.proxy -i openapi.yaml -o zilla.yaml
$ zilla config -t asyncapi.http.proxy -i asyncapi.yaml -o zilla.yaml
$ zilla config -t asyncapi.something.etc -i asyncapi.yaml -o zilla.yaml
```

We could call this argument something else, but I actually like the word template as it suggests what actually happens, we can generate configs to some predefined use cases and we fill in some values based on the input files.

I added the `@AllowedValues` annotation.

incubator/command-config/src/test/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGeneratorTest.java
```diff
@@ -0,0 +1,42 @@
...
+public class OpenApiHttpProxyConfigGeneratorTest
+{
+    @Test
+    public void shouldGenerateConfig() throws Exception
+    {
+        // GIVEN
+        Path expectedPath = Path.of(getClass().getResource("zilla.yaml").toURI());
+        String expectedResult = Files.readString(expectedPath);
+        Path inputPath = Path.of(getClass().getResource("openapi.yaml").toURI());
+        OpenApiHttpProxyConfigGenerator generator = new OpenApiHttpProxyConfigGenerator(inputPath);
```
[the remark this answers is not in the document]
attilakreiner: I changed the input from `Path` to `InputStream` as requested. The output is a `String`, the command writes it to the file system with `Files.writeString()`, and the test compares the actual result with the expected `String`. Pls LMK if it's OK.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGenerator.java
attilakreiner: Yes, I think it would make sense, as it'd be consistent with the generated output, that applies the shortcut syntax now.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGenerator.java
attilakreiner: Yep. I guess the code would look nicer that way. So, the argument here would be a function reference that builds and returns the object to inject, right? What if we need to insert more than one element, could we handle that case, too? 🤔

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,366 @@
...
+                                .pattern("Bearer {credentials}")
+                                .build()
+                            .build()
+                        .build()
+                    .build();
+
+        for (RouteConfig route : generateRoutes("http_client0", guardedRoutes))
+        {
+            httpServer0BindingBuilder.route(route);
+        }
```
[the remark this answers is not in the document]
attilakreiner: done, pls chk

incubator/command-config/src/test/resources/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/plain/zilla.yaml
```diff
@@ -0,0 +1,51 @@
...
+  http_client0:
+    type: http
+    kind: client
+    exit: tcp_client0
+  tcp_client0:
+    type: tcp
+    kind: client
+    options:
+      host: "${{env.TCP_CLIENT_HOST}}"
+      port: "${{env.TCP_CLIENT_PORT}}"
```
jfallows, line 51: I don't think we need to quotes for this one since `port` is normally integer type.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,519 @@
...
+        OpenApi openApi = null;
+        try
+        {
+            Jsonb jsonb = JsonbBuilder.create();
+            openApi = jsonb.fromJson(inputStream, OpenApi.class);
+            jsonb.close();
+        }
+        catch (Exception ex)
+        {
+            ex.printStackTrace();
```
jfallows, line 103: This should really be handled in `ZillaConfigCommand` if needed, not local to catching exception.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,519 @@
...
+            .inject(namespace -> injectTlsServer(namespace))
+            .binding()
+                .name("http_server0")
+                .type("http")
+                .kind(SERVER)
+                .options(HttpOptionsConfig::builder)
+                    .access()
+                        .policy(CROSS_ORIGIN)
+                        .build()
+                    .inject(options -> injectHttpServerOptions(options))
```
jfallows, line 184: Use `.inject(this::injectHttpServerOptions)` instead.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,519 @@
...
+            .binding()
+                .name("tcp_client0")
+                .type("tcp")
+                .kind(CLIENT)
+                .options(TcpOptionsConfig::builder)
+                    .host("") // env
+                    .ports(new int[]{0}) // env
+                    .build()
+                .build()
+            .inject(namespace -> injectVaults(namespace, isTlsEnabled))
```
jfallows, line 204: No need for `isTlsEnabled` parameter, already available to method via field.
Use `this::injectVaults` instead.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,519 @@
...
+                .name("tcp_server0")
+                .type("tcp")
+                .kind(SERVER)
+                .options(TcpOptionsConfig::builder)
+                    .host("0.0.0.0")
+                    .ports(allPorts)
+                    .build()
+                .inject(binding -> injectTcpRoutes(binding, plainExit, tlsExit))
+                .build();
+        return namespace;
```
jfallows, line 245: Move `return` to first line, since `namespace.binding()....build()` returns `namespace`.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,519 @@
...
+        }
+        else if (isTlsEnabled)
+        {
+            binding
+                .route()
+                    .exit(tlsExit)
+                    .build();
+        }
+        return binding;
+    }
```
jfallows, line 284: Inject these `http` (plain) and `https` (tls) routes using separate methods, which are no-ops if not needed.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,519 @@
...
+        }
+        return result;
+    }
+
+    private NamespaceConfig createNamespace()
+    {
+        return NamespaceConfig.builder()
+            .name("example")
+            .inject(namespace -> injectGuard(namespace))
+            .inject(namespace -> injectTcpServer(namespace, "http_server0", "tls_server0"))
```
jfallows, line 174: If these exit names are fixed, then we don't need them as parameters.

In general, most if not all of the inject lambdas can be method references of the form `this::injectX` with a single parameter, and potentially also driven by instance state.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,519 @@
...
+            for (Map<String, List<String>> securityItem : security)
+            {
+                for (String securityItemLabel : securityItem.keySet())
+                {
+                    if (isJwtEnabled && "jwt".equals(securitySchemes.get(securityItemLabel)))
+                    {
+                        route
+                            .guarded()
+                            .name("jwt0")
+                            .role("echo:stream")
```
jfallows, line 345: Perhaps this should be omitted since it doesn't have meaning in the context of the OpenAPI source?

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,519 @@
...
+        this.allPorts = resolveAllPorts();
+        this.httpPorts = resolvePortsForScheme("http");
+        this.httpsPorts = resolvePortsForScheme("https");
+        this.isPlainEnabled = httpPorts != null;
+        this.isTlsEnabled = httpsPorts != null;
+        this.securitySchemes = resolveSecuritySchemes();
+        this.isJwtEnabled = !securitySchemes.isEmpty();
+        this.namespace = createNamespace();
+        this.envVarsPatch = createEnvVarsPatch();
+        this.configWriter = new ConfigWriter(null);
```
jfallows, line 83: Recommend we do the majority of the work in the `generate` method instead of in the constructor.

jfallows (changes requested):

incubator/command-config/src/test/resources/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/plain/zilla.yaml
attilakreiner: We build the json object, then apply the json patch, then generate the yaml from the json result. So, this output is generated by `YAMLMapper` and it quotes these env var strings because they have all kinds of funky characters. We have a few options to tweak, but I don't think we can fine tune the yaml generation to this extent. However, if this actually turns out to be a problem for zilla, then we'll need to find a way to solve it after all. 🤔

https://github.com/attilakreiner/zilla/blob/2814d370eb0f9de900bfdd500af810048ef4643b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/ConfigWriter.java#L116-L119

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: fixed

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: done

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: done

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: fixed

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: Removing the role results in this output:
```
    routes:
    - exit: http_client0
      when:
      - headers:
          :path: /items
          :method: GET
      guarded:
        jwt0: []
```
Pls confirm if this is OK.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: done

incubator/command-config/src/test/resources/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/plain/zilla.yaml
attilakreiner: So it turns out that the quoting doesn't cause any problems and zilla correctly parses this to an int array anyway. As this would be pretty cumbersome to fix (e.g. applying some custom regexp to the generated output), my preference is to leave it as is, pls LMK.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
jfallows: Yes, this means they must be a valid user, but no specific roles are required.

Looks like this is the official syntax we are looking for to drive the generation, agree?
https://github.com/OAI/OpenAPI-Specification/pull/2515/files

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/airline/ZillaConfigCommand.java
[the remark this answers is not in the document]
jfallows: What is the output of the help text now that you have annotated the `AllowedValues`?

incubator/command-config/src/test/resources/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/complete/zilla.yaml
```diff
@@ -0,0 +1,124 @@
...
+      - "${{env.TLS_CLIENT_SNI}}"
+      alpn:
+      - "${{env.TLS_CLIENT_ALPN}}"
+    exit: tcp_client0
+  tcp_client0:
+    type: tcp
+    kind: client
+    options:
+      host: "${{env.TCP_CLIENT_HOST}}"
+      port: "${{env.TCP_CLIENT_PORT}}"
```
jfallows, line 93: Let's remove the quotes here, should be an integer.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,505 @@
...
+        }
+        return options;
+    }
+
+    private BindingConfigBuilder<NamespaceConfigBuilder<NamespaceConfig>> injectHttpServerRoutes(
+        BindingConfigBuilder<NamespaceConfigBuilder<NamespaceConfig>> binding)
+    {
+        for (String path : openApi.paths.keySet())
+        {
+            PathItem2 item = PathItem2.of(openApi.paths.get(path));
```
jfallows, line 310: If we are going to use `model2` classes, then perhaps it would be helpful to massage their representation to better suit iteration for generation of zilla.yaml.

For example, perhaps the concept of operation:
```
Operation
{
    String operationId
    String method
    String path
    List<String> roles
}
```

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,505 @@
...
+
+    public OpenApiHttpProxyConfigGenerator(
+        InputStream inputStream)
+    {
+        this.inputStream = inputStream;
+    }
+
+    public String generate()
+    {
+        this.openApi = parseOpenApi(inputStream);
```
jfallows, line 77: Does `openApi` need to be a field, or can it be passed to other methods resolving lookups?

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,505 @@
...
+                    .build()
+                .route()
+                    .when(TcpConditionConfig::builder)
+                        .ports(httpsPorts)
+                        .build()
+                    .exit("tls_server0")
+                    .build();
+        }
+        return binding;
+    }
```
jfallows, line 260: Suggest we keep it simple, and always include the port-specific route for `http` or`https`.
So no need for 3 cases of `tls && !plain`, `!tls && plain` and `tls && plain`, just 2 cases `(optional) plain` and `(optional) tls`.

jfallows (changes requested):

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: done

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: I checked this approach but didn't seem to work out.

- Using `Operation2` with `method` and `path` fields didn't result in any simplification of the code in `injectHttpServerRoutes`. We would need to create an extra object to contain and read values that we already have anyway, so this change did not seem to have a point.
- Moving the complex logic from `injectHttpServerRouteGuarded` to `Operation2` could be nice, however that code needs to access the instance variable `securitySchemes`, so we'd need to pass it when creating the `Operation2` object which looked a bit contrived to me.

I've finally undone all my changes, I'd prefer to keep it as is (at least for now); pls LMK if you insist.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,505 @@
...
+            ops.put("/vaults/client/options/trust/store", "${{env.TRUSTSTORE_PATH}}");
+            ops.put("/vaults/client/options/trust/type", "${{env.TRUSTSTORE_TYPE}}");
+            ops.put("/vaults/client/options/trust/password", "${{env.TRUSTSTORE_PASSWORD}}");
+            // server vault
+            ops.put("/vaults/server/options/keys/store", "${{env.KEYSTORE_PATH}}");
+            ops.put("/vaults/server/options/keys/type", "${{env.KEYSTORE_TYPE}}");
+            ops.put("/vaults/server/options/keys/password", "${{env.KEYSTORE_PASSWORD}}");
+        }
+
+        JsonArrayBuilder patch = Json.createArrayBuilder();
```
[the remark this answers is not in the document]
attilakreiner: Done. I created a helper method `addOp`, so we avoid building a map, but the code stays easy to read. Pls chk.

incubator/command-config/src/test/resources/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/complete/zilla.yaml
attilakreiner: Done, pls chk. We need to do another round of string transformation after the yaml output has been generated to remove the unnecessary quotes. If you have an idea for a more efficient approach pls LMK.

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
jfallows: Perhaps an inner class of the generator instead of a separate model2 package?

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,495 @@
...
+            addOp(patch, "/vaults/client/options/trust/store", "${{env.TRUSTSTORE_PATH}}");
+            addOp(patch, "/vaults/client/options/trust/type", "${{env.TRUSTSTORE_TYPE}}");
+            addOp(patch, "/vaults/client/options/trust/password", "${{env.TRUSTSTORE_PASSWORD}}");
+            // server vault
+            addOp(patch, "/vaults/server/options/keys/store", "${{env.KEYSTORE_PATH}}");
+            addOp(patch, "/vaults/server/options/keys/type", "${{env.KEYSTORE_TYPE}}");
+            addOp(patch, "/vaults/server/options/keys/password", "${{env.KEYSTORE_PASSWORD}}");
+        }
+
+        return JsonProvider.provider().createPatch(patch.build());
```
jfallows, line 475: Please use code similar to the following, instead of using an intermediate `JsonArrayBuilder` to build the raw patch document.

```
JsonPatchBuilder patch = Json.createPatchBuilder();

patch.replace("/bindings/tcp_client0/options/host", "${{env.TCP_CLIENT_HOST}}");
patch.replace("/bindings/tcp_client0/options/port", "${{env.TCP_CLIENT_PORT}}");

if (isJwtEnabled)
{
    ...
}

if (isTlsEnabled)
{
    ...
}

return patch.build();
```

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/ConfigWriter.java
```diff
@@ -49,20 +57,41 @@ public void write(
...
+        write0(namespace, writer, patch, unquotedEnvVars);
         return writer.toString();
     }
 
     private void write0(
         NamespaceConfig namespace,
-        Writer writer)
+        Writer writer,
+        JsonPatch patch,
+        List<String> unquotedEnvVars)
```
jfallows, line 94: Let's not leak the concept of unquoted env vars into the config writer.
Instead please post-process the resulting YAML string in the config command.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/ConfigWriter.java
```diff
@@ -79,14 +108,31 @@ private void write0(
...
                 .enable(MINIMIZE_QUOTES)
                 .build();
-            mapper.writeValue(writer, json);
+            String yaml = mapper.writeValueAsString(json);
+
+            for (String envVar : unquotedEnvVars)
+            {
+                yaml = yaml.replaceAll(
+                    Pattern.quote(String.format("\"${{env.%s}}\"", envVar)),
+                    String.format("\\${{env.%s}}", envVar)
```
jfallows, line 131: Btw, is this only needed because `MINIMIZE_QUOTES` is not removing the quotes generically?
Can we check why that's not already happening, since it is not typically necessary to have quotes on strings in yaml values.

incubator/command-config/src/test/resources/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/complete/zilla.yaml
```diff
@@ -0,0 +1,124 @@
...
+            headers:
+              authorization: "Bearer {credentials}"
+    routes:
+    - exit: http_client0
+      when:
+      - headers:
+          :path: /items
+          :method: GET
+      guarded:
+        jwt0: []
```
jfallows, line 48: Need roles too, please see comment above.

jfallows (changes requested):

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,495 @@
...
+        {
+            for (Map<String, List<String>> securityItem : security)
+            {
+                for (String securityItemLabel : securityItem.keySet())
+                {
+                    if (isJwtEnabled && "jwt".equals(securitySchemes.get(securityItemLabel)))
+                    {
+                        route
+                            .guarded()
+                                .name("jwt0")
```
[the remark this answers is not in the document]
attilakreiner: OK, pls chk

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,495 @@
...
+                    .build();
+        }
+        else
+        {
+            binding
+                .route()
+                    .exit("tcp_client0")
+                    .build();
+        }
+        return binding;
```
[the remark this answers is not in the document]
attilakreiner: done

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: oops, looks much better, thx 😅

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/ConfigWriter.java
attilakreiner: Created the method `unquoteEnvVars` in `OpenApiHttpProxyConfigGenerator`, I hope it's OK.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/ConfigWriter.java
attilakreiner: My observation is that it quotes the strings that contain curly braces.

incubator/command-config/src/test/resources/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/complete/zilla.yaml
attilakreiner: yep, pls chk

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,496 @@
...
+        return namespace;
+    }
+
+    private NamespaceConfigBuilder<NamespaceConfig> injectVaults(
+        NamespaceConfigBuilder<NamespaceConfig> namespace)
+    {
+        if (isTlsEnabled)
+        {
+            namespace
+                .vault()
```
jfallows, line 412: Indentation

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,496 @@
...
+
+    private GuardedConfigBuilder<RouteConfigBuilder<BindingConfigBuilder<NamespaceConfigBuilder<NamespaceConfig>>>>
+        injectGuardedRoles(
+        GuardedConfigBuilder<RouteConfigBuilder<BindingConfigBuilder<NamespaceConfigBuilder<NamespaceConfig>>>> guarded,
+        List<String> roles)
+    {
+        for (String role : roles)
+        {
+            guarded
+                .role(role);
```
jfallows, line 345: Single line is fine here. 😄

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
```diff
@@ -0,0 +1,496 @@
...
+                }
+            }
+        }
+        return route;
+    }
+
+    private GuardedConfigBuilder<RouteConfigBuilder<BindingConfigBuilder<NamespaceConfigBuilder<NamespaceConfig>>>>
+        injectGuardedRoles(
+        GuardedConfigBuilder<RouteConfigBuilder<BindingConfigBuilder<NamespaceConfigBuilder<NamespaceConfig>>>> guarded,
+        List<String> roles)
```
jfallows, line 340: Perhaps we can simplify this signature using:
```
private <C> GuardedConfigBuilder<C> injectGuardedRoles(
    GuardedConfigBuilder<C> guarded,
    List<String> roles)
```

jfallows (changes requested):

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/config/ConfigWriterTest.java
```diff
@@ -93,4 +101,56 @@ public void shouldWriteNamespace()
...
+                    .build()
+                .build();
+        JsonArray ops = Json.createArrayBuilder()
+            .add(Json.createObjectBuilder()
+                .add("op", "replace")
+                .add("path", "/bindings/test0/type")
+                .add("value", "newType")
+                .build())
+            .build();
+        JsonPatch patch = JsonProvider.provider().createPatch(ops);
```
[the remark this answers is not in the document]
attilakreiner: done

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: fixed

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: fixed

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: done

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/openapi/http/proxy/OpenApiHttpProxyConfigGenerator.java
attilakreiner: I wouldn't move the models to inner classes as we will have multiple generators (for asyncapi anyway, but potentially for openapi as well), so it makes sense that various generators can reuse the same model classes.

jfallows (approved):

PR #327, opened 2023-08-08
Include member count as part of group data ex

akrambek:
## Description

Fixes # (issue)

### DIFF
--- a/specs/binding-kafka.spec/src/main/java/io/aklivity/zilla/specs/binding/kafka/internal/KafkaFunctions.java
+++ b/specs/binding-kafka.spec/src/main/java/io/aklivity/zilla/specs/binding/kafka/internal/KafkaFunctions.java
@@ -1597,10 +1597,10 @@ public final class KafkaFunctions
                 return this;
             }
 
-            public KafkaGroupDataExBuilder memberCount(
-                int memberCount)
+            public KafkaGroupDataExBuilder members(
+                int members)
             {
-                groupDataExRW.memberCount(memberCount);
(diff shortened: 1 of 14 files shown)


specs/binding-kafka.spec/src/main/java/io/aklivity/zilla/specs/binding/kafka/internal/KafkaFunctions.java
```diff
@@ -2722,6 +2729,7 @@ public final class KafkaGroupDataExMatchBuilder
         {
             private String16FW leaderId;
             private String16FW memberId;
+            private int members;
```
jfallows, line 2732: This should be `Integer` so we can avoid attempting to match if not set, right?

specs/binding-kafka.spec/src/main/java/io/aklivity/zilla/specs/binding/kafka/internal/KafkaFunctions.java
```diff
@@ -2765,6 +2781,12 @@ private boolean matchMemberId(
             {
                 return memberId == null || memberId.equals(groupDataEx.memberId());
             }
+
+            private boolean matchmembers(
+                final KafkaGroupDataExFW groupDataEx)
+            {
+                return members == groupDataEx.members();
```
jfallows, line 2788: `members != null && members == groupDataEx.members();` ?

jfallows (changes requested):

jfallows (approved):

PR #329, opened 2023-08-10
Add hashKey support to merged stream

bmaidics:


specs/binding-kafka.spec/src/main/resources/META-INF/zilla/kafka.idl
```diff
@@ -229,12 +229,13 @@ scope kafka
...
             int32 deferred = 0;         // INIT only (TODO: move to DATA frame)
             int64 timestamp = 0;        // INIT only
-            int64 filters = -1;          // INIT only
+            int64 filters = -1;         // INIT only
             KafkaOffset partition;      // INIT only
             KafkaOffset[] progress;     // INIT only
             KafkaKey key;               // INIT only
             KafkaDelta delta;           // INIT + FIN
             KafkaHeader[] headers;      // INIT + FIN (produce), INIT only (fetch)
+            KafkaKey hashKey;           // INIT only
```
jfallows, line 238: Please group this with other `INIT only` members, right below `key`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
```diff
@@ -1220,10 +1221,13 @@ private KafkaOffsetType asMaximumOffset(
...
 
         private int nextPartitionData(
+            KafkaKeyFW hashKey,
             KafkaKeyFW key)
         {
             final int partitionCount = leadersByPartitionId.size();
-            final int keyHash = key.length() != -1 ? defaultKeyHash(key) : nextNullKeyHashData++;
+            int keyHash = (hashKey.length() != -1) ? defaultKeyHash(hashKey) :
+                (key.length() != -1) ? defaultKeyHash(key) :
+                    nextNullKeyHashData++;
```
jfallows, line 1230: `final` can be kept as we don't mutate `keyHash`.
Parentheses around `...length() != -1` can be removed (x2), agree?

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/merged/merged.produce.message.values.dynamic.hash.key/client.rpt
```diff
@@ -0,0 +1,99 @@
...
+                               .build()}
+
+connected
+
+write zilla:data.ext ${kafka:dataEx()
+                              .typeId(zilla:id("kafka"))
+                              .merged()
+                                  .timestamp(newTimestamp)
+                                  .partition(-1, -1)
+                                  .hashKey("key7")
```
jfallows, line 40: I think this test needs to specify both `key` and `hashKey` such that they would naturally hash to different partitions, and then prove that `hashKey` is being selected because that is the partition used.

jfallows (changes requested):

jfallows (approved):

PR #367, opened 2023-08-16
Mqtt kafka will message delivery

bmaidics:


jfallows (changes requested):

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/client.rpt
```diff
@@ -0,0 +1,400 @@
...
+                                .groupId("mqtt-will-consumer-group")
+                                .filter()
+                                    .header("type", "will-signal")
+                                    .build()
+                                .build()
+                            .build()}
+
+connected
+
+write notify FIRST_WILL_SIGNAL_SENT
```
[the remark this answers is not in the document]
bmaidics: If you look at the server.rpt, we use write notify FIRST_WILL_SIGNAL_SENT after we sent the will signal from the state kafka stream. This barrier is responsible for not sending to the zilla specific will stream this signal, until we've sent it from mqtt-kafka session stream.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
```diff
@@ -0,0 +1,380 @@
...
+                           .typeId(zilla:id("kafka"))
+                           .merged()
+                               .deferred(0)
+                               .partition(-1, -1)
+                               .key("client-1#will")
+                               .hashKey("client-1")
+                               .build()
+                           .build()}
+
+read zilla:data.empty
```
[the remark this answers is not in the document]
bmaidics: We're not receiving it, we are removing the will by producing a tombstone.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
```diff
@@ -0,0 +1,380 @@
...
+read zilla:begin.ext ${kafka:matchBeginEx()
+                            .typeId(zilla:id("kafka"))
+                            .merged()
+                                .capabilities("PRODUCE_AND_FETCH")
+                                .topic("mqtt_sessions")
+                                .filter()
+                                    .key("client-1#will-signal")
+                                    .build()
+                                .build()
+                            .build()}
```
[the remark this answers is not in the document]
bmaidics: Why do we need the groupId here?

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
```diff
@@ -0,0 +1,380 @@
...
+                               .group()
+                                   .groupId("client-1")
+                                   .protocol("highlander")
+                                   .timeout(1000)
+                                   .build()
+                               .build()}
+
+connected
+
+# wait so that we won't be the leader too early, before the lifetimeId arrives
```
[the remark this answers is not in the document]
bmaidics: When we become the leader, that's when we're supposed to send the will message, so that's when we're sending the window back to mqtt stream. If we become the leader early, before we could fetch the lifetimeId, it will be empty, which might be wrong.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
```diff
@@ -0,0 +1,380 @@
...
+
+read ${mqtt:willSignal()
+         .clientId("client-1")
+         .delay(1)
+         .deliverAt(-1)
+         .lifetimeId("1")
+         .willId("1")
+     .build()}
+
+write notify FIRST_WILL_SIGNAL_SENT
```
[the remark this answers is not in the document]
bmaidics: This is not true, as we've sent it, not receive from mqtt-kafka binding perspective, right?

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
```diff
@@ -0,0 +1,380 @@
...
+         .clientId("client-1")
+         .delay(1)
+         .deliverAt(-1)
+         .lifetimeId("1")
+         .willId("1")
+     .build()}
+write flush
+
+# we received  abort from mqtt
+read await SECOND_WILL_SIGNAL_SENT
```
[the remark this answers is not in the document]
bmaidics: This is not true, as we've sent it, not receive from mqtt-kafka binding perspective, right?

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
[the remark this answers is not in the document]
jfallows: Understood. How does that related to _migrate_ signal though, not will message or will signal?

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/client.rpt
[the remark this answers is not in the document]
jfallows: Right, but `notify` means that something has happened _locally_ in this script, not somewhere else.

In general, consider barrier names to be local variables for use only within the same script.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
```diff
@@ -0,0 +1,380 @@
...
+                               .deferred(0)
+                               .partition(-1, -1)
+                               .key("wills/one")
+                               .header("zilla:topic", "wills")
+                               .header("zilla:topic", "one")
+                               .header("zilla:format", "TEXT")
+                               .build()
+                           .build()}
+
+read "client one session expired"
```
bmaidics, line 377: Change it to: read "client-1 disconnected abruptly"

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/client.rpt
```diff
@@ -0,0 +1,344 @@
...
+
+read zilla:data.ext ${kafka:matchDataEx()
+                             .typeId(zilla:id("kafka"))
+                             .group()
+                                 .leaderId("member-1")
+                                 .memberId("member-1")
+                                 .members(1)
+                                 .build()
+                             .build()}
+read notify RECEIVED_LEADER_DATA
```
jfallows, line 178: Suggest rename to `RECEIVED_GROUP_MEMBERS_LEADER`.

In more complex scenarios, we can use `RECEIVED_GROUP_MEMBERS_NOT_LEADER` too.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/client.rpt
```diff
@@ -0,0 +1,344 @@
...
+
+write close
+
+
+connect await RECEIVED_LEADER_DATA
```
jfallows, line 183: `RECEIVED_GROUP_MEMBERS_LEADER`

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/client.rpt
```diff
@@ -0,0 +1,344 @@
...
+                                .filter()
+                                    .key("client-1#migrate")
+                                    .headerNot("sender-id", "sender-1")
+                                    .build()
+                                .build()
+                            .build()}
+
+connected
+
+read advised zilla:flush
```
jfallows, line 205: Comment above
```
# no will signals
# no session state
```

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/server.rpt
```diff
@@ -0,0 +1,339 @@
...
+                           .merged()
+                               .deferred(0)
+                               .partition(-1, -1)
+                               .key("client-1#will")
+                               .hashKey("client-1")
+                               .build()
+                           .build()}
+read zilla:data.null
+
+# non-clean start -> fetch lifetimeId
```
jfallows, line 100: Change comment to `# non-clean start`

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/server.rpt
```diff
@@ -0,0 +1,339 @@
...
+write ${mqtt:willSignal()
+            .clientId("client-1")
+            .delay(1)
+            .deliverAt(2)
+            .lifetimeId("1")
+            .willId("1")
+        .build()}
+write flush
+
+#Once the will is delivered -> remove will message + signal
```
jfallows, line 76: Change comment to `# cleanup will signal`

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/server.rpt
```diff
@@ -0,0 +1,339 @@
...
+                               .deferred(0)
+                               .partition(-1, -1)
+                               .key("client-1#will-signal")
+                               .hashKey("client-1")
+                               .header("type", "will-signal")
+                               .build()
+                           .build()}
+read zilla:data.null
+
+read zilla:data.ext ${kafka:matchDataEx()
```
jfallows, line 89: Add comment to `# cleanup will message`

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/server.rpt
```diff
@@ -0,0 +1,339 @@
...
+connected
+
+# no will signals
+write advise zilla:flush
+
+read closed
+write close
+
+
+# send initial migrate signal
```
jfallows, line 123: Change comment to `receive migrate signal from sender-1` and move after `connected`.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/server.rpt
```diff
@@ -0,0 +1,339 @@
...
+                               .group()
+                                   .groupId("client-1")
+                                   .protocol("highlander")
+                                   .timeout(1000)
+                                   .build()
+                               .build()}
+
+connected
+
+write zilla:data.ext ${kafka:dataEx()
```
jfallows, line 169: Add comment `# send group members (leader)`

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/server.rpt
```diff
@@ -0,0 +1,339 @@
...
+                                    .headerNot("sender-id", "sender-1")
+                                    .build()
+                                .build()
+                            .build()}
+
+connected
+
+# no session state
+# no migrate signals
+
```
jfallows, line 203: Remove blank line, as comment describes meaning of `zilla:flush`.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/client.rpt
```diff
@@ -0,0 +1,344 @@
...
+           .payload("client-1 disconnected abruptly")
+           .lifetimeId("1")
+           .willId("1")
+       .build()}
+
+write close
+read closed
+
+write notify WILL_MESSAGE_FETCHED
+
```
jfallows, line 309: Move `write notify` before `write close`, and change to `read notify` as it is after `read`, then rename to follow pattern, i.e. `SENT_SESSION_WILL_MESSAGE`.

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.close.deliver.will.message/client.rpt
```diff
@@ -0,0 +1,344 @@
...
+
+connect await WILL_MESSAGE_FETCHED
```
jfallows, line 311: `SENT_SESSION_WILL_MESSAGE`

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
```diff
@@ -0,0 +1,373 @@
...
+                               .group()
+                                   .groupId("client-1")
+                                   .protocol("highlander")
+                                   .timeout(1000)
+                                   .build()
+                               .build()}
+
+connected
+
+write zilla:data.ext ${kafka:dataEx()
```
jfallows, line 169: Add comment `# send group members (leader)`

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
```diff
@@ -0,0 +1,373 @@
...
+                                .topic("mqtt_sessions")
+                                .filter()
+                                    .key("client-1#will")
+                                    .build()
+                                .build()
+                            .build()}
+
+connected
+
+# saved will message for client-1
```
jfallows, line 287: Change comment to `# send session will message for client-1`

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.will.message.abort.deliver.will.retain/server.rpt
```diff
@@ -0,0 +1,373 @@
...
+
+# deliver will message for client-1
+read zilla:data.ext ${kafka:matchDataEx()
+                           .typeId(zilla:id("kafka"))
+                           .merged()
+                               .deferred(0)
+                               .partition(-1, -1)
+                               .key("obituaries")
+                               .header("zilla:topic", "wills")
+                               .header("zilla:topic", "one")
```
jfallows, line 334: `zilla:topic` `obituaries`

jfallows (changes requested):
At the moment, we are modeling `session` stream `abort` and `close` as the same behavior, then special casing `close` with `endEx` to handle the different behaviors.

It feels like we are working too hard to get the effect, and could instead take advantage of `abort` (potentially send will message) vs `close` (do not send will message), and trigger those accordingly from `mqtt` binding on `session` stream.

Potentially no need for `mqtt:endEx`, use `mqtt:abortEx` instead if we want to convey exact error code, and `close` implies normal disconnect.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -526,124 +840,68 @@ private void doMqttReset(
...
         {
             if (!MqttKafkaState.initialClosed(state))
             {
-                initialSeq = delegate.initialSeq;
-                initialAck = delegate.initialAck;
-                initialMax = delegate.initialMax;
                 state = MqttKafkaState.closeInitial(state);
 
-                doEnd(kafka, originId, routedId, initialId, initialSeq, initialAck, initialMax, traceId, authorization);
+                doEnd(kafka, originId, routedId, initialId, initialSeq, 0, 0, traceId, authorization);
```
jfallows, line 904: I think this was correct before the change, since we need to honor the values sent to use for `initialAck` and `initialMax` when sending the initial frames. There is no guarantee that it starts at `zero`.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -721,9 +972,7 @@ private void onKafkaBegin(
             state = MqttKafkaState.openingReply(state);
 
             assert replyAck <= replySeq;
-
-            delegate.doMqttBegin(traceId, authorization, affinity);
-            doKafkaWindow(traceId, authorization, 0, 0);
+            doKafkaWindow(traceId, authorization, 0, 0, 0);
```
jfallows, line 975: Is it necessary to send this `kafka` reply window proactively, or can we send it in reaction to `mqtt` reply window instead?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -741,30 +991,132 @@ private void onKafkaData(
...
+                final KafkaKeyFW key = kafkaMergedDataEx != null ? kafkaMergedDataEx.key() : null;
+
+                fetchWill:
+                if (key != null)
+                {
+                    if (payload == null)
+                    {
+                        //TODO: more elegant way for extracting clientId?
+                        final String clientId0 = key.value().get((b, o, m) -> b.getStringWithoutLengthUtf8(o, m - o))
+                            .split(WILL_SIGNAL_KEY_POSTFIX)[0];
```
jfallows, line 1016: Perhaps something like this?
```
final String clientId0 = key.value()
    .get((b, o, m) -> b.getStringWithoutLengthUtf8(o, m - o - WILL_SIGNAL_KEY_POSTFIX.length()));
```

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -741,30 +991,132 @@ private void onKafkaData(
...
+                progress.forEach(p -> newPartitions.add(p.partitionId()));
+                if (!newPartitions.equals(partitions))
+                {
+                    instanceId.regenerate();
+                    partitions = newPartitions;
+                }
+            }
+            else
+            {
+                //TODO: this means we lost all partitions during the assignment, so we need to regenerate, right?
```
jfallows, line 1085: I think this means we received a `FLUSH` with different or missing `KafkaMergedFlushEx` extension, right?

Perhaps we can afford to be more strict here, and not try to tolerate an edge case we don't need,

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -741,30 +991,132 @@ private void onKafkaData(
...
+            {
+                if (reconnectAt != NO_CANCEL_ID)
+                {
+                    signaler.cancel(reconnectAt);
+                }
+
+                reconnectAt = signaler.signalAt(
+                    currentTimeMillis() + SECONDS.toMillis(reconnectDelay),
+                    SIGNAL_INITIATE_WILL_KAFKA_STREAM,
+                    MqttKafkaSessionFactory.this::onWillKafkaStreamInitializationSignal);
```
jfallows, line 1118: Suggest hoisting this to a member field on the session factory, such as `handleReconnectWillStream`.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -783,41 +1135,1198 @@ private void onKafkaAbort(
...
+                        .partition(p -> p.partitionId(-1).partitionOffset(-1))
+                        .key(b -> b.length(key.length())
+                            .value(key.value(), 0, key.length()))
+                        .hashKey(b -> b.length(clientId.length())
+                            .value(clientId.value(), 0, clientId.length())))
+                    .build();
+
+                delegate.doKafkaData(traceId, authorization, kafkaWillDataEx);
+
+                String16FW willSignalKey = new String16FW(clientId.asString() + WILL_SIGNAL_KEY_POSTFIX);
```
jfallows, line 1535: Reuse flyweight instead of allocation.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -783,41 +1135,1198 @@ private void onKafkaAbort(
...
+            int padding,
+            int capabilities)
+        {
+            replyMax = bufferPool.slotCapacity();
+
+            doWindow(kafka, originId, routedId, replyId, replySeq, replyAck, replyMax,
+                traceId, authorization, budgetId, padding, replyPad, capabilities);
+        }
+
+        public void releaseBufferSlot(
```
jfallows, line 1506: `private`?

jfallows: Feel like this should happen naturally when trying to flush the buffered data when send is complete, or else stream aborts / resets and we need to cleanup the buffer.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -783,41 +1135,1198 @@ private void onKafkaAbort(
...
+                kafkaHeadersRW.item(h ->
+                {
+                    h.nameLen(helper.kafkaTimeoutHeaderName.sizeof());
+                    h.name(helper.kafkaTimeoutHeaderName);
+                    h.valueLen(4);
+                    h.value(expiryBuffer, 0, expiryBuffer.capacity());
+                });
+            }
+
+            if (will.contentType().asString() != null)
```
jfallows, line 1824: Suggest checking length for `-1` instead of doing a full utf-8 decode.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -783,41 +1135,1198 @@ private void onKafkaAbort(
...
+        return byteAt;
+    }
+
+    private static boolean isSetWillFlag(
+        int flags)
+    {
+        return (flags & MqttSessionFlags.WILL.value() << 1) != 0;
+    }
+
+    private static boolean isCleanStart(
```
jfallows, line 1984: Do we want to use `isSet...` for all flags, giving `isSetCleanStart(flags)`?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -783,41 +1135,1198 @@ private void onKafkaAbort(
...
+                break;
+            }
+        }
+        return byteAt;
+    }
+
+    private static boolean isSetWillFlag(
+        int flags)
+    {
+        return (flags & MqttSessionFlags.WILL.value() << 1) != 0;
```
jfallows, line 1981: Does this behave the way you want, with bit shift happening before bitwise and?

jfallows (changes requested):

incubator/binding-mqtt.spec/src/main/resources/META-INF/zilla/mqtt.idl
```diff
@@ -90,6 +105,8 @@ scope mqtt
         string16 contentType = null;
         MqttPayloadFormat format = BINARY;
         string16 responseTopic = null;
+        string16 lifetimeId = null;
+        string16 willId = null;
```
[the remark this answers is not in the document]
bmaidics: It's not used anywhere else. MqttMessage = MqttWillMessage. It was MqttWillMessage originally, but I remember we decided to rename it. I don't know for what reason though :)

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -301,29 +433,155 @@ private void onMqttData(
...
+                        .merged(m -> m
+                            .deferred(0)
+                            .timestamp(now().toEpochMilli())
+                            .partition(p -> p.partitionId(-1).partitionOffset(-1))
+                            .key(b -> b.length(clientId.length())
+                                .value(clientId.value(), 0, clientId.length())))
+                        .build();
+
+                    kafkaPayload = mqttSessionStateRO.tryWrap(buffer, offset, limit);
+                    if (kafkaPayload == null)
```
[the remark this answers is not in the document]
bmaidics: We're sending an empty payload from MQTT binding when we have clean start, so no, it's a valid payload right now.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -301,29 +433,155 @@ private void onMqttData(
...
+                    {
+                        kafkaPayload = EMPTY_OCTETS;
+                    }
+                    final int payloadSize = kafkaPayload == null ? 0 : kafkaPayload.sizeof();
+                    session.doKafkaData(traceId, authorization, budgetId,
+                        payloadSize, flags, kafkaPayload, kafkaDataEx);
```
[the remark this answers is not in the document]
bmaidics: For STATE type, yes. But for the WILL payload, no. We're generating the lifetimeId + willId and add those to the payload, so the size increases.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -341,8 +599,54 @@ private void onMqttEnd(
 
             assert initialAck <= initialSeq;
 
-            session.doKafkaEnd(traceId, initialSeq, authorization);
-            group.doKafkaEnd(traceId, initialSeq, authorization);
+            if (isSetWillFlag(flags))
+            {
+                // Cleanup will message + will signal
+                String16FW key = new String16FW(clientId.asString() + WILL_KEY_POSTFIX + lifetimeId);
```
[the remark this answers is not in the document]
bmaidics: What do you mean here by reusing the flyweight? We need to concat the key postfix and the lifetimeId. Or is there a better way for that with only flyweights?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
bmaidics: Actually git playing tricks here :). This is not a change, this a completely new code in the core0 will signal stream. I think we can assume that it's always 0, as there's no initial side for that. Or am I wrong?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
bmaidics: Nice, I like it.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
bmaidics: Yes. I did it this way so we don't need to change the flags coming from the network protocol, we can just pass it through the mqtt-kafka binding.
WILL_FLAG_MASK = 0b0000_0100;
CLEAN_START_FLAG_MASK = 0b0000_0010;
So in the mqtt idl: CLEAN_START(1), WILL(2)
Actually maybe it'd be better to provide the actual value, so we don't need shifting? So in this case 2 and 4? What do you think?

jfallows: I think it's already good like this. 👍

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
jfallows: As currently implemented, it appears as though we are counting `WINDOW` frames to determine when to release the buffer, `wasOpen` vs `!wasOpen`. But `WINDOW` frames can be logically split or combined into any number of frames, as long as the aggregate `seq`, `ack` are consistent, for example.

Therefore, I think we would need to treat it differently. Reacting to receiving `ack` with high enough value to release the buffer.

However, I think we can instead safely release the buffer a bit sooner, when the will message is fully delivered, perhaps on both `mqtt_messages` and `mqtt_retained` in the most complex case.

This would most easily be coded by detecting progress through the buffer independently on each stream, then notifying the delegate when complete, and having an `acquire` count on the delegate that is initialized once for each `mqtt_messages` and (optional) `mqtt_retained` stream, then decrements and does the actual `release` when the count reaches `zero`, so it doesn't matter which in order `mqtt_messages` vs `mqtt_retained` complete, or even if `mqtt_retained` is needed.

If an error occurs while delivering the will message to `mqtt_messages` and/or `mqtt_retained`, I think we would want to simply cascade that error to the `will` stream, `mqtt_messages` and `mqtt_retained`, triggering a reconnect for the `will` stream which will replay the `will` message.

Note that the `ack` to the `will` stream should be deferred until both `mqtt_messages` and optionally `mqtt_retained` (if will retained) have acknowledged.

jfallows (changes requested):

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -150,14 +241,45 @@ public MessageConsumer newStream(
...
         long bindingId)
     {
         sessionIds.remove(bindingId);
+
+        if (willProxy != null)
+        {
+            willProxy.doKafkaEnd(supplyTraceId.get(), 0);
+            signaler.cancel(reconnectAt);
+            reconnectAt = NO_CANCEL_ID;
+        }
```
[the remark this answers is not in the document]
bmaidics: One thing to add here: I can't use the remove in onDetached since we remove the binding in the MqttKafkaProxyFactory. This would require some restructuring of the code: moving the bindings in the individual factories like MqttKafkaSessionFactory etc.. Let me know if you think it's worth to change it.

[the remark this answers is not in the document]
bmaidics: Other then this I applied this feedback

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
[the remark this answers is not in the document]
bmaidics: Ohh I see. I changed the code a little bit. One thing though: how could we get a malformed sessionState? Can't we assume that since we're the one that's creating the sessionState in the mqtt-binding, it's never malformed?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/MqttKafkaConfiguration.java
```diff
@@ -70,29 +70,34 @@ public MqttKafkaConfiguration(
...
 
-    public Supplier<String> sessionId()
+    @FunctionalInterface
+    public interface StringSupplier extends Supplier<String>
     {
-        return () -> SESSION_ID.get(this);
     }
 
-    public Supplier<String> willId()
+    public StringSupplier sessionId()
```
jfallows, line 78: This is very close. The return type should be the more general `Supplier<String>` and the strongly typed `StringSupplier` should be `private`. Please see `KAFKA_CLIENT_SASL_SCRAM_NONCE` in `KafkaConfiguration` as an example.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/MqttKafkaConfiguration.java
```diff
@@ -105,106 +110,105 @@ public int willStreamReconnectDelay()
...
         Configuration config,
         String fullyQualifiedMethodName)
     {
-        Supplier<String> supplier = null;
+        StringSupplier supplier = null;
 
         try
         {
-            MethodType signature = MethodType.methodType(String.class);
+            MethodType signature = MethodType.methodType(StringSupplier.class);
```
jfallows, line 121: The referenced method is a `StringSupplier`, so it should return `String`, not `StringSupplier`.

Then the lambda returns `value` (String), not `value.get()`.

Same feedback for `decodeLongSupplier`.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/MqttKafkaConfiguration.java
```diff
@@ -105,106 +110,105 @@ public int willStreamReconnectDelay()
...
-        return supplier.getAsLong();
+        return supplier;
     }
 
-    private static String defaultInstanceId(
+    private static StringSupplier defaultInstanceId(
         Configuration config)
     {
-        return String.format("%s-%s", "zilla", UUID.randomUUID());
+        return () -> String.format("%s-%s", "zilla", UUID.randomUUID());
```
jfallows, line 188: Previous was correct. The default method is a supplier, does not return a supplier.
Same for other `StringSupplier` and `LongSupplier` defaults.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaBindingConfig.java
```diff
@@ -29,6 +30,7 @@ public class MqttKafkaBindingConfig
     public final KindConfig kind;
     public final MqttKafkaOptionsConfig options;
     public final List<MqttKafkaRouteConfig> routes;
+    public MqttKafkaSessionFactory.KafkaWillProxy willProxy;
```
jfallows, line 33: Suggest adding a blank line before this to separate from final fields, improves readability.

incubator/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
```diff
@@ -292,28 +373,28 @@ public void shouldReconnectWillStreamOnKafkaReset() throws Exception
...
     {
-        return "zilla-1";
+        return () -> "zilla-1";
     }
 
-    public static long supplyTime()
+    public static LongSupplier supplyTime()
     {
-        return 1000L;
+        return () -> 1000L;
```
jfallows, line 398: These static method changes can be reverted after applying `MqttKafkaConfiguration` feedback.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -104,7 +106,8 @@ public class MqttKafkaSessionFactory implements MqttKafkaStreamFactory
     private static final OctetsFW EMPTY_OCTETS = new OctetsFW().wrap(new UnsafeBuffer(new byte[0]), 0, 0);
     private static final int DATA_FLAG_COMPLETE = 0x03;
     private static final int SIGNAL_DELIVER_WILL_MESSAGE = 1;
-    private static final int SIGNAL_CONNECT_WILL_STREAM = 2;
+    public static final int SIGNAL_CONNECT_WILL_STREAM = 2;
+    public static final int SIZE_OF_UUID = 38;
```
jfallows, line 110: `private`

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -879,9 +874,13 @@ private KafkaWillProxy(
             this.replyId = supplyReplyId.applyAsLong(initialId);
             this.willFetchers = new Object2ObjectHashMap<>();
             this.partitions = new IntHashSet();
+            this.reconnectAt = signaler.signalAt(
+                currentTimeMillis(),
+                SIGNAL_CONNECT_WILL_STREAM,
+                this::onSignalConnectWillStream);
```
jfallows, line 880: This behavior should be moved out of the constructor to a separate method, as it is not really part of the object construction.

Suggest calling it something like `doKafkaBeginAt` and pass `currentTimeMillis()` as a parameter.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -1764,10 +1779,10 @@ private void onKafkaWindow(
                     signaler.signalAt(deliverAt, originId, routedId, initialId, SIGNAL_DELIVER_WILL_MESSAGE, 0);
                 willDeliverIds.computeIfAbsent(delegate.clientId, k -> new LongArrayList()).add(signalId);
             }
-            else
+            if (initialAck == delegate.messageSlotReserved)
```
jfallows, line 1782: This assumes `initialSeq` is always `0`, okay to assume?

jfallows (changes requested):

jfallows (approved):

PR #376, opened 2023-08-21
Feature/m1 docker build support

vordimous:
## Description

The primary issue I found was the `eclipse-temurin:20-alpine` issue only shows support for the `amd64` platform.

These changes are meant to improve the usability and cross-platform support for the zilla docker images.

The cross-platform changes are based on the [maven plugin docs](https://dmp.fabric8.io/#build-buildx) to add a property for the supported platforms. I replaced the alpine build image for the `eclipse-temurin:20` based on `ubuntu:jammy` a more suited build environment. The build image doesn't need to be as lean and is best to match the build purpose. This made building on both platforms able to use the same dockerfile. It isn't an industry standard to require apline images for all stages when using multistage builds.

command when running a build on an M1 architecture:
`./mvnw -DskipTests clean install -Ddocker.platforms=linux/arm64`

I included some metadata I have seen to be best practice. For now, a maintainer label and version env var following the [nginx image template](https://github.com/nginxinc/docker-nginx/blob/master/Dockerfile-debian.template). There are other recommended labels, but I didn't want to add too much.

The remaining challenge is when the alpine image copies the necessary binaries, they are not able to be found but the running container. I am not sure if it is an Ubuntu -> Alpine issue or a user ownership issue. If the base image for the app container is `ubuntu:jammy`, matching the build image, everything works as expected when both build and app images are alpine versions. 

Fixes # (issue)


attilakreiner:
> The remaining challenge is when the alpine image copies the necessary binaries, they are not able to be found but the running container. I am not sure if it is an Ubuntu -> Alpine issue or a user ownership issue. If the base image for the app container is ubuntu:jammy, matching the build image, everything works as expected when both build and app images are alpine versions.

I tried this approach and reached the same conclusion, unfortunately it doesn't work this way; the zilla built with zpm on one distro can't be executed on the other. I don't have a deep enough understanding to give a detailed explanation, but I think it has to do with the fact that the files/libraries required to run java are laid out differently on Ubuntu vs Alpine (even when using the same java distro/version).

vordimous:
example of multi OS build management: https://github.com/emqx/emqx-builder

vordimous:
After looking into how other repos manage this, I believe the solution is to publish a [multi-platform image](https://docs.docker.com/build/building/multi-platform/) as the base image tagged with the semver [x.x.x]. This would mean using the base images from this PR as the default images.

Then users who want to use the `alpine` version, which `eclipse-temurin:alpine` is only `linux/x86_64` compatible, could still use that build specifying the alpine suffix [x.x.x-alpine].

Implementing that would mean keeping both versions of docker files in this PR and specifying one as the default image and one as the alpine image. Then when building the default image, we would specify both platforms `-Ddocker.platforms=linux/arm64,linux/x86_64`, and when building the alpine version only the `-Ddocker.platforms=linux/x86_64`

@jfallows @attilakreiner what do you think?

jfallows:
This may also help if we decide to perform release of different platform docker images via GitHub Actions workflows.

https://github.com/github/roadmap/issues/528

vordimous:
arm support for alpine discussion: https://github.com/adoptium/containers/issues/158

vordimous:
Attemed a the suggested alternate image [bellsoft/liberica-openjdk-alpine-musl:20](https://hub.docker.com/layers/bellsoft/liberica-openjdk-alpine-musl/20/images/sha256-96e1607df14a8e58d826635688042ef8848b111ae9e090f9215f8bbd860af4c5?context=explore) and ran into a module not found error:  `Module jdk.jdwp.agent not found`

Looking at how adoptium builds their alpine images, they use their own java binaries and don't have an `aarch64_alpine` binary built.  https://github.com/adoptium/temurin20-binaries/releases

pom.xml
```diff
@@ -580,6 +580,8 @@
 
       <properties>
         <docker.context>incubator</docker.context>
+        <docker.platforms>linux/x86_64,linux/arm64</docker.platforms>
+        <docker.tagAlt>edge</docker.tagAlt>
```
vordimous, line 584: This naming can be updated, I just picked what I normally see in other docker repos. `edge` is a more common tag for the head of develop vs `developer-SNAPSHOT` which is specific to java.

pom.xml
```diff
@@ -614,6 +616,81 @@
...
+          <id>github</id>
+          <url>https://maven.pkg.github.com/aklivity/zilla/</url>
+        </snapshotRepository>
+      </distributionManagement>
+
+      <properties>
+        <docker.context>alpine</docker.context>
+        <docker.platforms>linux/x86_64</docker.platforms>
+        <docker.tagSuffix>-alpine</docker.tagSuffix>
+        <docker.tagAlt>alpine</docker.tagAlt>
```
vordimous, line 644: The alpine version is the only one to need a `tagSuffix` because it will share the same semver. When we want to start pushing the incubator build the `tagSuffix` can be set on build using a unique id like the commit hash or something.

pom.xml
```diff
@@ -637,6 +714,8 @@
 
       <properties>
         <docker.context>release</docker.context>
+        <docker.platforms>linux/x86_64,linux/arm64</docker.platforms>
+        <docker.tagAlt>latest</docker.tagAlt>
```
vordimous, line 718: This gets set here instead of manually setting it in the `docker-images` pom

vordimous:
@jfallows I think this PR is ready for review, but the builds are failing. I am not sure what to look for since the build works locally. Please advise.

vordimous:
The incubator `edge` alternate tag is to match a more docker friendly naming scheme for the latest development work. 

I am not in love with having the `incubator-alpine` directory but at least it is consistent with release? Local builds with then have both image version options if needed:
image

Any ideas on how to setup the pom differently are welcome.

cloud/docker-image/pom.xml
```diff
@@ -254,17 +254,89 @@
...
+                            <include>org/antlr/**</include>
+                            <include>org/sonatype/oss/**</include>
+                            <include>com/fasterxml/oss-parent/**</include>
+                            <include>com/fasterxml/jackson/**</include>
+                            <include>org/yaml/snakeyaml/**</include>
+                            <include>org/junit/**</include>
+                          </includes>
+                        </fileSet>
+                      </fileSets>
+                    </inline>
```
jfallows, line 321: This `inline` duplication is getting unwieldy. Lets hoist it out to a descriptor file in `src/main/assembly` instead that can be reused by each docker image build.

cloud/docker-image/pom.xml
```diff
@@ -254,17 +254,89 @@
...
             <image>
               <alias>zilla</alias>
-              <name>ghcr.io/aklivity/zilla:%v</name>
+              <name>ghcr.io/aklivity/zilla</name>
               <build>
                 <contextDir>${docker.context}</contextDir>
                 <tags>
-                  <tag>latest</tag>
+                  <tag>${docker.tag}</tag>
+                  <tag>${docker.tagAlt}</tag>
```
jfallows, line 333: We can specify `latest` tag explicitly, but only for `release` `ubuntu` image.

cloud/docker-image/pom.xml
```diff
@@ -254,17 +254,89 @@
...
+                      </fileSets>
+                    </inline>
+                  </assembly>
+                </assemblies>
+              </build>
+            </image>
             <image>
               <alias>zilla</alias>
-              <name>ghcr.io/aklivity/zilla:%v</name>
+              <name>ghcr.io/aklivity/zilla</name>
```
jfallows, line 328: We can restore the `%v` now given the feedback for the top level pom proposed changes I think.

cloud/docker-image/pom.xml
```diff
@@ -254,17 +254,89 @@
         <configuration>
           <filter>${*}</filter>
           <images>
```
jfallows, line 256: I think the main issue we need to figure out is how to control the merging of plugin configuration when release vs incubator profiles are activated.

By default, we have a single image built in the main pom, which could be the release or incubator images depending on local snapshot build vs performing the release build.

Then we want additional images to be included in the docker plugin configuration only for release, so those would be added in the release profile, but we need to use these maven hints to make sure they are merged as needed rather than replacing entirely.

See `advanced configuration inheritance` under https://maven.apache.org/pom.html#plugins, specifically `combine.self` and `combine.children`.

When testing the configuration, there's no need to execute the `docker-image` build, instead you can use the `help` plugin to see the effective pom observed by Maven.

```
./mvnw help:effective-pom
```
This will output the effective pom it uses to execute on the current project, which will let you try out different `combine.self` and `combine.children` hints, setting `-Drelease` from the command line to activate the release profile and disable the incubator profile to ensure all combinations look right before we do a full test.

jfallows (changes requested):

cloud/docker-image/pom.xml
jfallows: Thanks @vordimous.

There is usually only one pom per Maven project, and each Maven project has a `src/main` directory, so we'll need to get rid of the poms down in `src/main/docker/incubator` and `src/main/docker/release`.

Instead I think we can move the majority of the content of those poms into corresponding incubator and release profiles of the `docker-image` pom.

Also, I realized that docker build only lets files be copied from a directory local to the `Dockerfile`, so to avoid duplication of `zpm.json.template` and `zilla.properties` for each image, we might need to have more than one `Dockerfile` in the `release` directory, such as `Dockerfile` and `Dockerfile-alpine` perhaps.

I'll look into updating the pom and also see if we can override the name of the file used for `Dockerfile` in the `docker` maven plugin.

vordimous: > There is usually only one pom per Maven project, and each Maven project has a src/main directory, so we'll need to get rid of the poms down in src/main/docker/incubator and src/main/docker/release.

Yeah I was fairly certain you didn't want me to turn those into maven projects to have the child inheritance. I wasn't sure if there was a different way to do it or not. 

> Also, I realized that docker build only lets files be copied from a directory local to the Dockerfile, so to avoid duplication of zpm.json.template and zilla.properties for each image, we might need to have more than one Dockerfile in the release directory, such as Dockerfile and Dockerfile-alpine perhaps.

The plug-in does support filename `dockerFile` config, so we can use alpine.Dockerfile per the [naming convention](https://docs.docker.com/build/building/packaging/#filename)

jfallows (approved):

PR #387, opened 2023-09-01
Session expiry

bmaidics:


incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2073,12 +2200,72 @@ protected void doKafkaData(
             assert initialSeq <= initialAck + initialMax;
         }
 
+        protected void cancelExpirySignal(
```
jfallows, line 2203: `protected final` if we don't expect to override in subclass

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2073,12 +2200,72 @@ protected void doKafkaData(
...
+                            .name(TYPE_HEADER_NAME.value(), 0, TYPE_HEADER_NAME.length())
+                            .valueLen(EXPIRY_SIGNAL_NAME.length())
+                            .value(EXPIRY_SIGNAL_NAME.value(), 0, EXPIRY_SIGNAL_NAME.length())))
+                .build();
+
+            doKafkaData(traceId, authorization, 0, 0, DATA_FLAG_COMPLETE,
+                null, expirySignalKafkaDataEx);
+        }
+
+        protected void sendExpirySignal(
```
jfallows, line 2232: `protected final` if we don't expect to override in subclass

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2073,12 +2200,72 @@ protected void doKafkaData(
...
+        {
+            String16FW expirySignalKey = new String16FW.Builder()
+                .wrap(sessionSignalKeyBuffer, 0, sessionSignalKeyBuffer.capacity())
+                .set(delegate.clientId.asString() + EXPIRY_SIGNAL_KEY_POSTFIX, StandardCharsets.UTF_8).build();
+            Flyweight expirySignalKafkaDataEx = kafkaDataExRW
+                .wrap(extBuffer, 0, extBuffer.capacity())
+                .typeId(kafkaTypeId)
+                .merged(m -> m
+                    .deferred(0)
+                    .timestamp(now().toEpochMilli())
```
jfallows, line 2245: We have been using `System.currentTimeMillis()` everywhere, let's keep that consistent here too.
Please check for other occurrences of `Instant.now().toEpochMillis()` so they can use `System.currentTimeMillis()` instead.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2098,21 +2285,21 @@ private void sendWillSignal(
...
-                    .willId(delegate.willId)
-                    .instanceId(instanceId.instanceId())
+            final MqttSessionSignalFW willSignal =
+                mqttSessionSignalRW.wrap(sessionSignalBuffer, 0, sessionSignalBuffer.capacity())
+                    .will(w -> w.clientId(delegate.clientId)
+                        .delay(delegate.delay)
+                        .deliverAt(supplyTime.getAsLong() + delegate.delay)
+                        .lifetimeId(delegate.lifetimeId)
+                        .willId(delegate.willId)
+                        .instanceId(instanceId.instanceId()))
```
jfallows, line 2295: `.build()` should be aligend with `.instanceId()` above it.
Also suggest, moving `.clientId(...)` to next line for readability.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2098,21 +2285,21 @@ private void sendWillSignal(
...
                     .build();
 
             doKafkaData(traceId, authorization, 0, willSignal.sizeof(), DATA_FLAG_COMPLETE,
                 willSignal, willSignalKafkaDataEx);
         }
 
-        private void doKafkaData(
+        protected void doKafkaData(
```
jfallows, line 2302: `protected final` if we don't expect to override in subclass

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2354,7 +2544,7 @@ private void onKafkaReset(
             assert acknowledge <= sequence;
             assert acknowledge >= delegate.initialAck;
 
-            delegate.initialAck = acknowledge;
+            delegate.initialAck = acknowledge - delegate.sessionPadding;
```
jfallows, line 2547: This does not seem right.
Padding is used to increase the amount of reserved sent with DATA, and that is fully ack'd.
What is the rationale here? Perhaps trying to solve a different problem?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2516,7 +2706,7 @@ protected void handleKafkaData(
                 kafkaDataEx != null && kafkaDataEx.kind() == KafkaDataExFW.KIND_MERGED ? kafkaDataEx.merged() : null;
             final KafkaKeyFW key = kafkaMergedDataEx != null ? kafkaMergedDataEx.key() : null;
 
-            if (key != null)
+            if (key != null && payload != null)
             {
                 if (key.length() == (delegate.clientId.length()))
```
jfallows, line 2711: Please remove redundant parens. I'm surprised this didn't show up via checkstyle in local build.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2516,7 +2706,7 @@ protected void handleKafkaData(
...
                 {
```
jfallows, line 2712: Lets pull out `int keyLen = key.length()` once instead of calling it multiple times in the `if-else-if`.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2842,7 +3048,24 @@ private void onKafkaBegin(
 
             assert replyAck <= replySeq;
 
-            delegate.doMqttBegin(traceId, authorization, affinity);
+            Flyweight mqttBeginEx = EMPTY_OCTETS;
+
+            //TODO: decide where to do this? How will the group stream reject the invalid timeout?
```
jfallows, line 3053: The group stream will not reject the invalid timeout, instead it will send back a `KafkaGroupBeginEx` with a timeout that works, ideally the same as the one requested, but possibly updated to fall with the supported range of kafka consumer group session timeout values.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2842,7 +3048,24 @@ private void onKafkaBegin(
...
+            final int sessionExpiryMillisNew =
```
jfallows, line 3054: `sessionExpiryMillisInRange` or `sessionExpiryMillisBounded`

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2946,7 +3169,7 @@ private void onKafkaReset(
             assert acknowledge <= sequence;
             assert acknowledge >= delegate.initialAck;
 
-            delegate.initialAck = acknowledge;
+            delegate.initialAck = acknowledge - delegate.sessionPadding;
```
jfallows, line 3172: Same feedback as above.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -3451,6 +3682,7 @@ private MessageConsumer newGroupStream(
         final KafkaBeginExFW kafkaBeginEx =
             kafkaBeginExRW.wrap(writeBuffer, BeginFW.FIELD_OFFSET_EXTENSION, writeBuffer.capacity())
                 .typeId(kafkaTypeId)
+                // TODO: this will be rejected if if sessionExpiryMs > max group timeout in Kafka
```
jfallows, line 3685: As mentioned above, this will not be rejected, reply `KafkaGroupBeginEx` will provide a value in the supported range.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1259,7 +1259,8 @@ private final class MqttServer
         private boolean connected;
 
         private short topicAliasMaximum = 0;
-        private int sessionExpiryInterval = 0;
+        private int connectSessionExpiryInterval = 0;
+        private int sessionExpiryIntervalLimit = Integer.MAX_VALUE;
```
jfallows, line 1263: Can we remove this now, as it is effectively unbounded?

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1835,7 +1835,7 @@ private void resolveSession(
                 .typeId(mqttTypeId)
                 .session(sessionBuilder -> sessionBuilder
```
jfallows, line 1836: Please prefer single letter variables for lambdas, so `s -> s` here.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1835,7 +1835,7 @@ private void resolveSession(
...
                     .flags(flags)
-                    .expiry(sessionExpiryInterval)
+                    .expiry(connectSessionExpiryInterval)
```
jfallows, line 1838: The previous name `sessionExpiryInterval` seems better, agree?

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2244,22 +2244,74 @@ private void onDecodeDisconnect(
...
+                final MqttPropertyFW mqttProperty = mqttPropertyRO.wrap(decodeBuffer, decodeProgress, decodeLimit);
+                switch (mqttProperty.kind())
+                {
+                case KIND_SESSION_EXPIRY:
+                    if (isSetSessionExpiryInterval(propertyMask))
+                    {
+                        reasonCode = PROTOCOL_ERROR;
+                        break decode;
+                    }
+                    this.propertyMask |= CONNECT_SESSION_EXPIRY_INTERVAL_MASK;
```
jfallows, line 2296: Fields used to track decode state should be prefixed, `decodableXxx`, so this should be `decodablePropertyMask`.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2244,22 +2244,74 @@ private void onDecodeDisconnect(
...
+                    final int sessionExpiryInterval = (int) mqttProperty.sessionExpiry();
+                    if (sessionExpiryInterval > sessionExpiryIntervalLimit)
+                    {
+                        reasonCode = PROTOCOL_ERROR;
+                        break decode;
+                    }
```
jfallows, line 2302: This block can be removed now, right?

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2244,22 +2244,74 @@ private void onDecodeDisconnect(
...
+                switch (mqttProperty.kind())
+                {
+                case KIND_SESSION_EXPIRY:
+                    if (isSetSessionExpiryInterval(propertyMask))
+                    {
+                        reasonCode = PROTOCOL_ERROR;
+                        break decode;
+                    }
+                    this.propertyMask |= CONNECT_SESSION_EXPIRY_INTERVAL_MASK;
+                    final int sessionExpiryInterval = (int) mqttProperty.sessionExpiry();
```
jfallows, line 2297: Perhaps let's call it `sessionExpiry` as `sessionExpiryInterval` is a bit of a mouthful and we already made this decision on the method name `.sessionExpiry()`.

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -3203,7 +3255,17 @@ private void onSessionBegin(
...
+
+                final OctetsFW extension = begin.extension();
+                if (extension.sizeof() > 0)
+                {
+                    final MqttBeginExFW mqttBeginEx = extension.get(mqttBeginExRO::tryWrap);
+
+                    assert mqttBeginEx.kind() == MqttBeginExFW.KIND_SESSION;
+                    final MqttSessionBeginExFW mqttSessionBeginEx = mqttBeginEx.session();
+
+                    sessionExpiryIntervalLimit = mqttSessionBeginEx.expiry();
```
jfallows, line 3267: This is just the `sessionExpiry`, right? Do we need to track it as a limit separately for some reason?

jfallows (changes requested):

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/MqttKafkaConfiguration.java
```diff
@@ -61,6 +63,8 @@ public class MqttKafkaConfiguration extends Configuration
             MqttKafkaConfiguration::decodeLongSupplier, MqttKafkaConfiguration::defaultTime);
         WILL_AVAILABLE = config.property("will.available", true);
         WILL_STREAM_RECONNECT_DELAY = config.property("will.stream.reconnect", 2);
+        SESSION_EXPIRY_INTERVAL_MAX = config.property("session.expiry.interval.max", 30000);
+        SESSION_EXPIRY_INTERVAL_MIN = config.property("session.expiry.interval.min", 1000);
```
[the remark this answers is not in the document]
bmaidics: I think we discussed that until it's ready we should have configs.

incubator/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
```diff
@@ -60,11 +62,46 @@ public class MqttKafkaSessionProxyIT
...
+    @Configure(name = SESSION_EXPIRY_INTERVAL_MAX_NAME, value = "30000")
+    @Configure(name = SESSION_EXPIRY_INTERVAL_MIN_NAME, value = "2000")
+    @Specification({
+        "${mqtt}/session.connect.override.min.session.expiry/client",
+        "${kafka}/session.connect.override.min.session.expiry/server"})
+    public void shouldConnectServerOverridesSessionExpiryTooSmall() throws Exception
+    {
+        k3po.finish();
+    }
+
```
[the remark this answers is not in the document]
bmaidics: Exactly, we can change this test to that once it's ready

incubator/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/session.cancel.session.expiry/server.rpt
```diff
@@ -0,0 +1,73 @@
...
+                               .partition(-1, -1)
+                               .key("client-1#will-signal")
+                               .header("type", "will-signal")
+                               .build()
+                           .build()}
+write ${mqtt:sessionSignal()
+            .expiry()
+            .clientId("client-1")
+            .delay(1000)
+            .expireAt(expireAt)
```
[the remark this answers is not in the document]
bmaidics: Ohh, you're right. It was just a mistake to put 1000 as delay. expireAt doesn't need to be reconfigured, I just created the property so we can use it in the client side of the script.

incubator/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
```diff
@@ -198,6 +247,56 @@ public void shouldGroupStreamReceiveServerSentReset() throws Exception
...
+        value = "io.aklivity.zilla.runtime.binding.mqtt.kafka.internal.stream.MqttKafkaSessionProxyIT::supplyInstanceId")
+    @Configure(name = TIME_NAME,
+        value = "io.aklivity.zilla.runtime.binding.mqtt.kafka.internal.stream.MqttKafkaSessionProxyIT::supplyTime")
+    @Specification({
+        "${kafka}/session.cancel.session.expiry/server"})
+    public void shouldCancelSessionExpiry() throws Exception
+    {
+        k3po.start();
+        Thread.sleep(1000);
+        k3po.notifyBarrier("WAIT_1_SECOND");
```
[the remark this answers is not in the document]
bmaidics: By this I'm just simulating the client reconnecting after 1 second. But TBH we can just simulate the reconnection right after the expiry signal was sent. Do you think that's better?

incubator/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/network/disconnect.invalid.session.expiry/server.rpt
```diff
@@ -0,0 +1,43 @@
...
+      [0x00]                  # flags = none
+      [0x00]                  # reason code
+      [0x05]                  # properties
+      [0x27] 66560            # maximum packet size = 66560
+
+read  [0xe0 0x02]             # DISCONNECT
+      [0x00]                  # normal disconnect
+      [0x00]                  # properties = none
+
+read closed
```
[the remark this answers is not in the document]
bmaidics: Actually, this test is not needed anymore (I think). It supposed to test that if the client sends the session expiry in the CONNECT as 0, but sends non-zero in the disconnect it's a protocol error. Since we'll always override the 0 expiry in the CONNACK, I'm not sure if this test has any value. If we'd like to cover this scenario, we'd need to remember the original expiry, not just the "final" one.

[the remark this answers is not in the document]
bmaidics: Removed this test so far, let me know what you think.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
bmaidics: Nice, by reject I meant restrict, but it was the wrong word phrase for sure. It'll be an easy change once it's ready

incubator/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
bmaidics: Nope, fixed it

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
bmaidics: When we receive a reset, we set the delegate.initialAck to the reset frame ack, which is the greater then what mqtt sent. E.g: we don't have will message in the mqtt begin frame, so we only send the expiry cancellation + expire later signal, which is 32 bytes. This will be acked by Kafka, so onKafkaReset the resetl.acknowledge is 32. But mqtt didn't send any data, so delegate.initialSeq = 0. So maybe in reality, we'd need to remove the "delegate.initialAck = acknowledge;" assignment?

jfallows: Yeah, I think this raises a more significant point about the relationship between the `mqtt` session stream and the multiple `kafka` streams that see the same `mqtt` session stream as their `delegate`.

In general, each stream should be managing its own `initialSeq`, `initialAck`, etc locally to the stream.

In special cases, say when we are proxying 1-to-1 between streams, then it is feasible to track `initialSeq`, `initalAck`, etc on one stream only and use the same values for both.

However, in this case, it seems like with the 1-to-many relationship from `mqtt` session stream to `kafka` streams, we are no longer proxying in the special case, so we are trying to handle the divergence on a case by case basis.

I realize the design has evolved a bit during implementation. Perhaps we want to take one of those `kafka` streams and treat them as the primary, similar to your suggestion of not updating the delegate in the non-primary `kafka` stream.  If that is sufficient, then we still need to track `initialSeq`, `initialAck`, etc separately in the non-primary `kafka` stream as it will likle yneed to use its own values for sending its frames.

Let's perhaps also rethink the naming, so that a stream is not named as a `...Proxy` if it is not behaving as a proxy. In some cases, perhaps we should be naming such classes `...Stream` instead. There may be other good alternative name suffixes too, that might more clearly help to convey the intent of the design.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -697,10 +696,10 @@ private void onMqttAbort(
             }
             final MqttSessionSignalFW expirySignal =
                 mqttSessionSignalRW.wrap(sessionSignalBuffer, 0, sessionSignalBuffer.capacity())
-                    .expiry(w -> w.clientId(clientId)
+                    .expiry(w -> w.instanceId(instanceId.instanceId())
```
jfallows, line 699: Please move `.instanceId(...)` to next line, same as later in this class.

incubator/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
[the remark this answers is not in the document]
jfallows: Yes, we should aim to avoid adding delays into test execution during the build if avoidable please.

jfallows (changes requested):
Looks great! Just a few minor comments, plus the reply about flow control.

incubator/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/network/disconnect.invalid.session.expiry/server.rpt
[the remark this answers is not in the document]
bmaidics: TODO: Put back the test.

jfallows (approved):

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -1026,49 +1106,95 @@ private void onKafkaData(
...
                 final KafkaKeyFW key = kafkaMergedDataEx != null ? kafkaMergedDataEx.key() : null;
 
-                fetchWill:
+                reactToSignal:
                 if (key != null)
                 {
                     if (payload == null)
                     {
+                        final OctetsFW typeHeaderNameOctets =
+                            new OctetsFW().wrap(TYPE_HEADER_NAME.value(), 0, TYPE_HEADER_NAME.length());
```
jfallows, line 1115: Reuse instead of allocate `new OctetsFW()`.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -1026,49 +1106,95 @@ private void onKafkaData(
...
+                        final String type = kafkaMergedDataEx.headers().matchFirst(h -> h.name().equals(typeHeaderNameOctets))
+                            .value().get((b, o, m) -> b.getStringWithoutLengthUtf8(o, m - o));
+
+                        final String keyPostfix = type.equals(WILL_SIGNAL_NAME.asString()) ?
+                            WILL_SIGNAL_KEY_POSTFIX : EXPIRY_SIGNAL_KEY_POSTFIX;
```
jfallows, line 1120: `type` can be `OctetsFW` and `WILL_SIGNAL_NAME` can be preallocated as `OctetsFW` so that .equals still works, no need for `.asString()`.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -1026,49 +1106,95 @@ private void onKafkaData(
...
+
                         final String clientId0 = key.value()
-                            .get((b, o, m) -> b.getStringWithoutLengthUtf8(o, m - o - WILL_SIGNAL_KEY_POSTFIX.length()));
-                        String16FW clientId = new String16FW(clientId0);
-                        if (willDeliverIds.containsKey(clientId))
+                            .get((b, o, m) -> b.getStringWithoutLengthUtf8(o, m - o - keyPostfix.length()));
+                        final String16FW clientId = new String16FW(clientId0);
```
jfallows, line 1124: If the header values we need to compare to are `OctetsFW`, then we should be storing things in terms of `OctetsFW`, not `StringFW`, etc?

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -1026,49 +1106,95 @@ private void onKafkaData(
...
                             {
-                                deliverAt = supplyTime.getAsLong() + willSignal.delay();
+                                if (!instanceId.instanceId().equals(willSignal.instanceId()))
+                                {
+                                    deliverAt = supplyTime.getAsLong() + willSignal.delay();
+                                }
+                                else
+                                {
+                                    break reactToSignal;
+                                }
```
jfallows, line 1164: Probably we can simplify this with putting the break first (with inverted condition), then the `deliverAt = ...` doesn't need an `else` around it.

Actually, if you do that then the if conditions can be combined, so we don't need nested ifs either.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -1026,49 +1106,95 @@ private void onKafkaData(
...
                             {
-                                break fetchWill;
+                                if (!instanceId.instanceId().equals(expirySignal.instanceId()))
+                                {
+                                    expireAt = supplyTime.getAsLong() + expirySignal.delay();
+                                }
+                                else
+                                {
+                                    break reactToSignal;
+                                }
```
jfallows, line 1187: This can be simplified similar to above nested ifs.

incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -1026,49 +1106,95 @@ private void onKafkaData(
...
                             }
-                        }
 
-                        KafkaFetchWillProxy willFetcher = new KafkaFetchWillProxy(originId, routedId, this, sessionsTopic,
-                            clientId, willSignal.willId().asString(), willSignal.lifetimeId().asString(), deliverAt);
-                        willFetcher.doKafkaBegin(traceId, authorization, 0, willSignal.lifetimeId());
-                        willFetchers.put(clientId, willFetcher);
+                            final int contextId = System.identityHashCode(expiryClientId.asString().intern());
```
jfallows, line 1190: Let's avoid calling `String.intern()` on values specified externally, like clientIds.
Instead we can use a counter stored at the factory, or similar.

jfallows (changes requested):

jfallows (approved):

PR #399, opened 2023-09-08
Validator

attilakreiner:
## Description

Fixes # (issue)

### DIFF
--- a/specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
+++ b/specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
@@ -237,6 +237,114 @@
                                 {
                                     "type": "string"
                                 }
+                            },
+                            "requests":
+                            {
+                                "type": "array",
+                                "items":
+                                {
+                                    "type": "object",
+                                    "properties":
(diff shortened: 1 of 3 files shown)


specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
```diff
@@ -237,6 +237,114 @@
...
+                                            ]
+                                        },
+                                        {
+                                            "required":
+                                            [
+                                                "path",
+                                                "content"
+                                            ]
+                                        }
+                                    ],
```
jfallows, line 345: Looks like `path` is just always required, right?

jfallows (changes requested):

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
attilakreiner: I meant to express **path and (headers or params or content)** is required. We can simplify this to path is required ➡️ in that case people will be able to specify just a **path**, I guess it will not cause any issues, it will just be a no-op. Pls LMK.

PR #407, opened 2023-09-12
Fix kafka binding jpms module dependencies

jfallows:
## Description

Fix issue with kafka binding module dependencies for zpm install.


jfallows:
Use #411 shaded approach instead.

PR #412, opened 2023-09-14
Schema syntax validation for validator config 

attilakreiner:
## Description

Introduce schema syntax validation for validator config in the engine and use this in the kafka binding.

Fixes # (issue)


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/config/SchemaConfig.java
```diff
@@ -0,0 +1,36 @@
...
+ * under the License.
+ */
+package io.aklivity.zilla.runtime.binding.kafka.internal.validator.config;
+
+public class SchemaConfig
+{
+    public String schema;
+    public String strategy;
+    public String version;
+    public int id;
```
jfallows, line 23: Please make these fields `final`.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/config/SchemaConfig.java
attilakreiner: done

jfallows (approved):

PR #413, opened 2023-09-14
Produce Validator

aDaemonThread:


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -511,8 +511,8 @@ public void writeEntryFinish(
             if (type.value != null)
             {
                 OctetsFW value = headEntry.value();
-                if (value != null &&
-                    !type.value.validate(value.value(), value.offset(), value.sizeof()))
+                if (value == null ||
+                    !type.value.read(value.value(), value.offset(), value.sizeof()))
```
jfallows, line 515: This is saying that `value == null` is a validation failure, but `tombstone` messages have `null` value in Kafka, so log compacted topics would always fail validation.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -630,6 +630,29 @@ public void writeProduceEntryFin(
...
+        OctetsFW value = headEntry.value();
+        if (value == null ||
+            !type.value.write(value.value(), value.offset(), value.sizeof()))
+        {
+            status = false;
+            System.out.println("Value Validation failed");
+        }
+
+        return status;
+    }
```
jfallows, line 654: Where is key validation during produce?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientProduceFactory.java
```diff
@@ -681,6 +688,22 @@ private void onClientInitialData(
...
+                        if (entryKey != null ||
+                            !keyValidator.write(entryKey.value(), entryKey.offset(), entryKey.sizeof()))
+                        {
+                            System.out.println("Key Validation failed");
+                            error = ERROR_INVALID_RECORD;
+                            break init;
+                        }
+                    }
+                }
+
```
jfallows, line 706: This should be in `partition.writeProduceEntryStart` or else something similar to `partition.validProduceEntry` instead of spreading the validation both inside and outside the cache.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
```diff
@@ -2070,6 +2071,14 @@ private void onProducePartitionLeaderError(
                     produceStreams.remove(leader);
                 }
             }
+            else if (error == ERROR_INVALID_RECORD)
+            {
+                doMergedInitialReset(traceId, ex -> ex.set((b, o, l) -> kafkaResetExRW.wrap(b, o, l)
+                        .typeId(kafkaTypeId)
+                        .error(error)
+                        .build()
+                        .sizeof()));
```
jfallows, line 2080: This is a change of behavior on error, such that the same checks in `doMergedInitialResetIfNecessary()` are being bypassed when we call `doMergedInitialReset()` directly.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/config/AvroValidatorConfig.java
```diff
@@ -22,11 +22,14 @@
 public final class AvroValidatorConfig extends ValidatorConfig
 {
     public final List<KafkaCatalogConfig> catalogs;
+    public final String topic;
```
jfallows, line 25: This should be `name` not `topic`, right?

jfallows (changes requested):

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/config/cache.client.options.value.type.avro.yaml
```diff
@@ -0,0 +1,48 @@
...
+# under the License.
+#
+
+---
+name: test
+catalogs:
+  test0:
+    type: test
+    options:
+      schema: "{\"fields\":[{\"name\":\"id\",\"type\":\"string\"},{\"name\":\"status\",\"type\":\"string\"}],\"name\":\"Event\",\"namespace\":\"io.aklivity.example\",\"type\":\"record\"}"
```
jfallows, line 23: Please use YAML mult-line string syntax to make this more readable.
See https://yaml-multiline.info/

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/AvroValidator.java
```diff
@@ -33,13 +32,17 @@
...
 
 public final class AvroValidator implements Validator
 {
     private static final byte MAGIC_BYTE = 0x0;
 
     private final List<KafkaCatalogConfig> catalogs;
+    private final KafkaCatalogConfig catalog;
     private final Long2ObjectHashMap<CatalogHandler> handlersById;
+    private final String keySubject;
+    private final String valueSubject;
```
jfallows, line 45: When we associate the validator via kafka binding options, we do so in the context of the key or the value, not both at once.

Shouldn't we have a single subject here, populated with the appropriate name, then no need for `isKey` in any validator method signature to split it back apart again?

Remember that the validator abstraction is generic, so it should not have any Kafka-specific concepts like `key` in the method signatures, as that doesn't make sense for other non-kafka protocol bindings.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/AvroValidator.java
```diff
@@ -73,25 +79,77 @@ public boolean validate(
...
+            if (schemaId > 0)
             {
-                System.out.println("Unknown magic byte!");
-                break tryValidate;
+                parsedSchema = handler.resolve(schemaId);
+
+            }
+            else if (catalog.strategy.equals("topic"))
+            {
+                parsedSchema = handler.resolve(isKey ? keySubject : valueSubject, catalog.version);
```
jfallows, line 125: We discussed having `handler.resolve(subject, version)` return the `schemaId` so that `handler.resolve(schemaId)` would be used in all cases.

If we do that, then it seems we don't need `ParsedSchema`, agree?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/validator/config/AvroValidatorConfig.java
```diff
@@ -22,11 +22,14 @@
 public final class AvroValidatorConfig extends ValidatorConfig
 {
     public final List<KafkaCatalogConfig> catalogs;
+    public final String name;
```
jfallows, line 25: What's the rationale for this change?

jfallows (changes requested):

PR #414, opened 2023-09-14
Fix consumer assignment causing decoding issue

akrambek:
## Description

Fixes # (issue)


jfallows (approved):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheServerConsumerFactory.java
```diff
@@ -1031,17 +1031,19 @@ private void doPartitionAssignment(
...
                         String consumerId = members.get(member);
                         List<TopicPartition> topicPartitions = consumers.computeIfAbsent(
                             member, tp -> new ArrayList<>());
                         IntHashSet partitions = new IntHashSet();
 
-                        IntHashSet.IntIterator iterator = p.iterator();
                         for (; partitionIndex < newPartitionPerTopic; partitionIndex++)
                         {
-                            partitions.add(iterator.next());
+                            final Integer partitionId = iterator.next();
```
jfallows, line 1045: Please use `.nextValue()` instead of `.next()` on `IntIterator` to avoid boxing to `Integer`.

jfallows (changes requested):

jfallows (approved):

PR #424, opened 2023-09-17
Buffer fragmented kafka session signal messages

bmaidics:


incubator/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -912,6 +914,8 @@ public final class KafkaSignalStream
         private long replyAck;
         private int replyMax;
         private long reconnectAt;
+        private int sessionSignalDecodeSlot = NO_SLOT;
+        private int sessionSignalDecodeSlotOffset;
```
jfallows, line 918: Rename to `decodeSlot` and `decodeSlotOffset` as we are in scope of `KafkaSignalStream`.

jfallows (changes requested):

jfallows (approved):

PR #434, opened 2023-09-20
Produce Validator

aDaemonThread:
Contain mentioned changes::

1. Spec and implementation changes to support both karapace and schema-registry
2. Produce validator Spec and implementation
3. Validator refactoring
4. Docker build fix
5. Minor refactoring of `KafkaMergedFactory`


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaTopicConfigAdapter.java
```diff
@@ -96,7 +97,12 @@ public KafkaTopicConfig adaptFromJson(
 
         if (key != null)
         {
-            keyConfig = validator.adaptFromJson(key);
+            JsonObjectBuilder keyObject = Json.createObjectBuilder();
+
+            key.forEach(keyObject::add);
+            keyObject.add(SUBJECT, name + "-key");
```
jfallows, line 103: This is a helpful way to handle it, appending the correct suffix in the context of key vs value.
However, this is no longer round-tripping correctly after adapting to and from JSON.
Instead, let's apply the suffix where the value is used, and keep the parsing code generic, reading and writing the values verbatim.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaTopicConfigAdapter.java
jfallows: After digging into this a bit more, I realized the fix has deeper impact than we need at the moment, so we can go with this approach for now.

jfallows (approved):

jfallows (approved):

PR #436, opened 2023-09-20
Add affinity to mqtt server and client binding

bmaidics:


specs/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/application/subscribe.qos0.publish.retained.no.replay/client.rpt
```diff
@@ -17,6 +17,7 @@
 connect "zilla://streams/app0"
          option zilla:window 8192
          option zilla:transmission "duplex"
+         option zilla:affinity 0xa1
```
jfallows, line 20: Suggest we use `0xc1`, `0xc2` etc for different "clients".
In kafka, we used `0xb1`, `0xb2` etc for different "brokers". 😄 

These different values would be triggered by different client connections on network side, is that consistent with those scripts' behavior?

jfallows (changes requested):

jfallows (approved):

PR #439, opened 2023-09-21
Mqtt subscription handling bugfix

bmaidics:


runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2037,11 +2037,15 @@ private void onDecodeSubscribe(
                             break;
                         }
 
+                        if (!containsSubscriptionId)
+                        {
+                            subscriptionId = supplySubscriptionId.getAsInt();
```
jfallows, line 2042: There is potential for this generated subscriptionId to collide with client specified subscriptionIds.

Given that subscriptionId is variable byte integer, which has a range of `0 - 268435455`, we need to use values outside that range for ephemeral subscriptionIds generated by us for internal use only.

Since `268435455` is `0x7FFF_FFFF`, we can mask in some high bits to avoid the collision and also tell that its system generated.

To keep this simple, we can mask in `0x70` to set bits 30, 29 and 28 - leaving bit 31 unset so the value remains a positive integer in Java signed int type.

We also need to decide how to handle system generated subscriptionIds in `mqtt client` where they need to be on the wire. These would not get encoded, since the originally received protocol did not include them, so proxying through without a subscriptionId makes sense.

jfallows (changes requested):

jfallows (approved):

PR #447, opened 2023-09-22
Cache support for Schema Registry

aDaemonThread:


incubator/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/SchemaRegistryCatalogHandler.java
```diff
@@ -70,8 +78,18 @@ public int register(
...
+        String schema;
+        if (cache.containsKey(schemaId))
+        {
+            schema = cache.get(schemaId);
+        }
+        else
+        {
+            String response = sendHttpRequest(MessageFormat.format(SCHEMA_PATH, schemaId));
+            schema = response != null ? request.resolveSchemaResponse(response) : null;
+            cache.put(schemaId, schema);
```
jfallows, line 90: `null` values are not supported by `Int2ObjectCache`, so need to make sure not to put `null` value.
`get` returns `null` if cache miss

```
String schema = cache.get(schemaId);

if (schema == null)
{
    String response = sendHttpRequest(MessageFormat.format(SCHEMA_PATH, schemaId));
    if (response != null)
    {
        schema = request.resolveSchemaResponse(response);
        cache.put(schemaId, schema);
    }
}

return schema;
```

jfallows (changes requested):

jfallows (approved):

PR #467, opened 2023-09-24
Include validators in the mqtt config generator

### LINKED ISSUE #461: Generate `mqtt` server publish `validators` from `AsyncAPI` specification
Requires #453 and #454 for `inline` catalog with `json` validator.

attilakreiner:
## Description

Include validators in the `mqtt` config generator.

Fixes #461


incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/asyncapi/model/Channel.java
```diff
@@ -14,11 +14,14 @@
...
 package io.aklivity.zilla.runtime.command.config.internal.asyncapi.model;
 
+import java.util.LinkedHashMap;
+
 import jakarta.json.bind.annotation.JsonbProperty;
 
 public class Channel
 {
     public String address;
+    public LinkedHashMap<String, Message> messages;
```
jfallows, line 24: This should just be `Map<String, Message>` right?

jfallows (changes requested):

incubator/command-config/src/main/java/io/aklivity/zilla/runtime/command/config/internal/asyncapi/model/Channel.java
attilakreiner: It's necessary to maintain the order here for a deterministic output. 

- That's needed for the consistent run of the unit tests.
- I believe it's a good practice for the production code, too.

jfallows (approved):

PR #473, opened 2023-09-25
Json Validator and Inline Schema fix

aDaemonThread:


incubator/catalog-inline/src/main/java/io/aklivity/zilla/runtime/catalog/inline/config/InlineOptionsConfigAdapter.java
```diff
@@ -27,6 +27,7 @@ public class InlineOptionsConfigAdapter implements OptionsConfigAdapterSpi, Json
     private static final String SUBJECTS_NAME = "subjects";
     private static final String VERSION_NAME = "version";
     private static final String SCHEMA_NAME = "schema";
+    private static final String LATEST = "latest";
```
jfallows, line 30: Use `VERSION_DEFAULT` as constant name instead of `LATEST`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/SchemaConfigAdapter.java
```diff
@@ -26,6 +26,7 @@ public class SchemaConfigAdapter implements JsonbAdapter<SchemaConfig, JsonObjec
     private static final String STRATEGY_NAME = "strategy";
     private static final String SUBJECT_NAME = "subject";
     private static final String VERSION_NAME = "version";
+    private static final String LATEST = "latest";
```
jfallows, line 29: Use `VERSION_DEFAULT` for constant name.

jfallows (changes requested):

jfallows (approved):

PR #493, opened 2023-10-05
Mqtt topic space

bmaidics:


specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/subscribe.topic.space/server.rpt
```diff
@@ -0,0 +1,76 @@
...
+                                    .headers("zilla:filter")
+                                        .sequence("device")
+                                        .skip(1)
+                                        .build()
+                                    .build()
+                                .evaluation("EAGER")
+                                .build()
+                            .build()}
+
+connected
```
jfallows, line 76: We'll need to verify subscription ids work properly as messages are delivered from different kafka topics in the topic space, potentially matching topic filters within a specific kafka topic, as well as a higher level filter pattern spanning more than one kafka topic.

This will also come up for `retained` later too as we'll need to retrieve retained messages across the topic space matching the topic filter when we need to replay retained messages on subscribe.

specs/binding-mqtt-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/mqtt/kafka/config/SchemaTest.java
```diff
@@ -48,4 +48,12 @@ public void shouldValidateProxyWithOptions()
 
         assertThat(config, not(nullValue()));
     }
+
+    @Test
+    public void shouldValidateProxyWhenMqttTopicWithKafkaTopic()
```
jfallows, line 53: This might simplify to `shouldValidateProxyWhenTopicWithMessages` if we rename `with` `topic` to `messages` as mentioned above.

jfallows (changes requested):

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/schema/mqtt.kafka.schema.patch.json
```diff
@@ -89,6 +131,12 @@
                             "exit",
                             "options"
                         ]
+                    },
+                    {
+                        "required":
+                        [
+                            "routes"
```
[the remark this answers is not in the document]
bmaidics: Right, also makes sense to promote outside of anyOf

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaConditionMatcher.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+
+    private static Matcher asMatcher(
+        String wildcard)
+    {
+        return Pattern.compile(
+                wildcard.replace(".", "\\.")
+                    .replace("*", ".*")
+                    .replaceAll("\\{([a-zA-Z_]+)\\}", "(?<$1>.+)"))
```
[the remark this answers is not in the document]
bmaidics: Not right now. But we're using the exact same wildcard pattern in many other bindings (http-kafka, sse-kafka, grpc etc.) Should I add some tests for this (and maybe for the other bindings as well)?

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaWithConfigAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public WithConfig adaptFromJson(
+        JsonObject object)
+    {
+        String topic = object.containsKey(TOPIC_NAME)
+            ? object.getString(TOPIC_NAME)
+            : null;
+
+        return new MqttKafkaWithConfig(topic);
```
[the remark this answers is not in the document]
bmaidics: I can add a builder, but I thought we'll only add builder if that particular binding will ever be generated using the config generator. AFAIK we don't generate mqtt-kafka config, right? Also, grpc-kafka and http-kafka doesn't use builders as well, and I was trying to be consistent with those. 
(BTW: I can easily add one if needed, as it's a couple of minutes, just trying to understand the reason)

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/config/proxy.when.mqtt.topic.with.kafka.topic.yaml
```diff
@@ -28,11 +28,11 @@ bindings:
...
         with:
-          topic: mqtt-sensors
+          messages: mqtt-sensors
       - exit: kafka0
         when:
           - topic: device/*
+          - topic: sensor/two
         with:
-          topic: mqtt-devices
+          messages: mqtt-devices
```
jfallows, line 38: Looks good, let's rename the scenario to something like `proxy.when.topic.with.messages.yaml` as we no longer have the collision between `mqtt` topic and `kafka` topic in the config.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaWithConfigAdapter.java
```diff
@@ -37,13 +37,13 @@ public String type()
...
-        MqttKafkaWithConfig mqttKafkaWith = (MqttKafkaWithConfig) with;
+        MqttKafkaWithConfig config = (MqttKafkaWithConfig) with;
 
         JsonObjectBuilder object = Json.createObjectBuilder();
 
-        if (mqttKafkaWith.topic != null)
+        if (config.topic != null)
         {
-            object.add(TOPIC_NAME, mqttKafkaWith.topic);
+            object.add(TOPIC_NAME, config.topic);
```
jfallows, line 46: This should change to `MESSAGE_NAME` and `config.messages`, agree?

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -136,17 +136,24 @@ public MessageConsumer newStream(
...
 
-        final MqttKafkaRouteConfig resolved = binding != null ? binding.resolve(authorization) : null;
+        final MqttKafkaRouteConfig resolved = binding != null ?
+            binding.resolve(authorization, mqttPublishBeginEx.topic().asString()) : null;
         MessageConsumer newStream = null;
 
         if (resolved != null)
         {
             final long resolvedId = resolved.id;
+            final String16FW kafkaTopic = resolved.with.isPresent() ? resolved.with.get().topic() : binding.messagesTopic();
```
jfallows, line 154: Let's rename this `messagesTopic`.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaWithConfigAdapter.java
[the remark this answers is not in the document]
jfallows: We will need builders for all of them, but it's okay if we defer this for now, as you say we're not generating `mqtt-kafka` yet from `zilla generate`.

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -166,21 +167,26 @@ public MessageConsumer newStream(
...
+        final MqttSubscribeBeginExFW mqttSubscribeBeginEx = mqttBeginEx.subscribe();
 
-        final MqttKafkaRouteConfig resolved = binding != null ?
-            binding.resolve(authorization) : null;
+        final MqttKafkaBindingConfig binding = supplyBinding.apply(routedId);
+        final List<String> topicFilters = new ArrayList<>();
+        mqttSubscribeBeginEx.filters().forEach(f -> topicFilters.add(f.pattern().asString()));
 
+        final List<MqttKafkaRouteConfig> routes = binding != null ?
+            binding.resolve(authorization, topicFilters) : null;
```
jfallows, line 181: Suggest renaming this method as `resolveAll` indicating it is finding all matches.
Also, suggest passing in `MqttSubscribeBeginExFW` as a parameter instead of building up the topic filter list externally. We may later add support for working with `StringFW` and patterns more directly, so that becomes an implementation detail.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -166,21 +167,26 @@ public MessageConsumer newStream(
...
+        if (routes != null && !routes.isEmpty())
         {
-            final long resolvedId = resolved.id;
-            final String16FW kafkaMessagesTopic = binding.messagesTopic();
-            final String16FW kafkaRetainedTopic = binding.retainedTopic();
-            newStream = new MqttSubscribeProxy(mqtt, originId, routedId, initialId, resolvedId,
-                kafkaMessagesTopic, kafkaRetainedTopic)::onMqttMessage;
+            final String16FW defaultMessagesTopic = binding.messagesTopic();
+            final String16FW retainedTopic = binding.retainedTopic();
+            newStream = new MqttSubscribeProxy(mqtt, originId, routedId, initialId, routes,
```
jfallows, line 188: Each route should be able to answer which messages or retained topic to use, no need to implement defaulting here, that should be encapsulated inside the route config or with config implementation.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -193,7 +199,7 @@ private final class MqttSubscribeProxy
         private final long routedId;
         private final long initialId;
         private final long replyId;
-        private final KafkaMessagesProxy messages;
+        private final Object2ObjectHashMap<MqttKafkaRouteConfig, KafkaMessagesProxy> messages;
```
jfallows, line 202: Do we need to key by route config here?
I think each route knows its own index, so using the route index and/or a bitmask of matching route indexes might be lighter.

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: Good idea. I guess route.order is what I'm looking for

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaRouteConfig.java
```diff
@@ -58,4 +60,29 @@ boolean matches(
...
+    {
+        return when.isEmpty() || when.stream().anyMatch(m -> matchTopicFilter(topicFilter, m.topic));
+    }
+
+    private boolean matchTopicFilter(
+        String topicFilter,
+        String topic)
+    {
+        Matcher topicFilterMatcher = asTopicFilterMatcher(topicFilter);
+        return topicFilterMatcher.reset(topic).matches();
```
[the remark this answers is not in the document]
bmaidics: I'm not sure if I understand what you mean by "perhaps the opposite way".
We only know what we want to match against once the we receive the subscribe filter, which is either in mqtt begin, or flush, so I'm not sure if we can optimise here very easily.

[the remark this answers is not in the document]
jfallows: Let's take an example to see if there's any opportunity to pre-create a matcher here.

If the routes have subscribe when topic is `/some/hierarchical/topic/name`, then it would match the following topic filter patterns (i think)
 - `/(some|+)/(hierarchical|+)/(topic|+)/(name|+)`
 - `/(some|+)/(hierarchical|+)/(topic|+)/(name|+)/#`
 - `/(some|+)/(hierarchical|+)/(topic|+)/#`
 - `/(some|+)/(hierarchical|+)/#`
 - `/(some|+)/#`
 - `/#`
 - `#`

This can be combined into a single regex as it either terminates with `#` or continues to be more specific.

Then the topic-derived regex can be used to match the topic filter, which is opposite from how we normally match the topic to the topic filter.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -706,14 +691,12 @@ final class KafkaMessagesProxy
...
-            long routedId,
-            String16FW topic,
             MqttKafkaRouteConfig routeConfig,
             MqttSubscribeProxy mqtt)
         {
             this.originId = originId;
-            this.routedId = routedId;
-            this.topic = topic;
+            this.routedId = routeConfig.id;
+            this.topic = routeConfig.messages;
```
jfallows, line 699: Use `route` instead of `routeConfig` here, as it is easier to read and there is no ambiguity.

runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaConditionMatcherTest.java
```diff
@@ -0,0 +1,76 @@
...
+    @Test
+    public void shouldNotMatchSimpleConditions()
+    {
+        MqttKafkaConditionConfig condition = new MqttKafkaConditionConfig("/some/hierarchical/topic/name");
+        MqttKafkaConditionMatcher matcher = new MqttKafkaConditionMatcher(condition);
+
+        assertFalse(matcher.matches("/some/+"));
+        assertFalse(matcher.matches("/some/hierarchical/+"));
+        assertFalse(matcher.matches("/some/+"));
+        assertFalse(matcher.matches("/some/hierarchical/topic/name/something"));
```
jfallows, line 49: Include negative checks that omit leading `/` too.

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaConditionMatcher.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private boolean matchTopic(
+        CharSequence topic)
+    {
+        return this.topicMatcher == null || this.topicMatcher.reset(topic).matches();
+    }
+
+
+    public static String generateRegexPattern(
```
jfallows, line 47:
```suggestion
    private static String asRegexPattern(
```

jfallows: Also please move to end of the class, as it is just a supporting method, not the main aspect.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -307,11 +305,6 @@ private void onMqttBegin(
...
             filters.forEach(filter ->
             {
-                int subscriptionId = (int) filter.subscriptionId();
-                if (!messagesSubscriptionIds.contains(subscriptionId))
-                {
-                    messagesSubscriptionIds.add(subscriptionId);
-                }
                 if ((filter.flags() & SEND_RETAIN_FLAG) != 0)
                 {
                     retainAvailable = true;
```
jfallows, line 310:
```
filters.forEach(f -> retainAvailable |= (f.flags() & SEND_RETAIN_FLAG) != 0);
```

jfallows (changes requested):

jfallows (approved):

PR #498, opened 2023-10-06
Better handle request with same group id

### LINKED ISSUE #500: Group stream with same group id may get hang
**Describe the bug**
The following bugs have been identified while trying to make mqtt pub request with the same client id

- Due to the wrong decoder assignment we may get into a bad state
- Not being able to handle unknown member id in heartbeat response

**To Reproduce**

1. Checkout https://github.com/aklivity/zilla-examples/tree/main/mqtt.kafka.broker and start the stack
2. watch -n 1  "mosquitto_pub -V 'mqttv5' --topic 'zilla' --message 'Hello, world' --debug --insecure -i my-client-id-1"

akrambek:
## Description

The following bugs have been identified while trying to make mqtt pub request with the same client id

Due to the wrong decoder assignment we may get into a bad state
Not being able to handle unknown member id in heartbeat response

Fixes #500


jfallows (dismissed):

jfallows (dismissed):

jfallows (dismissed):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientGroupFactory.java
```diff
@@ -4243,6 +4244,11 @@ private void onJoinGroupResponse(
...
                     .build()
                     .sizeof()));
+
+            if (!encoders.isEmpty())
+            {
+                signaler.signalNow(originId, routedId, initialId, traceId, SIGNAL_NEXT_REQUEST, 0);
+            }
         }
 
         private void onSynGroupRebalance(
```
jfallows, line 4254: `onSyncGroupRebalance`

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientConnectionPool.java
```diff
@@ -909,6 +951,9 @@ private void onSignal(
             case SIGNAL_STREAM_INITIAL_RESET:
                 doStreamReset(traceId);
                 break;
+            case SIGNAL_STREAM_REPLY_WINDOW:
+                doStreamWindow(traceId);
+                break;
             default:
                 doSignal(sender, originId, routedId, initialId, initialSeq,
                     initialAck, connection.initialMax, traceId, signalId, payload);
```
jfallows, line 959: Missing `break` from `default` block.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientConnectionPool.java
```diff
@@ -909,6 +951,9 @@ private void onSignal(
             case SIGNAL_STREAM_INITIAL_RESET:
                 doStreamReset(traceId);
                 break;
+            case SIGNAL_STREAM_REPLY_WINDOW:
+                doStreamWindow(traceId);
```
jfallows, line 955: This method is not a `doXXX` method as it doesn't write a frame for stream.

Please rename `doStreamWindow` to `onStreamSignalReplyWindow`.

Also, given that the signals are used to defer something that happened, it might be better to name them for where they originated, rather than how you want them to react.

For example, `SIGNAL_STREAM_BEGIN` (triggered from `onStreamBegin`) and then the reaction can either be `doStreamBegin` directly or `doStreamBegin` called from `onStreamSignalBegin`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientConnectionPool.java
```diff
@@ -1306,7 +1380,16 @@ private void onConnectionAbort(
...
-            streams.forEach(s -> streamsByInitialIds.remove(s).cleanup(traceId));
+            requests.clear();
+            responses.clear();
+            responseAcks.clear();
+            responseRsts.clear();
+            streams.forEach(s ->
+            {
+                KafkaClientStream stream = streamsByInitialIds.get(s);
+                stream.cleanup(traceId);
+                streamsByInitialIds.remove(s);
```
jfallows, line 1391: Do we need to remove this one by one, or perhaps this instead?

```
...
streams.forEach(s -> streamsByInitialIds.get(s).cleanup(traceId));
streamsByInitialIds.clear();
...
```

Please rename `streamsByInitialIds` to `streamsByInitialId`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientConnectionPool.java
```diff
@@ -1446,10 +1526,18 @@ private void cleanupBudgetCreditorIfNecessary()
...
+            long traceId,
+            long authorization,
             long streamId)
         {
+            responseRsts.add(streamId);
             streams.remove(streamId);
-            streamsByInitialIds.remove(streamId);
+
+            KafkaClientStream stream = streamsByInitialIds.get(streamId);
+            if (stream.replyAck == stream.replySeq)
```
jfallows, line 1537: Should we not always set `stream.replyAck = stream.replySeq` so that a partial stream cannot block progress of connection window?

jfallows (changes requested):

jfallows (changes requested):

jfallows (approved):

PR #499, opened 2023-10-06
converter implementation

aDaemonThread:


specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/engine.schema.json
```diff
@@ -372,6 +372,14 @@
...
                         "properties":
                         {
+                            "expect":
+                            {
+                                "type": "string",
+                                "enum":
+                                [
+                                    "json"
+                                ]
+                            },
```
jfallows, line 382: We need to put `validator` implementation-specific properties in the validator-specific schema patches so that engine doesn't need global knowledge of the possible properties used by all the validator implementations.

This currently includes
  - encoding
  - expect

Please refactor them out of the engine schema.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/validator/TestValidator.java
```diff
@@ -22,28 +22,20 @@
...
 {
     @Override
-    public boolean read(
+    public DirectBuffer read(
         DirectBuffer data,
         int index,
         int length)
     {
-        return validate(data, index, length);
+        return length == 18 ? data : null;
```
jfallows, line 30: This is not quite right because it loses information about index and length.

Take a look at `MessageConsumer`, `MessageFunction` and `MessagePredicate` in `io.aklivity.zilla.runtime.engine.binding.function` package.

I think we might need the following in `io.aklivity.zilla.runtime.engine.validator.function` package:
 - `ToIntFragmentFunction`

```
ToIntFragmentFunction
  int applyAsInt(int flags, DirectBuffer buffer, int index, int length)
```

```
Validator
  int validate(int flags, DirectBuffer buffer, int index, int length, ToIntFragmentFunction next)
```

The return value from `applyAsInt` is the progress towards `length`, and the implementation would write the validated (converted) value to `next`.

```
public class TestValidator implements Validator
{
    public DirectBuffer validate(
        DirectBuffer data,
        int index,
        int length,
        ToIntFragmentFunction next)
    {
        return length != 18 ? -1 : next.applyAsInt(flags, buffer, index, length);
    }
}
```

So the lambda we pass in for `next` would actually write the value into the next step in the pipeline.

A better implementation would track the state of progress towards 18 bytes in this case, but that makes it stateful, which is another level of complexity we need to sort out after getting the interfaces to support streaming.

What do you think?

jfallows (changes requested):

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/validator/TestValidator.java
jfallows: Alternatively, we can consider using `MessageConsumer` more directly in validator, if that makes it simpler to propagate valid `DATA` frames.

```
public class TestValidator implements Validator
{
    public int validate(
        int msgTypeId, // DataFW.TYPE_ID
        DirectBuffer data,
        int index,
        int length,
        MessageConsumer next)
    {
        return length != 18 ? -1 : next.applyAsInt(msgTypeId, buffer, index, length);
    }
}
```

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValidator.java
```diff
@@ -114,14 +116,15 @@ public DirectBuffer read(
             convertResponse(data, value, payloadBytes, schema);
         }
 
-        return value;
+        return value.capacity() == 0 ? -1 : next.applyAsInt(value, index, value.capacity());
```
jfallows, line 119: Now that we don't have to return the `value`, the allocations for `value`, `payloadBytes`, `byteBuf`, `valBytes` can be eliminated, or reuse shared instances as necessary that are only needed on stack during method invocation, so no chance of concurrent reuse.

incubator/validator-avro/src/test/java/io/aklivity/zilla/runtime/validator/avro/AvroValidatorTest.java
```diff
@@ -45,6 +46,8 @@ public class AvroValidatorTest
             "{\"name\":\"status\",\"type\":\"string\"}]," +
             "\"name\":\"Event\",\"namespace\":\"io.aklivity.example\",\"type\":\"record\"}";
 
+    private final ToIntValueFunction fragmentFunction = (buffer, index, length) -> length;
```
jfallows, line 49: This should be `static` too, right?
Perhaps called `NEXT_NOP` ?

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValidator.java
jfallows: Note the return value here should be how much of `length` was processed, whereas the return value of `next` is an indication of how much of what was passed into `next` was processed.

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/IntegerValidator.java
```diff
@@ -26,21 +27,23 @@ public IntegerValidator(IntegerValidatorConfig config)
...
-    public DirectBuffer read(
+    public int read(
         DirectBuffer data,
         int index,
-        int length)
+        int length,
+        ToIntValueFunction next)
     {
-        return validate(data, index, length) ? data : null;
+        return !validate(data, index, length) ? -1 : next.applyAsInt(data, index, length);
```
jfallows, line 36: This one is all good to return the progress from `next` because there is no transformation, so the progress in `next` is equivalent to the progress in this validator.

jfallows (changes requested):

PR #503, opened 2023-10-09
Integrate http binding with validators

### LINKED ISSUE #455: Integrate `http` binding with `validators`
Support `validators` for `http` request `path` params, `query` params, `headers` and `content`.

attilakreiner:
## Description

Use the validators to validate the headers (header fields, path parameters, query parameters) as well as the content (payload) of requests for both http/1.1 and h2.

Fixes #455


runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -1051,7 +1061,8 @@ else if (!isCorsRequestAllowed(server.binding, headers))
                         }
 
                         HttpRouteConfig route = binding.resolve(exchangeAuth, headers::get);
-                        if (route != null)
+                        server.selectRequest(headers::get);
+                        if (route != null && server.validate(beginEx, headers))
```
jfallows, line 1065: The `validate` method can also move to `HttpBindingConfig`, then the `selectRequest` / `resolveRequest` and supporting parse methods would be internal with only on-stack state (no fields) needed.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -164,6 +191,241 @@ private Function<Function<String, String>, String> asAccessor(
...
+            String path = headerByName.apply(":path");
+            String method = headerByName.apply(":method");
+            String contentType = headerByName.apply("content-type");
+            for (HttpRequestType request : requests)
+            {
+                boolean isMatch = true;
+                if (method != null && !method.equals(request.method.name()))
+                {
+                    isMatch = false;
+                }
```
jfallows, line 256: Let's simplify this logic as follows.
```
isMatch |= method == null || method.equals(request.method.name());
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -164,6 +191,241 @@ private Function<Function<String, String>, String> asAccessor(
...
+                if (isMatch && contentType != null && !request.contentType.contains(contentType))
+                {
+                    isMatch = false;
+                }
```
jfallows, line 260:
```
isMatch |= contentType == null || request.contentType.contains(contentType);
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -164,6 +191,241 @@ private Function<Function<String, String>, String> asAccessor(
...
+                if (isMatch)
+                {
+                    isMatch = parseParams(request, path);
+                }
```
jfallows, line 264:
```
isMatch &= parseParams(request, path);
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -164,6 +191,241 @@ private Function<Function<String, String>, String> asAccessor(
...
+
+    private boolean parseParams(
+        HttpRequestType request,
+        String path)
+    {
+        boolean result = false;
+        String pattern =
+            "^" +
+            request.path.replaceAll("\\{([a-zA-Z_]+)\\}", "(?<$1>.+?)") +
+            "/?(\\?(?<query0>.*))?$";
```
jfallows, line 283: Use `String.format(...)` instead of `+` concatenation.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -164,6 +191,241 @@ private Function<Function<String, String>, String> asAccessor(
...
+    {
+        boolean isValid = true;
+        if (request != null)
+        {
+            boolean isValidHeader = validateHeaderValues(request, beginEx);
+            boolean isValidPathParams = validatePathParams(request);
+            boolean isValidQueryParams = validateQueryParams(request);
+            isValid = isValidHeader && isValidPathParams && isValidQueryParams;
+        }
+        return isValid;
```
jfallows, line 345:
```
return request == null ||
    validateHeaderValues(request, beginEx) &&
    validatePathParams(request) &&
    validateQueryParams(request);
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -164,6 +191,241 @@ private Function<Function<String, String>, String> asAccessor(
...
+    }
+
+    private boolean validateHeaderValues(
+        HttpRequestType request,
+        HttpBeginExFW beginEx)
+    {
+        AtomicBoolean isValid = new AtomicBoolean(true);
```
jfallows, line 352: Use `MutableBoolean` instead of `AtomicBoolean`.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -2205,7 +2218,7 @@ private void onDecodeCorsPreflight(
             assert exchange == null;
         }
 
-        private void onDecodeHeaders(
+        private DirectBuffer onDecodeHeaders(
```
jfallows, line 2221: In general we don't return values from `onDecodeXXX` methods if possible.
Why not doEncodeHeaders here instead?

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -4847,11 +4877,23 @@ else if (!isCorsRequestAllowed(binding, headers))
...
                                     .build();
 
-                            exchange.doRequestBegin(traceId, beginEx);
+                            HttpRequestType request = binding.resolveRequest(headers::get);
+
+                            final Http2Exchange exchange = new Http2Exchange(originId, routedId, NO_REQUEST_ID, streamId,
+                                exchangeAuth, traceId, policy, origin, contentLength, request);
 
-                            if (endRequest)
+                            boolean isValid = binding.validateHeader(request, beginEx);
```
jfallows, line 4885: This should be called `validateHeaders` (plural).

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -4847,11 +4877,23 @@ else if (!isCorsRequestAllowed(binding, headers))
                                     .headers(hs -> headers.forEach((n, v) -> hs.item(h -> h.name(n).value(v))))
                                     .build();
 
-                            exchange.doRequestBegin(traceId, beginEx);
+                            HttpRequestType request = binding.resolveRequest(headers::get);
```
jfallows, line 4880: Try not to lean too heavily on the map of headers we create here as it is undesirable and we'll likely remove it as some point..

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -5057,21 +5099,30 @@ private int onDecodeData(
...
+                                {
+                                    exchange.doRequestEnd(traceId, EMPTY_OCTETS);
+                                }
+                            }
                         }
                         else
                         {
-                            exchange.doRequestEnd(traceId, EMPTY_OCTETS);
+                            doEncodeHeaders(traceId, authorization, streamId, headers400, true);
+                            exchange.doRequestAbort(traceId, EMPTY_OCTETS);
```
jfallows, line 5125: This code restructuring prevents validation of empty payload and treats empty payload as a special case.
Let's keep the simpler codepath where we don't introduce the special case for empty payload.

jfallows (changes requested):

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
attilakreiner: done

attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
attilakreiner: fixed to not use the map, pls chk

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -49,9 +69,19 @@ public final class HttpBindingConfig
     public final List<HttpRouteConfig> routes;
     public final ToLongFunction<String> resolveId;
     public final Function<Function<String, String>, String> credentials;
+    public final List<HttpRequestType> requests;
+    public Map<String, String8FW> pathParams;
+    public Map<String, String8FW> queryParams;
```
jfallows, line 74: All field on binding should be final with no stateful side effects of calling methods like `resolveRequest`.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -164,6 +195,225 @@ private Function<Function<String, String>, String> asAccessor(
...
+    }
+
+    private boolean parseParams(
+        HttpRequestType request,
+        String path)
+    {
+        boolean result = false;
+        String pattern = String.format("^%s/?(\\?(?<query0>.*))?$",
+            request.path.replaceAll("\\{([a-zA-Z_]+)\\}", "(?<$1>.+?)"));
+        Matcher matcher = Pattern.compile(pattern).matcher(path);
```
jfallows, line 277: This `pattern` and `matcher` should be precomputed once on `HttpRequestType` instead of recomputing on each call to `resolveRequest`.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -164,6 +195,225 @@ private Function<Function<String, String>, String> asAccessor(
...
+                String[] keyValue = paramPair.split("=");
+                if (keyValue.length == 2)
+                {
+                    String key = URLDecoder.decode(keyValue[0], UTF_8);
+                    String8FW value = new String8FW(URLDecoder.decode(keyValue[1], UTF_8));
+                    queryParams.put(key, value);
+                }
+            }
+        }
+        return queryParams;
```
jfallows, line 323: There's a lot of object allocation going on in these parse methods. Let's aim to eliminate it.

jfallows (changes requested):

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -205,6 +203,9 @@ private List<HttpRequestType> createRequests(
             {
                 HttpRequestType requestType = new HttpRequestType();
                 requestType.path = request.path;
+                String pattern = String.format("^%s/?(\\?(?<query0>.*))?$",
+                    requestType.path.replaceAll("\\{([a-zA-Z_]+)\\}", "(?<$1>.+?)"));
+                requestType.pathPattern = Pattern.compile(pattern);
```
jfallows, line 208: Let's go further and instantiate the `Matcher` instead.

```
requestType.pathMatcher = Pattern.compile(pattern).matcher("");
```

Then later we can use `requestType.pathMatcher.reset(path)` to use the matcher.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
jfallows: We should be able to simplify this loop to use a Pattern Matcher for each query parameter, calling `.find()` for each one, then extracting the parameter name and value from the matcher via named groups.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -308,17 +307,25 @@ private Map<String, String8FW> parseQueryParams(
...
-                    String key = URLDecoder.decode(keyValue[0], UTF_8);
-                    String8FW value = new String8FW(URLDecoder.decode(keyValue[1], UTF_8));
-                    queryParams.put(key, value);
+                    String key = query.substring(ampersandIndex + 1, equalsIndex);
+                    String value = nextAmpersandIndex == -1 ?
+                        query.substring(equalsIndex + 1) :
+                        query.substring(equalsIndex + 1, nextAmpersandIndex);
+                    if (!key.isEmpty())
+                    {
+                        queryParams.put(URLDecoder.decode(key, UTF_8), new String8FW(URLDecoder.decode(value, UTF_8)));
```
jfallows, line 323: We only need to extract the query parameters that we are actually going to validate, right?

jfallows, line 323: We only need to extract the query parameters that we are actually going to validate, right?

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRequestType.java
```diff
@@ -29,6 +30,11 @@ public class HttpRequestType
     public HttpRequestConfig.Method method;
     public List<String> contentType;
 
+    // path parsing helpers
+    public Pattern pathPattern;
+    public Map<String, String8FW> pathParamValues;
+    public Map<String, String8FW> queryParamValues;
```
jfallows, line 36: The fields in this class are shared by all requests, except for `pathParamValues` and `queryParamValues` which are storing parsed values from an individual request, so should not be here.

jfallows (changes requested):

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -193,57 +194,65 @@ private Function<Function<String, String>, String> asAccessor(
...
 
-    public HttpRequestType resolveRequest(
+    // visible for testing
+    int urlDecodedComparator(
+        String input1,
+        String input2)
+    {
+        String decoded1 = URLDecoder.decode(input1, UTF_8);
+        String decoded2 = URLDecoder.decode(input2, UTF_8);
+        return decoded1.compareTo(decoded2);
```
jfallows, line 252: Please extract this into a separate class, such as `PercentEncodableStringComparator` that implements `Comparator<String, String>`.

The purpose of the implementation is to avoid allocations, just readonly operations to compare. This implementation does not do that yet.

Instead, we want to walk down each string in parallel, char by char, advancing either 1 char (no URL encoding) or 3 chars (% plus 2 chars) for each, until we get an effective mismatch of char vs char, we discover effective decoded length mismatch, or both strings are equivalent.

Then no allocation will occur during compare.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -193,57 +194,65 @@ private Function<Function<String, String>, String> asAccessor(
         return accessor;
     }
 
-    private List<HttpRequestType> createRequests(
+    private List<HttpRequestType> createRequestTypes(
         BiFunction<ValidatorConfig, ToLongFunction<String>, Validator> createValidator)
     {
         List<HttpRequestType> result = new LinkedList<>();
```
jfallows, line 200: Rename `result` to `requestTypes`.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -252,173 +261,142 @@ public HttpRequestType resolveRequest(
...
                 {
-                    Validator validator = request.headers.get(header.name());
+                    Validator validator = requestType.headers.get(header.name());
                     if (validator != null)
                     {
                         String16FW value = header.value();
                         if (!validator.read(value.value(), value.offset(), value.length()))
                         {
-                            isValid.set(false);
+                            valid.value = false;
```
jfallows, line 327:
```
String16FW value = header.value();
valid.value &= validator.read(value.value(), value.offset(), value.length());
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -252,173 +261,142 @@ public HttpRequestType resolveRequest(
...
+        String path)
     {
-        boolean isValid = true;
-        for (String name : request.pathParamValues.keySet())
+        Matcher matcher = requestType.pathMatcher.reset(path);
+        boolean matches = matcher.matches();
+        assert matches;
+
+        boolean valid = true;
+        for (String name : requestType.pathParams.keySet())
```
jfallows, line 345: Make sure we are not creating a new iterator every time.
See `Object2ObjectHashMap` if needed.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -252,173 +261,142 @@ public HttpRequestType resolveRequest(
...
-        HttpRequestType request)
+        HttpRequestType requestType,
+        String path)
     {
-        boolean isValid = true;
-        for (String name : request.queryParamValues.keySet())
+        Matcher matcher = requestType.queryMatcher.reset(path);
+        boolean valid = true;
+        boolean found = matcher.find();
+        while (found)
```
jfallows, line 369:
```suggestion
        while (valid && matcher.find())
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -252,173 +261,142 @@ public HttpRequestType resolveRequest(
...
+            String name = matcher.group(1);
+            Validator validator = requestType.queryParams.get(name);
             if (validator != null)
             {
-                String8FW value = request.queryParamValues.get(name);
+                String8FW value = new String8FW(matcher.group(2));
                 if (!validator.read(value.value(), value.offset(), value.length()))
                 {
-                    isValid = false;
+                    valid = false;
```
jfallows, line 378:
```
value &= validator.read(value.value(), value.offset(), value.length());
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -252,173 +261,142 @@ public HttpRequestType resolveRequest(
...
                     break;
                 }
             }
+            found = matcher.find();
```
jfallows, line 382: Remove this line after adjusting `while` condition above.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -252,173 +261,142 @@ public HttpRequestType resolveRequest(
...
+        boolean valid = true;
+        if (requestType != null && requestType.content != null)
         {
-            Validator validator = request.content;
-            isValid = validator.read(buffer, index, length);
+            Validator validator = requestType.content;
+            valid = validator.read(buffer, index, length);
         }
-        return isValid;
+        return valid;
```
jfallows, line 399:
```
return requestType != null &&
  requestType.content != null &&
  requestType.content.read(buffer, index, length);
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRequestType.java
```diff
@@ -17,27 +17,130 @@
...
 
-public class HttpRequestType
+public final class HttpRequestType
 {
+    private static final String PATH_FORMAT = "^%s/?(?:\\?.*)?$";
+    private static final String PATH_REGEX = "\\{([a-zA-Z0-9_-]+)\\}";
+    private static final String PATH_REPLACEMENT = "(?<$1>.+?)";
+    private static final String QUERY_REGEX = "(?<=[?&])([^&=]+)=([^&]+)(?=&|$)";
+    private static final Pattern QUERY_PATTERN = Pattern.compile(QUERY_REGEX);
+    private static final String EMPTY = "";
```
jfallows, line 34:
```suggestion
    private static final String EMPTY_INPUT = "";
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -2251,9 +2251,9 @@ private DirectBuffer onDecodeHeaders(
             String origin,
             HttpBeginExFW beginEx)
         {
-            boolean isValid = binding.validateHeaders(request, beginEx);
+            boolean valid = binding.validateHeaders(requestType, beginEx);
```
jfallows, line 2254: Rename `valid` to `heasdersValid`.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -2291,9 +2291,9 @@ private int onDecodeBody(
             int limit,
             Flyweight extension)
         {
-            boolean isValid = binding.validateContent(request, buffer, 0, limit - offset);
+            boolean valid = binding.validateContent(requestType, buffer, 0, limit - offset);
```
jfallows, line 2294: Rename `valid` to `contentValid`.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -4901,13 +4901,13 @@ else if (!isCorsRequestAllowed(binding, headers))
...
-                            HttpRequestType request = binding.resolveRequest(beginEx);
+                            HttpRequestType requestType = binding.resolveRequestType(beginEx);
 
                             final Http2Exchange exchange = new Http2Exchange(originId, routedId, NO_REQUEST_ID, streamId,
-                                exchangeAuth, traceId, policy, origin, contentLength, request);
+                                exchangeAuth, traceId, policy, origin, contentLength, requestType);
 
-                            boolean isValid = binding.validateHeaders(request, beginEx);
-                            if (isValid)
+                            boolean valid = binding.validateHeaders(requestType, beginEx);
```
jfallows, line 4909: Rename `valid` to `headersValid`.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -5120,8 +5120,8 @@ private int onDecodeData(
                 else
                 {
                     final int payloadLength = payload.capacity();
-                    boolean isValid = binding.validateContent(exchange.request, payload, 0, payloadLength);
-                    if (isValid)
+                    boolean valid = binding.validateContent(exchange.request, payload, 0, payloadLength);
```
jfallows, line 5123: Rename `valid` to `contentValid`.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRequestType.java
```diff
@@ -17,27 +17,130 @@
...
+            return this;
+        }
+
+        public HttpRequestType build()
+        {
+            String pathPattern = String.format(PATH_FORMAT, path.replaceAll(PATH_REGEX, PATH_REPLACEMENT));
+            Matcher pathMatcher = Pattern.compile(pathPattern).matcher(EMPTY);
+            Matcher queryMatcher = QUERY_PATTERN.matcher(EMPTY);
+            return new HttpRequestType(path, method, contentType, pathMatcher, queryMatcher, headers, pathParams, queryParams,
+                content);
```
jfallows, line 143: Looks good! :+1:

runtime/binding-http/src/test/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfigTest.java
```diff
@@ -0,0 +1,197 @@
...
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.hamcrest.Matchers.lessThan;
+
+import org.junit.Test;
+
+import io.aklivity.zilla.runtime.engine.config.BindingConfig;
+import io.aklivity.zilla.runtime.engine.config.KindConfig;
+
+public class HttpBindingConfigTest
```
jfallows, line 28: Change this to `HttpQueryStringComparatorTest` to test `HttpQueryStringComparator` in isolation, and please name each method for its purpose, perhaps of the form `shouldCompare...EqualTo...()`, `shouldCompare...GreaterThan...()`, `shouldCompare...LessThan...()`.

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/streams/application/rfc7230/validation/valid/client.rpt
```diff
@@ -39,9 +38,11 @@ read zilla:begin.ext ${http:beginEx()
 
 read "response1"
 read closed
-read notify RESPONSE_ONE_RECEIVED
+read notify RESPONSE_1_RECEIVED
```
jfallows, line 41: I think we use words for barriers everywhere by convention, so let's keep the same here.

jfallows (changes requested):

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRequestType.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
attilakreiner: done

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/streams/application/rfc7230/validation/valid/client.rpt
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: IntelliJ pointed out that `valid` is always false here so it can be simplified to
`valid = validator.read(value.value(), value.offset(), value.length());`

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: done

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: I don't think this code is identical to the previous. I went with
```
return requestType == null ||
            requestType.content == null ||
            requestType.content.read(buffer, index, length);
```
The intent is to accept things as valid if have no intention to validate them. Pls LMK.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: I replaced `HashMap` with `Object2ObjectHashMap`, LGTM.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: OK, pls chk.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -365,21 +352,15 @@ private boolean validateQueryParams(
...
             Validator validator = requestType.queryParams.get(name);
             if (validator != null)
             {
                 String8FW value = new String8FW(matcher.group(2));
-                if (!validator.read(value.value(), value.offset(), value.length()))
-                {
-                    valid = false;
-                    break;
-                }
+                valid = validator.read(value.value(), value.offset(), value.length());
```
jfallows, line 362: This should be `valid &= ...` to better convey the intent.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/PercentEncodableStringComparator.java
```diff
@@ -0,0 +1,99 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.aklivity.zilla.runtime.binding.http.internal.config;
+
+import java.util.Comparator;
+
+public class PercentEncodableStringComparator implements Comparator<String>
```
jfallows, line 20: Apologies for the dueling naming feedback.
Lets rename this to `HttpQueryStringComparator`. (all classes in this project start with `Http`)

runtime/binding-http/src/test/java/io/aklivity/zilla/runtime/binding/http/internal/config/PercentEncodableStringComparatorTest.java
```diff
@@ -0,0 +1,286 @@
...
+package io.aklivity.zilla.runtime.binding.http.internal.config;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.hamcrest.Matchers.lessThan;
+
+import org.junit.Test;
+
+public class PercentEncodableStringComparatorTest
```
jfallows, line 25: Lets rename this to `HttpQueryStringComparatorTest`.

runtime/binding-http/src/test/java/io/aklivity/zilla/runtime/binding/http/internal/config/PercentEncodableStringComparatorTest.java
```diff
@@ -0,0 +1,286 @@
...
+
+        // WHEN
+        int result = comparator.compare("hello1", "hello");
+
+        // THEN
+        assertThat(result, greaterThan(0));
+    }
+
+    @Test
+    public void shouldBeGreaterThanZeroWhenComparesLongerAndShorterWhereSecondIsEncoded()
```
jfallows, line 249: `shouldCompareLongerUnencodedGreaterThanShorterEncoded`
Let's rename the others to a similar form.

jfallows (changes requested):

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
attilakreiner: OK

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/PercentEncodableStringComparator.java
attilakreiner: OK

runtime/binding-http/src/test/java/io/aklivity/zilla/runtime/binding/http/internal/config/PercentEncodableStringComparatorTest.java
attilakreiner: OK

runtime/binding-http/src/test/java/io/aklivity/zilla/runtime/binding/http/internal/config/PercentEncodableStringComparatorTest.java
attilakreiner: OK

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -5149,13 +5149,15 @@ private int onDecodeData(
                         if (!HttpState.replyOpened(exchange.state))
                         {
                             doEncodeHeaders(traceId, authorization, streamId, headers400, true);
+                            doNetworkEnd(traceId, authorization);
```
jfallows, line 5152: Calling `doNetworkEnd` will close the underlying `tcp` connection, which is heavy handed because there can be other HTTP/2 requests in flight over the same connection.

The HTTP/2 equivalent of this is `doEncodeRstStream`, so it seems that we need to call that here as well.

In fact, I think we just need to make the `doEncodeRstStream` call below unconditional and remove the call to `doNetworkEnd` here.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
jfallows: Yes, returning `boolean` is better than `DirectBuffer`.

runtime/binding-http/src/test/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpQueryStringComparatorTest.java
```diff
@@ -0,0 +1,286 @@
...
+    public void shouldCompareSameUnencodedEqualsToUnencoded()
+    {
+        // GIVEN
+        HttpQueryStringComparator comparator = new HttpQueryStringComparator();
+
+        // WHEN
+        int result = comparator.compare("hello", "hello");
+
+        // THEN
+        assertThat(result, equalTo(0));
```
jfallows, line 37: Use [`comparator`](https://hamcrest.org/JavaHamcrest/javadoc/2.0.0.0/org/hamcrest/comparator/ComparatorMatcherBuilder.html#comparedBy-java.util.Comparator-) support in matchers.

```
assertThat("hello", comparedBy(comparator).comparesEqualTo("hello"));
```

Similar feedback for other test method implementations.

jfallows (changes requested):

jfallows (approved):

PR #507, opened 2023-10-12
Client topic space

bmaidics:


specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/config/proxy.when.client.topic.space.yaml
```diff
@@ -0,0 +1,53 @@
...
+        exit: kafka0
+      - when:
+          - subscribe:
+              - topic: /sensor-clients/#
+        with:
+          messages: mqtt-devices
+        exit: kafka0
+      - when:
+          - subscribe:
+              - topic: /device-clients/#
```
jfallows, line 50: Let's move second `subscribe` alongside `subscribe` above to avoid re-stating `with` and `exit`.

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/config/proxy.when.topic.with.messages.yaml
```diff
@@ -13,25 +13,26 @@
 # specific language governing permissions and limitations under the License.
 #
 
-#TODO: this should be the final config
```
jfallows, line 16: Some adjustments needed here for `publish` and `subscribe` syntax under `when` condition.

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/schema/mqtt.kafka.schema.patch.json
```diff
@@ -79,15 +88,87 @@
...
+                                                "type": "array",
+                                                "items":
+                                                {
+                                                    "topic":
+                                                    {
+                                                        "title": "Topic",
+                                                        "type": "string"
+                                                    }
+                                                }
+                                            }
```
jfallows, line 128: We definitely don't need to do it now, but if we get requests for simpler syntax to avoid specifying topic once each for publish and subscribe, we can potentially add `topic` as a peer of `publish` and `subscribe` (later) that would imply both `publish` and `subscribe`.

I'd like to wait and see if it turns out we need more properties than just `topic` under `publish` and `subscribe` first, because that would impact the simplified syntax, if we need it.

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/publish.client.topic.space/client.rpt
```diff
@@ -0,0 +1,85 @@
...
+                            .typeId(zilla:id("kafka"))
+                            .merged()
+                                .deferred(0)
+                                .partition(-1, -1)
+                                .key("/clients/client-2/sensors/one")
+                                .hashKey("client-2")
+                                .header("zilla:filter", "clients")
+                                .header("zilla:filter", "client-2")
+                                .header("zilla:filter", "sensors")
+                                .header("zilla:filter", "one")
```
jfallows, line 43: Is this missing a leading `.header("zilla:filter", "")` to represent the empty segment before the leading `/` in the mqtt topic name?

Perhaps it would be more readable to include the leading `/` for each segment in the `zilla:filter` header value, then omit the leading `/` when not present at the start of the mqtt topic name to differentiate instead of needing a leading empty `""` header value?  That would be a backwards incompatible change, but if we want to do it we should do it sooner than later. Let's discuss.

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/publish.client.topic.space/client.rpt
```diff
@@ -0,0 +1,85 @@
...
+                                .header("zilla:local", "client-1")
+                                .header("zilla:format", "TEXT")
```
jfallows, line 45: We probably don't need explicit format in these tests, as it is outside the scope of what we are testing.

jfallows (changes requested):

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/schema/mqtt.kafka.schema.patch.json
bmaidics: Actually, I was thinking of this. And in the regular topic space PR the syntax is already done.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaBindingConfig.java
```diff
@@ -16,29 +16,47 @@
...
+    private static final Function<String, String> DEFAULT_CLIENTS = x -> null;
+
     public final long id;
     public final KindConfig kind;
     public final MqttKafkaOptionsConfig options;
     public final List<MqttKafkaRouteConfig> routes;
 
     public MqttKafkaSessionFactory.KafkaSignalStream willProxy;
+    public List<MqttKafkaSubscribeFactory.KafkaMessagesBootstrap> bootstrapStreams;
+    public final List<Function<String, String>> clients;
```
jfallows, line 45: `clients` should move up with other final fields.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaBindingConfig.java
```diff
@@ -50,6 +68,26 @@ public MqttKafkaRouteConfig resolve(
...
+            .orElse(null);
+    }
+
+    public List<MqttKafkaRouteConfig> resolveAll(
+        long authorization,
+        Array32FW<MqttTopicFilterFW> filters)
+    {
+        return routes.stream()
+            .filter(r -> r.authorized(authorization) && filters.anyMatch(f -> r.matchesSubscribe(f.pattern().asString())))
+            .distinct()
```
jfallows, line 87: Why is `.distinct()` used here?

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaBindingConfig.java
```diff
@@ -64,4 +102,58 @@ public String16FW retainedTopic()
...
+    {
+        final List<MqttKafkaRouteConfig> bootstrapRoutes = new ArrayList<>();
+        routes.forEach(r ->
+        {
+            if (options.clients.stream().anyMatch(r::matchesClient))
+            {
+                bootstrapRoutes.add(r);
+            }
+        });
+        return bootstrapRoutes;
```
jfallows, line 116: This method is allocating on each call, suggest pre-allocating once in constructor instead if viable.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaBindingConfig.java
```diff
@@ -64,4 +102,58 @@ public String16FW retainedTopic()
...
+        if (clients != null)
+        {
+            for (String client : clients)
+            {
+                Matcher topicMatch =
+                    Pattern.compile(client.replace("{identity}", "(?<identity>[^\\s/]+)").replace("#", ".*"))
+                        .matcher("");
+
+                Function<String, String> accessor = DEFAULT_CLIENTS;
+                accessor = orElseIfNull(accessor, topic ->
```
jfallows, line 133: Why not just assign `accessor` to the lambda starting with `topic -> ` directly?

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaProxyFactory.java
```diff
@@ -55,7 +56,7 @@ public MqttKafkaProxyFactory(
             config, context, bindings::get);
 
         final MqttKafkaSubscribeFactory subscribeFactory = new MqttKafkaSubscribeFactory(
-            config, context, bindings::get);
+            config, context, bindings::get, BOOTSTRAP_STREAM_RECONNECT_DELAY);
```
jfallows, line 59: Suggest calling `config.bootstrapStreamReconnectDelay()` from constructor instead of passing property to constructor.

Equivalent feedback for `MqttKafkaSessionFactory` and `WILL_STREAM_RECONNECT_DELAY` property.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -163,6 +172,7 @@ private final class MqttPublishProxy
         private final KafkaRetainedProxy retained;
         private final String16FW kafkaMessagesTopic;
         private final String16FW kafkaRetainedTopic;
+        private final List<Function<String, String>> clients;
```
jfallows, line 175: We should be storing `kafkaMessagesTopic` on `KafkaMessagesProxy` and `kafkaRetainedTopic` on `KafkaRetainedProxy` instead.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -388,6 +429,13 @@ private void onMqttData(
                     .timestamp(now().toEpochMilli())
                     .partition(p -> p.partitionId(-1).partitionOffset(-1))
                     .key(b -> b.set(key))
+                    .hashKey(b ->
+                    {
+                        if (hashKey != null)
+                        {
+                            b.set(hashKey);
+                        }
+                    })
```
jfallows, line 438: Create a private method to use instead of inlining here since `hashKey` is a member field.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -83,6 +90,7 @@ public class MqttKafkaSubscribeFactory implements MqttKafkaStreamFactory
     private static final int SEND_RETAIN_FLAG = 1 << SEND_RETAINED.ordinal();
     private static final int RETAIN_FLAG = 1 << RETAIN.ordinal();
     private static final int RETAIN_AS_PUBLISHED_FLAG = 1 << RETAIN_AS_PUBLISHED.ordinal();
+    private static final int SIGNAL_CONNECT_BOOTSTRAP_STREAM = 1;
```
jfallows, line 93: `DATA_FIN_FLAG` (below) should be private.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -124,33 +132,77 @@ public class MqttKafkaSubscribeFactory implements MqttKafkaStreamFactory
...
+        this.reconnectDelay = reconnectDelay.getAsInt(config);
+    }
+
+    @Override
+    public void onAttached(
+        long bindingId)
+    {
+        MqttKafkaBindingConfig binding = supplyBinding.apply(bindingId);
+        List<MqttKafkaRouteConfig> bootstrap = binding.bootstrapRoutes();
+        if (bootstrapAvailable && bootstrap != null)
```
jfallows, line 181: `bootstrap` is never null by implementation.

Perhaps use configuration to return empty list of bootstrap routes if bootstrap not available?

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -124,33 +132,77 @@ public class MqttKafkaSubscribeFactory implements MqttKafkaStreamFactory
...
+    {
+        MqttKafkaBindingConfig binding = supplyBinding.apply(bindingId);
+
+        if (binding.bootstrapStreams != null)
+        {
+            for (KafkaMessagesBootstrap stream : binding.bootstrapStreams)
+            {
+                stream.doKafkaEnd(supplyTraceId.get(), 0);
+            }
+            binding.bootstrapStreams.clear();
```
jfallows, line 204: Bootstrap streams are scoped by binding id, but only relevant to subscribe factory, so it feels like a cleaner abstraction to keep them stored locally here in subscribe factory, keyed by binding id.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -575,11 +599,17 @@ private void onMqttWindow(
...
+            else
             {
-                messages.flushData(traceId, authorization, budgetId);
+                messages.values().forEach(m ->
+                {
+                    if (m.messageSlotOffset != m.messageSlotLimit)
+                    {
+                        m.flushData(traceId, authorization, budgetId);
+                    }
+                });
```
jfallows, line 610: Test this condition inside a method on `KafkaMessagesProxy` instead so this becomes a simpler one line lambda.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -684,6 +707,244 @@ private int replyWindow()
         }
     }
 
+    public final class KafkaMessagesBootstrap
```
jfallows, line 710: After moving the storage of bootstrap streams from binding to subscribe factory, this class can become private.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -732,13 +1002,29 @@ private void doKafkaBegin(
         {
             if (!MqttKafkaState.initialOpening(state))
             {
+                filtersRW.wrap(filterBuffer, 0, filterBuffer.capacity());
```
jfallows, line 1005: Assign return value to local variable and use instead of `filtersRW`.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -1212,7 +1538,7 @@ private void doKafkaBegin(
             replyAck = 0;
             replyMax = 0;
 
-            sendRetainedFiltersRW.wrap(retainFilterBuffer, 0, retainFilterBuffer.capacity());
+            filtersRW.wrap(filterBuffer, 0, filterBuffer.capacity());
```
jfallows, line 1541: Assign return value to local variable and use instead of `filtersRW`.

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -303,12 +305,39 @@ private void onMqttBegin(
...
+            String clientHashKey = null;
+            if (clients != null)
+            {
+                for (Function<String, String> client : clients)
+                {
+                    clientHashKey = client.apply(topicName);
+                    break;
+                }
             }
+            return new String16FW(clientHashKey);
```
jfallows, line 340: Return `null` when `clients` is `null` to avoid unused `String16FW` allocation. Call site then needs to check for `null` directly instead of for `.value()` buffer.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -421,6 +452,14 @@ private void onMqttData(
             }
         }
 
+        private void setHashKey(KafkaKeyFW.Builder b)
```
jfallows, line 455: Parameter formatting and use full name `builder` for method parameter.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaProxyFactory.java
jfallows: This feedback for `BOOTSTRAP_STREAM_RECONNECT_DELAY` has not yet been applied.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -126,18 +134,26 @@ public class MqttKafkaSubscribeFactory implements MqttKafkaStreamFactory
...
+    private final List<KafkaMessagesBootstrap> bootstrapStreams;
+
+    private int reconnectAttempt;
 
     public MqttKafkaSubscribeFactory(
         MqttKafkaConfiguration config,
         EngineContext context,
-        LongFunction<MqttKafkaBindingConfig> supplyBinding)
+        LongFunction<MqttKafkaBindingConfig> supplyBinding,
+        IntPropertyDef reconnectDelay)
```
jfallows, line 156: This constructor parameter would be removed when feedback above has been applied.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -655,6 +698,244 @@ private int replyWindow()
...
+            MqttKafkaRouteConfig route)
+        {
+            this.originId = originId;
+            this.routedId = route.id;
+            this.topic = route.messages;
+            this.initialId = supplyInitialId.applyAsLong(routedId);
+            this.replyId = supplyReplyId.applyAsLong(initialId);
+        }
+
+        private void doKafkaBegin(
```
jfallows, line 732: Rename to `doKafkaBeginAt` to avoid confusion over method intent.

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaConditionMatcher.java
```diff
@@ -14,42 +14,79 @@
...
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import io.aklivity.zilla.runtime.binding.mqtt.kafka.config.MqttKafkaConditionConfig;
 
 public class MqttKafkaConditionMatcher
 {
-    private final Matcher topicMatcher;
-    public final String topic;
+    public final List<Matcher> subscribeMatchers;
```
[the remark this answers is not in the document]
bmaidics: We're checking if null in matchesClient which is in the route config, so that we only check try to bootstrap in routes where there is at least one subscribe "when".
Should I replace it with a getter though?

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaRouteConfig.java
```diff
@@ -60,15 +60,23 @@ boolean authorized(
...
+            .filter(m -> m.kind == MqttKafkaConditionMatcher.MqttKafkaConditionKind.SUBSCRIBE)
+            .allMatch(m -> m.matchesSubscribe(client));
     }
 
-    public boolean matchesTopicFilter(
+    public boolean matchesPublish(
         String topic)
     {
-        return when.isEmpty() || when.stream().anyMatch(m -> m.matches(topic));
+        return when.isEmpty() || when.stream().anyMatch(m -> m.matchesPublish(topic));
```
jfallows, line 80: This should filter by kind `PUBLISH`?

Now that we have `kind` do we need separate `matchesPublish` and `matchesSubscribe` methods? Perhaps just `matches` method and filter by kind here instead?

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaConditionMatcher.java
```diff
@@ -14,42 +14,81 @@
...
+    private final List<Matcher> subscribeMatchers;
+    private final List<Matcher> publishMatchers;
+    public final MqttKafkaConditionKind kind;
 
     public MqttKafkaConditionMatcher(
         MqttKafkaConditionConfig condition)
     {
-        this.topic = condition.topic;
-        this.topicMatcher = condition.topic != null ? asTopicMatcher(condition.topic) : null;
+        this.subscribeMatchers =
```
jfallows, line 34: Given that `MqttKafkaConditionConfig` can only be `subscribe` or `publish`, then we need `kind` on `MqttKafkaConditionConfig` and a generic `topics` field, instead of separate storage for `subscribes` and `publishes` on `MqttKafkaConditionConfig`, but without exposing `kind` as an explicit property in the JSON schema patch.

Then the kind can propagate here

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -146,11 +160,44 @@ public MqttKafkaSubscribeFactory(
...
+        this.reconnectDelay = config.bootstrapStreamReconnectDelay();
+        this.bootstrapStreams = new ArrayList<>();
+    }
+
+    @Override
+    public void onAttached(
+        long bindingId)
+    {
+        MqttKafkaBindingConfig binding = supplyBinding.apply(bindingId);
+        List<MqttKafkaRouteConfig> bootstrap = binding.bootstrapRoutes();
```
jfallows, line 180: These 2 lines should be inside the `if (bootstrapAvailable)` check.

jfallows (changes requested):

jfallows (approved):

PR #511, opened 2023-10-13
Add extraEnv to Deployment in the helm chart

### LINKED ISSUE #520: Support `extraEnv` in helm chart
Using the `env` variable in the `values.yaml` requires some unusual quoting in some cases (e.g. numeric values have to be double quoted like `'"42"'`); other helm charts use a different implementation which results in easier usage. This could be implemented by adding support for an `extraEnv` variable without introducing a breaking change.

We should implement an `extraEnv` variable in the `values.yaml` for more natural quoting syntax when providing values for environment variables in the `Deployment` definition.

**Describe alternatives you've considered**
n/a

**Additional context**
For example:
```
env:
  - name: PORT
    value: '"42"'
```
vs
```
extraEnv:
  - name: PORT
    value: "42"
```

attilakreiner:
## Description

Add `extraEnv` option to offer an easier way of overriding environment variables.

Fixes #520

### DIFF
--- a/cloud/helm-chart/src/main/helm/templates/deployment.yaml
+++ b/cloud/helm-chart/src/main/helm/templates/deployment.yaml
@@ -56,6 +56,9 @@ spec:
             - name: {{ $env.name }}
               value: {{ $env.value }}
           {{- end }}
+          {{- if .Values.extraEnv }}
+          {{- toYaml .Values.extraEnv | nindent 12 }}
+          {{- end }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           ports:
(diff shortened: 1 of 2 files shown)


vordimous:
This LGTM, IT doesn't break any existing `values.yaml` that have implemented a workaround and removes the `env` attribute from the default `values.yaml` meaning new users will see and use `extraEnv`.

jfallows:
@attilakreiner this looks great!

Please file a `zilla` enhancement issue indicating the `env` syntax we have now (weird escaping) and the improved `extraEnv` syntax we want to support without needing the same escaping (implemented by this change).

Then add the issue number to `Fixes #` in the description above to link this PR to the issue so it closes automatically when this PR is merged.

vordimous (approved):

jfallows (approved):

PR #519, opened 2023-10-16
Propagate gRPC status code when not ok

### LINKED ISSUE #504: gRPC method call doesn't respond when status code is not OK
**Describe the bug**
I have a custom gRPC service, the response of its methods sometimes return status codes different to "0 OK" (NotFound, AlreadyExists, PermissionDenied, Unauthenticated). Zilla doesn't respond when the status code of my gRPC is different to "0 OK".

**To Reproduce**
1. Checkout https://github.com/fgarcia-code/examples/tree/main/zilla/echo and start.
2. Use Postman to call the gRPC EchoStatus service with status "OK":
   image
3. Call with status "Not Found":
   image
4. Review UI for Apache Kafka, the service publish in `echo-requests`, but not in `echo-responses`.
   image

**Expected behavior**
I think that Zilla should responds in the same way with the status code.

I don't know if I'm doing something wrong.

jfallows:
## Description

When gRPC response status code arrives in HTTP response headers (not trailers), typically due to being an error of some kind (not OK status), then make sure it is conveyed consistently back to the gRPC client.

Fixes #504


runtime/binding-grpc/src/main/java/io/aklivity/zilla/runtime/binding/grpc/internal/stream/GrpcClientFactory.java
```diff
@@ -450,21 +465,49 @@ private void doAppData(
...
-            if (!GrpcState.replyClosed(state))
+            GrpcAbortExFW abortEx = grpcStatus != null
+                    ? grpcAbortExRW.wrap(extBuffer, 0, extBuffer.capacity())
+                            .typeId(grpcTypeId)
+                            .status(grpcStatus)
+                            .build()
+                    : grpcAbortExRW.wrap(extBuffer, 0, extBuffer.capacity())
+                            .typeId(grpcTypeId)
+                            .status(HEADER_VALUE_GRPC_INTERNAL_ERROR)
+                            .build();
```
akrambek, line 494: Seems like the only diff is status. Wouldn't it be shorter if we just hoist that value only?

runtime/binding-grpc/src/main/java/io/aklivity/zilla/runtime/binding/grpc/internal/stream/GrpcClientFactory.java
```diff
@@ -450,21 +465,49 @@ private void doAppData(
...
-        private void doAppAbort(
+        private void doAppAbortDeferring(
+            long traceId,
+            long authorization,
+            String16FW grpcStatus)
+        {
+            this.grpcStatus = grpcStatus != null ? grpcStatus.asString() : null;
+            this.state = GrpcState.closingReply(state);
+
+            if (GrpcState.replyOpened(state))
```
akrambek, line 476: We are setting the state to closing; the next condition is checking if it is still open. Is my understanding correct?

jfallows: Yes, state is not a single value, it flips bits for different parts of the overall state, so both are true at the same time.

runtime/binding-grpc/src/main/java/io/aklivity/zilla/runtime/binding/grpc/internal/stream/GrpcClientFactory.java
jfallows: Right, but they are different types so that doesn't work. 

We need to have the grpcStatus live longer than a method call, so we keep it as a String, whereas the common case precomputes the value, not as a String.

akrambek: Oh I see I missed that thanks

specs/binding-grpc-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/kafka/streams/kafka/produce/unary.rpc.error/client.rpt
```diff
@@ -0,0 +1,110 @@
...
+                                .build()
+                            .build()}
+
+write flush
+
+write close
+write notify SENT_ASYNC_REQUEST
+read closed
+
+connect await SENT_ASYNC_REQUEST
```
akrambek, line 70: Why it is called `ASYNC` as far as I know there is no such terminology in grpc as well

jfallows: Async over a pair of Kafka topics.

PR #533, opened 2023-10-25
Validator Interface Update & Converter Changes

aDaemonThread:
For previous review comments refer: https://github.com/aklivity/zilla/pull/499


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -560,7 +576,24 @@ public void writeProduceEntryStart(
...
+        }
+        else
+        {
+            final ValueConsumer writeValue = (buffer, index, length) ->
+            {
+                Varint32FW newLength = varIntRW.set(length).build();
+                logFile.appendBytes(newLength);
+                logFile.appendBytes(buffer, index, length);
+            };
+            OctetsFW value = key.value();
```
jfallows, line 594: This line should be hoisted before `if-else` block as it is called twice; here and in the `if` condition.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -560,7 +576,24 @@ public void writeProduceEntryStart(
...
+            validated = valueValidator.validate(value.buffer(), value.offset(), value.sizeof(), writeValue);
```
jfallows, line 595: If validation is unsuccessful, then we probably don't want to be doing the rest of the work below in this method, agree?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -560,7 +576,24 @@ public void writeProduceEntryStart(
...
+            logFile.appendBytes(key);
+        }
+        else
+        {
+            final ValueConsumer writeValue = (buffer, index, length) ->
+            {
+                Varint32FW newLength = varIntRW.set(length).build();
+                logFile.appendBytes(newLength);
+                logFile.appendBytes(buffer, index, length);
+            };
```
jfallows, line 593: This `writeKey` lambda can be hoisted to the constructor and reused.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -581,12 +614,15 @@ public void writeProduceEntryStart(
...
+        return validated;
     }
 
-    public void writeProduceEntryContinue(
+    public int writeProduceEntryContinue(
         Node head,
         MutableInteger position,
-        OctetsFW payload)
+        OctetsFW payload,
+        ValueValidator valueValidator)
```
jfallows, line 625: Rename parameter to `validateValue`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -415,13 +440,25 @@ public void writeEntryContinue(
         final int logRequired = payload.sizeof();
         assert logAvailable >= logRequired;
 
-        logFile.appendBytes(payload.buffer(), payload.offset(), payload.sizeof());
+        if (payload != null)
+        {
+            final ValueConsumer writeValue = (buffer, index, length) ->
+            {
+                logFile.appendBytes(buffer, index, length);
+            };
```
jfallows, line 448:
```suggestion
            final ValueConsumer writeValue = logFile::appendBytes;
```

Also, this can be hoisted to constructor and reused.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -415,13 +440,25 @@ public void writeEntryContinue(
...
+            int validated = valueValidator.validate(payload.buffer(), payload.offset(), payload.sizeof(), writeValue);
+            if (validated == -1)
+            {
+                // For Fetch Validation failure, we still push the event to Cache
+                logFile.appendBytes(payload.buffer(), payload.offset(), payload.sizeof());
```
jfallows, line 453: Suggest calling `writeValue.accept(...)` instead, for clarity.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -341,7 +344,8 @@ public void writeEntryStart(
         int valueLength,
         KafkaCacheEntryFW ancestor,
         int entryFlags,
-        KafkaDeltaType deltaType)
+        KafkaDeltaType deltaType,
+        ValueValidator valueValidator)
```
jfallows, line 348: Rename parameter to `validateKey`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -535,7 +550,8 @@ public void writeProduceEntryStart(
         long keyHash,
         int valueLength,
         ArrayFW<KafkaHeaderFW> headers,
-        int trailersSizeMax)
+        int trailersSizeMax,
+        ValueValidator valueValidator)
```
jfallows, line 554: Rename parameter to `validateValue`.

incubator/validator-json/src/test/java/io/aklivity/zilla/runtime/validator/json/JsonValueValidatorTest.java
```diff
@@ -119,6 +122,34 @@ public void shouldVerifyInvalidJsonData()
...
+        CatalogConfig catalogConfig = new CatalogConfig("test0", "test", new TestCatalogOptionsConfig(SCHEMA));
+        LongFunction<CatalogHandler> handler = value -> context.attach(catalogConfig);
+        JsonValueValidator validator = new JsonWriteValueValidator(config, resolveId, handler);
+
+        DirectBuffer data = new UnsafeBuffer();
+
+        String payload = "{" +
+                "\"id\": \"123\"," +
+                "\"status\": \"OK\"" +
+                "}";
```
jfallows, line 145: Fix indentation for readability to more closely align with formatted JSON.
```
        String payload =
            "{" +
                "\"id\": \"123\"," +
                "\"status\": \"OK\"" +
            "}";
```

incubator/validator-json/src/test/java/io/aklivity/zilla/runtime/validator/json/JsonValidatorFactoryTest.java
```diff
@@ -26,28 +26,38 @@
...
+        // THEN
+        assertThat(jsonValueValidator, instanceOf(JsonValueValidator.class));
+    }
+
+    @Test
+    public void shouldCreateWriteValidator()
+    {
         // WHEN
-        Validator jsonValidator = factory.create(validator, resolveId, supplyCatalog);
+        ValueValidator jsonValueValidator = factory.createValueWriter(validator, resolveId, supplyCatalog);
```
jfallows, line 58: Rename to `writer`.

incubator/validator-json/src/test/java/io/aklivity/zilla/runtime/validator/json/JsonValidatorFactoryTest.java
```diff
@@ -26,28 +26,38 @@
...
 
         // THEN
-        assertThat(jsonValidator, instanceOf(JsonValidator.class));
+        assertThat(jsonValueValidator, instanceOf(JsonValueValidator.class));
```
jfallows, line 61: Let's check for the more specific subclass `JsonWriteValueValidator` instead.

incubator/validator-json/src/test/java/io/aklivity/zilla/runtime/validator/json/JsonValueValidatorTest.java
```diff
@@ -59,6 +60,8 @@ public class JsonValidatorTest
                     "]" +
                 "}";
 
+    private final ValueConsumer valueFunction = (buffer, index, length) -> {};
```
jfallows, line 63: Rename to `consumer`.

incubator/validator-json/src/main/java/io/aklivity/zilla/runtime/validator/json/JsonValueValidator.java
```diff
@@ -66,54 +68,30 @@ public JsonValidator(
...
                 schema = handler.resolve(schemaId);
             }
         }
-
-        return schema != null && validate(schema, payloadBytes);
+        return schema;
     }
 
-    private boolean validate(
+    boolean validate(
```
jfallows, line 94: A lot of work is done in this method on every invocation, even though much of it is guaranteed to be identical if the schemaId is constant.

Lets aim to optimize this a bit more by hoisting more of the predictable work to be done just once if possible.

incubator/validator-json/src/main/java/io/aklivity/zilla/runtime/validator/json/JsonReadValueValidator.java
```diff
@@ -0,0 +1,85 @@
...
+            schema = fetchSchema(schemaId);
+            if (schema != null && validate(schema, payloadBytes))
+            {
+                int size = payloadBytes.length;
+                value.wrap(data);
+                next.accept(value, index, size);
+                valLength = size;
+            }
+        }
+        return valLength;
```
jfallows, line 83: Lets try to eliminate as much of the instantiation and copying as possible from this method.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/StringValueValidatorTest.java
```diff
@@ -14,121 +14,123 @@
...
 import org.agrona.concurrent.UnsafeBuffer;
 import org.junit.Test;
 
+import io.aklivity.zilla.runtime.engine.validator.function.ValueConsumer;
 import io.aklivity.zilla.runtime.validator.core.config.StringValidatorConfig;
 
-public class StringValidatorTest
+public class StringValueValidatorTest
 {
+    private final ValueConsumer valueFunction = (buffer, index, length) -> {};
```
jfallows, line 30: Rename to `consumer`, perhaps define `NOP` on `ValueConsumer` instead?

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/StringValidatorFactoryTest.java
```diff
@@ -25,25 +25,59 @@
...
+    ToLongFunction<String> resolveId = mock(ToLongFunction.class);
+    LongFunction<CatalogHandler> supplyCatalog = mock(LongFunction.class);
+    StringValidatorFactory factory = new StringValidatorFactory();
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void shouldCreateValueReader()
+    {
+        // WHEN
+        ValueValidator valueValidator = factory.createValueReader(validator, resolveId, supplyCatalog);
```
jfallows, line 45: Rename to `reader`.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/StringValidatorFactoryTest.java
```diff
@@ -25,25 +25,59 @@
...
+        // THEN
+        assertThat(valueValidator, instanceOf(StringValueValidator.class));
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void shouldCreateValueWriter()
+    {
+        // WHEN
+        ValueValidator valueValidator = factory.createValueWriter(validator, resolveId, supplyCatalog);
```
jfallows, line 56: Rename to `writer`.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/StringValidatorFactoryTest.java
```diff
@@ -25,25 +25,59 @@
...
-    public void shouldCreate()
+    public void shouldCreateFragmentReader()
     {
-        // GIVEN
-        ValidatorConfig validator = new StringValidatorConfig("utf_8");
-        ToLongFunction<String> resolveId = mock(ToLongFunction.class);
-        LongFunction<CatalogHandler> supplyCatalog = mock(LongFunction.class);
-        StringValidatorFactory factory = new StringValidatorFactory();
+        // WHEN
+        FragmentValidator fragmentValidator = factory.createFragmentReader(validator, resolveId, supplyCatalog);
```
jfallows, line 67: Rename to `reader`.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/StringValidatorFactoryTest.java
```diff
@@ -25,25 +25,59 @@
...
+        assertThat(fragmentValidator, instanceOf(StringFragmentValidator.class));
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void shouldCreateFragmentWriter()
+    {
         // WHEN
-        Validator stringValidator = factory.create(validator, resolveId, supplyCatalog);
+        FragmentValidator fragmentValidator = factory.createFragmentWriter(validator, resolveId, supplyCatalog);
```
jfallows, line 78: Rename to `writer`.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/StringFragmentValidatorTest.java
```diff
@@ -0,0 +1,86 @@
...
+public class StringFragmentValidatorTest
+{
+    private final FragmentConsumer fragmentConsumer = (flags, buffer, index, length) -> {};
+    private static final int FLAGS_INIT = 0x02;
+    private static final int FLAGS_FIN = 0x01;
+
+    @Test
+    public void shouldVerifyCompleteAndValidMessage()
+    {
+        StringValidatorConfig config = new StringValidatorConfig("utf_8");
```
jfallows, line 37: We need config builder for string validator config.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/LongValueValidatorTest.java
```diff
@@ -14,19 +14,20 @@
...
+import io.aklivity.zilla.runtime.engine.validator.function.ValueConsumer;
 import io.aklivity.zilla.runtime.validator.core.config.LongValidatorConfig;
 
-public class LongValidatorTest
+public class LongValueValidatorTest
 {
     private final LongValidatorConfig config = new LongValidatorConfig();
-    private final LongValidator validator = new LongValidator(config);
+    private final LongValueValidator validator = new LongValueValidator(config);
+    private final ValueConsumer valueFunction = (buffer, index, length) -> {};
```
jfallows, line 30: Rename to `consumer`, perhaps define `NOP` on `ValueConsumer` instead?

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/LongValidatorFactoryTest.java
```diff
@@ -25,25 +25,36 @@
...
+        assertThat(longValueValidator, instanceOf(LongValueValidator.class));
+    }
 
+    @Test
+    @SuppressWarnings("unchecked")
+    public void shouldCreateWriteValidator()
+    {
         // WHEN
-        Validator longValidator = factory.create(validator, resolveId, supplyCatalog);
+        ValueValidator longValueValidator = factory.createValueWriter(validator, resolveId, supplyCatalog);
```
jfallows, line 55: Rename to `writer`.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/LongValidatorFactoryTest.java
```diff
@@ -25,25 +25,36 @@
...
-    public void shouldCreate()
+    public void shouldCreateReadValidator()
     {
-        // GIVEN
-        ValidatorConfig validator = new LongValidatorConfig();
-        ToLongFunction<String> resolveId = mock(ToLongFunction.class);
-        LongFunction<CatalogHandler> supplyCatalog = mock(LongFunction.class);
-        LongValidatorFactory factory = new LongValidatorFactory();
+        // WHEN
+        ValueValidator longValueValidator = factory.createValueReader(validator, resolveId, supplyCatalog);
```
jfallows, line 44: Rename to `reader`.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/LongValidatorFactoryTest.java
```diff
@@ -25,25 +25,36 @@
...
 import io.aklivity.zilla.runtime.validator.core.config.LongValidatorConfig;
 
 public class LongValidatorFactoryTest
 {
+    // GIVEN
+    ValidatorConfig validator = new LongValidatorConfig();
+    ToLongFunction<String> resolveId = mock(ToLongFunction.class);
+    LongFunction<CatalogHandler> supplyCatalog = mock(LongFunction.class);
+    LongValidatorFactory factory = new LongValidatorFactory();
+
```
jfallows, line 38: The `GIVEN` section should be inside the test method.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/IntegerValueValidatorTest.java
```diff
@@ -14,19 +14,20 @@
...
+import io.aklivity.zilla.runtime.engine.validator.function.ValueConsumer;
 import io.aklivity.zilla.runtime.validator.core.config.IntegerValidatorConfig;
 
-public class IntegerValidatorTest
+public class IntegerValueValidatorTest
 {
     private final IntegerValidatorConfig config = new IntegerValidatorConfig();
-    private final IntegerValidator validator = new IntegerValidator(config);
+    private final IntegerValueValidator validator = new IntegerValueValidator(config);
+    private final ValueConsumer writeValue = (buffer, index, length) -> {};
```
jfallows, line 30: Rename to `consumer`, perhaps define `NOP` on `ValueConsumer` instead?

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/IntegerValidatorFactoryTest.java
```diff
@@ -25,25 +25,36 @@
...
+        assertThat(integerValueValidator, instanceOf(IntegerValueValidator.class));
+    }
 
+    @Test
+    @SuppressWarnings("unchecked")
+    public void shouldCreateWriteValidator()
+    {
         // WHEN
-        Validator integerValidator = factory.create(validator, resolveId, supplyCatalog);
+        ValueValidator integerValueValidator = factory.createValueWriter(validator, resolveId, supplyCatalog);
```
jfallows, line 55: Rename to `writer`.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/IntegerValidatorFactoryTest.java
```diff
@@ -25,25 +25,36 @@
...
-    public void shouldCreate()
+    public void shouldCreateReadValidator()
     {
-        // GIVEN
-        ValidatorConfig validator = new IntegerValidatorConfig();
-        ToLongFunction<String> resolveId = mock(ToLongFunction.class);
-        LongFunction<CatalogHandler> supplyCatalog = mock(LongFunction.class);
-        IntegerValidatorFactory factory = new IntegerValidatorFactory();
+        // WHEN
+        ValueValidator integerValueValidator = factory.createValueReader(validator, resolveId, supplyCatalog);
```
jfallows, line 44: Rename to `reader`.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/IntegerValidatorFactoryTest.java
```diff
@@ -25,25 +25,36 @@
...
 import io.aklivity.zilla.runtime.validator.core.config.IntegerValidatorConfig;
 
 public class IntegerValidatorFactoryTest
 {
+    // GIVEN
+    ValidatorConfig validator = new IntegerValidatorConfig();
+    ToLongFunction<String> resolveId = mock(ToLongFunction.class);
+    LongFunction<CatalogHandler> supplyCatalog = mock(LongFunction.class);
+    IntegerValidatorFactory factory = new IntegerValidatorFactory();
+
```
jfallows, line 38: `GIVEN` section should be inside the test method.

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/StringValueValidator.java
```diff
@@ -18,14 +18,15 @@
...
+import io.aklivity.zilla.runtime.engine.validator.function.ValueConsumer;
 import io.aklivity.zilla.runtime.validator.core.config.StringValidatorConfig;
 
-public final class StringValidator implements Validator
+public final class StringValueValidator implements ValueValidator
 {
     private Predicate<byte[]> predicate;
 
-    public StringValidator(
+    public StringValueValidator(
```
jfallows, line 29: Define a static `ENCODING_VALIDATORS`map of string (encoding) to Predicate<byte[]> and make isValidXXX method static too.
Define a static catch-all case, `INVALID_ENCODING`, with lambda `b -> false`.
Then in constructor use `ENCODING_VALIDATORS.getOrDefault(config.encoding, INVALID_ENCODING)` to lookup correct predicate.

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/StringValidatorFactory.java
```diff
@@ -39,11 +40,38 @@ public URL schema()
     }
 
     @Override
-    public Validator create(
+    public ValueValidator createValueReader(
```
jfallows, line 43: Now that validator is created by engine, can engine automatically assign `id` on `ValidatorConfig` instead of passing `resolveId` around?

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/StringFragmentValidator.java
```diff
@@ -0,0 +1,153 @@
...
+                break;
+            }
+            else
+            {
+                i += 2;
+            }
+            status = true;
+        }
+        return status;
+    }
```
jfallows, line 152: This code should not be duplicated.

Perhaps define an enum with each supported encoding constant and a validate method implementing what's needed for that encoding?

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/LongValueValidator.java
```diff
@@ -16,39 +16,30 @@
...
-        int length)
-    {
-        return length == 8;
+        int valLength = -1;
+        if (length == 8)
+        {
+            next.accept(data, index, length);
+            valLength = length;
+        }
+        return valLength;
```
jfallows, line 43:
```suggestion
        boolean value = length == 8

        if (valid)
        {
            next.accept(data, index, length);
        }

        return valid ? length : -1;
```

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/IntegerValueValidator.java
```diff
@@ -16,38 +16,29 @@
...
-        int length)
-    {
-        return length == 4;
+        int valLength = -1;
+        if (length == 4)
+        {
+            next.accept(data, index, length);
+            valLength = length;
+        }
+        return valLength;
```
jfallows, line 42:
```suggestion
        boolean valid = length == 4;

        if (valid)
        {
            next.accept(data, index, length);
        }

        return valid ? length : -1;
```

incubator/validator-core.spec/src/main/scripts/io/aklivity/zilla/specs/validator/core/schema/string.schema.patch.json
```diff
@@ -3,5 +3,40 @@
...
+                    "type":
+                    {
+                        "const": "string"
+                    },
+                    "encoding":
+                    {
+                        "type": "string",
+                        "enum":
+                        [
+                            "utf_8"
```
jfallows, line 35: Also `utf_16` here?
Should it be `utf_8` or `utf-8`, etc?

incubator/validator-avro/src/test/java/io/aklivity/zilla/runtime/validator/avro/AvroValueValidatorTest.java
```diff
@@ -78,55 +80,146 @@ public void shouldVerifyValidAvroEvent()
...
+        DirectBuffer expected = new UnsafeBuffer();
+
+        byte[] bytes = {0x00, 0x00, 0x00, 0x00, 0x01, 0x06, 0x69, 0x64,
+            0x30, 0x10, 0x70, 0x6f, 0x73, 0x69, 0x74, 0x69, 0x76, 0x65};
+        expected.wrap(bytes, 0, bytes.length);
+
+        String payload = "{" +
+                "\"id\":\"id0\"," +
+                "\"status\":\"positive\"" +
+                "}";
```
jfallows, line 218: Use indentation aligned with literal JSON layout, including indentation.
See similar feedback with example in JSON validator.

incubator/validator-avro/src/test/java/io/aklivity/zilla/runtime/validator/avro/AvroValidatorFactoryTest.java
```diff
@@ -17,35 +17,52 @@
...
-    public void shouldCreate()
+    public void shouldCreateReadValidator()
     {
-        // GIVEN
-        ValidatorConfig validator = new AvroValidatorConfig(List.of(new CatalogedConfig("test0", List.of())), "test-value");
-        ToLongFunction<String> resolveId = i -> 0L;
-        LongFunction<CatalogHandler> supplyCatalog = i -> new TestCatalogHandler(new TestCatalogOptionsConfig("schema0"));
-        AvroValidatorFactory factory = new AvroValidatorFactory();
+        // WHEN
+        ValueValidator avroValueValidator = factory.createValueReader(validator, resolveId, supplyCatalog);
```
jfallows, line 53: Rename to `reader`.

incubator/validator-avro/src/test/java/io/aklivity/zilla/runtime/validator/avro/AvroValidatorFactoryTest.java
```diff
@@ -17,35 +17,52 @@
...
+        // THEN
+        assertThat(avroValueValidator, instanceOf(AvroValueValidator.class));
+    }
+
+    @Test
+    public void shouldCreateWriteValidator()
+    {
         // WHEN
-        Validator avroValidator = factory.create(validator, resolveId, supplyCatalog);
+        ValueValidator avroValueValidator = factory.createValueWriter(validator, resolveId, supplyCatalog);
```
jfallows, line 63: Rename to `writer`.

incubator/validator-avro/src/test/java/io/aklivity/zilla/runtime/validator/avro/AvroValidatorFactoryTest.java
```diff
@@ -17,35 +17,52 @@
...
 
         // THEN
-        assertThat(avroValidator, instanceOf(AvroValidator.class));
+        assertThat(avroValueValidator, instanceOf(AvroValueValidator.class));
```
jfallows, line 66: Test for more specific subclass type.

incubator/validator-avro/src/test/java/io/aklivity/zilla/runtime/validator/avro/AvroValidatorFactoryTest.java
```diff
@@ -17,35 +17,52 @@
...
-        // GIVEN
-        ValidatorConfig validator = new AvroValidatorConfig(List.of(new CatalogedConfig("test0", List.of())), "test-value");
-        ToLongFunction<String> resolveId = i -> 0L;
-        LongFunction<CatalogHandler> supplyCatalog = i -> new TestCatalogHandler(new TestCatalogOptionsConfig("schema0"));
-        AvroValidatorFactory factory = new AvroValidatorFactory();
+        // WHEN
+        ValueValidator avroValueValidator = factory.createValueReader(validator, resolveId, supplyCatalog);
 
+        // THEN
+        assertThat(avroValueValidator, instanceOf(AvroValueValidator.class));
```
jfallows, line 56: Test for more specific subclass type.

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValueValidator.java
```diff
@@ -14,128 +14,89 @@
...
+    final List<CatalogedConfig> catalogs;
+    final SchemaConfig catalog;
+    final Long2ObjectHashMap<CatalogHandler> handlersById;
+    final CatalogHandler handler;
+    final DecoderFactory decoder;
+    final EncoderFactory encoder;
+    final String subject;
+    final String expect;
+    DatumReader reader;
+    DatumWriter writer;
```
jfallows, line 49: These fields are intended to be `protected`, used by subclasses only?

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroReadValueValidator.java
```diff
@@ -0,0 +1,132 @@
...
+                new GenericDatumWriter<>(record.getSchema());
+            writer.write(record, jsonEncoder);
+            jsonEncoder.flush();
+            outputStream.close();
+        }
+        catch (Exception e)
+        {
+        }
+        return outputStream.toByteArray();
+    }
```
jfallows, line 131: Lets aim to avoid repeated work per invocation that can be hoisted and done just once instead, then reused.

jfallows (changes requested):

incubator/validator-core.spec/src/main/scripts/io/aklivity/zilla/specs/validator/core/schema/string.schema.patch.json
aDaemonThread: I was trying to align this with `StandardCharsets`. For example: `Charset UTF_8`

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaBindingConfig.java
```diff
@@ -120,4 +139,16 @@ public KafkaOffsetType supplyDefaultOffset(
...
+        String topic)
+    {
+        return readers != null ? readers.get(topic) : new KafkaTopicType(ValueValidator.NONE, ValueValidator.NONE);
+    }
+
+    public KafkaTopicType resolveWriteValidator(
+        String topic)
+    {
+        return writers != null ? writers.get(topic) : new KafkaTopicType(ValueValidator.NONE, ValueValidator.NONE);
+    }
```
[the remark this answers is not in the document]
jfallows: Alternatively, resolve to return ValueValidator instead of KafkaTopicType.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttBindingConfig.java
```diff
@@ -44,8 +45,9 @@ public final class MqttBindingConfig
     public final MqttOptionsConfig options;
     public final List<MqttRouteConfig> routes;
     public final Function<String, String> credentials;
-    public final Map<String, Validator> topics;
+    public final Map<String, ValueValidator> topics;
     public final ToLongFunction<String> resolveId;
+    public final GuardHandler guard;
```
jfallows, line 50: Is this field used?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/OptionsConfig.java
```diff
@@ -15,6 +15,21 @@
...
+import java.util.Collections;
+import java.util.List;
+
 public class OptionsConfig
 {
+    public final List<ValidatorConfig> validators;
+
+    public OptionsConfig()
+    {
+        this(Collections.emptyList());
```
jfallows, line 27: Use static import for `emptyList()`.

runtime/engine/pom.xml
```diff
@@ -27,7 +27,7 @@
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
     <jacoco.coverage.ratio>0.70</jacoco.coverage.ratio>
-    <jacoco.missed.count>4</jacoco.missed.count>
+    <jacoco.missed.count>6</jacoco.missed.count>
```
jfallows, line 30: Let's get this back down to `4` if possible.

jfallows (changes requested):

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttBindingConfig.java
aDaemonThread: yes, in MqttServerFactory class

`sessionAuth = guard.reauthorize(initialId, credentialsMatch);`

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValueValidator.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+        else if (catalog != null)
+        {
+            schemaId = handler.resolve(subject, catalog.version);
+            if (schemaId > 0)
+            {
+                schema = handler.resolve(schemaId);
+            }
+        }
+        return schema != null ? new Schema.Parser().parse(schema) : null;
```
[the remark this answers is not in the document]
aDaemonThread: unable to reuse the schema parser.

`org.apache.avro.SchemaParseException: Can't redefine: io.aklivity.example.Event

	at org.apache.avro.Schema$Names.put(Schema.java:1604)
	at org.apache.avro.Schema$Names.add(Schema.java:1598)
	at org.apache.avro.Schema.parse(Schema.java:1723)
	at org.apache.avro.Schema$Parser.parse(Schema.java:1471)
	at org.apache.avro.Schema$Parser.parse(Schema.java:1458)`

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValueValidator.java
```diff
@@ -0,0 +1,101 @@
...
+        boolean status = false;
+        try
+        {
+            reader = new GenericDatumReader(schema);
+            reader.read(null, decoder.binaryDecoder(bytes, offset, length, null));
+            status = true;
+        }
+        catch (Exception e)
+        {
+        }
```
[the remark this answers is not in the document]
aDaemonThread: updated the catch block with specific exception.

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroReadValidator.java
```diff
@@ -0,0 +1,138 @@
...
+                new SpecificDatumWriter<>(record.getSchema()) :
+                new GenericDatumWriter<>(record.getSchema());
+            writer.write(record, jsonEncoder);
+            jsonEncoder.flush();
+            outputStream.close();
+        }
+        catch (IOException e)
+        {
+        }
+        return outputStream.toByteArray();
```
jfallows, line 136: Note that `.toByteArray()` is creating a copy of the underlying array so there is more left to optimize, but this will likely be the last one we sort out.

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValidator.java
```diff
@@ -14,132 +14,99 @@
...
+        int schemaId)
     {
-        boolean status = false;
-        byte[] payloadBytes = new byte[length];
-        data.getBytes(0, payloadBytes);
-        ByteBuffer byteBuf = ByteBuffer.wrap(payloadBytes);
+        Schema schema = null;
 
-        if (byteBuf.get() == MAGIC_BYTE)
+        if (schemaId == 0)
```
jfallows, line 74: `NO_SCHEMA_ID`

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValidator.java
```diff
@@ -14,132 +14,99 @@
...
-    {
-        boolean status = false;
-        String schema = null;
-        int schemaId = catalog != null ? catalog.id : 0;
-
-        byte[] payloadBytes = new byte[length];
-        data.getBytes(0, payloadBytes);
-
-        if (schemaId > 0)
+        if (cache.containsKey(schemaId))
```
jfallows, line 79: Use `cache.computeIfAbsent(schemaId, this::resolveSchema)` and add a private method called `Schema resolveSchema(int schemaId)` that uses the `handler` to resolve the `schemaId` then parses and returns the `Schema`, or `null` if not resolved by `handler`.

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValidator.java
```diff
@@ -14,132 +14,99 @@
...
+        int offset,
+        int length)
     {
         boolean status = false;
         try
         {
-            reader = new GenericDatumReader(parser.parse(schema));
-            reader.read(null, decoder.binaryDecoder(payloadBytes, null));
+            reader = new GenericDatumReader<>(schema);
+            reader.read(null, decoder.binaryDecoder(bytes, offset, length, null));
```
jfallows, line 106: Use the same optimizations here as read validator for `reader` and `decoderFactory` /  `decoder`.

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValidator.java
```diff
@@ -14,132 +14,99 @@
...
             status = true;
         }
-        catch (Exception e)
+        catch (IOException | AvroRuntimeException e)
```
jfallows, line 109: Rename to `ex` and add `ex.printStackTrace()` to catch block body.

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroWriteValidator.java
```diff
@@ -0,0 +1,143 @@
...
+            writer = new GenericDatumWriter<>(schema);
+            BinaryEncoder binaryEncoder = encoder.binaryEncoder(outputStream, null);
+            writer.write(record, binaryEncoder);
+            binaryEncoder.flush();
+            outputStream.close();
+        }
+        catch (IOException e)
+        {
+        }
+        return outputStream.toByteArray();
```
jfallows, line 141: Apply equivalent feedback here as in read validator.

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/IntegerValidatorFactory.java
```diff
@@ -39,9 +39,32 @@ public URL schema()
...
+    @Override
+    public FragmentValidator createFragmentReader(
+        ValidatorConfig config,
+        LongFunction<CatalogHandler> supplyCatalog)
+    {
+        return new IntegerValidator(IntegerValidatorConfig.class.cast(config));
+    }
+
+    @Override
+    public FragmentValidator createFragmentWriter(
```
jfallows, line 66: Follow same pattern with private `createReader` and `createWriter` methods, delegating from public methods accordingly.

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/LongValidatorFactory.java
```diff
@@ -39,9 +39,32 @@ public URL schema()
...
+    @Override
+    public FragmentValidator createFragmentReader(
+        ValidatorConfig config,
+        LongFunction<CatalogHandler> supplyCatalog)
+    {
+        return new LongValidator(LongValidatorConfig.class.cast(config));
+    }
+
+    @Override
+    public FragmentValidator createFragmentWriter(
```
jfallows, line 66: Follow same pattern with private `createReader` and `createWriter` methods, delegating from public methods accordingly.

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/StringEncoding.java
```diff
@@ -0,0 +1,154 @@
...
+    UTF_8
+    {
+        @Override
+        public boolean validate(
+            DirectBuffer data,
+            int index,
+            int length)
+        {
+            byte[] bytes = new byte[length];
+            data.getBytes(index, bytes);
```
jfallows, line 30: Remove these 2 lines and use `DirectBuffer` API to access bytes instead of `bytes[...]` array syntax.

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/StringEncoding.java
```diff
@@ -0,0 +1,154 @@
...
+    UTF_16
+    {
+        @Override
+        public boolean validate(
+            DirectBuffer data,
+            int index,
+            int length)
+        {
+            byte[] bytes = new byte[length];
+            data.getBytes(index, bytes);
```
jfallows, line 80: Same as above.

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/IntegerValidator.java
```diff
@@ -16,38 +16,55 @@
...
 
-import io.aklivity.zilla.runtime.engine.validator.Validator;
+import io.aklivity.zilla.runtime.engine.validator.FragmentValidator;
+import io.aklivity.zilla.runtime.engine.validator.ValueValidator;
+import io.aklivity.zilla.runtime.engine.validator.function.FragmentConsumer;
+import io.aklivity.zilla.runtime.engine.validator.function.ValueConsumer;
 import io.aklivity.zilla.runtime.validator.core.config.IntegerValidatorConfig;
 
-public class IntegerValidator implements Validator
+public class IntegerValidator implements ValueValidator, FragmentValidator
```
jfallows, line 25: Need to support ascii format integers for OpenAPI query params etc.

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/StringValidatorFactory.java
```diff
@@ -39,9 +39,32 @@ public URL schema()
...
+    @Override
+    public FragmentValidator createFragmentReader(
+        ValidatorConfig config,
+        LongFunction<CatalogHandler> supplyCatalog)
+    {
+        return new StringValidator(StringValidatorConfig.class.cast(config));
+    }
+
+    @Override
+    public FragmentValidator createFragmentWriter(
```
jfallows, line 66: Follow same pattern with private `createReader` and `createWriter` methods, delegating from public methods accordingly.

incubator/validator-core/src/test/java/io/aklivity/zilla/runtime/validator/core/StringValidatorTest.java
```diff
@@ -14,121 +14,187 @@
...
+import io.aklivity.zilla.runtime.engine.validator.function.FragmentConsumer;
+import io.aklivity.zilla.runtime.engine.validator.function.ValueConsumer;
 import io.aklivity.zilla.runtime.validator.core.config.StringValidatorConfig;
 
 public class StringValidatorTest
 {
+    private static final int FLAGS_INIT = 0x02;
+
     @Test
     public void shouldVerifyValidUTF8()
```
jfallows, line 35: In class names and method names, use `InitCaps` for acronyms, so `shouldVerifyValidUtf8()`.

incubator/validator-json/src/main/java/io/aklivity/zilla/runtime/validator/json/JsonReadValidator.java
```diff
@@ -0,0 +1,91 @@
...
+        if (data.getByte(index) == MAGIC_BYTE)
+        {
+            progress += BitUtil.SIZE_OF_BYTE;
+            schemaId = data.getInt(index + progress);
+            progress += BitUtil.SIZE_OF_INT;
+        }
+        else
+        {
+            schemaId = catalog != null ? catalog.id : 0;
+        }
```
jfallows, line 81: Same feedback as `AvroReadValidator` here.

jfallows (changes requested):

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -4727,6 +4709,15 @@ private int calculateSubscribeFlags(
             return flags;
         }
 
+        private boolean validContent(
+            String topic,
+            OctetsFW payload)
+        {
+            final ValueValidator contentValueValidator = supplyValidator.apply(topic);
```
jfallows, line 4716: Just `validator`.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -4727,6 +4709,15 @@ private int calculateSubscribeFlags(
...
+            return contentValueValidator == null ||
```
jfallows, line 4717: Should we introduce a `ValueValidator.NOP` to return from `supplyValidator` so we can avoid these `null` checks?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/catalog/CatalogHandler.java
```diff
@@ -18,6 +18,11 @@
 public interface CatalogHandler
 {
     int NO_SCHEMA_ID = 0;
+    String TEST = "test"; // Added for unit test & IT purpose
+    String SCHEMA_REGISTRY = "schema-registry";
+    String INLINE = "inline";
```
jfallows, line 23: These cannot go here, caused by abstraction leak and this is a public API for engine.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/validator/TestValidator.java
```diff
@@ -16,34 +16,90 @@
...
+    public int maxPadding(
         DirectBuffer data,
         int index,
         int length)
     {
-        return validate(data, index, length);
+        int padding = 0;
+        if (appendPrefix)
+        {
+            padding = MAX_PADDING_LEN; // TODO: fetch this from catalog
```
jfallows, line 54: Let's defer this to the catalog handler, if present.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/merged/merged.fetch.message.value.convert/server.rpt
```diff
@@ -0,0 +1,46 @@
...
+
+write zilla:data.ext ${kafka:dataEx()
+                              .typeId(zilla:id("kafka"))
+                              .merged()
+                                  .fetch()
+                                   .timestamp(newTimestamp)
+                                   .partition(0, 1, 2)
+                                   .build()
+                              .build()}
+write [0x00] [0x00 0x00 0x00 0x00 0x01] ${kafka:varint(3)} "id0" ${kafka:varint(8)} "positive"
```
jfallows, line 45: I thought it was 1 byte (magic) followed by int32 (schemaId) then payload, no?
This has 6-byte prefix, not 5 bytes.

jfallows (changes requested):

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
aDaemonThread: we already have ValueValidator no op, null check here is redundant. removed it.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/merged/merged.fetch.message.value.convert/server.rpt
aDaemonThread: that's true, but I have added this script to mimic the scenario when the prefix can be more than five byte. So that the padding can be dynamic.

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroWriteValidator.java
```diff
@@ -141,11 +119,12 @@ private byte[] serializeJsonRecord(
         encoded.reset();
         try
         {
-            Schema schema = schemas.get(schemaId);
-            record = records.computeIfAbsent(schemaId, this::supplyRecord);
+            Schema schema = supplySchema(schemaId);
+            GenericDatumReader<GenericRecord> reader = supplyReader(schemaId);
+            GenericDatumWriter<GenericRecord> writer = supplyWriter(schemaId);
+            GenericRecord record = supplyRecord(schemaId);
```
jfallows, line 125: Can the json decoder below be precreated and reused instead of creating afresh each time?

incubator/validator-core/src/main/java/io/aklivity/zilla/runtime/validator/core/StringEncoding.java
```diff
@@ -26,41 +26,24 @@ public boolean validate(
...
+                for (int j = 1; j < charByteCount; j++)
                 {
-                    if (bytesIndex + j >= length || (data.getByte(bytesIndex + j) & 0b11000000) != 0b10000000)
+                    if (index + j >= limit || (data.getByte(index + j) & 0b11000000) != 0b10000000)
                     {
                         break;
                     }
                 }
-                bytesIndex += numBytes;
+                index += charByteCount;
```
jfallows, line 44: We are calculating `index + j` more than once here and also `j` is not descriptive.
```
final int charByteLimit =. index + charByteCount;
for (int charByteIndex = index + 1; charByteIndex < charByteLimit; charByteIndex++)
{
    if (charByteIndex >= limit || (data.getByte(charByteIndex) & 0b11000000) != 0b10000000)
    {
        break validate;
    }
}
```
I think the `break` above needs to break from the outer `while` loop, not just the `for` loop, so we'll need to add a `validate:` label to the `while` loop, agree?

incubator/validator-json/src/main/java/io/aklivity/zilla/runtime/validator/json/JsonValidator.java
```diff
@@ -95,31 +89,40 @@ protected boolean validate(
...
+    private JsonSchema supplySchema(
+        int schemaId)
+    {
+        return schemas.computeIfAbsent(schemaId, this::resolveSchema);
+    }
+
+    private JsonProvider supplyProvider(
+        int schemaId)
+    {
+        return providers.computeIfAbsent(schemaId, id -> createProvider(supplySchema(id)));
```
jfallows, line 101: Suggest we simplify by making `createProvider` take `schemaId` and calling `supplySchema` from `createProvider` implementation.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/catalog/CatalogHandler.java
```diff
@@ -35,4 +34,16 @@ String resolve(
...
+    default int enrich(
+        int schemaId,
+        ValueConsumer next)
+    {
+        return NO_ENRICHMENT;
+    }
+
+    default int maxPadding()
+    {
+        return ZERO_PADDING;
```
jfallows, line 47: We can return `0` here instead of a constant.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogHandler.java
```diff
@@ -15,23 +15,61 @@
...
+
 import io.aklivity.zilla.runtime.engine.catalog.CatalogHandler;
 import io.aklivity.zilla.runtime.engine.test.internal.catalog.config.TestCatalogOptionsConfig;
+import io.aklivity.zilla.runtime.engine.validator.function.ValueConsumer;
 
 public class TestCatalogHandler implements CatalogHandler
 {
+    private static final int MAX_PADDING_LEN = 10;
+    private static final byte MAGIC_BYTE = 0x0;
+    private static final int ENRICHED_LENGTH = 5;
```
jfallows, line 31: Suggest renaming to `PREFIX_LENGTH`.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogHandler.java
```diff
@@ -15,23 +15,61 @@
...
+        int schemaId,
+        ValueConsumer next)
     {
-        this.schema = options.schema;
+        int length = 0;
+        if (embed)
+        {
+            prefixRO.putByte(0, MAGIC_BYTE);
+            prefixRO.putInt(1, schemaId, ByteOrder.BIG_ENDIAN);
+            next.accept(prefixRO, 0, 5);
```
jfallows, line 59:
```suggestion
            next.accept(prefixRO, 0, PREFIX_LENGTH);
```

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogHandler.java
```diff
@@ -15,23 +15,61 @@
...
+import org.agrona.MutableDirectBuffer;
+import org.agrona.concurrent.UnsafeBuffer;
+
 import io.aklivity.zilla.runtime.engine.catalog.CatalogHandler;
 import io.aklivity.zilla.runtime.engine.test.internal.catalog.config.TestCatalogOptionsConfig;
+import io.aklivity.zilla.runtime.engine.validator.function.ValueConsumer;
 
 public class TestCatalogHandler implements CatalogHandler
 {
+    private static final int MAX_PADDING_LEN = 10;
```
jfallows, line 29: Suggest renaming to `MAX_PADDING_LENGTH` for consistency of naming.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogHandler.java
```diff
@@ -15,23 +15,61 @@
...
+        this.prefixRO = new UnsafeBuffer(new byte[5]);
+        this.id = config.id;
+        this.embed = config.embed;
+        this.exclude = config.exclude;
+    }
+
+    @Override
+    public int enrich(
+        int schemaId,
+        ValueConsumer next)
```
jfallows, line 52: What is the intention of the return value of `enrich`?

It seems like it is trying to say how much of the input should be ignored, while also potentially writing something to `next`.

So in the case of `embed` where we are adding the prefix bytes, the `PREFIX_LENGTH` would already be passed to `next`, so perhaps this case should return `0`?

And when we are doing `exclude`, meaning we are stripping off the `PREFIX_LENGTH` bytes, we would not call `next` so we need to return `PREFIX_LENGTH` to indicate how many bytes we skipped over.

The implementation below is returning `PREFIX_LENGTH` for both cases, is it intentional?

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogHandler.java
```diff
@@ -15,23 +15,61 @@
...
+            prefixRO.putByte(0, MAGIC_BYTE);
+            prefixRO.putInt(1, schemaId, ByteOrder.BIG_ENDIAN);
+            next.accept(prefixRO, 0, 5);
+            length = ENRICHED_LENGTH;
+        }
+        else if (exclude)
+        {
+            length = ENRICHED_LENGTH;
+        }
+        return length;
```
jfallows, line 66: Does it make sense to have booleans for `embed` and `exclude` or is this more a funciton of whether the validator is read vs write?

If it is a property of the validator, then perhaps we need to pass this context from the validator to the catalog handler, either via a parameter or by having 2 methods on catalog handler?

incubator/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/SchemaRegistryCatalogHandler.java
```diff
@@ -49,12 +57,7 @@ public SchemaRegistryCatalogHandler(
         this.crc32c = new CRC32C();
         this.cache = new Int2ObjectCache<>(1, 1024, i -> {});
         this.schemaIdCache = new Int2ObjectCache<>(1, 1024, i -> {});
-    }
-
-    @Override
-    public String type()
-    {
-        return SCHEMA_REGISTRY;
+        this.prefixRO = new UnsafeBuffer(new byte[5]);
```
jfallows, line 60: We should probably have this structure generated from an internal `idl` instead for better readability, much like we do for the kafka cache entry descriptions and other protocol codec flyweights.

jfallows (changes requested):

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroWriteValidator.java
aDaemonThread: I think we can do that, however it requires to handle IOException so the AvroValidator constructor will have try/catch or throw exception handled. If that works, we can initialise this with a dummy schema & value.

aDaemonThread: this doesn't seems possible as we don't have a method to take reuse JsonDecoder.

aDaemonThread: but I was able to optimise BinaryEncoder & reuse the same instance.

incubator/validator-avro/src/main/java/io/aklivity/zilla/runtime/validator/avro/AvroValidator.java
```diff
@@ -42,21 +45,21 @@
...
     protected final SchemaConfig catalog;
     protected final CatalogHandler handler;
     protected final DecoderFactory decoderFactory;
     protected final EncoderFactory encoderFactory;
     protected final BinaryDecoder decoder;
+    protected final BinaryEncoder encoder;
     protected final String subject;
     protected final String format;
-    protected final ByteArrayOutputStream encoded;
+    protected final ExpandableDirectBufferOutputStream encoded;
```
jfallows, line 62: Suggest renaming this to `expandable` as it is used for both encode and decode.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/catalog/TestCatalogHandler.java
```diff
@@ -95,4 +97,63 @@ public String resolve(
...
+        else if (catalog.id != NO_SCHEMA_ID)
+        {
+            schemaId = catalog.id;
+        }
+        else
+        {
+            schemaId = resolve(subject, catalog.version);
+        }
+        return schemaId;
+    }
```
jfallows, line 158: Let's see if we can simplify this to not require embedding magic byte and schema id in payload.

incubator/catalog-inline/src/test/java/io/aklivity/zilla/runtime/catalog/inline/internal/InlineIT.java
```diff
@@ -55,4 +60,75 @@ public void shouldResolveSchemaViaSchemaId()
...
+        byte[] bytes = payload.getBytes();
+        data.wrap(bytes, 0, bytes.length);
+
+        int schemaId = catalog.resolve(data, 0, data.capacity(), schema, "test-value");
+
+        assertEquals(42, schemaId);
+    }
+
+    @Test
+    public void shouldVerifyEnrichedData()
```
jfallows, line 119: enriched -> encoded

jfallows (changes requested):

jfallows (approved):

PR #568, opened 2023-11-08
Fix flow conrol bug + indexoutofbound exception

bmaidics:


runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -1220,7 +1220,9 @@ else if (type.equals(EXPIRY_SIGNAL_NAME_OCTETS) && sessionExpiryIds.containsKey(
                         case MqttSessionSignalFW.KIND_EXPIRY:
                             final MqttExpirySignalFW expirySignal = sessionSignal.expiry();
                             long expireAt = expirySignal.expireAt();
-                            final String16FW expiryClientId = expirySignal.clientId();
+                            final String16FW clientId = expirySignal.clientId();
+                            final String16FW expiryClientId =
+                                new String16FW().wrap(clientId.buffer(), clientId.offset(), clientId.limit());
```
jfallows, line 1225: Creating a new flyweight does not make a copy of the underlying buffer.
```suggestion
                            final String16FW clientId = new String16FW(expirySignal.clientId().asString());
```

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientProduceFactory.java
```diff
@@ -1318,8 +1318,7 @@ private void onClientInitialFlush(
 
             assert initialAck <= initialSeq;
 
-            final int noAck = (int) (initialSeq - initialAck);
-            doClientInitialWindow(traceId, noAck, noAck + initialBudgetMax);
+            doClientInitialWindow(traceId, 0, initialBudgetMax);
```
jfallows, line 1321: Are you sure this is correct?
Would expect the reaction to `DATA` and `FLUSH` to be the same with regards to flow control, given they can both advance the `sequence` and indicate `reserved`.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientProduceFactory.java
bmaidics: If we check the previous code, noAck = initialSeq - initialAck.
So in the doClienttInitialWindow: 
`final long newInitialAck = Math.max(initialSeq - minInitialNoAck, initialAck);`
But in reality this is:
`final long newInitialAck = Math.max(initialAck, initialAck);`
initialAck is 0 in the beginning, so basically it'll never be increased, since newInitialAck > initialAck condition will never be true. I think when we're doing the data path we're not increasing the initialAck in the doClientInitialWindow call which we're doing from the onClientInitialData, but when the actual window frame is coming back from the network.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -209,6 +207,7 @@ public class MqttKafkaSessionFactory implements MqttKafkaStreamFactory
 
     private String serverRef;
     private int reconnectAttempt;
+    private int contextCounter;
```
jfallows, line 210: Suggest renaming to `nextContextId` instead.

jfallows (changes requested):

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/merged/unmerged.produce.flush.dynamic.hashed/client.rpt
```diff
@@ -141,6 +141,15 @@ write zilla:data.ext ${kafka:dataEx()
...
 
+write zilla:data.ext ${kafka:dataEx()
+                              .typeId(zilla:id("kafka"))
+                              .produce()
+                                  .sequence(-1)
+                                  .ackMode("LEADER_ONLY")
+                                  .build()
+                              .build()}
+write zilla:data.empty
+write flush
```
jfallows, line 152: It would be better if these script changes where not needed.
Perhaps it is only necessary to have this behavior if `reserved > 0` on `FLUSH`?
Then we can treat it as an isolated case, instead of modifying every script that sends a `FLUSH`.

jfallows (changes requested):

jfallows (approved):

PR #570, opened 2023-11-13
MQTT 3.1.1 support - specs

bmaidics:


specs/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/network/3.1.1/connect.successful.fragmented/client.rpt
```diff
@@ -0,0 +1,33 @@
+#
```
jfallows, line 1: I think we need to revise the terminology fragmented vs unfragmented for these scenarios.

Lets remove fragmented suffix from the normal case and treat "unfragmented" as the special case.

TLS False Start assumes successful TLS handshake and starts sending encrypted TLS records while confirmation of the successful (or unsuccessful) handshake is still in flight.
https://webopedia.com/definitions/tls-false-start/

Let's reuse that terminology here, giving `connect.subscribe.false.start` as the scenario name, since we send `SUBSCRIBE` assuming a successful `mqtt` handshake while `CONNACK` is still in flight.

jfallows (changes requested):

specs/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/network/3.1.1/connect.successful.fragmented/client.rpt
bmaidics: The normal case (connect.successful) mean that the connect packet came as a whole -> unfragmented.
So when you say: "Lets remove fragmented suffix from the normal case and treat "unfragmented" as the special case." the normal case doesn't have a fragmented suffix. 
For connect.subscribe, I renamed the scripts. Let me know if I understood the first part correctly.

runtime/binding-mqtt/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/client/ConnectionIT.java
```diff
@@ -34,7 +34,7 @@
 public class ConnectionIT
```
jfallows, line 34: These `ITs` will also need to move to ` .v5` subpackage, then `v4` subpackage will test mqtt 3.1.1 specification.

runtime/binding-mqtt/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/server/v5/ConnectionIT.java
```diff
@@ -442,9 +442,9 @@ public void shouldConnectWithServerDefinedKeepAlive() throws Exception
     @Test
     @Configuration("server.yaml")
     @Specification({
-        "${net}/connect.subscribe.unfragmented/client",
+        "${net}/connect.subscribe.false.start/client",
```
jfallows, line 445: Looks good! 👍

specs/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/network/v4/publish.one.message.subscribe.unfragmented/client.rpt
```diff
@@ -21,40 +21,28 @@ connect "zilla://streams/net0"
```
jfallows, line 21: Suggest `.batched` instead of `.unfragmented`.

Also do we need to say `.one.message`? Perhaps `client.publish.subscribe.batched` is sufficient?

Since `.false.start` is specific to `CONNECT` +packet, and this is the more general case, perhaps use `.batched` in place of `.false.start` too, such as `connect.subscribe.batched`?

jfallows (changes requested):

jfallows (approved):

PR #582, opened 2023-11-20
MQTT 3.1.1 implementation

bmaidics:


runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttClientFactory.java
```diff
@@ -274,6 +275,7 @@ public final class MqttClientFactory implements MqttStreamFactory
     private final MqttPublishHeader mqttPublishHeaderRO = new MqttPublishHeader();
 
     private final MqttConnectV5FW.Builder mqttConnectV5RW = new MqttConnectV5FW.Builder();
+    private final MqttConnectHeaderFW.Builder mqttConnectHeaderRW = new MqttConnectHeaderFW.Builder();
```
jfallows, line 278: 1. Use `MutableInteger` instead of `AtomicInteger` below and `.value = ...` instead of `.set(...)`.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttClientFactory.java
```diff
@@ -2282,14 +2284,21 @@ private void doEncodeConnect(
...
                     .protocolName(MQTT_PROTOCOL_NAME)
                     .protocolVersion(MQTT_PROTOCOL_VERSION)
                     .flags(flags)
                     .keepAlive((int) MILLISECONDS.toSeconds(keepAliveMillis))
+                    .build();
+
+            doNetworkData(traceId, authorization, 0L, connectHeader);
+
+            final MqttConnectV5FW connect =
+                mqttConnectV5RW.wrap(writeBuffer, FIELD_OFFSET_PAYLOAD, writeBuffer.capacity())
```
jfallows, line 2301: Let's send the combined connect header plus properties plus client-id in a single call to `doNetworkData(...)` passing the aggregated write buffer, from beginning of connect header to end of connect.

runtime/binding-mqtt/src/main/zilla/protocol.idl
```diff
@@ -94,26 +94,26 @@
...
             uint16 keepAlive;
+        }
+
+        struct MqttConnectV4
+        {
             string16 clientId;
         }
 
-        struct MqttConnectV5 extends MqttPacketHeader
+        struct MqttConnectV5
```
jfallows, line 110: Perhaps `MqttConnectV4` and `MqttConnectV5` can extend `MqttConnectHeader` so it is not necessary to think about the separate parts when sending, such as from mqtt client.

specs/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/network/v5/session.connect.payload.fragmented/server.rpt
```diff
@@ -22,15 +22,14 @@ accept "zilla://streams/net0"
...
+read  [0x10 0x38]                                      # CONNECT
       [0x00 0x04] "MQTT"                               # protocol name
       [0x05]                                           # protocol version
       [0x26]                                           # flags = will retain, will flag, clean start
       [0x00 0x0a]                                      # keep alive = 10s
       [0x00]                                           # properties
       [0x00 0x03] "one"                                # client id
-      [0x02]                                           # will properties
-      [0x01 0x01]                                      # format = utf-8
+      [0x000]                                           # will properties
```
jfallows, line 32: typo

jfallows (commented):
Please use separate decodeXXXvN methods instead of switch statement to decode all versions in same decodeXXX method.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttClientFactory.java
bmaidics: I'm not sure if you wanted to add this comment here.

runtime/binding-mqtt/src/main/zilla/protocol.idl
jfallows: Yes, that is understood.

However you can have both the separate `MqttConnect` struct and the `MqttConnectV5 extends MqttConnect` struct (also `MqttConnectV4 extends MqttConnect` struct) to eliminate the unnatural creating of the `MqttConnectV5` flyweight in 2 parts when sending from mqtt client.

runtime/binding-mqtt/src/main/zilla/protocol.idl
```diff
@@ -22,20 +22,34 @@
...
+            UNSUBSCRIBEV5,
+            UNSUBACK4,
+            UNSUBACKV5,
             PINGREQ,
+            PINGREQ,
+            PINGRESP,
             PINGRESP,
-            DISCONNECT,
+            DISCONNECTV4,
+            DISCONNECTV5,
```
jfallows, line 52: I wasn't expecting this change, we only need the super set of all packet types as they have the same values across protocols, right?

Then the decoders would be registered separately for v4 vs v5 instead of trying to put them all in the same map, agree?

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttClientFactory.java
```diff
@@ -2295,16 +2296,14 @@ private void doEncodeConnect(
                     .keepAlive((int) MILLISECONDS.toSeconds(keepAliveMillis))
                     .build();
 
-            doNetworkData(traceId, authorization, 0L, connectHeader);
-
             final MqttConnectV5FW connect =
-                mqttConnectV5RW.wrap(writeBuffer, FIELD_OFFSET_PAYLOAD, writeBuffer.capacity())
+                mqttConnectV5RW.wrap(writeBuffer, connectHeader.limit(), writeBuffer.capacity())
```
jfallows, line 2300: Please use extends in `protocol.idl` so that `MqttConnectV5` can be created here using a single flyweight.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -320,14 +321,19 @@ public final class MqttServerFactory implements MqttStreamFactory
 
     private final MqttServerDecoder decodeInitialType = this::decodeInitialType;
     private final MqttServerDecoder decodePacketType = this::decodePacketType;
```
jfallows, line 323: We need separate `decodePacketTypeV4` and `decodePacketTypeV5`.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -339,15 +345,20 @@ public final class MqttServerFactory implements MqttStreamFactory
...
+        decodersByPacketType.put(MqttPacketType.SUBSCRIBEV5, decodeSubscribeV5);
+        decodersByPacketType.put(MqttPacketType.UNSUBSCRIBEV4, decodeUnsubscribeV4);
+        decodersByPacketType.put(MqttPacketType.UNSUBSCRIBEV5, decodeUnsubscribeV5);
         decodersByPacketType.put(MqttPacketType.PINGREQ, decodePingreq);
-        decodersByPacketType.put(MqttPacketType.DISCONNECT, decodeDisconnect);
+        decodersByPacketType.put(MqttPacketType.DISCONNECTV4, decodeDisconnectV4);
+        decodersByPacketType.put(MqttPacketType.DISCONNECTV5, decodeDisconnectV5);
         // decodersByPacketType.put(MqttPacketType.AUTH, decodeAuth);
         this.decodersByPacketType = decodersByPacketType;
     }
```
jfallows, line 364: We need separate `decodersByPacketTypeV4` and `decodersByPacketTypeV5`, and no versions in the `MqttPacketType` constants.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -753,9 +764,9 @@ private int decodeInitialType(
         if (packet != null)
         {
             final int length = packet.remainingLength();
-            final MqttPacketType packetType = MqttPacketType.valueOf(packet.typeAndFlags() >> 4);
+            final MqttPacketType packetType = MqttPacketType.valueOf((packet.typeAndFlags() >> 4) * 2 + server.typeOffset);
 
-            if (packetType != MqttPacketType.CONNECT)
+            if (packetType != MqttPacketType.CONNECTV5 && packetType != MqttPacketType.CONNECTV4)
```
jfallows, line 769: This `decodeInitialType` can become the generic `decodeConnect` to determine the version, then reject or set decoder to `decodePacketTypeV4` or `decodePacketTypeV5`.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -753,9 +764,9 @@ private int decodeInitialType(
         if (packet != null)
         {
             final int length = packet.remainingLength();
-            final MqttPacketType packetType = MqttPacketType.valueOf(packet.typeAndFlags() >> 4);
+            final MqttPacketType packetType = MqttPacketType.valueOf((packet.typeAndFlags() >> 4) * 2 + server.typeOffset);
```
jfallows, line 767: This approach is too cryptic with the `* 2 + server.typeOffset`, let's revert as it's not needed if we keep the decoders separate by version, as they should be anyway.

jfallows (changes requested):

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -947,34 +956,34 @@ private int decodeConnectV5(
...
+                reasonCode = decodeConnectProtocol(mqttConnect);
                 if (reasonCode != SUCCESS)
                 {
                     break decode;
                 }
 
                 if (server.decodeablePacketBytes > maximumPacketSize)
                 {
-                    server.onDecodeError(traceId, authorization, PACKET_TOO_LARGE, mqttConnectHeader.protocolVersion());
+                    server.onDecodeError(traceId, authorization, PACKET_TOO_LARGE, mqttConnect.protocolVersion());
```
jfallows, line 986: Version is fixed at `v5` here.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2090,7 +2096,7 @@ private int onDecodeConnect(
             DirectBuffer buffer,
             int progress,
             int limit,
-            MqttConnectHeaderFW connectHeader,
+            MqttConnectFW connect,
```
jfallows, line 2099: Pass the parts of `connect` as individual parameters so this can be called easily from `decodeConnectV4` and `decodeConnectV5`.

jfallows (changes requested):

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -853,34 +853,34 @@ private int decodeConnectV4(
         {
             int reasonCode = SUCCESS;
 
-            final MqttConnectFW mqttConnect = mqttConnectRO.tryWrap(buffer, offset, limit);
+            final MqttConnectV4FW mqttConnectV4 = mqttConnectV4RO.tryWrap(buffer, progress, limit);
```
jfallows, line 856: Local variable is just `mqttConnect`, we are in the context of `decodeConnectV4` method so no need to respecify context.

jfallows (changes requested):
We need separate `decodePacketTypeV4` and `decodePacketTypeV5` decoders.

Then `decodeInitialType` should decode the generic connect fields to extract the version and switch to the either `decodePacketTypeV4` or `decodePacketTypeV5`.

There should be no need to keep checking the version in `decodePacketTypeXX` methods.

Suggest having an `Int2ObjectHashMap<MqttServerDecoder>` from version to `decodePacketTypeXX` methods.

Then at the end of a decodeXXX method that is identical for both versions, instead of setting the decoder to `decodePacketType` you can set it to `decodePacketTypeByVersion.get(server.version)` instead.

jfallows (commented):
Still need to apply `decodePacketTypeVx` feedback.

jfallows (approved):

Issue #581, opened 2023-11-17
`group.min.session.timeout.ms` is null using zilla in front of redpanda


vordimous:
**Describe the bug**
the `group.min.session.timeout.ms` config option is null instead of a string value

Stacktrace:
```
org.agrona.concurrent.AgentTerminationException: java.lang.NumberFormatException: Cannot parse null string
    at io.aklivity.zilla.runtime.engine/io.aklivity.zilla.runtime.engine.internal.registry.DispatchAgent.doWork(DispatchAgent.java:707)
    at org.agrona.core/org.agrona.concurrent.AgentRunner.doDutyCycle(AgentRunner.java:291)
    at org.agrona.core/org.agrona.concurrent.AgentRunner.run(AgentRunner.java:164)
    at java.base/java.lang.Thread.run(Thread.java:1623)
Caused by: java.lang.NumberFormatException: Cannot parse null string
    at java.base/java.lang.Integer.parseInt(Integer.java:627)
    at java.base/java.lang.Integer.parseInt(Integer.java:781)
    at io.aklivity.zilla.runtime.binding.kafka/io.aklivity.zilla.runtime.binding.kafka.internal.stream.KafkaClientGroupFactory$DescribeClient.onDecodeDescribeResponse(KafkaClientGroupFactory.java:3055)
    at io.aklivity.zilla.runtime.binding.kafka/io.aklivity.zilla.runtime.binding.kafka.internal.stream.KafkaClientGroupFactory.decodeDescribeResponse(KafkaClientGroupFactory.java:789)
    at io.aklivity.zilla.runtime.binding.kafka/io.aklivity.zilla.runtime.binding.kafka.internal.stream.KafkaClientGroupFactory$DescribeClient.decodeNetwork(KafkaClientGroupFactory.java:2926)
    at io.aklivity.zilla.runtime.binding.kafka/io.aklivity.zilla.runtime.binding.kafka.internal.stream.KafkaClientGroupFactory$DescribeClient.onNetworkData(KafkaClientGroupFactory.java:2562)
    at io.aklivity.zilla.runtime.binding.kafka/io.aklivity.zilla.runtime.binding.kafka.internal.stream.KafkaClientGroupFactory$DescribeClient.onNetwork(KafkaClientGroupFactory.java:2477)
[...]

PR #584, opened 2023-11-21
Consumer group session timeout defaults

jfallows:
## Description

Default consumer group session timeout properties if not returned in describe response from kafka broker


vordimous:
fixes https://github.com/aklivity/zilla/issues/581

vordimous (approved):

PR #586, opened 2023-11-22
Include validation in the `openapi.http.proxy` generator

### LINKED ISSUE #459: Generate `http` server request `validators` from `OpenAPI` specification
Requires #453 and #454 for `inline` catalog with `json` validator.

attilakreiner:
## Description

Include validation in the generate command in the `openapi.http.proxy` template

Fixes #459


incubator/command-generate/src/main/java/io/aklivity/zilla/runtime/command/generate/internal/asyncapi/view/SchemaView.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+        else if (ARRAY_TYPE.equals(schema.type) && schema.items != null && schema.items.ref != null)
+        {
+            schema.items = resolveRef(schema.items.ref);
+        }
+        this.schemas = schemas;
+        this.schema = schema;
+    }
+
+    public String getType()
```
attilakreiner, line 55: I am using `jsonb` for serialization. I didn't find a workaround to change the method names from `get...()`.  The `@JsonbProperty` annotation seems to work only on methods that follow the naming convention.

incubator/command-generate/src/main/moditect/module-info.java
```diff
@@ -29,13 +29,15 @@
 
     requires com.fasterxml.jackson.dataformat.yaml;
     requires com.fasterxml.jackson.databind;
+    requires org.apache.commons.collections4;
```
jfallows, line 32: We typically try to use either built-in JDK collections or Agrona collections instead of commons collections, what part of commons collections are we using in this implementation?

(trying to minimize dependency sprawl)

jfallows (changes requested):

incubator/command-generate/src/main/moditect/module-info.java
attilakreiner: I use the helper methods `MapUtils.isNotEmpty` and `CollectionUtils.isNotEmpty`, they are shortcuts for checking if the argument is not null and is not empty. It came handy in some places and apache commons was already added as a dependency so I assumed it's fine to use. But I can expand the null+emptiness checks, and remove the use of apache commons, if you prefer, not an issue. Pls LMK.

jfallows (commented):
Let's keep our code clear of using the commons-collections transitive dependency.

attilakreiner:
> Let's keep our code clear of using the commons-collections transitive dependency.
Sure, done.

jfallows (approved):

PR #629, opened 2023-12-08
WebSocket inbound ping frames support

### LINKED ISSUE #606: WebSocket inbound `ping` frames are rejected
Per [RFC 6455, section 5.5.2](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.2)
> Upon receipt of a Ping frame, an endpoint MUST send a Pong frame in
> response, unless it already received a Close frame.  It SHOULD
> respond with Pong frame as soon as is practical.

Binding type `ws` , kind `server` currently rejects inbound `ping` frames (opcode `0x09`).
Binding type `ws` , kind `client` currently rejects inbound `ping` frames (opcode `0x09`).

Expect `ws` binding to support `ping` frames according to spec for both `server` and `client`.

akrambek:
## Description

> Per [RFC 6455, section 5.5.2](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.2)
> 
> Upon receipt of a Ping frame, an endpoint MUST send a Pong frame in
> response, unless it already received a Close frame. It SHOULD
> respond with Pong frame as soon as is practical.

Binding type ws , kind server currently rejects inbound ping frames (opcode 0x09).
Binding type ws , kind client currently rejects inbound ping frames (opcode 0x09).

Expect ws binding to support ping frames according to spec for both server and client.

Fixes #606


jfallows (changes requested):
Need to remove `runtime/binding-ws/hs_err_pid8124.log`

jfallows (dismissed):

jfallows (approved):

PR #633, opened 2023-12-12
Support ability to connect to specific kafka cluster node hostname

akrambek:
## Description

Support ability to connect to specific kafka cluster node hostname

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientConnectionPool.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientConnectionPool.java
@@ -1280,6 +1280,7 @@ public final class KafkaClientConnectionPool extends KafkaClientSaslHandshaker
                             .destination(host)
                             .sourcePort(0)
                             .destinationPort(port)))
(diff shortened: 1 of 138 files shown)


jfallows (dismissed):

jfallows (approved):

jfallows (approved):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientConnectionPool.java
```diff
@@ -1280,6 +1280,7 @@ private void doConnectionBegin(
                             .destination(host)
                             .sourcePort(0)
                             .destinationPort(port)))
+                        .infos(i -> i.item(ii -> ii.authority(host)))
```
jfallows, line 1283: This should be coming from `ProxyBeginEx` `info` `authority`, not `host`, and `address` key for shared connections should include `authority` in addition to `host` and `port`, agree?

Otherwise, two streams differing only by `authority` would be treated as equal, even though their TLS handshakes would have a different `sni`.

jfallows (changes requested):

jfallows (approved):

PR #634, opened 2023-12-12
Improve server sent DISCONNECT reasonCodes

bmaidics:


runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -133,6 +134,35 @@ public class MqttKafkaSessionFactory implements MqttKafkaStreamFactory
...
+    public static final int INVALID_DESCRIBE_CONFIG_ERROR_CODE = 35;
+    public static final int INVALID_SESSION_TIMEOUT_ERROR_CODE = 26;
+    public static final byte MQTT_NOT_AUTHORIZED = (byte) 0x87;
+    public static final byte MQTT_IMPLEMENTATION_SPECIFIC_ERROR = (byte) 0x83;
+    public static final String MQTT_INVALID_SESSION_TIMEOUT_REASON = "Invalid session expiry interval";
+    private static final String16FW EMPTY_STRING = new String16FW("");
+
+    static
+    {
+        final Int2IntHashMap reasonCodes = new Int2IntHashMap(-1);
```
jfallows, line 149: Please us a constant for default error code returned when there is no entry in the map for the requested key.

i.e. `-1` should be a constant representing the default mqtt error code we want to send, then the `& 0xff` below likely becomes unnecessary.

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
bmaidics: & 0xff is still necessary, as we defined the error codes as hex, and they're overflowing and become negative values. E.g: `byte MQTT_NOT_AUTHORIZED = (byte) 0x87;`

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -3292,10 +3323,24 @@ private void onKafkaReset(
...
+            final KafkaResetExFW kafkaResetEx = extension.get(kafkaResetExRO::tryWrap);
+            final int error = kafkaResetEx != null ? kafkaResetEx.error() : -1;
+
+            Flyweight mqttResetEx = EMPTY_OCTETS;
+            if (error != -1)
+            {
+                mqttResetEx = mqttSessionResetExRW.wrap(sessionExtBuffer, 0, sessionExtBuffer.capacity())
+                    .typeId(mqttTypeId)
+                    .reasonCode(MQTT_REASON_CODES.get(kafkaResetEx.error()) & 0xff)
+                    .reason(MQTT_REASONS.getOrDefault(kafkaResetEx.error(), EMPTY_STRING))
```
jfallows, line 3340: Why do we default to `EMPTY_STRING` here?
Should be `null`, and then `reason` string should be omitted from the encoded mqtt packet when `null`, right?

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
jfallows: `0xff` is an `int`
`MQTT_NOT_AUTHORIZED && 0xff` is `byte && int` which gives `int`
`MqttResetEx.reasonCode` takes an `int` and does the `uint8` range checking in the generated method body.
```
    public Builder reasonCode(int value) {
      if (value < 0) {
        throw new IllegalArgumentException(String.format("Value %d too low for field \"reasonCode\"", value));
      }
      if (value > 0XFF) {
        throw new IllegalArgumentException(String.format("Value %d too high for field \"reasonCode\"", value));
      }
```

So we need an `int`, but we're storing these constants as `byte`, causing the wrapping-to-become-negative problem and forcing the `& 0xff`.

Please change the type of the constants from `byte` to `int` and remove the `& 0xff` as it is then no longer needed.

jfallows: > Please us a constant for default error code returned when there is no entry in the map for the requested key.

I think you might have missed this feedback. The `missingValue` passed to the constructor is the value returned when the key is not present in the map. So if we get an error code other than the explicit ones listed, then we'll get that `missingValue` back from the map.

Currently we are masking that to `0xff` so we'd return a default mqtt error code of `0xff`.

Please define a constant for the default mqtt error code we want to return and pass it to the `Int2IntHashMap` constructor instead of `-1`.

If this default mqtt error code matches some of the explicit mappings below, such as `MQTT_IMPLEMENTATION_SPECIFIC_ERROR`, then they do not also need to be listed explicitly.

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -3334,11 +3332,16 @@ private void onKafkaReset(
...
-                    .reasonCode(MQTT_REASON_CODES.get(kafkaResetEx.error()) & 0xff)
-                    .reason(MQTT_REASONS.getOrDefault(kafkaResetEx.error(), EMPTY_STRING))
-                    .build();
+                    .reasonCode(MQTT_REASON_CODES.get(kafkaResetEx.error()));
+                final String16FW reason = MQTT_REASONS.get(kafkaResetEx.error());
+                if (reason != null)
+                {
+                    mqttResetBuilder.reason(reason);
+                }
+                mqttResetEx = mqttResetBuilder.build();
```
jfallows, line 3344: We already have `error` above, so no need to call `kafkaResetEx.error()` again multiple times.
I think it's okay to pass `null` to `.reason(...)` on the builder, so probably no need to split the builder and make the code more complicated.

We likely need just a very small diff from the previous code.

jfallows (changes requested):

jfallows (approved):

PR #635, opened 2023-12-12
Add jumbograms and proxy extension parsing to dump command

attilakreiner:
## Description

This commit adds these features to the `dump` command:

- support of IPv6 Jumbograms to accommodate large frames
- adds the worker index and streams buffer offset
- adds the ability to dissect proxy type extensions
- fixes a bug with DATA frames
- adds the option to install the wireshark plugin
- adds version info to the wireshark plugin


incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
```diff
@@ -275,13 +297,18 @@ public void run()
...
                 }
                 idleStrategy.idle(workCount);
             } while (workCount != exitWorkCount);
         }
-        catch (IOException ex)
+        catch (Exception ex)
         {
+            if (exceptions)
+            {
+                ex.printStackTrace();
```
attilakreiner, line 310: I added the `-e` switch for consistency with the `start` command. The engine is not started so it can't print exceptions, `ZillaMain` doesn't print the either, so I am printing them here,

incubator/command-dump/src/main/java/io/aklivity/zilla/runtime/command/dump/internal/airline/ZillaDumpCommand.java
```diff
@@ -198,13 +223,37 @@ public final class ZillaDumpCommand extends ZillaCommand
...
     }
 
     @Override
     public void run()
     {
+        if (pluginDirectory != null)
+        {
+            try
+            {
+                InputStream is = getClass().getResourceAsStream("/wireshark/zilla.lua");
```
jfallows, line 237: Please move `zilla.lua` so it is in a subdirectory matching the package of this class, then this becomes `getClass().getResourceAsStream("zilla.lua")`.

i.e. `src/main/resources/io/aklivity/zilla/runtime/command/dump/internal/airline/zila.lua`

jfallows (changes requested):

jfallows (approved):

PR #640, opened 2023-12-13
Implement mqtt message expiry

bmaidics:


runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -1439,13 +1441,28 @@ private void onKafkaData(
                 final OctetsFW key = kafkaMergedDataEx != null ? kafkaMergedDataEx.fetch().key().value() : null;
                 final long filters = kafkaMergedDataEx != null ? kafkaMergedDataEx.fetch().filters() : 0;
                 final KafkaOffsetFW partition = kafkaMergedDataEx != null ? kafkaMergedDataEx.fetch().partition() : null;
+                final long timestamp = kafkaMergedDataEx != null ? kafkaMergedDataEx.fetch().timestamp() : 0;
```
jfallows, line 1444: What if expired message is fragmented? We need a test scenario for that.

Then `onKafkaData` gets called per fragment, but `timestamp` is only available on the `init` fragment.

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/subscribe.expire.message/client.rpt
```diff
@@ -0,0 +1,70 @@
...
+                                .capabilities("FETCH_ONLY")
+                                .topic("mqtt-messages")
+                                .partition(0, 0, 1, 1)
+                                .build()
+                            .build()}
+
+connected
+
+
+read notify SEND_DATA
```
jfallows, line 49: Barrier names should notify things that have already happened, independent of the anticipated reaction, such as sending some data.

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/subscribe.expire.message/server.rpt
```diff
@@ -0,0 +1,74 @@
...
+                                    .timestamp(timestamp)
+                                    .filters(1)
+                                    .partition(0, 2, 2)
+                                    .progress(0, 3)
+                                    .progress(1, 1)
+                                    .key("sensor/one")
+                                    .header("zilla:filter", "sensor")
+                                    .header("zilla:filter", "one")
+                                    .header("zilla:local", "client")
+                                    .headerInt("zilla:timeout-ms", 1000)
```
jfallows, line 66: We've been using `mqtt` naming concepts such as `zilla:filter` (topic filter), `zilla:local` (no local), `zilla:format` (format), so it seems like we should name this `zilla:expires` or similar (expiry interval) instead of timeout.

Also, per spec "If present, the Four Byte value is the lifetime of the Application Message in seconds" so seems like the unit should be seconds not milliseconds, to align with the mqtt expiry interval concept.

jfallows (changes requested):

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/subscribe.expire.message/server.rpt
bmaidics: When we initially discussed this we agreed that we should stick to Zilla terminology, not the mqtt ones. Also I'm not sure what it was but there was a reason that we changed the timeout to ms. Maybe because in Kafka we use milliseconds?

bmaidics: Let me know if we should change this.

jfallows: Yes, we should change `zilla:timeout-ms` to `zilla:expiry` please.

These header names are all derived from `mqtt` specific concepts except `timeout-ms` which sticks out as being inconsistent at the moment.

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/publish.one.message.changed.topic.name/client.rpt
```diff
@@ -39,7 +39,7 @@ write zilla:data.ext ${kafka:dataEx()
                                .header("zilla:filter", "sensor")
                                .header("zilla:filter", "one")
                                .header("zilla:local", "client")
-                               .headerInt("zilla:timeout-ms", 15000)
+                               .headerInt("zilla:expiry", 15000)
```
jfallows, line 42: The unit should also be adjusted to `seconds`, not `milliseconds`, per previous feedback please.

See https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901112

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/subscribe.expire.message.fragmented/client.rpt
```diff
@@ -0,0 +1,72 @@
...
+                                .capabilities("FETCH_ONLY")
+                                .topic("mqtt-messages")
+                                .partition(0, 0, 1, 1)
+                                .build()
+                            .build()}
+
+connected
+
+
+read notify WAIT_TIME_ELAPSED
```
jfallows, line 49: Think of this barrier name as an event that has already happened.
Suggest renaming to `MESSAGE_EXPIRED`.

runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeProxyIT.java
```diff
@@ -591,7 +613,21 @@ public void shouldExpireMessage() throws Exception
     {
         k3po.start();
         Thread.sleep(1500);
-        k3po.notifyBarrier("SEND_DATA");
+        k3po.notifyBarrier("WAIT_TIME_ELAPSED");
```
jfallows, line 616: If possible, let's avoid putting sleeps into the build as it doesn't take long for these to pile up and make the build a lot slower than necessary.

Perhaps we can take a different approach by initializing the message timestamp to be 1sec in the past, then the message is already expired and we don't need to wait?

runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeProxyIT.java
```diff
@@ -591,7 +613,21 @@ public void shouldExpireMessage() throws Exception
...
+    @Configuration("proxy.yaml")
+    @Configure(name = WILL_AVAILABLE_NAME, value = "false")
+    @Specification({
+        "${mqtt}/subscribe.expire.message/client",
+        "${kafka}/subscribe.expire.message.fragmented/server"})
+    public void shouldExpireMessageFragmented() throws Exception
+    {
+        k3po.start();
+        Thread.sleep(1500);
+        k3po.notifyBarrier("WAIT_TIME_ELAPSED");
```
jfallows, line 630: Same feedback as above about trying to eliminate the sleep from the build.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
```diff
@@ -1455,6 +1461,7 @@ private void onKafkaData(
                         expireInterval = timestamp + helper.timeout - now().toEpochMilli();
                         if (expireInterval < 0)
                         {
+                            expiredMessage = true;
```
jfallows, line 1464:
```
long expireInterval;
if (helper.timeout != -1)
{
   ...
}
else
{
    expireInterval = helper.timeout;
}
```
can this be simplified to:
```
long expireInterval = -1;
if (helper.timeout != -1)
{
   ...
}
```

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java
bmaidics: Originally I wanted to write this like you showed. But as we're using the expireInterval in a lambda we'd get:
"Variable used in lambda expression should be final or effectively final".

jfallows (approved):

PR #651, opened 2023-12-22
Support MQTT fragmented messages

### LINKED ISSUE #282: Handle data fragmentation for MQTT binding
This should also include handling large messages when encoding connect frame, and we have a large will message.

bmaidics:
Fixed #282


specs/binding-mqtt.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/streams/application/publish.10k/server.rpt
```diff
@@ -0,0 +1,72 @@
...
+                                .topic("sensor/one")
+                                .build()
+                             .build()}
+
+connected
+
+read zilla:data.ext ${mqtt:matchDataEx()
+                            .typeId(zilla:id("mqtt"))
+                            .publish()
+                                .build()
```
jfallows, line 63: Should we be tracking `deferred(2048)` here?

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/mqtt/publish.retained.10k/server.rpt
```diff
@@ -0,0 +1,45 @@
...
+                                .topic("sensor/one")
+                                .build()
+                             .build()}
+
+connected
+
+read zilla:data.ext ${mqtt:matchDataEx()
+                            .typeId(zilla:id("mqtt"))
+                            .publish()
+                                .build()
```
jfallows, line 36: Should we be tracking `deferred(2048)` here?

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/mqtt/publish.10k/server.rpt
```diff
@@ -0,0 +1,45 @@
...
+                                .topic("sensor/one")
+                                .build()
+                             .build()}
+
+connected
+
+read zilla:data.ext ${mqtt:matchDataEx()
+                            .typeId(zilla:id("mqtt"))
+                            .publish()
+                                .build()
```
jfallows, line 36: Should we be tracking `deferred(2048)` here?

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttClientFactory.java
```diff
@@ -2949,49 +2983,53 @@ private void onSessionData(
...
-                        subscription.qos = filter.qos();
-                        newSubscribeState.add(subscription);
-                    });
+                    case WILL:
+                        MqttWillMessageFW willMessage = mqttWillMessageRO.tryWrap(buffer, offset, limit);
+                        final OctetsFW willPayload = payloadRO.wrap(buffer, willMessage.limit(), limit);
+                        client.doEncodeConnect(traceId, authorization, client.clientId, client.flags,
+                            client.sessionExpiry, willMessage, willPayload);
+                        client.doSignalConnackTimeout(traceId);
+                        break;
```
jfallows, line 2996: Please refactor to separate `onSessionWillData` method.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttClientFactory.java
```diff
@@ -2949,49 +2983,53 @@ private void onSessionData(
...
+                        if (newSubscriptions.size() > 0)
+                        {
+                            client.doEncodeSubscribe(traceId, authorization, newSubscriptions, packetId);
+                        }
+                        if (oldSubscriptions.size() > 0)
+                        {
+                            client.doEncodeUnsubscribe(traceId, authorization, oldSubscriptions, packetId);
+                        }
+                        client.sessionStream.subscriptions.addAll(newSubscriptions);
+                        client.sessionStream.subscriptions.removeAll(oldSubscriptions);
```
jfallows, line 3030: Please refactor to separate `onSessionStateData` method.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -375,76 +380,82 @@ private void onMqttData(
...
+                    .merged(m -> m.produce(mp -> mp
+                        .deferred(deferred)
+                        .timestamp(now().toEpochMilli())
+                        .partition(p -> p.partitionId(-1).partitionOffset(-1))
+                        .key(b -> b.set(key))
+                        .hashKey(this::setHashKey)
+                        .headers(kafkaHeadersRW.build())))
+                    .build();
 
+                retainedData = (mqttPublishDataEx.flags() & 1 << MqttPublishFlags.RETAIN.value()) != 0;
```
jfallows, line 450: Please hoist `1 << MqttPublishFlags.RETAIN.value()` as `PUBLISH_FLAGS_RETAINED_MASK`.
Please remember `publishFlags` instead of `retainedData` during fragmented message publish.
Please add a static method `boolean hasPublishFlagRetained(int publishFlags)` which is used for `if (retainedData)` condition below.

This will help the readability and better convey the intent I think.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -614,11 +583,57 @@ private void onMqttData(
 
                     kafkaPayload = payload.sizeof() > 0 ? mqttSessionStateRO.wrap(buffer, offset, limit) : EMPTY_OCTETS;
 
-                    session.doKafkaData(traceId, authorization, budgetId,
-                        reserved, flags, kafkaPayload, kafkaDataEx);
+                    session.doKafkaData(traceId, authorization, budgetId, reserved,
+                        sessionPadding, flags, kafkaPayload, kafkaDataEx);
                     break;
                 }
```
jfallows, line 589: Same feedback here for this switch across session will and state - please refactor to separate methods.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1343,70 +1380,108 @@ private int decodePublishV5(
...
-                final String topic = mqttPublishHeader.topic;
-                final long topicKey = topicKey(topic, qos);
+                final String topic = mqttPublishHeaderRO.topic;
+                final long topicKey = topicKey(topic, mqttPublishHeaderRO.qos);
                 MqttServer.MqttPublishStream publisher = server.publishes.get(topicKey);
 
                 if (publisher == null)
                 {
-                    publisher = server.resolvePublishStream(traceId, authorization, topic, qos);
+                    publisher = server.resolvePublishStream(traceId, authorization, topic, mqttPublishHeaderRO.qos);
```
jfallows, line 1398: These should use `mqttPublishHeader` (local) instead of `mqttPublishHeader` (field), same instance, right?

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -3018,56 +3143,112 @@ else if (mqttPublishHeaderRO.retained && !retainAvailable(capabilities))
...
+            long authorization,
+            int reserved,
+            int packetId,
+            OctetsFW payload,
+            int offset,
+            int limit)
+        {
+            int reasonCode = SUCCESS;
+
+            if (mqttPublishHeaderRO.payloadFormat.equals(MqttPayloadFormat.TEXT) && invalidUtf8(payload))
```
jfallows, line 3174: What if `mqttPublishHeaderRO` is reused by a different client, can it cause a bug?
Prefer not to refer to factory-scoped state from client-specific methods other than on stack use.

jfallows (changes requested):
Btw, although we have removed maximum packet size as a limitation, it may still be relevant to reject messages that are too large, for example when Kafka rejects the message as exceeding maximum message size.

So we probably need to detect this case, and make sure the MQTT client understands that the publish is failing because the message is too large, by receiving a suitable error code at the MQTT client.

What happens when the message in Kafka is larger than the maximum packet size would permit to send the MQTT client?

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -1383,19 +1383,19 @@ private int decodePublishV5(
                 publishLimit = publish.limit();
             }
 
-            final MqttPublishHeader mqttPublishHeader = mqttPublishHeaderRO.reset();
+            final MqttPublishHeader mqttPublishHeader = new MqttPublishHeader();
```
jfallows, line 1386: This does not seem to be following the same coding pattern we use for other server decoders.

I think this `MqttPublishHeader` is being kept to store information about the currently decoded `PUBLISH` frame so that it can be used later on in the decode.

However, the approach we usually take is to have `decodedXXX` or `decodeableXXX` fields describing fields either already observed (e.g. topic, qos, etc) or anticipated (e.g. remaining bytes). This can work because we decode one packet at a time, so even a large `PUBLISH` packet is decoded fully before moving on to the next packet.

Then these fields are used directly or indirectly as needed to pass parameters to `onDecodeXXX` methods.

Please let me know if there's any reason why we might need to take a different approach here for MQTT decode.

runtime/binding-mqtt/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/server/v5/ConnectionIT.java
```diff
@@ -414,6 +414,16 @@ public void shouldRejectWillRetainWithoutWillFlag() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Configuration("server.yaml")
+    @Specification({
+        "${net}/connect.max.packet.size.exceeded/client",
+        "${app}/connect.max.packet.size.exceeded/server"})
+    public void shouldIgnorePublishPacketBiggerThanMaxPacketSize() throws Exception
```
jfallows, line 422: Please name method for script scenario name, i.e. `shouldRejectMaxPacketSizeExceeded`.

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/kafka/publish.reject.large.message/server.rpt
```diff
@@ -0,0 +1,40 @@
...
+                                .build()
+                            .build()}
+
+
+connected
+
+write zilla:reset.ext ${kafka:resetEx()
+                           .typeId(zilla:id("kafka"))
+                           .error(18)
+                           .build()}
```
jfallows, line 38: We currently send
RECORD_LIST_TOO_LARGE | 18
-- | --

but might also possibly get

MESSAGE_TOO_LARGE | 10
-- | --

so please create 2 streams here and make sure both cause the same `mqtt` error about published message being too large.

jfallows (changes requested):

runtime/binding-mqtt-kafka/hs_err_pid34510.log
jfallows, line 1: Please remove from PR. 😄

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #658, opened 2023-12-26
Catalog cache TTL implementation

aDaemonThread:
Refer: https://github.com/aDaemonThread/zilla/pull/4

### DIFF
--- a/incubator/catalog-schema-registry.spec/src/main/scripts/io/aklivity/zilla/specs/catalog/schema/registry/schema/schema.registry.schema.patch.json
+++ b/incubator/catalog-schema-registry.spec/src/main/scripts/io/aklivity/zilla/specs/catalog/schema/registry/schema/schema.registry.schema.patch.json
@@ -40,9 +40,10 @@
                                 "type": "string",
                                 "default": "default"
                             },
-                            "cacheTTL":
+                            "max-age":
                             {
-                                "type": "integer",
+                                "title": "Max Age",
+                                "type": "number",
                                 "default": 300
                             }
                         },
--- a/incubator/catalog-schema-registry.spec/src/main/scripts/io/aklivity/zilla/specs/catalog/schema/registry/schema/schema.registry.schema.patch.json
+++ b/incubator/catalog-schema-registry.spec/src/main/scripts/io/aklivity/zilla/specs/catalog/schema/registry/schema/schema.registry.schema.patch.json
@@ -39,6 +39,11 @@
                             {
                                 "type": "string",
                                 "default": "default"
+                            },
+                            "cacheTTL":
+                            {
+                                "type": "integer",
+                                "default": 300
                             }
                         },
                         "additionalProperties": false
--- a/incubator/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/SchemaRegistryCatalogHandler.java
+++ b/incubator/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/SchemaRegistryCatalogHandler.java
@@ -50,7 +50,7 @@ public class SchemaRegistryCatalogHandler implements CatalogHandler
     private final CRC32C crc32c;
     private final Int2ObjectCache<CachedSchema> schemas;
     private final Int2ObjectCache<CachedSchemaId> schemaIds;
-    private final long cacheTtl;
+    private final long maxAgeMillis;
 
     public SchemaRegistryCatalogHandler(
         SchemaRegistryOptionsConfig config)
@@ -61,7 +61,7 @@ public class SchemaRegistryCatalogHandler implements CatalogHandler
         this.crc32c = new CRC32C();
         this.schemas = new Int2ObjectCache<>(1, 1024, i -> {});
         this.schemaIds = new Int2ObjectCache<>(1, 1024, i -> {});
-        this.cacheTtl = config.cacheTtl * 1000;
+        this.maxAgeMillis = config.maxAge.toMillis();
     }
 
     @Override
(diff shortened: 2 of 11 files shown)


incubator/catalog-schema-registry.spec/src/main/scripts/io/aklivity/zilla/specs/catalog/schema/registry/schema/schema.registry.schema.patch.json
```diff
@@ -39,6 +39,11 @@
                             {
                                 "type": "string",
                                 "default": "default"
+                            },
+                            "cacheTTL":
```
jfallows, line 43: Let's rename this to `max-age` and parse it as a `Duration` same as `max-age` in `http` binding options and `challenge` in `jwt` guard options.

See https://github.com/aklivity/zilla/blob/develop/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpOptionsConfigAdapter.java#L359 as an example.

incubator/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/SchemaRegistryCatalogHandler.java
```diff
@@ -58,8 +59,9 @@ public SchemaRegistryCatalogHandler(
         this.client = HttpClient.newHttpClient();
         this.request = new RegisterSchemaRequest();
         this.crc32c = new CRC32C();
-        this.cache = new Int2ObjectCache<>(1, 1024, i -> {});
-        this.schemaIdCache = new Int2ObjectCache<>(1, 1024, i -> {});
+        this.schemas = new Int2ObjectCache<>(1, 1024, i -> {});
+        this.schemaIds = new Int2ObjectCache<>(1, 1024, i -> {});
+        this.cacheTtl = config.cacheTtl * 1000;
```
jfallows, line 64: Let's rename this field to `maxAgeMillis` and then it will be `config.maxAge.toMillis()` as we are parsing `max-age` as a `Duration`.

jfallows (changes requested):

incubator/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/config/SchemaRegistryOptionsConfigAdapter.java
```diff
@@ -78,6 +92,15 @@ public OptionsConfig adaptFromJson(
...
             }
+
+            if (object.containsKey(MAX_AGE_NAME))
+            {
+                options.maxAge(Duration.ofSeconds(object.getJsonNumber(MAX_AGE_NAME).longValue()));
+            }
+            else
+            {
+                options.maxAge(Duration.ofSeconds(300));
+            }
```
jfallows, line 103: Handle defaults via Constantine type Duration instead of hardcoding.
Also defaulting should be handled in the builder and the builder is called conditionally from the config adapter so value defaults if not present.

incubator/catalog-schema-registry/src/test/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/SchemaRegistryCatalogFactoryTest.java
```diff
@@ -45,8 +47,11 @@ public void shouldLoadAndCreate()
         CatalogContext context = catalog.supply(mock(EngineContext.class));
         assertThat(context, instanceOf(SchemaRegistryCatalogContext.class));
 
-        SchemaRegistryOptionsConfig catalogConfig =
-            new SchemaRegistryOptionsConfig("http://localhost:8081", "default");
+        SchemaRegistryOptionsConfig catalogConfig = SchemaRegistryOptionsConfig.builder()
+            .url("http://localhost:8081")
+            .context("default")
+            .maxAge(Duration.ofSeconds(300))
```
jfallows, line 53: Use value not matching default to test.

jfallows (changes requested):

incubator/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/config/SchemaRegistryOptionsConfigAdapter.java
```diff
@@ -78,6 +87,15 @@ public OptionsConfig adaptFromJson(
...
             }
+
+            if (object.containsKey(MAX_AGE_NAME))
+            {
+                options.maxAge(Duration.ofSeconds(object.getJsonNumber(MAX_AGE_NAME).longValue()));
+            }
+            else
+            {
+                options.maxAge();
+            }
```
jfallows, line 98: This else clause should be removed, then defaulting happens in the builder build method as needed.

jfallows (changes requested):

incubator/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/config/SchemaRegistryOptionsConfigBuilder.java
```diff
@@ -53,9 +57,20 @@ public SchemaRegistryOptionsConfigBuilder<T> context(
...
+    }
+
     @Override
     public T build()
     {
-        return mapper.apply(new SchemaRegistryOptionsConfig(url, context));
+        if (maxAge == null)
+        {
+            this.maxAge = Duration.ofSeconds(MAX_AGE_DEFAULT);
+        }
```
jfallows, line 73: Defaulting should not mutate the builder state, this should be a local variable instead, initialized to the field, tested for null and defaulted if needed, then passing the local variable to the constructor instead of the field.

jfallows (changes requested):

jfallows (approved):

PR #694, opened 2024-01-03
Optimize memory allocation for mqtt-kafka offset tracking

bmaidics:


specs/binding-mqtt.spec/src/test/java/io/aklivity/zilla/specs/binding/mqtt/internal/MqttFunctionsTest.java
```diff
@@ -1272,16 +1273,18 @@ public void shouldEncodeMqttOffsetMetadata()
...
-        DirectBuffer buffer = new UnsafeBuffer(BitUtil.fromHex(state));
-        MqttOffsetMetadataFW offsetMetadata = new MqttOffsetMetadataFW().wrap(buffer, 0, buffer.capacity());
+        final IntArrayList metadataList = new IntArrayList();
+        int offset = 0;
+        final DirectBuffer buffer = new String16FW(state).value();
+        byte version = buffer.getByte(offset++);
+        for (; offset < buffer.capacity(); offset += BitUtil.SIZE_OF_SHORT)
+        {
+            metadataList.add((int) buffer.getShort(offset));
+        }
```
jfallows, line 1283: Needs a version prefix in case we choose to evolve the structure over time.
Needs BIG_ENDIAN as we are putting into an external system so could be read by any endian environment, not just local native endianness.

Also, let's use `int16[length]` support in `.idl` to generate a flyweight instead.
See examples in flyweight test resources at https://github.com/aklivity/zilla/blob/develop/build/flyweight-maven-plugin/src/test/resources/test-project/test.idl#L123.

jfallows (changes requested):

jfallows (approved):

PR #733, opened 2024-01-15
Qos2 idempotent producer

bmaidics:


runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -227,24 +258,59 @@ private final class MqttPublishProxy
...
+                    this.retained = new KafkaRetainedProxy(originId, resolvedId, routedId, affinity, this, kafkaRetainedTopic);
+                }
+            }
+            else
+            {
+                this.messages =
+                    new KafkaMessagesProxy(originId, resolvedId, routedId, affinity, this, kafkaMessagesTopic);
+                this.retained =
+                    new KafkaRetainedProxy(originId, resolvedId, routedId, affinity, this, kafkaRetainedTopic);
+            }
```
jfallows, line 313: I'd like to better understand why this complexity is needed.
Seems like we just want to defer the offset commit to session, everything else stays the same, no?

jfallows (changes requested):
The cyclomatic complexity of the code has gone up quite a bit with these proposed changes.

```
MqttKafkaSessionFactory
  MqttSessionProxy
  -> KafkaOffsetFetchStream
  -> KafkaInitProducerStream
  -> KafkaOffsetCommitStream
     -> MqttKafkaPublishFactory.MqttPublishProxy
     -> MqttKafkaPublishFactory.KafkaProxy
  -> MqttKafkaPublishFactory.KafkaMessagesProxy
  -> MqttKafkaPublishFactory.KafkaRetainedProxy

MqttKafkaPublishFactory
  MqttPublishProxy
  -> MqttKafkaSessionFactory.MqttSessionProxy
```

In general, we don't want to reach across stream factories for behavior, even doing so for state should be kept to a minimum.

In this case, we had discussed a minor change to have session handle offset commit instead of publish, but this change seems to have more inter-stream behavioral dependencies than that.

Perhaps we should have a discussion about it to determine how best to simplify if possible.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -444,6 +436,12 @@ public final class MqttSessionProxy
         private int sessionPadding;
         private String willId;
         private int delay;
+        private int unfetchedKafkaTopics;
+        private Long2ObjectHashMap<PublishOffsetMetadata> offsets;
+        private Int2ObjectHashMap<KafkaTopicPartition> partitions;
+        private Int2ObjectHashMap<KafkaTopicPartition> retainedPartitions;
+        private long affinity;
```
jfallows, line 443: `affinity` should be `final`, passed to constructor from `newStream`

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -4754,6 +4927,28 @@ private String16FW offsetMetadataToString(
             offsetMetadata.offset(), offsetMetadata.limit()));
     }
 
+    public static final class PublishClientMetadata
```
jfallows, line 4930: If the state is being shared across different two stream factories, then the class representing the state should be defined outside of both stream factories to make the non-encapsulation clear.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -1761,6 +2000,30 @@ private void doKafkaWindow(
...
+        mqttOffsetMetadataRW.producerId(metadata.producerId);
+        mqttOffsetMetadataRW.producerEpoch(metadata.producerEpoch);
+
+        if (metadata.packetIds != null)
+        {
+            metadata.packetIds.forEach(p -> mqttOffsetMetadataRW.appendPacketIds(p.shortValue()));
+        }
+        final MqttPublishOffsetMetadataFW offsetMetadata = mqttOffsetMetadataRW.build();
+        return new String16FW(BitUtil.toHex(offsetMetadata.buffer().byteArray(),
+            offsetMetadata.offset(), offsetMetadata.limit()));
```
jfallows, line 2017: We'll want to optimize this at some point to avoid allocation during qos2 ack.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java
```diff
@@ -1329,10 +1312,24 @@ private void onKafkaWindow(
 
             assert initialAck <= initialSeq;
 
-            if (delegate != null)
+            if (wasOpen)
+            {
+                delegate.doMqttWindow(traceId, authorization, budgetId, padding, capabilities);
+            }
+            else if (delegate.qos < MqttQoS.EXACTLY_ONCE.value())
```
jfallows, line 1319: These `if` conditions can be combined via `||` since the behavior for each is identical, agree?

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
bmaidics: We don't even need to store the affinity in the session stream, as it's only adding the metadata, but never queries. At the beginning. At mqttBegin we already have it.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -3515,106 +3802,1263 @@ private void doKafkaData(
...
+                                delegate.offsets.values().forEach(o ->
+                                    o.packetIds.forEach(p -> sessionBuilder.appendPacketIds(p.shortValue())));
+                            }).build();
+
+                        delegate.doMqttBegin(traceId, authorization, 0, mqttBeginEx);
+                        delegate.session = new KafkaSessionStateProxy(originId, routedId, delegate);
+                        delegate.session.doKafkaBeginIfNecessary(traceId, authorization, 0);
+                    }
+                }
+            }
```
jfallows, line 4828: This section is quite complex. Let's try moving some of the implementation reaction to `delegate.onXX` methods to let it know what happened and then let it react.

Try also to determine what are the invariants here in both sides of the if condition, to help improve the abstraction instead of treating each side as an independent linear codepath.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -3515,106 +3802,1263 @@ private void doKafkaData(
...
+
+
+        private KafkaPartition(
+            String topic,
+            int partitionId)
+        {
+            this.partitionId = partitionId;
+            this.topic = topic;
+        }
+    }
```
jfallows, line 4961: Is this a duplicate of the `KafkaPartition` in the metadata class?

Let's see if we can move what's necessary to top level classes, or perhaps this one is not needed after we enhance the abstraction for state tracking a little bit.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -3515,106 +3802,1263 @@ private void doKafkaData(
...
+        mqttOffsetMetadataRW.producerEpoch(metadata.producerEpoch);
+
+        if (metadata.packetIds != null)
+        {
+            metadata.packetIds.forEach(p -> mqttOffsetMetadataRW.appendPacketIds(p.shortValue()));
+        }
+        final MqttPublishOffsetMetadataFW offsetMetadata = mqttOffsetMetadataRW.build();
+        return new String16FW(BitUtil.toHex(offsetMetadata.buffer().byteArray(),
+            offsetMetadata.offset(), offsetMetadata.limit()));
+    }
```
jfallows, line 4945: There is a lot of allocation going on here. Let's see if we can reduce it since it seems to be happening on the data path.

Perhaps we can move the encode/decode logic closer to the metadata abstraction to reuse the data structures to start.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -5216,12 +5204,50 @@ private void onSessionBegin(
 
                     sessionExpiry = mqttSessionBeginEx.expiry();
                     capabilities = mqttSessionBeginEx.capabilities();
-                    maximumQos = mqttSessionBeginEx.qosMax();
+                    subscribeMaximumQos = mqttSessionBeginEx.subscribeQosMax();
```
jfallows, line 5207:
```suggestion
                    subscribeQosMax = mqttSessionBeginEx.subscribeQosMax();
```

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -5216,12 +5204,50 @@ private void onSessionBegin(
...
+                    if (mqttSessionBeginEx.packetIds() != null)
+                    {
+                        mqttSessionBeginEx.packetIds().forEachRemaining((IntConsumer) p -> unreleasedPacketIds.add(p));
+                    }
```
jfallows, line 5211: Please avoid calling `mqttSessionBeginEx.packetIds()` twice.

specs/binding-mqtt.spec/src/main/resources/META-INF/zilla/mqtt.idl
```diff
@@ -254,8 +264,17 @@ scope mqtt
...
-            int16[length] packetIds;
+            int8 length;
+            int16[length] packetIds = null;
+        }
+
+        struct MqttPublishOffsetMetadata
+        {
+            uint8 version = 1;
+            int64 producerId = 0;
+            int16 producerEpoch = 0;
```
jfallows, line 275: We can't have (or need) `kafka` concepts in `mqtt` metadata.

Should these be in `internal.idl` of `mqtt-kafka` binding instead?

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -5216,12 +5204,50 @@ private void onSessionBegin(
...
+                    unreleasedPacketIds.removeInt(packetId);
+                    switch (version)
+                    {
+                    case 4:
+                        doEncodePubcompV4(traceId, authorization, packetId);
+                        break;
+                    case 5:
+                        doEncodePubcompV5(traceId, authorization, packetId);
+                        break;
+                    }
```
jfallows, line 5247: We will need to address all these switch statements for version 4 vs version 5 in the code each time an encoding decision is needed, even though the version is already decided just once up front.

We can use an inner class MqttVersion enum with methods like `doEncodePubcomp(...)` where the `v4` constant delegates to `doEncodePubcompV4` and the `v5` constant delegates to `doEncodePubcompV5`, while here we would just call `version.doEncodePubcomp(...)`. Similar approach for other `switch (version)` statements.

This feedback can be handled in a different PR though.

specs/binding-mqtt.spec/src/test/java/io/aklivity/zilla/specs/binding/mqtt/streams/application/PublishIT.java
```diff
@@ -226,6 +226,15 @@ public void shouldPublishQoS2MessageAckWithReasoncode() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${app}/publish.qos2.recover/client",
+        "${app}/publish.qos2.recover/server"})
+    public void shouldReleaseQos2PacketIdAtRecover() throws Exception
```
jfallows, line 233:
```suggestion
    public void shouldReleaseQos2PacketIdDuringRecovery() throws Exception
```

specs/binding-mqtt.spec/src/test/java/io/aklivity/zilla/specs/binding/mqtt/streams/application/PublishIT.java
```diff
@@ -226,6 +226,15 @@ public void shouldPublishQoS2MessageAckWithReasoncode() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${app}/publish.qos2.recover/client",
+        "${app}/publish.qos2.recover/server"})
```
jfallows, line 232:
```suggestion
        "${app}/publish.qos2.recovery/client",
        "${app}/publish.qos2.recovery/server"})
```

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -3464,7 +3752,7 @@ private void onKafkaReset(
                     mqttSessionResetExRW.wrap(sessionExtBuffer, 0, sessionExtBuffer.capacity())
                     .typeId(mqttTypeId)
                     .reasonCode(MQTT_REASON_CODES.get(error))
-                    .reason(MQTT_REASONS.get(error))
+                    .reason(MQTT_REASONS.getOrDefault(error, EMPTY_STRING))
```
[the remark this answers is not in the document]
bmaidics: Yes, it does return null which leads to NPE in String16FW.set as we're calling length() on it.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
bmaidics: Create a Helper class, add non-static fields for the FW. Add encode/decode

specs/binding-mqtt-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/KafkaIT.java
```diff
@@ -857,8 +857,8 @@ public void shouldPublishQoS2MessageRetained() throws Exception
 
     @Test
     @Specification({
-        "${kafka}/publish.qos2.recover/client",
-        "${kafka}/publish.qos2.recover/server"})
+        "${kafka}/publish.qos2.recovery/client",
+        "${kafka}/publish.qos2.recovery/server"})
     public void shouldPublishQoS2MessageRecover() throws Exception
```
jfallows, line 862:
```suggestion
    public void shouldPublishQoS2MessageDuringRecovert() throws Exception
```

specs/binding-mqtt-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/mqtt/kafka/streams/MqttIT.java
```diff
@@ -681,8 +681,8 @@ public void shouldPublishQoS2MessageRetained() throws Exception
 
     @Test
     @Specification({
-        "${mqtt}/publish.qos2.recover/client",
-        "${mqtt}/publish.qos2.recover/server"})
+        "${mqtt}/publish.qos2.recovery/client",
+        "${mqtt}/publish.qos2.recovery/server"})
     public void shouldPublishQoS2MessageRecover() throws Exception
```
jfallows, line 686:
```suggestion
    public void shouldPublishQoS2MessageDuringRecovery() throws Exception
```

runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishProxyIT.java
```diff
@@ -305,8 +305,8 @@ public void shouldSendMessageQos2Retained() throws Exception
     @Configuration("proxy.yaml")
     @Configure(name = WILL_AVAILABLE_NAME, value = "false")
     @Specification({
-        "${mqtt}/publish.qos2.recover/client",
-        "${kafka}/publish.qos2.recover/server"})
+        "${mqtt}/publish.qos2.recovery/client",
+        "${kafka}/publish.qos2.recovery/server"})
     public void shouldSendMessageQos2Recover() throws Exception
```
jfallows, line 310:
```suggestion
    public void shouldSendMessageQos2DuringRecovery() throws Exception
```

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -1722,6 +1906,70 @@ private void onKafkaReset(
             }
         }
 
+        private void onSignal(SignalFW signal)
```
jfallows, line 1909: Formatting

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2933,46 +3181,20 @@ private void onKafkaData(
...
-            initialAck = acknowledge;
-            initialMax = maximum;
-            state = MqttKafkaState.openInitial(state);
-
-            assert initialAck <= initialSeq;
+            }
+            else
+            {
+                handleKafkaData(data);
+            }
```
jfallows, line 3188: Check all `onKafkaFlush` to set `replySeq = sequence + reserved;` not just `replySeq = sequence;` which will handle the case when `reserved` is non-zero as well.

Suggest renaming to `onKafkaDataImpl(DataFW)` instead of `handleKafkaData`.

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -4874,6 +4881,37 @@ private void onKafkaReset(
...
+            if (!wasOpen)
+            {
+                delegate.onOffsetCommitOpened(traceId, authorization, budgetId);
+            }
+            else
+            {
+                delegate.onOffsetCommitAck(traceId, authorization);
+            }
+        }
+
```
jfallows, line 4914: Please review the method ordering of all the inner classes in the factories.

Suggest grouping the `onMqttXxx` and `doMqttXxx` together, also group `onKafkaXxx` and `doKafkaXxx` together, with additional supporting methods at the end of each inner class.

jfallows (changes requested):
Looks better!

jfallows (dismissed):

PR #735, opened 2024-01-15
Fragment validator interface & implementation

aDaemonThread:


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -170,7 +173,7 @@ public final class Engine implements Collector, AutoCloseable
             DispatchAgent agent =
                 new DispatchAgent(config, tasks, labels, errorHandler, tuning::affinity,
                         bindings, exporters, guards, vaults, catalogs, metricGroups, converterFactory,
-                    this, coreIndex, readonly);
+                    validatorFactory, this, coreIndex, readonly);
```
jfallows, line 176: If possible, converters and validators should follow the same pattern as the others, passing a collection instead of a factory.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineContext.java
```diff
@@ -139,6 +141,9 @@ Converter createReader(
     Converter createWriter(
         ConverterConfig converter);
 
+    Validator supplyValidator(
+        ValidatorConfig converter);
+
```
jfallows, line 146: Please move this up next to other `supplyXXX` methods for guard, vault and catalog.
Also, shouldn't this be returning a `ValidatorHandler` (per core) instead of a `Validator` (global) given that it is on `EngineContext` which is per core?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/validator/ValidatorFactorySpi.java
```diff
@@ -0,0 +1,30 @@
...
+import io.aklivity.zilla.runtime.engine.catalog.CatalogHandler;
+import io.aklivity.zilla.runtime.engine.config.ValidatorConfig;
+
+public interface ValidatorFactorySpi
+{
+    String type();
+
+    Validator create(
+        ValidatorConfig config,
+        LongFunction<CatalogHandler> supplyCatalog);
```
jfallows, line 29: This seems inconsistent, as `Validator` is global, whereas `CatalogHandler` is per core.

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/config/StringValidatorConfigAdapter.java
```diff
@@ -0,0 +1,81 @@
...
+            assert false;
+        }
+        return result;
+    }
+
+    @Override
+    public String type()
+    {
+        return "string";
+    }
```
jfallows, line 80: Please move this to first method position for clarity.

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/config/StringValidatorConfigAdapter.java
```diff
@@ -0,0 +1,81 @@
...
+        {
+            result = StringValidatorConfig.builder().build();
+        }
+        else if (value instanceof JsonObject)
+        {
+            JsonObject object = (JsonObject) value;
+            String encoding = object.containsKey(ENCODING_NAME)
+                ? object.getString(ENCODING_NAME)
+                : null;
+            result = new StringValidatorConfig(encoding);
```
jfallows, line 67: Use builder for both cases.
Also use enum returned from `value.getValueType()` in a switch instead of if-else with instanceof.

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/config/StringValidatorConfigAdapter.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+        else
+        {
+            assert false;
+        }
```
jfallows, line 72: No need for this assert, covered by schema.

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidator.java
```diff
@@ -0,0 +1,54 @@
...
+    public String name()
+    {
+        return StringValidator.NAME;
+    }
+
+    @Override
+    public ValidatorContext supply(
+        EngineContext context)
+    {
+        return new StringValidatorContext();
```
jfallows, line 46: Usually we pass the engine context in the constructor.

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidator.java
```diff
@@ -0,0 +1,54 @@
...
+import io.aklivity.zilla.runtime.engine.validator.ValidatorContext;
+import io.aklivity.zilla.runtime.types.core.config.StringValidatorConfig;
+
+public class StringValidator implements Validator
+{
+    public static final String NAME = "string";
+
+    public StringValidator(
+        StringValidatorConfig config,
+        LongFunction<CatalogHandler> supplyCatalog)
```
jfallows, line 32: No need to pass `supplyCatalog` here if we can call it from `EngineContext` passed to supply method instead?

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidatorHandler.java
```diff
@@ -0,0 +1,107 @@
...
+import io.aklivity.zilla.runtime.engine.validator.ValidatorHandler;
+import io.aklivity.zilla.runtime.types.core.config.StringValidatorConfig;
+
+public class StringValidatorHandler implements ValidatorHandler
+{
+    private boolean initialize;
+    private int incompleteCharCount;
+    private int incompleteChar;
+
+    public StringValidatorHandler(StringValidatorConfig config)
```
jfallows, line 29: Formatting and naturally need to pull out the encoding from the config.

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidatorHandler.java
```diff
@@ -0,0 +1,107 @@
...
+    {
+    }
+
+    @Override
+    public boolean validate(
+        int flags,
+        DirectBuffer data,
+        int index,
+        int length,
+        ValueConsumer next)
```
jfallows, line 39: Let's discuss what additional changes we need to make to the interface to support stateful validation of multi-byte chars across fragmentation boundaries.

jfallows (changes requested):

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidatorHandler.java
jfallows:
```suggestion
    public StringValidatorHandler(
        StringValidatorConfig config)
```

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidatorHandler.java
```diff
@@ -38,32 +38,28 @@ public boolean validate(
...
 
         final int limit = index + length;
         validate:
-        while (index < limit && initialize)
+        while (index < limit)
         {
             final int charByte0 = data.getByte(index);
             final int charByteCount = (charByte0 & 0b1000_0000) != 0
                     ? Integer.numberOfLeadingZeros((~charByte0 & 0xff) << 24)
                     : 1;
```
jfallows, line 54: `charByteCount` should move to the `else` clause as it is not needed for the `if` clause.

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidatorHandler.java
```diff
@@ -77,31 +73,19 @@ public boolean validate(
                     }
                 }
                 index += charByteCount;
```
jfallows, line 75: Feels like we are double counting the char bytes when fragmented, since `index++` will happen for each pending byte and `index += charByteCount` will happen for complete chars.

Perhaps we can simplify this algorithm to work from the perspective of `pendingBytes` always, and handle complete chars as a special case, where we need to reinitialize the number of `pendingBytes` from the first character byte?

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidatorHandler.java
```diff
@@ -77,31 +73,19 @@ public boolean validate(
...
         {
-            valid = initialize && incompleteCharCount == 0 && index == limit;
-            initialize = false;
+            valid = pendingCharBytes == 0 && index == limit;
         }
         else
         {
-            valid = initialize && index == limit;
+            valid = index + pendingCharBytes == limit;
         }
```
jfallows, line 88: Suggest simplifying to a ternary operator.
```
value = (flags & FLAGS_FIN) == 0x00
    ? index + pendingCharBytes == limit
    : pendingCharBytes == 0 && index == limit;
```

Btw, why is `index + pendingCharBytes == limit` the correct expression for fragmented messages? The `index` and `length` represent a single fragment, so why does that relate to `pendingBytes`?

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/config/StringValidatorConfigAdapter.java
```diff
@@ -53,29 +58,23 @@ public JsonValue adaptToJson(
     public StringValidatorConfig adaptFromJson(
         JsonValue value)
     {
+        JsonValue.ValueType type = value.getValueType();
         StringValidatorConfig result = null;
-        if (value instanceof JsonString)
+        switch (type)
```
jfallows, line 63:
```suggestion
        switch (value.valueType())
```

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/config/StringValidatorConfigAdapter.java
```diff
@@ -53,29 +58,23 @@ public JsonValue adaptToJson(
     public StringValidatorConfig adaptFromJson(
         JsonValue value)
     {
+        JsonValue.ValueType type = value.getValueType();
         StringValidatorConfig result = null;
```
jfallows, line 62: Use `config` instead of `result`?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/validator/ValidatorContext.java
```diff
@@ -19,14 +19,9 @@
 
 public interface ValidatorContext
 {
-    default ValidatorHandler attach(
+    default ValidatorHandler supplyHandler(
```
jfallows, line 22: I think we need to remove the `default` and `{ return null; }` as this must always be implemented.

jfallows (changes requested):

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidatorEncoding.java
```diff
@@ -0,0 +1,109 @@
...
+                }
+            }
+
+            return (flags & FLAGS_FIN) == 0x00
+                ? index == limit
+                : pendingCharBytes == 0 && index == limit;
+        }
+    },
+
+    INVALID
```
jfallows, line 79: If schema requires `utf_8` as only possibility, this can be removed, agree?

incubator/types-core/src/main/java/io/aklivity/zilla/runtime/types/core/StringValidatorEncoding.java
```diff
@@ -0,0 +1,109 @@
...
+ * License at
+ *
+ *   https://www.aklivity.io/aklivity-community-license/
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package io.aklivity.zilla.runtime.types.core;
```
jfallows, line 15: All non-public code should be in `internal` subpackage like other modules.

jfallows (changes requested):

jfallows (approved):

PR #751, opened 2024-01-19
Support  expression for primitive type in json schema

### LINKED ISSUE #583: Schema validation fails before the `${{env.*}}` parameters have been removed
**Describe the bug**
the json schema does not check for the presence of env variables `${{env.*}}` in the zilla.yaml and invalidates fields that aren't expecting strings. 

**To Reproduce**
1. open a zilla.yaml in vscode with `${{env.*}}` defined in a number field and it will have an error shown

akrambek:
## Description

Support  expression for primitive type in json schema.

Fixes #583


jfallows (commented):
Remember to remove expression configuration property from EngineConfiguraiton.

Also please include test to verify validation failure for replacing expression with expression such that text looking like expression remains after replacing actual expressions.

jfallows (approved):

PR #760, opened 2024-01-24
Limit sharding to mqtt 5

bmaidics:


specs/binding-mqtt.spec/src/main/resources/META-INF/zilla/mqtt.idl
```diff
@@ -18,7 +18,8 @@ scope mqtt
     enum MqttSessionFlags (uint8)
     {
         CLEAN_START(1),
-        WILL(2)
+        WILL(2),
+        SHARDING(3)
```
jfallows, line 22: I think this would be better moved to `MqttServerCapabilities` and renamed.

```
        enum MqttServerCapabilities (uint8)
        {
            RETAIN (0),
            WILDCARD (1),
            SUBSCRIPTION_IDS (2),
            SHARED_SUBSCRIPTIONS (3),
            REDIRECT(4)
        }
```
where `REDIRECT` describes a capability of the `mqtt` server, meaning we can send the `RESET` with `MqttResetEx` saying to redirect to a different `serverRef`, and as it turns out we then happen to use that for `sharding` as an implementation detail at `mqtt-kafka`.

jfallows (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2309,10 +2310,10 @@ private static int indexOfByte(
         return byteAt;
     }
 
-    private static boolean isSetShardingFlag(
+    private static boolean redirectCapability(
```
jfallows, line 2313:
```suggestion
    private static boolean hasRedirectCapability(
```

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -2309,10 +2310,10 @@ private static int indexOfByte(
...
         int flags)
     {
-        return (flags & 1 << MqttSessionFlags.SHARDING.value()) != 0;
+        return (flags & 1 << MqttServerCapabilities.REDIRECT.value()) != 0;
```
jfallows, line 2316:
```suggestion
        return (flags & REDIRECT_AVAILABLE_MASK) != 0;
```

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttOptionsConfigAdapter.java
```diff
@@ -45,6 +44,8 @@ public class MqttOptionsConfigAdapter implements OptionsConfigAdapterSpi, JsonbA
     private static final String AUTHORIZATION_CREDENTIALS_CONNECT_NAME = "connect";
     private static final String TOPICS_NAME = "topics";
     private static final String VERSIONS_NAME = "versions";
+    private static final String MQTT_VERSION_4 = "v3.1.1";
+    private static final String MQTT_VERSION_5 = "v5";
```
jfallows, line 48: Please rename the constants to better align with the updated values.
Perhaps have an `MqttVersion` enum with enum constants named for the specification versions, with an attribute on each for the corresponding protocol version to better document what's happening here via the code.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttOptionsConfigAdapter.java
```diff
@@ -110,7 +111,7 @@ public JsonObject adaptToJson(
         {
             JsonArrayBuilder versions = Json.createArrayBuilder();
             mqttOptions.versions
-                .forEach(versions::add);
+                .forEach(v -> versions.add(v == 4 ? MQTT_VERSION_4 : MQTT_VERSION_5));
```
jfallows, line 114: With the `MqttVersion` enum, you can keep the `versions::add`.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttOptionsConfigAdapter.java
```diff
@@ -166,7 +167,11 @@ public OptionsConfig adaptFromJson(
         if (object.containsKey(VERSIONS_NAME))
         {
             List<Integer> versions = object.getJsonArray(VERSIONS_NAME).stream()
-                .map(item -> ((JsonNumber) item).intValue())
+                .map(item ->
+                {
+                    String s = ((JsonString) item).getString();
+                    return s.equals(MQTT_VERSION_4) ? 4 : 5;
+                })
```
jfallows, line 174: With the `MqttVersion` enum, this can use a static method on `MqttVersion` to lookup by version string.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -2950,14 +2950,15 @@ else if (this.authField.equals(MqttConnectProperty.PASSWORD))
 
                 this.session = new MqttSessionStream(originId, resolved.id, 0);
 
-                final int shardingMask = protocolVersions.contains(MQTT_PROTOCOL_VERSION_5) && protocolVersions.size() == 1
-                    ? SHARDING_FLAG_MASK : 0;
+                final int redirectMask = protocolVersions.contains(MQTT_PROTOCOL_VERSION_5) && protocolVersions.size() == 1
```
jfallows, line 2953:
```suggestion
                final int capabilities = protocolVersions.contains(MQTT_PROTOCOL_VERSION_5) && protocolVersions.size() == 1
```

jfallows (changes requested):

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttOptionsConfigAdapter.java
bmaidics: We can simplify this, but not all the way to versions::add, since it's an unambiguous call (add(String), add(JsonValue))

runtime/binding-mqtt/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttOptionsConfigAdapterTest.java
```diff
@@ -103,20 +104,20 @@ public void shouldReadOptions()
...
         assertThat(topic.content.type, equalTo("test"));
 
-        int version1 = options.versions.get(0);
-        assertThat(version1, equalTo(4));
-        int version2 = options.versions.get(1);
-        assertThat(version2, equalTo(5));
+        MqttVersion version1 = options.versions.get(0);
+        assertThat(version1.versionNumber(), equalTo(4));
+        MqttVersion version2 = options.versions.get(1);
+        assertThat(version2.versionNumber(), equalTo(5));
```
jfallows, line 110: This doesn't seem correct. The `protocolVersion` (now `protocol`) should be the wire protocol, whereas the `versionNumber` (now `specification`) should be the spec version, i.e. v3.1.1, v5.

This seems to have mixed `versionNumber` with wire protocol version 4 and 5.

Also, these asserts can compare to `MqttVersion` enums directly, instead of attributes of those enums.

If we need a test to verify the attributes, then please add `MqttVersionTest` and let's do it once for the `MqttVersion` enum constants.

jfallows (approved):

PR #767, opened 2024-01-30
Model specific cache detect schema change update

aDaemonThread:


incubator/model-avro/src/main/java/io/aklivity/zilla/runtime/model/avro/internal/AvroModelHandler.java
```diff
@@ -217,4 +244,13 @@ private int calculatePadding(
         }
         return padding;
     }
+
+    private int generateCRC32C(
+        String schemaText)
+    {
+        byte[] bytes = schemaText.getBytes();
```
jfallows, line 251: This line is going to allocate a new byte[] on every call to validate.
Let's instead store the actual schemaText rather than a crc32c hash to compare.

incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
```diff
@@ -125,4 +149,13 @@ private JsonProvider createProvider(
         }
         return provider;
     }
+
+    private int generateCRC32C(
+        String schemaText)
+    {
+        byte[] bytes = schemaText.getBytes();
```
jfallows, line 156: Same feedback here as for `avro`.

incubator/model-protobuf/src/main/java/io/aklivity/zilla/runtime/model/protobuf/internal/ProtobufModelHandler.java
```diff
@@ -270,4 +297,13 @@ private DescriptorTree createDescriptorTree(
         }
         return tree;
     }
+
+    private int generateCRC32C(
+        String schemaText)
+    {
+        byte[] bytes = schemaText.getBytes();
```
jfallows, line 304: Same feedback here as for `avro`.

jfallows (changes requested):

jfallows (approved):

PR #774, opened 2024-02-01
 Skip invalid Kafka messages during Fetch

aDaemonThread:


jfallows (commented):
Looks good, we just need to improve the log messages.

`timestamp namespace.binding: Skipping invalid message on topic abc, partition N, offset M`

?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -510,7 +515,8 @@ public void writeEntryContinue(
                 if (converted == -1)
                 {
                     logFile.writeInt(entryMark.value + FIELD_OFFSET_FLAGS, CACHE_ENTRY_FLAGS_ABORTED);
-                    System.out.println("Invalid Value");
+                    System.out.println(String.format("%s %s: Skipping invalid message on topic %s, partition %d, offset %d",
+                        System.currentTimeMillis(), bindingName, topic, id, offset));
```
jfallows, line 519: Later, we're going to be logging to the `EngineContext`, and we usually don't pass strings around for bindings, just binding ids.

Let's change the method signature to put `EngineContext` and `bindingId` as the first 2 parameters. Note that `bindingId` is  already there as `routedId` in `KafkaCacheServerFetchFanout`.

```suggestion
                    System.out.printf("%s:%s %s: Skipping invalid message on topic %s, partition %d, offset %d\n",
                        System.currentTimeMillis(), context.supplyNamespace(bindingId), context.supplyLocalName(bindingId), topic, id, offset);
```

Also, let's put these printfs inside conditional if based on `EngineConfiguration.verbose()`.

jfallows (changes requested):

jfallows (approved):

PR #781, opened 2024-02-05
 Log validation failure of HTTP messages (stdout)

aDaemonThread:


runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/HttpConfiguration.java
```diff
@@ -122,4 +126,15 @@ public String16FW userAgentHeader()
     {
         return userAgentHeader;
     }
+
+    public boolean verbose()
+    {
+        return HTTP_VERBOSE.get(this);
+    }
+
+    private static boolean supplyVerbose(
```
jfallows, line 135:
```suggestion
    private static boolean verboseDefault(
```

jfallows (changes requested):

jfallows (changes requested):

jfallows (approved):

PR #790, opened 2024-02-09
Support TLSv1.3 handshake completion

### LINKED ISSUE #791: TLSv1.3 client handshake stall
**Describe the bug**
When `tls``client` binding is configured to use `TLSv1.3` (default) and remote TLS endpoint also supports TLSv1.3 then the TLS handshake can stall.

**To Reproduce**
Configure `zilla.yaml` to use `tls` `client` binding with `TLSv1.3` and attempt to complete TLS handshake.

**Expected behavior**
`TLSv1.3` handshake should complete successfully.

**Additional context**
As a workaround, explicitly configuring the `tls` `client` binding `version` to use `TLSv1.2` prevents the stall.

`SSLEngine` handshake status immediately after the handshake can be `NOT_HANDSHAKING` rather than `FINISHED` with `TLSv1.3`

See https://bugs.openjdk.org/browse/JDK-8240871 for details.

jfallows:
Fixes #791

### DIFF
--- a/runtime/binding-tls/src/test/java/io/aklivity/zilla/runtime/binding/tls/internal/streams/BridgeIT.java
+++ b/runtime/binding-tls/src/test/java/io/aklivity/zilla/runtime/binding/tls/internal/streams/BridgeIT.java
@@ -15,7 +15,6 @@
  */
 package io.aklivity.zilla.runtime.binding.tls.internal.streams;
 
-import static io.aklivity.zilla.runtime.engine.EngineConfiguration.ENGINE_DRAIN_ON_CLOSE;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.rules.RuleChain.outerRule;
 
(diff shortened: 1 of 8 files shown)


runtime/binding-tls/src/test/java/io/aklivity/zilla/runtime/binding/tls/internal/streams/BridgeIT.java
```diff
@@ -0,0 +1,70 @@
...
+        .addScriptRoot("bridge", "io/aklivity/zilla/specs/binding/tls/streams/bridge");
+
+    private final TestRule timeout = new DisableOnDebug(new Timeout(10, SECONDS));
+
+    private final EngineRule engine = new EngineRule()
+            .directory("target/zilla-itests")
+            .countersBufferCapacity(8192)
+            .configurationRoot("io/aklivity/zilla/specs/binding/tls/config")
+            .external("app1")
+            .configure(ENGINE_DRAIN_ON_CLOSE, false)
```
akrambek, line 45: Did you intentionally want to keep it?

akrambek (changes requested):

runtime/binding-tls/src/test/java/io/aklivity/zilla/runtime/binding/tls/internal/streams/BridgeIT.java
jfallows: No, it was carried from another IT, removed now, thanks!

akrambek (approved):

PR #797, opened 2024-02-14
Zilla is validating env vars before replacing them

### LINKED ISSUE #795: Zilla is validating `env` vars before replacing them.
**Describe the bug**
Zilla is validating `env` vars before replacing them. 

```
[23,31][/telemetry/exporters/prometheus_metric_exporter/options/endpoints/0/port] The value must be of integer type, but actual type is string.
error
io.aklivity.zilla.runtime.engine.config.ConfigException: Engine configuration failed
```

**To Reproduce**
```
docker run -v ./zilla.yaml:/etc/zilla/zilla.yaml --env PORT="7190" ghcr.io/aklivity/zilla:0.9.68 start -v
```
zilla.yaml
```
---
name: example
bindings:
  north_tcp_server:
    type: tcp
    kind: server
    options:
      host: 0.0.0.0
      port: 12345
    exit: north_echo_server
  north_echo_server:
    type: echo
    kind: server

telemetry:
  exporters:
    prometheus_metric_exporter:
      type: prometheus
      options:
        endpoints:
          - scheme: http
            path: /metrics
            port: ${{env.PORT}}
```

akrambek:
## Description

Zilla is validating env vars before replacing them

Fixes #795

### DIFF
--- a/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigAnnotator.java
+++ b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigAnnotator.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2021-2023 Aklivity Inc.
+ *
+ * Aklivity licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.aklivity.zilla.runtime.engine.config;
+
+import java.util.Deque;
+import java.util.LinkedList;
+
+import jakarta.json.Json;
+import jakarta.json.JsonArray;
+import jakarta.json.JsonArrayBuilder;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonObjectBuilder;
+import jakarta.json.JsonValue;
+
+import org.agrona.collections.MutableInteger;
+
+public final class EngineConfigAnnotator
+{
+    private final Deque<String> schemaKeys;
+    private final Deque<Integer> schemaIndexes;
+
+    public EngineConfigAnnotator()
+    {
+        this.schemaKeys = new LinkedList<>();
+        this.schemaIndexes = new LinkedList<>();
+    }
+
+    public JsonObject annotateJson(
+        JsonObject jsonObject)
+    {
+        schemaKeys.clear();
+        schemaIndexes.clear();
+
+        return (JsonObject) annotateJsonObject(jsonObject);
+    }
+
+    private JsonValue annotateJsonObject(
+        JsonObject jsonObject)
+    {
+        JsonObjectBuilder builder = Json.createObjectBuilder();
+
+        jsonObject.forEach((key, value) ->
+        {
+            schemaKeys.push(key);
+
+            parse:
+            if ("expression".equals(key))
+            {
+                builder.add(key, value);
+            }
+            else if (value.getValueType() == JsonValue.ValueType.OBJECT)
+            {
+                builder.add(key, annotateJsonObject(value.asJsonObject()));
+            }
+            else if (value.getValueType() == JsonValue.ValueType.ARRAY)
+            {
+                if (jsonObject.containsKey("type") && key.equals("enum"))
+                {
+                    break parse;
+                }
+                builder.add(key, annotateJsonArray(value.asJsonArray()));
+            }
+            else if (key.equals("type") && isPrimitiveType(value))
+            {
+                builder.add("anyOf", createAnyOfTypes(jsonObject));
+            }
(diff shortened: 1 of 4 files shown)


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigReader.java
```diff
@@ -190,7 +190,7 @@ private boolean validateAnnotatedSchema(
         validate:
         try
         {
-            final JsonObject annotatedSchemaObject = (JsonObject) annotateJsonObject(schemaObject);
+            final JsonObject annotatedSchemaObject = jsonAnnotator.apply(schemaObject);
```
jfallows, line 193:
```suggestion
final EngineConfigAnnotator annotator = new EngineConfigAnnotator();
final JsonObject annotatedSchemaObject = annotator.annotate(schemaObject);
```

Given the state needed by internals of annotator, suggest we just instantiate this here on stack and call the annotate method directly instead of having a field.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigAnnotator.java
```diff
@@ -0,0 +1,159 @@
...
+    private final Deque<String> schemaKeys;
+    private final Deque<Integer> schemaIndexes;
+
+    public EngineConfigAnnotator()
+    {
+        this.schemaKeys = new LinkedList<>();
+        this.schemaIndexes = new LinkedList<>();
+    }
+
+    public JsonObject annotateJson(
```
jfallows, line 41:
```suggestion
    public JsonObject annotate(
```

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigAnnotator.java
```diff
@@ -0,0 +1,159 @@
...
+import jakarta.json.Json;
+import jakarta.json.JsonArray;
+import jakarta.json.JsonArrayBuilder;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonObjectBuilder;
+import jakarta.json.JsonValue;
+
+import org.agrona.collections.MutableInteger;
+
+public final class EngineConfigAnnotator
```
jfallows, line 30: Please add `EngineConfigAnnotatorTest` to verify some of the important scenarios.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigAnnotator.java
akrambek: Sounds good just wanted to check if that what you mean

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigReader.java
akrambek: That's what I was thinking but then thought that you have resaon

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigAnnotator.java
```diff
@@ -0,0 +1,160 @@
...
+{
+    private final LinkedList<String> schemaKeys;
+    private final LinkedList<Integer> schemaIndexes;
+
```
akrambek, line 33: Used implementation because I need an interface from both the deque and the list. Let me know if this is wrong

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigReader.java
```diff
@@ -61,6 +57,8 @@ public final class EngineConfigReader
     private final Resolver expressions;
     private final Collection<URL> schemaTypes;
     private final Consumer<String> logger;
+    private final EngineConfigAnnotator annotator;
+
```
jfallows, line 61: Please remove this field.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigReader.java
```diff
@@ -72,6 +70,7 @@ public EngineConfigReader(
         this.expressions = expressions;
         this.schemaTypes = schemaTypes;
         this.logger = logger;
+        this.annotator = new EngineConfigAnnotator();
```
jfallows, line 73: Please remove this assignment.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigReader.java
```diff
@@ -190,7 +189,7 @@ private boolean validateAnnotatedSchema(
         validate:
         try
         {
-            final JsonObject annotatedSchemaObject = (JsonObject) annotateJsonObject(schemaObject);
+            final JsonObject annotatedSchemaObject = annotator.annotate(schemaObject);
```
jfallows, line 192:
```suggestion
            final EngineConfigAnnotator() annotator = new EngineConfigAnnotator();;
            final JsonObject annotatedSchemaObject = annotator.annotate(schemaObject);
```

This should be local because it has internal parse state, no need to carry any potential side effects of previous read to next read.

jfallows (changes requested):

jfallows (approved):

PR #830, opened 2024-03-05
Integer Validator improvement to support OpenAPI & AsyncAPI specs

aDaemonThread:
## Description

 - integer validation (text & binary)
     - int32, int64  
     - max and min value
     - multiple of N


incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/IntegerFormat.java
```diff
@@ -0,0 +1,184 @@
...
+import org.agrona.DirectBuffer;
+
+public enum IntegerFormat
+{
+    TEXT
+    {
+        private boolean number;
+        private boolean sign;
+        int decimalPlaces;
+        private int value;
```
jfallows, line 30: When we use enum as a strategy object, the implementation needs to be stateless because the same instance of the enum constant is reused by all callers, so any state stored on the instance of the enum constant is also shared by all callers, causing a collision either across threads or due to overlapping fragmented validate calls even on the same thread.

Suggest moving the validation to a different class that can have state, and keeping just the remaining necessary behavioral parts here in the enum that can perform the checks needed by the stateful validator.

jfallows (changes requested):

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32ValidatorHandler.java
```diff
@@ -0,0 +1,126 @@
...
+                }
+                valid = conditions(value, max, min, exclusiveMax, exclusiveMin, multiple);
+            }
+            else
+            {
+                valid = format.validateContinue(progress);
+            }
+        }
+        return valid;
+    }
```
jfallows, line 105: If we use `MutableInteger`, then enum can modify state without inadvertently sharing state across validations or workers.

Using `decoded` field to store incrementally decoded result across multiple calls to `format.decode(...)`.
Using `processed` field to store total processed bytes across multiple calls to `format.decode(...)`.

```suggestion
    @Override
    public boolean validate(
        int flags,
        DirectBuffer data,
        int index,
        int length,
        ValueConsumer next)
    {
        if ((flags & FLAGS_INIT) != 0x00)
        {
            decoded.value = 0;
            processed.value = 0;
        }

        int progress = format.decode(decoded, processed, data, index, length);
        boolean valid = progress != IntegerFormat.INVALID_INDEX;

        if ((flags & FLAGS_FIN) != 0x00)
        {
            valid &= format.valid(decoded, processed);
            valid &= check.test(value);
        }

        return valid;
    }
```

where the enum becomes something like...
```
public enum IntegerFormat
{
    public static final int INVALID_INDEX = -1;

    TEXT
    {
        @Override
        public int decode(
            MutableInteger decoded,
            MutableInteger processed,
            DirectBuffer data,
            int index,
            int length)
        {
            int progress = index;
            int limit = index + length;

            decode:
            for (; progress < limit; progress++)
            {
                int digit = data.getByte(progress);

                if (digit < '0' || '9' < digit)
                {
                    if (processed.value == 0))
                    {
                        switch (digit)
                        {
                        case '-':
                            decoded.value = Integer.MIN_VALUE;
                            processed.value++;
                            continue decode;
                        case '+':
                            decoded.value = Integer.MAX_VALUE;
                            processed.value++;
                            continue decode;
                        default:
                            break;
                        }
                    }

                    progress = INVALID_INDEX;
                    break decode;
                }

                int multipler = 10;

                if (processed.value == 1))
                {
                    switch (decoded.value)
                    {
                    case Integer.MIN_VALUE:
                        decoded.value = -1;
                        multipler = 1;
                        break;
                    case Integer.MAX_VALUE:
                        decoded.value = 1;
                        multipler = 1;
                        break;
                    default:
                        break;
                    }
                }

                decoded.value = decoded.value * multipler + (digit - '0');
                processed.value++;
            }
        }

        @Override
        public boolean valid(
            MutableInteger decoded,
            MutableInteger processed)
        {
            return processed.value > 1 || decoded.value != Integer.MIN_VALUE || decoded.value != Integer.MAX_VALUE;
        }
    },

    BINARY
    {
        private static final int INT32_SIZE = 4;

        @Override
        public int decode(
            MutableInteger decoded,
            MutableInteger processed,
            DirectBuffer data,
            int index,
            int length)
        {
            int progress = index;
            int limit = index + length;

            decode:
            for (; progress < limit; progress++)
            {
                int digit = data.getByte(progress);

                if (processed.value >= INT32_SIZE)
                {
                    progress = INVALID_INDEX;
                    break decode;
                }

                decoded.value <<= 8;
                decoded.value |= digit & 0xFF;
                processed.value++;
            }

            return progress;
        }

        @Override
        public boolean valid(
            MutableInteger decoded,
            MutableInteger processed)
        {
            return processed.value == INT32_SIZE;
        }
    }
```

jfallows (changes requested):

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/config/Int32ModelConfigAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+
+            int multiple = object.containsKey(MULTIPLE_NAME)
+                ? object.getInt(MULTIPLE_NAME)
+                : 0;
+
+            result = new Int32ModelConfig(format, max, min, exclusiveMax, exclusiveMin, multiple);
+        }
+        else
+        {
+            assert false;
```
jfallows, line 129: This should be an exception instead of an assert, we need to know for sure things are working or not at runtime.

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/config/Int32ModelConfigAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+
+        return builder.build();
+    }
+
+    @Override
+    public ModelConfig adaptFromJson(
+        JsonValue value)
+    {
+        Int32ModelConfig result = null;
+        if (value instanceof JsonString)
```
jfallows, line 94: Can we check the value type instead of instanceof?

jfallows (changes requested):

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32Format.java
```diff
@@ -0,0 +1,167 @@
...
+
+public enum Int32Format
+{
+
+    TEXT
+    {
+        @Override
+        public int decode(
+            MutableInteger decoded,
+            MutableInteger processed,
```
jfallows, line 28: Suggest we add a separate `Int32State` class and use instead of these two mutable integers.
```
public final class Int32State
{
    public int decoded;
    public int processed;
}
```
then pass an instance here as a parameter called `state`, and refer to the fields as needed below.

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32Format.java
```diff
@@ -0,0 +1,167 @@
...
+            int progress = index;
+            int limit = index + length;
+
+            for (; progress < limit; progress++)
+            {
+                int digit = data.getByte(progress);
+
+                if (digit < '0' || '9' < digit)
+                {
+                    if (processed.value == 0)
```
jfallows, line 42:
```suggestion
                    if (state.processed == 0)
```

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32Format.java
```diff
@@ -0,0 +1,167 @@
...
+                    {
+                        switch (digit)
+                        {
+                        case '-':
+                            decoded.value = Integer.MIN_VALUE;
+                            processed.value++;
```
jfallows, line 48:
```suggestion
                            state.decoded = Integer.MIN_VALUE;
                            state.processed++;
```

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32Format.java
```diff
@@ -0,0 +1,167 @@
...
+                            continue;
+                        case '+':
+                            decoded.value = Integer.MAX_VALUE;
+                            processed.value++;
```
jfallows, line 52:
```suggestion
                            state.decoded = Integer.MAX_VALUE;
                            state.processed++;
```

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32Format.java
```diff
@@ -0,0 +1,167 @@
...
+                        }
+                    }
+
+                    progress = INVALID_INDEX;
+                    break;
+                }
+
+                int multipler = 10;
+
+                if (processed.value == 1)
```
jfallows, line 65:
```suggestion
                if (state.processed == 1)
```

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32Format.java
```diff
@@ -0,0 +1,167 @@
...
+                    case Integer.MIN_VALUE:
+                        decoded.value = -1 * (digit - '0');
+                        break;
+                    case Integer.MAX_VALUE:
+                        decoded.value = digit - '0';
+                        break;
+                    default:
+                        decoded.value = decoded.value * multipler + (digit - '0');
+                        break;
+                    }
```
jfallows, line 78:
```suggestion
                    switch (state.decoded)
                    {
                    case Integer.MIN_VALUE:
                        state.decoded = -1 * (digit - '0');
                        break;
                    case Integer.MAX_VALUE:
                        state.decoded = digit - '0';
                        break;
                    default:
                        state.decoded = state.decoded * multipler + (digit - '0');
                        break;
                    }
```

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32Format.java
```diff
@@ -0,0 +1,167 @@
...
+                }
+                else
+                {
+                    decoded.value = decoded.value < 0
+                        ? decoded.value * multipler - (digit - '0')
+                        : decoded.value * multipler + (digit - '0');
```
jfallows, line 84:
```suggestion
                    state.decoded = state.decoded < 0
                        ? state.decoded * multipler - (digit - '0')
                        : state.decoded * multipler + (digit - '0');
```

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32Format.java
```diff
@@ -0,0 +1,167 @@
...
+                }
+                processed.value++;
```
jfallows, line 86:
```suggestion
                state.processed++;
```

incubator/model-core/src/main/java/io/aklivity/zilla/runtime/model/core/internal/Int32Format.java
```diff
@@ -0,0 +1,167 @@
...
+            return progress;
+        }
+
+        @Override
+        public boolean valid(
+            MutableInteger decoded,
+            MutableInteger processed)
+        {
+            return processed.value > 1 || decoded.value != Integer.MIN_VALUE || decoded.value != Integer.MAX_VALUE;
+        }
```
jfallows, line 98:
```suggestion
        public boolean valid(
            Int32State state)
        {
            return state.processed > 1 || state.decoded != Integer.MIN_VALUE || state.decoded != Integer.MAX_VALUE;
        }
```

jfallows (changes requested):

jfallows (approved):

PR #839, opened 2024-03-07
Add event logs to open telemetry exporter

### LINKED ISSUE #785: Support remote logging of events via `otlp`
As devops, I want to understand why zilla is not behaving as expected when integrating with external actors such as authenticating clients, Kafka broker authorization or API version support, or DNS servers, so that I can correct either zilla.yaml or the external system setup as needed.

**Acceptance criteria**
 - `otlp` exporter of named events

**Additional context**
Depends on aklivity/zilla#679

**Tasks**
 - [ ] Implement `otlp` telemetry exporter (events only)
 - [ ] Document `otlp` telemetry exporter (reference, events)

attilakreiner:
## Description

This change adds the feaure to send event logs to the open telemetry exporter.

Fixes #785


incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/config/OtlpOverridesConfig.java
```diff
@@ -19,10 +19,13 @@
 public class OtlpOverridesConfig
```
jfallows, line 19: Let's remove this (also from schema) and use `OltpConfiguration` for overrides instead.

incubator/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/application/event/server.rpt
```diff
@@ -0,0 +1,82 @@
...
+                    '"scope":{'
+                        '"name":"OtlpLogsSerializer",'
+                        '"version":"1.0.0"'
+                    '},'
+                    '"logRecords":['
+                        '{'
+                            '"timeUnixNano":42000000,'
+                            '"observedTimeUnixNano":42000000,'
+                            '"traceId":"8000000000000001",'
+                            '"spanId":"100000004",'
```
jfallows, line 54: Where did the values for `traceId` and `spanId` come from?
Perhaps we can leave these unset (if permitted) since we don't have a valid `otlp` trace or span to associate them with just yet.

incubator/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/network/event/client.rpt
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations under the License.
+#
+
+connect "zilla://streams/net0"
+    option zilla:window 8192
+
+connected
```
jfallows, line 19: Is this needed or is there a spec from engine that we should be using instead?
Also noticed there is no server.rpt for this client.rpt.

incubator/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/application/event/server.rpt
```diff
@@ -0,0 +1,82 @@
...
+                        '}'
+                    '}'
+                ']'
+            '},'
+            '"scopeLogs":['
+                '{'
+                    '"scope":{'
+                        '"name":"OtlpLogsSerializer",'
+                        '"version":"1.0.0"'
+                    '},'
```
jfallows, line 48: How does this relate to the `OtlpLogsSerializer` class in our internal implementation?
Is it a standard scope for logs, or our own creation?

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/OltpExporterHandler.java
```diff
@@ -104,17 +112,30 @@ public int export()
...
+            {
+                String logsJson = logsSerializer.serializeAll();
+                HttpRequest logsRequest = HttpRequest.newBuilder()
+                    .uri(logsEndpoint)
+                    .header("Content-Type", "application/json")
+                    .POST(HttpRequest.BodyPublishers.ofString(logsJson))
+                    .timeout(timeoutInterval)
+                    .build();
+                logsResponse = httpClient.sendAsync(logsRequest, HttpResponse.BodyHandlers.ofString());
+                logsResponse.thenAccept(responseHandler);
```
jfallows, line 138: Let's split this into `int exportMetrics()` and `int exportLogs()` called from `int export()`.

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/serializer/EventReader.java
```diff
@@ -0,0 +1,121 @@
...
+        addBody("body");
+        eventAttributesJson = Json.createArrayBuilder();
+        addStringAttribute("event.name", "test.test");
+        String extension = formatter.format(msgTypeId, buffer, index, length);
+        addStringAttribute("extension", extension);
+        //addIntAttribute("id", 42);
+        eventJson.add(ATTRIBUTES, eventAttributesJson);
+        eventsJson.add(eventJson);
+        System.out.println(eventJson);
+        System.out.printf("%s 0x%016x [%d]\n", qname, event.traceId(), event.timestamp());
```
jfallows, line 93: Some tidy left todo here.

jfallows (changes requested):

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/config/OtlpOverridesConfig.java
attilakreiner: done

incubator/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/application/event/server.rpt
attilakreiner: both are optional, removed for now

incubator/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/network/event/client.rpt
attilakreiner: yep, we can get this from the engine, fixed
also, I'm done adding the missing script pairs, should be good now

incubator/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/application/event/server.rpt
attilakreiner: This is just a scope that we have to define ourselves, it can really be an arbitrary string, but one scope is mandatory.
For metrics we went with `OtlpMetricsSerializer`, so `OtlpLogsSerializer` seemed to be the logical choice here.

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/OltpExporterHandler.java
attilakreiner: done

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/serializer/EventReader.java
attilakreiner: done

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/OltpConfiguration.java
```diff
@@ -58,4 +62,15 @@ public Duration warningInterval()
     {
         return OTLP_EXPORTER_WARNING_INTERVAL.get(this);
     }
+
+    // TODO: Ati
```
jfallows, line 66: TODO

jfallows (commented):
Apologies, I think I may have misinterpreted our intent for `overrides`, I think we need to revert that change.

I was thinking why we need to have overrides in the config, when we already have location, and it seemed to be for override like we do in other places.

However, it turns out the override is to give us flexibility in the path endpoint for metrics / events while specifying the location url once, while going with standard path defaults when only the location is specified.

incubator/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/OltpConfiguration.java
attilakreiner: oops, removed

jfallows (approved):

PR #844, opened 2024-03-12
Implement JoinGroup request as first class stream

### LINKED ISSUE #719: Zilla Quickstart gRPC RouteGuide service hangs after lots of messages
**Describe the bug**
Starting the zilla quickstart.

**To Reproduce**
1. Run the [quickstart](https://docs.aklivity.io/zilla/latest/tutorials/quickstart/kafka-proxies.html#zilla-docker-compose-stack)
2. Invoke the `ListFeature` endpoint in the gRPC RouteGuide collection one or many times
  - each invoke should stream ~50 messages
3. switch to and invoke the `RecordRoute` endpoint, send multiple messages, end the stream, and a message should return stating the number of messages recieved
4. repeat steps 2 and 3 until one doesn't finish correctly

akrambek:
## Description

Move join group request stream out of coordinator stream and use a separate connection if the group.initial.rebalance.delay.ms > 0 otherwise use the connection pool since there won't be any delay.

Fixes #719

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFactory.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFactory.java
@@ -15,7 +15,7 @@
  */
 package io.aklivity.zilla.runtime.binding.kafka.internal.stream;
 
-import java.util.function.Function;
+import java.util.function.IntFunction;
 import java.util.function.LongFunction;
 import java.util.function.UnaryOperator;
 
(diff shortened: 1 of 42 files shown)


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFactory.java
```diff
@@ -62,12 +63,20 @@ public KafkaClientFactory(
         final BindingHandler streamFactory = config.clientConnectionPool() ? connectionPool.streamFactory() :
                 context.streamFactory();
 
+        final Function<Integer, BindingHandler> streamFactorySupplier = d ->
```
jfallows, line 66: Avoid auto-boxing, use `IntFunction<BindingHandler>` instead, for all references too.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientGroupFactory.java
```diff
@@ -362,7 +381,7 @@ public MessageConsumer newStream(
             {
                 final long resolvedId = resolved.id;
                 final List<KafkaServerConfig> servers = binding.servers();
-                final KafkaSaslConfig sasl = resolveSasl.apply(binding.sasl());
+                final KafkaSaslConfig sasl = resolveSaslSupplier.apply(0).apply(binding.sasl());
```
jfallows, line 384: Please define a constant `private static final int NO_REBALANCE_DELAY = 0` and use it here instead of `0`.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientGroupFactory.java
```diff
@@ -413,6 +432,7 @@ public void onDetached(
 
     private MessageConsumer newStream(
         MessageConsumer sender,
+        int initialDelay,
```
jfallows, line 435: Given the meaning already associated with `initial` for streams, let's rename `initialDelay` to `rebalanceDelay` in this file.

jfallows (changes requested):

PR #851, opened 2024-03-14
Add key filter support in openapi asyncapi mapping

akrambek:
## Description

Add key filter support in openapi asyncapi mapping.

### DIFF
--- a/incubator/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/config/OpenapiAsyncCompositeBindingAdapter.java
+++ b/incubator/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/config/OpenapiAsyncCompositeBindingAdapter.java
@@ -15,7 +15,6 @@
 package io.aklivity.zilla.runtime.binding.openapi.asyncapi.internal.config;
 
 import static io.aklivity.zilla.runtime.engine.config.KindConfig.PROXY;
-import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
 
 import java.util.List;
@@ -194,15 +193,15 @@ public final class OpenapiAsyncCompositeBindingAdapter implements CompositeBindi
                     .build());
                 break merge;
             }
-
-            if (includeKey)
-            {
-                fetch.filters(List.of(HttpKafkaWithFetchFilterConfig.builder()
-                    .key("${params.id}")
-                    .headers(emptyList())
-                    .build()));
-            }
         }
+
+        if (includeKey)
+        {
+            fetch.filters(List.of(HttpKafkaWithFetchFilterConfig.builder()
+                .key("${params.id}")
+                .build()));
+        }
+
         return fetch;
     }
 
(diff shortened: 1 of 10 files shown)


incubator/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/config/OpenapiAsyncCompositeBindingAdapter.java
```diff
@@ -119,6 +122,8 @@ private <C> BindingConfigBuilder<C> injectHttpKafkaRoutes(
                         final AsyncapiChannelView channel = AsyncapiChannelView
                             .of(asyncapi.channels, asyncapiOperation.channel);
 
+                        final boolean includeKey = pathId.reset(item).find();
```
jfallows, line 125: I think this should not depend on the name of the parameter being `id`, for example `taxiId` should be fine too.

Also, the terminology `includeKey` is describing anticipated usage of the parameter, which is leaking the abstraction a bit. Instead, lets call this `hasParam`, or perhaps even passing the parameter name(s) so we are not so dependent on it being `{id}` in the path vs `{taxiId}` for example.

incubator/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/config/OpenapiAsyncCompositeBindingAdapter.java
```diff
@@ -181,11 +188,20 @@ private <C> HttpKafkaWithFetchConfigBuilder<C> injectHttpKafkaRouteFetchWith(
...
+                    .path("/-")
                     .build());
                 break merge;
             }
+        }
 
+        if (includeKey)
+        {
+            fetch.filters(List.of(HttpKafkaWithFetchFilterConfig.builder()
+                .key("${params.id}")
```
jfallows, line 201:
```suggestion
                .key(String.format("${params.%s}", paramName))
```

runtime/binding-http-kafka/src/test/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaWithConfigAdapterTest.java
```diff
@@ -129,11 +131,13 @@ public void shouldWriteWithFetchTopicAndFilters()
...
             .fetch(HttpKafkaWithFetchConfig.builder()
                 .topic("test")
-                .filters(singletonList(new HttpKafkaWithFetchFilterConfig(
-                        "fixed-key",
-                        singletonList(new HttpKafkaWithFetchFilterHeaderConfig(
-                            "tag",
-                            "fixed-tag")))))
+                .filters(singletonList(HttpKafkaWithFetchFilterConfig.builder()
+                    .key("fixed-key")
+                    .headers(singletonList(HttpKafkaWithFetchFilterHeaderConfig.builder()
```
jfallows, line 136: There should also be a `.header(String name, String value)` method on the builder that lazily instantiates the `headers` list and adds the header config.

jfallows (changes requested):

incubator/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/config/OpenapiAsyncCompositeBindingAdapter.java
```diff
@@ -49,8 +51,10 @@
 public final class OpenapiAsyncCompositeBindingAdapter implements CompositeBindingAdapterSpi
 {
     private static final Pattern JSON_CONTENT_TYPE = Pattern.compile("^application/(?:.+\\+)?json$");
+    private static final Pattern PARAMETER_PATTERN = Pattern.compile("\\{([^}]*)\\}");
```
jfallows, line 54:
```suggestion
    private static final Pattern PARAMETER_PATTERN = Pattern.compile("\\{([^}]+)\\}");
```
We require at least one character for parameter name, agree?

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #872, opened 2024-03-20
Schema fixes + avoiding duplicate reply begin on mqtt-kafka subscribe stream

bmaidics:


incubator/binding-asyncapi.spec/src/main/scripts/io/aklivity/zilla/specs/binding/asyncapi/schema/asyncapi.schema.patch.json
```diff
@@ -82,28 +94,7 @@
...
-                                            },
-                                            "retained":
-                                            {
-                                                "title": "Kafka Retained Channel",
-                                                "type": "string"
-                                            },
-                                            "additionalProperties": false
-                                        },
-                                        "additionalProperties": false
+                                        "$ref": "#/$defs/options/binding/mqtt-kafka/topics"
```
jfallows, line 97: I'm not sure about this one, since we're deliberately using `channels` asyncapi concept here, whereas mqtt-kafka deals with kafka-specific `topics` instead, so unifying this config causes asyncapi to switch to `topics` instead.

specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/schema/mqtt.kafka.schema.patch.json
```diff
@@ -43,34 +43,7 @@
...
-                                    "additionalProperties": false
-                                },
-                                "required":
-                                [
-                                    "sessions",
-                                    "retained",
-                                    "messages"
-                                ],
-                                "additionalProperties": false
+                                "$ref": "#/$defs/options/binding/mqtt-kafka/topics"
```
jfallows, line 46: See feedback on asyncapi options for `mqtt-kafka`.

jfallows (changes requested):

jfallows (approved):

PR #883, opened 2024-03-27
Support specific server in AsyncAPI spec in asyncapi binding

bmaidics:


incubator/binding-asyncapi.spec/src/main/scripts/io/aklivity/zilla/specs/binding/asyncapi/config/server.http.secure.yaml
```diff
@@ -27,6 +27,11 @@ bindings:
     options:
       specs:
         http_api: http/asyncapi.yaml
+      server: plain
```
jfallows, line 30: How is this syntax going to work when we support multiple specs as the server name has meaning only within the context of an individual spec, right?

Also, we have decided to optionally support multiple servers in openapi binding and will need to do the same here for parity.

There is also a need to support server variables, such that they can be replaced by valid values. Otherwise, server variables with no default would not be usable in the mapping.

jfallows (changes requested):

incubator/binding-asyncapi.spec/src/main/scripts/io/aklivity/zilla/specs/binding/asyncapi/config/server.http.secure.yaml
jfallows: There may well be value when multiple specs are expected to be deployed together on the same server, such that specifying the server(s) under options can then be applied to all specs.

This is acting like a filter on the specs to extract the servers that match for the generated configuration.

jfallows (approved):

jfallows (approved):

PR #889, opened 2024-04-01
zilla crash while using model-json and schema is not found

### LINKED ISSUE #817: Error in JsonValidatorHandler when the remote registry can't be accessed
**Describe the bug**
When you set up zilla with validation and schema-registry enabled, but the schema registry is inaccessible zilla crashes with a NPE. 

**To Reproduce**
1. Start zilla with a configuration where json validation and schema-registry is enabled e.g.
```
---
name: test
telemetry:
  exporters:
    stdout0:
      type: stdout
catalogs:
  catalog0:
    type: schema-registry
    options:
      url: http://localhost:8081
      context: default
      max-age: 30
bindings:
  north_tcp_server:
    type: tcp
    kind: server
    options:
      host: 0.0.0.0
      port: 7114
    exit: north_http_server
  north_http_server:
    type: http
    kind: server
    options:
      requests:
        - path: /hello
          method: POST
          content:
            model: json
            catalog:
              catalog0:
                - subject: item
    exit: south_http_client
  south_http_client:
    type: http
    kind: client
    exit: south_tcp_client
  south_tcp_client:
    type: tcp
    kind: client
    options:
      host: localhost
      port: 8080

```
2. Make sure the schema registry is **not available** at http://localhost:8081
3. Run the command `curl -v "localhost:7114/hello" -X POST -d "{'greeting':'hello'}"`
4. Zilla crashes at [this line](https://github.com/aklivity/zilla/blob/d0a8f21e5bcc2ac008554031818539b331a3f6e6/incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonValidatorHandler.java#L78) with a stack trace:
```
org.agrona.concurrent.AgentTerminationException: java.lang.NullPointerException: Cannot invoke "jakarta.json.spi.JsonProvider.createParser(java.io.InputStream)" because "provider" is null
	at io.aklivity.zilla.runtime.engine/io.aklivity.zilla.runtime.engine.internal.registry.EngineWorker.doWork(EngineWorker.java:761)
	at org.agrona.core/org.agrona.concurrent.AgentRunner.doDutyCycle(AgentRunner.java:291)
	at org.agrona.core/org.agrona.concurrent.AgentRunner.run(AgentRunner.java:164)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.NullPointerException: Cannot invoke "jakarta.json.spi.JsonProvider.createParser(java.io.InputStream)" because "provider" is null
	at io.aklivity.zilla.runtime.model.json/io.aklivity.zilla.runtime.model.json.internal.JsonValidatorHandler.validate(JsonValidatorHandler.java:78)
	at io.aklivity.zilla.runtime.engine/io.aklivity.zilla.runtime.engine.model.ValidatorHandler.validate(ValidatorHandler.java:41)
	at io.aklivity.zilla.runtime.binding.http/io.aklivity.zilla.runtime.binding.http.internal.stream.HttpServerFactory$HttpServer$HttpExchange.validateContent(HttpServerFactory.java:3231)
	at io.aklivity.zilla.runtime.binding.http/io.aklivity.zilla.runtime.binding.http.internal.stream.HttpServerFactory$HttpServer.onDecodeBody(HttpServerFactory.java:2300)
	at io.aklivity.zilla.runtime.binding.http/io.aklivity.zilla.runtime.binding.http.internal.stream.HttpServerFactory.decodeContent(HttpServerFactory.java:1425)
```

**Expected behavior**
1. Zilla shouldn't crash.
2. There is a decision to be made whether we deem inputs valid or invalid in the case the schema registry is not accessible. As discussed with @jfallows, we should follow the policy to deem inputs invalid if the schema registry is inaccessible.

**Screenshots**
N/A

**Additional context**

**IMPORTANT NOTE:** The repro steps above are pointing out the erratic behaviour in the case of **json** validation, but we should double check if the same problem exists in the case of **avro** and **protobuf** as well, and if so, those need to be fixed as well.

aDaemonThread:
Fixes #817

### DIFF
--- a/incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
+++ b/incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
@@ -76,8 +76,15 @@ public abstract class JsonModelHandler
         try
         {
             JsonProvider provider = supplyProvider(schemaId);
-            in.wrap(buffer, index, length);
-            provider.createReader(in).readValue();
+            if (provider != null)
+            {
+                in.wrap(buffer, index, length);
+                provider.createReader(in).readValue();
+            }
+            else
+            {
+                status = false;
+            }
         }
         catch (JsonValidatingException ex)
         {
--- a/incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
+++ b/incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
@@ -76,14 +76,11 @@ public abstract class JsonModelHandler
         try
         {
             JsonProvider provider = supplyProvider(schemaId);
-            if (provider != null)
+            status &= provider != null;
+            if (status)
             {
                 in.wrap(buffer, index, length);
                 provider.createReader(in).readValue();
-            }
-            else
-            {
-                status = false;
             }
         }
         catch (JsonValidatingException ex)
--- a/incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonValidatorHandler.java
+++ b/incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonValidatorHandler.java
@@ -75,10 +75,17 @@ public class JsonValidatorHandler extends JsonModelHandler implements ValidatorH
                     : handler.resolve(subject, catalog.version);
 
                 JsonProvider provider = supplyProvider(schemaId);
-                parser = provider.createParser(in);
-                while (parser.hasNext())
+                if (provider != null)
                 {
-                    parser.next();
+                    parser = provider.createParser(in);
+                    while (parser.hasNext())
+                    {
+                        parser.next();
+                    }
+                }
+                else
+                {
+                    status = false;
                 }
             }
         }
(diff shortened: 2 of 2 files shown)


incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
```diff
@@ -76,8 +76,15 @@ protected final boolean validate(
...
-            provider.createReader(in).readValue();
+            if (provider != null)
+            {
+                in.wrap(buffer, index, length);
+                provider.createReader(in).readValue();
+            }
+            else
+            {
+                status = false;
+            }
```
jfallows, line 87: Try to avoid using `else` clauses to assign boolean variables like this.

Instead, aim to take an approach that would also work for multiple conditions discovered incrementally.

```suggestion
            status &= provider != null;

            if (status)
            {
                in.wrap(buffer, index, length);
                provider.createReader(in).readValue();
            }
```

incubator/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonValidatorHandler.java
```diff
@@ -75,10 +75,17 @@ public boolean validate(
...
-                    parser.next();
+                    parser = provider.createParser(in);
+                    while (parser.hasNext())
+                    {
+                        parser.next();
+                    }
+                }
+                else
+                {
+                    status = false;
```
jfallows, line 88:
```suggestion
            status &= provider != null;

            if (status)
            {
                in.wrap(buffer, index, length);
                provider.createReader(in).readValue();
            }
```

jfallows (changes requested):

jfallows (approved):

PR #893, opened 2024-04-02
Support karapace catalog

bmaidics:


runtime/catalog-karapace/src/main/moditect/module-info.java
jfallows, line 1: This one needs a module name change too.

jfallows (commented):
Please revert the `common` changes.

jfallows (approved):

PR #901, opened 2024-04-08
Update helm chart

### LINKED ISSUE #884: Helm chart QoL improvements
Helm charts can have a wide range of usage patterns that are exposed through the `values.yaml` file. There are some common components and helm templates to add that can help deploy zilla in situations devs are more familiar with.

ENV vars are the main wait to dynamically modify the zilla.yaml and giving the user more options to configure them will be much more helpful. If we can just replicate what the Grafana [helm chart](https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L446) has it would make any deployment pattern easier. 

**Acceptance criteria**
 - use the structure of chart templates and values found in the default `helm create zilla` [template](https://helm.sh/docs/helm/helm_create/)
   - unique name use values generated in the `_helpers.tpl` file
   - keep optional ingress and service account
   - keep `values.yaml` structure and similar as possible with comments
 - remove license text from the `NOTES.txt` file
 - add comments in `values.yaml` for added parameters
 - add more ways to configure ENV vars

attilakreiner:
## Description

This change updates the zilla helm charts.

Fixes #884


cloud/helm-chart/src/main/helm/zilla/values.yaml
```diff
@@ -0,0 +1,191 @@
...
+
+args: ["start", "-v", "-e"]
+extraEnv: []
+configPath: /etc/zilla
+configMaps: {}
+secrets: {}
+
+## Extra environment variables that will be pass onto deployment pods
+##
+## to provide grafana with access to CloudWatch on AWS EKS:
```
vordimous, line 63: This and a few other places still reference `grafana`.

attilakreiner: yep, thx, fixed

attilakreiner: oops, I see there is some more to fix, thx for pointing out @vordimous

attilakreiner: done

cloud/helm-chart/src/main/helm/zilla/templates/ingress.yaml
```diff
@@ -0,0 +1,76 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations under the License.
+#
+
+{{ if .Values.ingress.enabled -}}
+{{- $fullName := include "zilla.fullname" . -}}
+{{- $svcPort := .Values.ingress.port -}}
```
vordimous, line 18: Is there a reason this isn't `.Values.ingress.port` and defaulted to `0`?

cloud/helm-chart/src/main/helm/zilla/values.yaml
```diff
@@ -0,0 +1,175 @@
...
+  enabled: false
+  className: ""
+  annotations: {}
+    # kubernetes.io/ingress.class: nginx
+    # kubernetes.io/tls-acme: "true"
+  hosts:
+    - host: chart-example.local
+      paths:
+        - path: /
+          pathType: ImplementationSpecific
```
vordimous, line 124: I don't think this should be enabled by default but I guess it doesn't hurt anything?
```suggestion
  hosts: []
  #  - host: chart-example.local
  #    paths:
  #      - path: /
  #        pathType: ImplementationSpecific
```

cloud/helm-chart/src/main/helm/zilla/templates/ingress.yaml
vordimous, line 18: There is also a slight syntax issue on the first line so maybe there was a copy past mixup
```suggestion
{{- if .Values.ingress.enabled -}}
{{- $fullName := include "zilla.fullname" . -}}
{{- $svcPort := .Values.service.port -}}
```

vordimous (changes requested):
I left a few comments and found one [missing file](https://github.com/grafana/helm-charts/blob/1e803f80bcb8ca5522824e5563c9cb804c564243/charts/grafana/templates/secret-env.yaml#L2) that is needed.

cloud/helm-chart/src/main/helm/zilla/templates/ingress.yaml
attilakreiner: discussed on slack

cloud/helm-chart/src/main/helm/zilla/templates/ingress.yaml
```diff
@@ -0,0 +1,76 @@
...
+# this file except in compliance with the License.  You may obtain a copy of the
+# License at
+#
+#   https://www.aklivity.io/aklivity-community-license/
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations under the License.
+#
```
vordimous, line 14: Most helm charts don't include license text in the templates directory since these files are to be parsed by the helm template parser. If we do need to add these then we need to use the proper [comment format](https://github.com/bitnami/charts/blob/main/bitnami/kafka/templates/secrets.yaml#L1).

vordimous (changes requested):

cloud/helm-chart/src/main/helm/zilla/templates/ingress.yaml
vordimous: @jfallows, can we remove the license text from everything under `templates/`? The license will still exist in the `Chart.yaml`, meaning when Helm makes the full deployment artifact, it will have the license comment in it.

cloud/helm-chart/src/main/helm/zilla/values.yaml
attilakreiner: as discussed, prefer to keep it uncommented, it's mandatory if we enable the ingress

cloud/helm-chart/src/main/helm/zilla/templates/ingress.yaml
attilakreiner: I believe `#` comments are valid yaml. Including or excluding the license text is a question to @jfallows, I guess.

vordimous: Correct, the `Chart.yaml` file is a valid yaml file, but everything in the `templates` folder gets parsed by the golang templating engine so the comments need to be in that format.

attilakreiner: will remove the comments, as agreed

cloud/helm-chart/src/main/helm/zilla/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    {{- include "zilla.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": test
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['wget']
+      args: ['{{ include "zilla.fullname" . }}:{{ .Values.service.port }}']
```
vordimous, line 14: I think this won't work because we don't define one `{{ .Values.service.port }}` right?

cloud/helm-chart/src/main/helm/zilla/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
...
+{{- end }}
+{{- else if contains "NodePort" .Values.service.type }}
+  export NODE_PORT=$(kubectl get --namespace {{ .Release.Namespace }} -o jsonpath="{.spec.ports[0].nodePort}" services {{ include "zilla.fullname" . }})
+  export NODE_IP=$(kubectl get nodes --namespace {{ .Release.Namespace }} -o jsonpath="{.items[0].status.addresses[0].address}")
+  echo http://$NODE_IP:$NODE_PORT
+{{- else if contains "LoadBalancer" .Values.service.type }}
+     NOTE: It may take a few minutes for the LoadBalancer IP to be available.
+           You can watch the status of by running 'kubectl get --namespace {{ .Release.Namespace }} svc -w {{ include "zilla.fullname" . }}'
+  export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ include "zilla.fullname" . }} --template "{{"{{ range (index .status.loadBalancer.ingress 0) }}{{.}}{{ end }}"}}")
+  echo http://$SERVICE_IP:{{ .Values.service.port }}
```
vordimous, line 16: Perhaps we need to print the range of `{{ .Values.service.ports }}` here?

vordimous (changes requested):

cloud/helm-chart/src/main/helm/zilla/templates/tests/test-connection.yaml
attilakreiner: fixed

cloud/helm-chart/src/main/helm/zilla/templates/NOTES.txt
attilakreiner: fixed

vordimous (approved):

cloud/helm-chart/pom.xml
```diff
@@ -44,7 +44,8 @@
               <tpl>SCRIPT_STYLE</tpl>
             </mapping>
             <excludes>
-              <exclude>src/main/helm/artifacthub-repo.yml</exclude>
+              <exclude>src/main/helm/zilla/artifacthub-repo.yml</exclude>
+              <exclude>src/main/helm/zilla/**</exclude>
```
jfallows, line 48: The second one is sufficient to encompass the first one, right?

jfallows (approved):

cloud/helm-chart/pom.xml
attilakreiner: Yep, true. Here's the fix: https://github.com/aklivity/zilla/pull/912

PR #910, opened 2024-04-11
Use per worker registration for composite namespaces

jfallows:
## Description

Eliminate race condition when registering composite namespaces across multiple workers.


incubator/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/AsyncapiBindingConfig.java
```diff
@@ -235,7 +239,9 @@ private void attachServerClientBinding(
         for (AsyncapiSchemaConfig config : configs)
         {
             Asyncapi asyncapi = config.asyncapi;
-            final NamespaceConfig composite = binding.attach.apply(namespaceGenerator.generate(binding, asyncapi));
+            final NamespaceConfig composite = namespaceGenerator.generate(binding, asyncapi);
+            composite.readURL = binding.readURL;
```
akrambek, line 243: I wonder if we still need this since we don't use anymore readURL

incubator/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/config/OpenapiAsyncapiBindingConfig.java
```diff
@@ -139,8 +143,9 @@ public void attach(
                     (e, n) -> e,
                     Object2ObjectHashMap::new));
 
-        this.composite = binding.attach.apply(namespaceGenerator.generate(binding, openapis,
-            asyncapis, openapiSchemaIdsByApiId::get));
+        this.composite = namespaceGenerator.generate(binding, openapis, asyncapis, openapiSchemaIdsByApiId::get);
+        this.composite.readURL = binding.readURL;
```
akrambek, line 147: the same goes here.

akrambek (dismissed):

incubator/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/AsyncapiBindingConfig.java
jfallows: Yep, per our discussion there is still an `assert` for non-null `readURL` on namespace in engine.
Looking forward to perhaps being able to remove this later when we only use catalog and no longer need direct local file access from bindings, but still needed for now.

incubator/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/config/OpenapiAsyncapiBindingConfig.java
jfallows: see reply above

akrambek (approved):

PR #967, opened 2024-04-18
Use format to construct get openapi operation for async rquest

akrambek:


specs/binding-openapi-asyncapi.spec/src/main/scripts/io/aklivity/zilla/specs/binding/openapi/asyncapi/config/proxy.yaml
```diff
@@ -29,7 +29,7 @@ catalogs:
               name: MIT
           servers:
             - url: http://localhost:9090
-          paths:
+          paths:
```
jfallows, line 32: Unintentional?

jfallows (changes requested):

jfallows (approved):

PR #977, opened 2024-04-24
unify caching across workers to maximize cache hits

### LINKED ISSUE #937: Resiliently handle `karapace` catalog unreachable
As devops, I want `zilla` to be resilient when `karapace` catalog is unreachable.

**Acceptance criteria**
 - resilient retries when catalog unreachable
 - handle catalog not reachable yet on first access
   - models treat missing schema using reject-messages-by-default strategy
   - bindings treat missing schema using reject-streams-by-default strategy; `openapi`, `asyncapi`, `grpc`
 - handle catalog not reachable after already caching a result
   - reuse cached result beyond cache expiration
   - log events indicating behavioral workarounds when unreachable, also when later reachable
   - unify caching across workers to maximize cache hits

**Additional context**
None

aDaemonThread:
Fixes #937


runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -54,11 +57,13 @@ public class KarapaceCatalogHandler implements CatalogHandler
     private final long maxAgeMillis;
     private final KarapaceEventContext event;
     private final long catalogId;
+    private final ConcurrentHashMap<Integer, CompletableFuture<String>> cache;
```
jfallows, line 60:
```suggestion
    private final ConcurrentMap<Integer, CompletableFuture<String>> cachedSchemas;
```

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -69,24 +74,41 @@ public KarapaceCatalogHandler(
         this.maxAgeMillis = config.maxAge.toMillis();
         this.event = new KarapaceEventContext(context);
         this.catalogId = catalogId;
+        this.cache = cache;
```
jfallows, line 77:
```suggestion
        this.cachedSchemas = cachedSchemas;
```

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -69,24 +74,41 @@ public KarapaceCatalogHandler(
...
+                schema = future.get();
+                if (schema != null)
+                {
+                    cache.put(schemaId, future);
+                    schemas.put(schemaId, schema);
+                }
+            }
+            catch (ExecutionException | InterruptedException e)
+            {
+                // TODO: log an event
```
jfallows, line 111: There is still a race here, between `cache.get(schemaId)` and `cache.put(schemaId, future)`.

To solve the race, we need to use one of the methods on `ConcurrentMap` that is not available on `Map`, such as:
```
CompletableFuture<> future = cache.get(schemaId);
if (future == null)
{
    CompletableFuture<> newFuture = new CompletableFuture<String>();
    future = cache.putIfAbsent(schemaId, newFuture);
    if (future == null)
    {
        // send request and complete newFuture
        future = newFuture;
    }
}
assert future != null;
schema = future.get();
if (schema != null)
{
    schemas.put(schemaId, schema);
}
```

runtime/catalog-karapace/src/test/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceIT.java
```diff
@@ -72,7 +73,7 @@ public void shouldResolveSchemaViaSchemaId() throws Exception
             "{\"name\":\"status\",\"type\":\"string\"}]," +
             "\"name\":\"Event\",\"namespace\":\"io.aklivity.example\",\"type\":\"record\"}";
 
-        KarapaceCatalogHandler catalog = new KarapaceCatalogHandler(config, context, 0L);
+        KarapaceCatalogHandler catalog = new KarapaceCatalogHandler(config, context, 0L, new ConcurrentHashMap<>());
```
jfallows, line 76: Suggest adding a convenience constructor that calls `this(config, context, catalogId, new ConcurrentHashMap<>())` instead of requiring the test to handle the complexity.

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
jfallows: Note, we also need to prevent unnecessary concurrent requests to resolve the schemaIds, such as `latest`, same as for resolving the schemas themselves from schemaIds.

jfallows (changes requested):

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -113,6 +135,10 @@ public int resolve(
             {
                 schemaIds.put(checkSum, new CachedSchemaId(System.currentTimeMillis(), schemaId));
             }
+            else if (schemaIds.containsKey(checkSum))
+            {
+                schemaId = schemaIds.get(checkSum).id;
+            }
```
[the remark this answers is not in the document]
aDaemonThread: yes.

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -88,21 +100,27 @@ public String resolve(
...
                 {
-                    String response = sendHttpRequest(MessageFormat.format(SCHEMA_PATH, schemaId));
-                    return response != null ? request.resolveSchemaResponse(response) : null;
-                });
+                    newFuture = CompletableFuture.supplyAsync(() ->
+                    {
+                        String response = sendHttpRequest(MessageFormat.format(SCHEMA_PATH, schemaId));
+                        return response != null ? request.resolveSchemaResponse(response) : null;
+                    });
+                    future = newFuture;
```
jfallows, line 115: I don't think we can reassign `newFuture` here, as the value being lost is the one that already made it into the concurrent map, so if we reassign it then the one in the map would not be completed, agree?

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -119,26 +137,80 @@ public int resolve(
...
+                future = cachedSchemaIds.putIfAbsent(schemaKey, newFuture);
+                if (future == null)
+                {
+                    newFuture = CompletableFuture.supplyAsync(() ->
+                    {
+                        String response = sendHttpRequest(MessageFormat.format(SUBJECT_VERSION_PATH, subject, version));
+                        return new CachedSchemaId(System.currentTimeMillis(),
+                            response != null ? request.resolveResponse(response) : NO_SCHEMA_ID);
+                    });
+                    future = newFuture;
```
jfallows, line 163: Same issue here as above, agree?

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -119,26 +137,80 @@ public int resolve(
...
+                                    return new CachedSchemaId(System.currentTimeMillis(),
+                                        response != null ? request.resolveResponse(response) : NO_SCHEMA_ID);
+                                });
+                                return id;
+                            }
+                            else
+                            {
+                                return existingFuture;
+                            }
+                        });
```
jfallows, line 195: I think we can perhaps simplify the logic by creating the candidate newFuture and calling:
```
CompletableFuture<> newFuture = new CompletableFuture<CachedSchemaId>();
CompletableFuture<> future = cachedSchemaIds.merge(schemaKey, newFuture, (v1, v2) -> v1 != null && !v1.isDone() ? v1 : v2);
if (future == newFuture)
{
    // send request
    // complete future
}
cachedSchemaId  = future.get();
```

jfallows (changes requested):

specs/catalog-karapace.spec/src/main/resources/META-INF/zilla/karapace.idl
```diff
@@ -18,7 +18,9 @@ scope karapace
     {
         enum KarapaceEventType (uint8)
         {
-            REMOTE_ACCESS_REJECTED (1)
+            REMOTE_ACCESS_REJECTED (1),
+            FUTURE_COMPLETED_EXCEPTIONALLY (2),
+            STALE_SCHEMA_SERVED (3)
```
jfallows, line 23: How do these differ in the information they convey to the log observer?

jfallows (changes requested):

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/CachedSchema.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class CachedSchema
+{
+    public static final String SCHEMA_PLACEHOLDER = "schema";
+    public static final CachedSchema IN_PROGRESS = new CachedSchema(SCHEMA_PLACEHOLDER, new AtomicInteger(0));
+
+    public String schema;
+    public AtomicInteger event;
+
```
jfallows, line 26: If these are both unmodifiable, then they should be marked `final` too.

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/CachedSchema.java
```diff
@@ -0,0 +1,34 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+package io.aklivity.zilla.runtime.catalog.karapace.internal;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class CachedSchema
+{
+    public static final String SCHEMA_PLACEHOLDER = "schema";
+    public static final CachedSchema IN_PROGRESS = new CachedSchema(SCHEMA_PLACEHOLDER, new AtomicInteger(0));
```
jfallows, line 22: No need to explicitly pass `0` to `AtomicInteger` constructor.

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/CachedSchema.java
```diff
@@ -0,0 +1,34 @@
...
+    public String schema;
+    public AtomicInteger event;
+
+    public CachedSchema(
+        String schema,
+        AtomicInteger event)
+    {
+        this.schema = schema;
+        this.event = event;
+    }
```
jfallows, line 33: Does it make sense to have a convenience constructor with just `schema` defauting to `new AtomicInteger()`.

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/CachedSchema.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class CachedSchema
+{
+    public static final String SCHEMA_PLACEHOLDER = "schema";
+    public static final CachedSchema IN_PROGRESS = new CachedSchema(SCHEMA_PLACEHOLDER, new AtomicInteger(0));
+
+    public String schema;
+    public AtomicInteger event;
```
jfallows, line 25: Not sure why this is called `event`.
Isn't this just tracking the number of `retries`, which is normally `0` when everything is working normally?

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/CachedSchemaId.java
```diff
@@ -14,26 +14,39 @@
...
 {
-    public static final CachedSchemaId IN_PROGRESS = new CachedSchemaId(Long.MAX_VALUE, NO_SCHEMA_ID);
+    public static final int PLACEHOLDER_SCHEMA_ID = -1;
+    public static final CachedSchemaId IN_PROGRESS = new CachedSchemaId(Long.MAX_VALUE, PLACEHOLDER_SCHEMA_ID,
+        new AtomicInteger(0), Long.MAX_VALUE);
 
     public long timestamp;
     public int id;
+    public AtomicInteger event;
+    public long retry;
```
jfallows, line 28: Should some or all of these be `final`?

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/CachedSchemaId.java
```diff
@@ -14,26 +14,39 @@
...
         long timestamp,
-        int id)
+        int id,
+        AtomicInteger event,
+        long retry)
     {
         this.timestamp = timestamp;
         this.id = id;
+        this.event = event;
+        this.retry = retry;
```
jfallows, line 39: I think this is saying `retryAfter`, right?

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -84,54 +88,77 @@ public KarapaceCatalogHandler(
...
+                        if (response == null && eventStatus.getAndIncrement() == 0)
+                        {
+                            event.unretrievableSchemaId(catalogId, schemaId);
+                        }
+                        else if (response != null && eventStatus.getAndSet(0) > 0)
+                        {
+                            event.retrievableSchemaId(catalogId, schemaId);
+                        }
+                        newFuture.complete(new CachedSchema(response != null ? request.resolveSchemaResponse(response) : null,
+                            eventStatus));
```
jfallows, line 142: We are repeating the check for response == or != null several times here, which makes the code a bit harder to read.

Suggest we have one block
```
if (response == null)
{
    if (retryAttempts.getAndIncrement() == 0)
    {
        event.unretrievableSchemaId(catalogId, schemaId);
    }
    newFuture.complete(new CachedSchema(null, retryAttempts));
}
else
{
    if (retryAttempts.getAndSet(0) > 0)
    {
        event.retrievableSchemaId(catalogId, schemaId);
    }
    newFuture.complete(new CachedSchema(request.resolveSchemaResponse(response), retryAttempts));
}
```

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceEventContext.java
```diff
@@ -39,40 +40,43 @@ public class KarapaceEventContext
...
+        this.staleSchemaID = context.supplyEventId("catalog.karapace.unretrievable.schema.subject.version.stale.schema");
+        this.unretrievableSchemaId = context.supplyEventId("catalog.karapace.unretrievable.schema.id");
+        this.retrievableSchemaSubjectVersionId = context.supplyEventId("catalog.karapace.retrievable.schema.subject.version");
+        this.retrievableSchemaId = context.supplyEventId("catalog.karapace.retrievable.schema.id");
         this.eventWriter = context.supplyEventWriter();
         this.clock = context.clock();
     }
 
-    public void remoteAccessRejected(
+    public void unretrievableSchemaSubjectVersion(
```
jfallows, line 64: I think it might be more readable if these methods were named `onXxx` instead of `xxx`.

So in the calling context it would be something like `events.onUnretrievableSchemaSubjectVersion(...)` (not `event...`), wdyt?

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -271,15 +334,10 @@ private String sendHttpRequest(
...
-            if (!success)
-            {
-                event.remoteAccessRejected(catalogId, httpRequest, httpResponse.statusCode());
-            }
             return responseBody;
         }
         catch (Exception ex)
         {
-            event.remoteAccessRejected(catalogId, httpRequest, 0);
         }
```
jfallows, line 341: Rather than having an empty catch block, I think it would better document the intent if `responseBody` is hoisted outside the `try/catch` and explicitly set to `null` in the `catch` before `return responseBody;` at the end of the method.

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -45,6 +46,9 @@ public class KarapaceCatalogHandler implements CatalogHandler
     private static final String REGISTER_SCHEMA_PATH = "/subjects/{0}/versions";
     private static final int MAX_PADDING_LENGTH = 5;
     private static final byte MAGIC_BYTE = 0x0;
+    public static final long RESET_RETRY_DELAY_MS_DEFAULT = 0L;
+    public static final long RETRY_INITIAL_DELAY_MS_DEFAULT = 1000L;
+    public static final int RETRY_MULTIPLER_DEFAULT = 2;
```
jfallows, line 51: Can these be `private`?

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -151,45 +178,81 @@ public int resolve(
...
+                        {
+                            retry = RETRY_INITIAL_DELAY_MS_DEFAULT;
+                            event.unretrievableSchemaSubjectVersion(catalogId, subject, version);
+                        }
+
+                        if (cachedSchemaId != null && cachedSchemaId.retry != RESET_RETRY_DELAY_MS_DEFAULT)
+                        {
+                            long current = cachedSchemaId.retry;
+                            long update = current * RETRY_MULTIPLER_DEFAULT;
+                            retry = update < maxAgeMillis ? update : current;
```
jfallows, line 221: This logic is trying to prevent the value from increasing if it would end up larger than `maxAgeMillis`, but that means `maxAgeMillis` is okay too.

Suggest it is easier to follow if you change to max out at `maxAgeMillis` instead of `current`.

```suggestion
retryAfter = Math.min(cachedSchemaId.retryAfter << 1, maxAgeMillis);
```

runtime/catalog-karapace/src/test/java/io/aklivity/zilla/runtime/catalog/karapace/internal/EventIT.java
```diff
@@ -64,22 +64,17 @@ public void shouldLogEvents() throws Exception
     }
 
     @Test
-    public void shouldLogFutureCompletedExceptionally()
+    public void shouldLogRemoteRegistryEvents()
     {
         EngineContext context = mock(EngineContext.class);
         when(context.clock()).thenReturn(Clock.systemUTC());
         when(context.supplyEventWriter()).thenReturn(mock(MessageConsumer.class));
         KarapaceEventContext event = new KarapaceEventContext(context);
```
jfallows, line 72: Suggest we call this `events` instead of `event`.

jfallows (changes requested):

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -97,21 +174,98 @@ public int resolve(
...
+            }
+            assert future != null;
+            try
+            {
+                cachedSchemaId = future.get();
+                schemaId = cachedSchemaId.id;
+                if (schemaId != NO_SCHEMA_ID)
+                {
+                    schemaIds.put(schemaKey, cachedSchemaId);
+                    if (cachedSchemaId.retryAttempts.compareAndSet(1, 2))
```
jfallows, line 260: This doesn't look right to me, because the value of `retryAttempts` should be independent of serving stale schema, right?

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceCatalogHandler.java
```diff
@@ -190,22 +344,18 @@ private String sendHttpRequest(
...
         {
             HttpResponse<String> httpResponse = client.send(httpRequest, HttpResponse.BodyHandlers.ofString());
             boolean success = httpResponse.statusCode() == 200;
-            String responseBody = success ? httpResponse.body() : null;
-            if (!success)
-            {
-                event.remoteAccessRejected(catalogId, httpRequest, httpResponse.statusCode());
-            }
-            return responseBody;
+            responseBody = success ? httpResponse.body() : null;
```
jfallows, line 352: We don't really need `success` any more, right?
```suggestionif (
responseBody = httpResponse.statusCode() == 200 ? httpResponse.body() : null;
```

runtime/catalog-karapace/src/test/java/io/aklivity/zilla/runtime/catalog/karapace/internal/EventIT.java
```diff
@@ -56,4 +62,19 @@ public void shouldLogEvents() throws Exception
...
+        when(context.clock()).thenReturn(Clock.systemUTC());
+        when(context.supplyEventWriter()).thenReturn(mock(MessageConsumer.class));
+        KarapaceEventContext events = new KarapaceEventContext(context);
+        events.onUnretrievableSchemaSubjectVersion(0L, "subject", "version");
+        events.onRetrievableSchemaSubjectVersion(0L, "subject", "version");
+        events.onUnretrievableSchemaSubjectVersionStaleSchema(0L, "subject", "version", 1);
+        events.onUnretrievableSchemaId(0L, 1);
+        events.onRetrievableSchemaId(0L, 1);
+    }
+
```
jfallows, line 79: This should be done as an IT with scripts verifying each different possible behavior of the remote Karapace system.

Is anything missing from `test` binding that would let the `karapace` catalog be tested that way?

Note: please see `filesystem` catalog EventIT.

runtime/catalog-karapace/src/test/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceIT.java
jfallows, line 1: Prefer this IT done via scripts not mocks.

Note: if using mocks then it is a unit test, not an integration test - unit tests with mocks are not stable after a refactor of the code as they also need to be changed, whereas integration tests remain stable, providing better confidence after changing implementation.

jfallows (changes requested):

runtime/catalog-karapace/src/test/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceIT.java
jfallows: There should be no need to drive any direct calls to `KarapaceCatalogHandler` via code and mocks.

Instead, this should be driven by `test` binding in `zilla.yaml` for the test. It may be necessary to add support for expected `schema` in `test` binding options so that it can verify the retrieved schema matches expectations.

Note: this would be for `test` binding only, as other bindings have no need to validate expectations.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -246,9 +250,16 @@ private void onInitialBegin(
...
                 {
-                    catalog.resolve(0);
+                    if (catalog.subject != null && catalog.version != null)
+                    {
+                        handler.resolve(catalog.subject, catalog.version);
+                    }
+                    else
+                    {
+                        handler.resolve(catalog.id);
+                    }
```
jfallows, line 262: This is excellent. As mentioned above, we may need to add support for verifying the expected schema here, then reject the initial begin with a reset if the schema doesn't match expectations, as that will cause the IT to fail.

The expected schemas may be a separate test binding option, independent of cataloged so as not to interfere with that generic definition used by other bindings that don't need to validate expectations.

incubator/catalog-filesystem.spec/src/main/scripts/io/aklivity/zilla/specs/catalog/filesystem/config/event.yaml
```diff
@@ -36,6 +36,7 @@ bindings:
     type: test
     kind: server
     options:
-      catalogs:
-        - catalog0
+      catalog:
+        catalog0:
+          - id: 1
```
jfallows, line 41: It would seem more readable and descriptive of the intent to put `subject: not-subject1` instead of `id: 1` here, agree?

jfallows (changes requested):

specs/catalog-karapace.spec/src/main/scripts/io/aklivity/zilla/specs/catalog/karapace/config/resolve/schema/id/retry/zilla.yaml
```diff
@@ -0,0 +1,46 @@
...
+  net0:
+    type: test
+    kind: server
+    options:
+      catalog:
+        catalog0:
+          - id: 9
+      catalog-assertion:
+        - schema: null
+        - schema: '{"fields":[{"name":"id","type":"string"},{"name":"status","type":"string"}],"name":"Event","namespace":"io.aklivity.example","type":"record"}'
```
jfallows, line 45: This looks great!

Suggest we generalize the concept of `assertions` as we'll potentially need that later for vault, etc.

```suggestion
      assertions:
        catalog:
          catalog0:
            - schema: null
            - schema: '{"fields":[{"name":"id","type":"string"},{"name":"status","type":"string"}],"name":"Event","namespace":"io.aklivity.example","type":"record"}'
```

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/binding/test.schema.patch.json
```diff
@@ -57,6 +57,29 @@
...
+                                    {
+                                        "id":
+                                        {
+                                            "type": "integer"
+                                        },
+                                        "schema":
+                                        {
+                                            "type": ["string", "null"]
+                                        },
+                                        "interval":
```
jfallows, line 76: Perhaps `delay` instead of `interval` since it is not repeating.

specs/catalog-karapace.spec/src/main/scripts/io/aklivity/zilla/specs/catalog/karapace/config/resolve/schema/id/zilla.yaml
```diff
@@ -0,0 +1,33 @@
...
+bindings:
+  net0:
+    type: test
+    kind: server
+    options:
+      catalog:
+        catalog0:
+          - id: 9
+      catalog-assertion:
+        - schema: '{"fields":[{"name":"id","type":"string"},{"name":"status","type":"string"}],"name":"Event","namespace":"io.aklivity.example","type":"record"}'
```
jfallows, line 32: Is ti possible to use yaml syntax magic to let this be pretty printed json in the file, but compact format when parsed?

jfallows (changes requested):

specs/catalog-karapace.spec/src/main/scripts/io/aklivity/zilla/specs/catalog/karapace/config/resolve/schema/id/zilla.yaml
aDaemonThread: I wasn't able to find a way to implement this feedback, tried using `> (Folded Block Scalar)` instead of `|`. But the result was still varying from the expected Schema string.

jfallows (approved):

PR #978, opened 2024-04-24
Fix multiple exporters issue

attilakreiner:
## Description

This change fixes an issue, so that multiple exporters can run correctly simultaneously.

### DIFF
--- a/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/layouts/EventsLayout.java
+++ b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/layouts/EventsLayout.java
@@ -30,7 +30,9 @@ import java.nio.file.StandardCopyOption;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.LinkedList;
 import java.util.List;
+import java.util.Queue;
 
 import org.agrona.CloseHelper;
 import org.agrona.DirectBuffer;
@@ -106,7 +108,7 @@ public final class EventsLayout implements AutoCloseable
             rethrowUnchecked(ex);
         }
         buffer = createRingBuffer(path, capacity);
-        accessors.forEach(a -> a.setBufferSpy(createRingBufferSpy()));
+        accessors.forEach(a -> a.addNextBufferSpy(createRingBufferSpy()));
     }
 
     private RingBufferSpy createRingBufferSpy()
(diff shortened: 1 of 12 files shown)


runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/stream/StdoutEventsStream.java
```diff
@@ -43,14 +45,15 @@ public StdoutEventsStream(
...
+        this.stdoutExporterTypeId = context.supplyTypeId(StdoutExporter.NAME);
         this.readEvent = context.supplyEventReader();
         this.formatter = context.supplyEventFormatter();
         this.out = out;
     }
 
     public int process()
     {
-        return readEvent.read(this::handleEvent, 1);
+        return readEvent.read(this::handleEvent, stdoutExporterTypeId, 1);
```
jfallows, line 56: This approach breaks if we have 2 exporter instances of the same type.

Let's instead remove the complexity of creating multiple buffer spies in the events layout requiring the change to the interface, and instead create a new spy when a new event reader is supplied. That is the signal to us that we need independent spy because we have an independent event reader, which needs to keep track of its own progress through the events.

runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterContext.java
```diff
@@ -73,4 +73,10 @@ public MessageReader supplyEventReader()
     {
         return context.supplyEventReader();
     }
+
+    public int supplyTypeId(
+        String name)
+    {
+        return context.supplyTypeId(name);
+    }
```
jfallows, line 81: We can revert this change when we create a new spy per event reader.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/layouts/EventsLayout.java
```diff
@@ -110,7 +106,15 @@ private void rotateFile()
             rethrowUnchecked(ex);
         }
         buffer = createRingBuffer(path, capacity);
-        bufferSpy = createRingBufferSpy(path);
+        accessors.forEach(a -> a.setBufferSpy(createRingBufferSpy()));
```
jfallows, line 109: This doesn't seem like the right timing to move all the accessors to the latest rotated buffer.

It makes sense to rotate the events layout itself during write, but each event accessor would need to independently determine when they have reached the end of the current buffer and move to the next one (even if it's not the latest).

jfallows (changes requested):

jfallows (approved):

PR #979, opened 2024-04-24
MqttKafka publish intern fix

bmaidics:


runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishMetadata.java
```diff
@@ -93,20 +93,35 @@ public static final class KafkaOffsetMetadata
...
         }
 
         KafkaOffsetMetadata(
+            String topic,
             long producerId,
             short producerEpoch,
             IntArrayList packetIds)
         {
             this.sequence = 1;
+            this.topic = topic;
```
jfallows, line 124: I think we'll have the same issue unless this is changed, agree?
```suggestion
            this.topic = topic.intern();
```

jfallows (changes requested):

jfallows (approved):

PR #982, opened 2024-04-25
Support multiple specs in asyncapi binding

bmaidics:


runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/AsyncapiProtocol.java
```diff
@@ -113,54 +111,41 @@ protected <C> CatalogedConfigBuilder<C> injectJsonSchemas(
...
-                result = server.url();
-                break;
+                for (String securitySchemeName : asyncapi.components.securitySchemes.keySet())
+                {
+                    String guardType = asyncapi.components.securitySchemes.get(securitySchemeName).bearerFormat;
+                    //TODO: change when jwt support added for mqtt in asyncapi
+                    //if ("jwt".equals(guardType))
+                    //{
+                    //    result.put(securitySchemeName, guardType);
+                    //}
```
jfallows, line 146: Is this TODO still relevant?

jfallows (changes requested):

runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/AsyncapiProtocol.java
bmaidics: I think yes. As asyncapi doesn't have JWT support added for MQTT yet and we use this to determine if we need to add the mqtt authorization options for the server. If we filter for "jwt" we'll never add those options.

jfallows (approved):

PR #990, opened 2024-05-01
`tls`  `client/server/echo` handshake benchmark 

### LINKED ISSUE #961: Integrate JMH into `tls` binding
Add microbenchmark for TLS handshake, including reconnect with session resumption.

Note: improved performance for session resumption may require #956.

akrambek:
`tls`  `client/server/echo` handshake benchmark 

Fixes #961


buffers0
jfallows, line 1: Please remove buffers0 from the PR. 😄

runtime/binding-tls/src/test/java/io/aklivity/zilla/runtime/binding/tls/internal/bench/TlsWorker.java
```diff
@@ -0,0 +1,728 @@
...
+
+        @Override
+        public long signalAt(
+            long timeMillis,
+            int signalId,
+            IntConsumer handler)
+        {
+            handler.accept(signalId);
+
+            return System.currentTimeMillis();
```
jfallows, line 585: I think we don't need to return current time here, perhaps `NO_CANCEL_ID` instead since these are executed on stack and therefore not cancelable?

Same for other `signalXXX` methods.

jfallows (changes requested):

jfallows (approved):

PR #992, opened 2024-05-02
Add MQTT client authentication

bmaidics:


runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttBindingConfig.java
```diff
@@ -226,6 +234,30 @@ private Function<String, String> asAccessor(
             });
         }
 
+        return injector;
```
jfallows, line 237: The local changes in this method should be reverted, right?

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttBindingConfig.java
```diff
@@ -226,6 +234,30 @@ private Function<String, String> asAccessor(
...
+            {
+                String result = null;
+                if (connect != null)
+                {
+                    result = config.pattern.replace("{credentials}", connect);
+                }
+                return result;
+            });
+        }
+
```
jfallows, line 260: This one should rename `accessor` to `injector`, right?
I think this was the intention of the change in `asAccessor` just not applied to this method.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttBindingConfig.java
```diff
@@ -226,6 +234,30 @@ private Function<String, String> asAccessor(
...
+        MqttCredentialsConfig credentials)
+    {
+        Function<String, String> accessor = DEFAULT_CREDENTIALS;
+        List<MqttPatternConfig> connectPatterns = credentials.connect;
+
+        if (connectPatterns != null && !connectPatterns.isEmpty())
+        {
+            MqttPatternConfig config = connectPatterns.get(0);
+
+            accessor = orElseIfNull(accessor, connect ->
```
jfallows, line 250: Do we actually need `orElseIfNull` here or above?

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -176,7 +176,7 @@ public long reauthorize(
                 .orElse(null);
 
             JwtSessionStore sessionStore = supplySessionStore(contextId);
-            session = sessionStore.supplySession(subject, roles);
+            session = sessionStore.supplySession(subject, credentials, roles);
```
jfallows, line 179: I think we only need to key the session by subject and roles, but retain the credentials as a field below, like other session attributes.

This will also ensure that latest credentials are in effect after reauthorizing an existing session, as older credentials may have expired.

jfallows (changes requested):

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/config/MqttBindingConfig.java
```diff
@@ -223,7 +223,7 @@ private Function<String, String> asAccessor(
                 Pattern.compile(config.pattern.replace("{credentials}", "(?<credentials>[^\\s]+)"))
                     .matcher("");
 
-            injector = orElseIfNull(injector, connect ->
+            accessor = orElseIfNull(accessor, connect ->
```
jfallows, line 226: Do we need `orElseIfNull` for the `accessor` either?
Suspect not, so let's update and remove `orElseIfNull` as well, to eliminate dead code.

jfallows (changes requested):

jfallows (approved):

PR #1005, opened 2024-05-07
Support multiple specs in openapi binding

bmaidics:


jfallows (commented):
We need an IT that configures the binding with multiple specs and verifies correctness during the build, not just via external end-to-end testing, right?

jfallows (approved):

Issue #1012, opened 2024-05-08
Option to structure Event message format as JSON


vordimous:
**Describe the desired outcome from the user's perspective**
As a DevOps observability admin, I want to parse all parts of log messages programmatically. Having a static structure for log messages enables static text parsing methods. Variable parts of log messages should be configurable to output as JSON instead of a flat test string.

**Acceptance criteria**
 - The `stdout` exporter has the option to print raw text or unformatted JSON.
 - The `otlp` exporter can send the Event message in the `body` as a flat string or JSON string.

**Additional context**
[Open Telemetry Logs Data model `body` field](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-body)

jfallows:
While building the `zilla logs` command (#2131 / PR #2136), we ended up implementing the same idea for its own output — `zilla logs --format text|json` — and it seems worth reconciling the naming/defaults here before this issue is picked up, so the CLI and the `stdout` exporter don't end up with divergent conventions for the same underlying event-formatting concern.

What we landed on for `zilla logs`:
- `--format text|json`, defaulting to `text` (the pre-existing plain-line output), so existing usage/scripts are unaffected unless they opt in.
- JSON is emitted as JSON Lines (one object per event), with fields `namespace`, `timestamp` (epoch millis, not a formatted string — friendlier for `jq` filtering/sorting), `traceId` (zero-padded hex **string**, not a JSON number — a trace ID is an unsigned 64-bit value and can exceed the safe-integer range for `jq`/JS consumers, so encoding it as a number risks silent precision loss), `event`, and `message`.
- Rendered via `jakarta.json` (`Json.createObjectBuilder()...build().toString()`) for correct escaping, per this repo's "no Jackson, use jakarta.json" convention.

Two things from the issue description that might be worth revisiting in light of that:
1. **Naming** — this issue calls the non-JSON option "raw"; we used "text" for `zilla logs`. Suggest picking one term and using it consistently across the `stdout` exporter's `options.format` and `zilla logs --format` so they read as the same concept.
2. **Default** — this issue proposes JSON as the new default for the `stdout` exporter. That's a behavior change for every existing config using `type: stdout` with no `format` specified (parsers/dashboards built against the current line format would break silently on upgrade). We defaulted `zilla logs` to `text`/`raw` for that reason. Worth confirming that tradeoff is intentional here too, or whether `raw`/`text` should stay the default with JSON opt-in, at least for `stdout`.

[...]

PR #1029, opened 2024-05-15
feat: replace static event name with dynamic based on event id

### LINKED ISSUE #1013: Use full Event ID and the event name
Log messages can use as much context as possible. This will help observability admins filter and track information more easily.

Update the log `message` fields to be more human-readable and self-descriptive. 

When logging Events, use the full event ID formatted as the event name. `model.protobuf.validation.failed` becomes `MODEL_PROTOBUF_VALIDATION_FAILED`.

vordimous:
## Description

This will use the defined internal `eventId` to format an observability-friendly in an all-caps snake case.

Fixes #1013

docs PR: 
- https://github.com/aklivity/zilla-docs/pull/230


runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/stream/StdoutEventsStream.java
```diff
@@ -61,8 +62,9 @@ private void handleEvent(
     {
         final EventFW event = eventRO.wrap(buffer, index, index + length);
         String qname = context.supplyQName(event.namespacedId());
+        String eventName = context.supplyLocalName(event.id()).replace('.', '_').toUpperCase();
```
jfallows, line 65: Looks like this code is going to repeat in two ways.
 - allocate strings for the same event id as multiple events are logged
 - same code in different exporters

Given, that we want these event names to be consistent, we should move the implementation into the engine for reuse and handle the caching of event names there too.

In engine `EngineContext`, let's add
```
public String supplyEventName(
  int eventId);
```
Then in `EngineWorker` (which implements `EngineContext`.
```
private final Int2ObjectHashMap<String> eventNames;

...(initialize in constructor)...

@Override
public String supplyEventName(
  int eventId)
{
    return eventNames.computeIfAbsent(eventId, this::computeEventName);
}

private String computeEventName(
  int eventId)
{
    return supplyLocalName(eventId).replace('.', '_').toUpperCase();
}
```

Then adjust `StdoutExporterContext` to expose `supplyEventName(int eventId)` _instead_ of `supplyLocalName` (now unused), and the above code becomes:

```
String eventName = context.supplyEventName(event.id());
```

runtime/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/serializer/EventReader.java
```diff
@@ -76,10 +76,11 @@ private void handleEvent(
         eventJson.add(TIME_UNIX_NANO, nanos);
         eventJson.add(OBSERVED_TIME_UNIX_NANO, nanos);
         String qname = context.supplyQName(event.namespacedId());
+        String eventName = context.supplyLocalName(event.id()).replace('.', '_').toUpperCase();
```
jfallows, line 79:
```suggestion
        String eventName = context.supplyEventName(event.id());
```

runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterContext.java
```diff
@@ -64,6 +64,12 @@ public String supplyQName(
         return context.supplyQName(namespacedId);
     }
 
+    public String supplyLocalName(
+            long namespacedId)
+    {
+        return context.supplyLocalName(namespacedId);
+    }
+
```
jfallows, line 72:
```suggestion
    public String supplyEventName(
            int eventId)
    {
        return context.supplyEventName(eventId);
    }

```

jfallows (changes requested):

attilakreiner:
LGTM apart from the test failures. It looks like there is test code (`EchoWorker` and `TlsWorker`) that also implement the `interface EngineContext`, hence the compile error.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/config/client.event.yaml
```diff
@@ -24,7 +24,8 @@ telemetry:
         events:
           - qname: test.app0
             id: binding.kafka.api.version.rejected
-            message: API_VERSION_REJECTED 32 0
+            name: BINDING_KAFKA_API_VERSION_REJECTED
+            message: 32 0
```
jfallows, line 28: Do you think we should map these API keys to the human readable equivalents for better readability?
See https://kafka.apache.org/protocol.html#protocol_api_keys

```suggestion
            message: DESCRIBE_CONFIGS (32) version 0
```

specs/model-avro.spec/src/main/scripts/io/aklivity/zilla/specs/model/avro/config/event.yaml
```diff
@@ -23,7 +23,8 @@ telemetry:
         events:
           - qname: test.net0
             id: model.avro.validation.failed
-            message: VALIDATION_FAILED java.io.EOFException
+            name: MODEL_AVRO_VALIDATION_FAILED
+            message: java.io.EOFException
```
jfallows, line 27: We'll need to clean this up as well, to eliminate the exposure of Java internals.

jfallows (changes requested):

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/config/client.event.yaml
vordimous: Yes definitely. Is there an enum or library with them already or do we need to map them manually?

vordimous: I just talked with @attilakreiner about the general design of these message outputs and we might want to address the human readability for all of them. Particulary since we are planning to allow a `json` format option. This will help with the `MODEL_AVRO_VALIDATION_FAILED` and `BINDING_TLS_TLS_FAILED` events. 

- https://github.com/aklivity/zilla/issues/1012

specs/model-avro.spec/src/main/scripts/io/aklivity/zilla/specs/model/avro/config/event.yaml
vordimous: Note: This is a valid exception `message` returned from the `AvroRuntimeException` class.

vordimous:
followup changes:

- `MqttKafkaResetMqttConnectionExFW` should include more properties about why a connection was reset, topic names, etc
- Remove internal only values from the event messages, ex: identity in the `MqttEventContext.onClientConnected` event
- more precisely handle validation exceptions by not passing all exception messages ex: catch (IOException | AvroRuntimeException ex)
- include the exception message for the TLS binding error events.
- `KafkaEventFormatter` needs a map of all ApiKey and ApiVersion codes to render the text version that is relevant to the user

specs/binding-mqtt-kafka.spec/src/main/resources/META-INF/zilla/mqtt_kafka.idl
```diff
@@ -40,7 +40,6 @@ scope mqtt_kafka
 
         struct MqttKafkaResetMqttConnectionEx extends core::stream::Extension
         {
-            string16 reason;
```
jfallows, line 43: This event type structure name is out of date. Let's rename it to `MqttKafkaNonCompactSessionsTopicEx`, for consistency with other components.

Then let's also add the topic name used by the binding when recording the event, and include that topic name in the output when formatting the message.

specs/guard-jwt.spec/src/main/scripts/io/aklivity/zilla/specs/guard/jwt/config/event.yaml
```diff
@@ -23,7 +23,8 @@ telemetry:
         events:
           - qname: test.net0
             id: guard.jwt.authorization.failed
-            message: AUTHORIZATION_FAILED user
+            name: GUARD_JWT_AUTHORIZATION_FAILED
+            message: No active session found for token identity (user).
```
jfallows, line 27: This message seems to imply we are looking for matching state in zilla and couldn't find it.

However, I think this event is logged when an attempt to authorize the user (by validating the token) fails, with no implication of state expected to be found at zilla.

Let's reword to capture the underlying behavior without implying expectation to find server-side state, agree?

specs/model-avro.spec/src/main/scripts/io/aklivity/zilla/specs/model/avro/config/event.yaml
```diff
@@ -23,7 +23,8 @@ telemetry:
         events:
           - qname: test.net0
             id: model.avro.validation.failed
-            message: VALIDATION_FAILED java.io.EOFException
+            name: MODEL_AVRO_VALIDATION_FAILED
+            message: A message payload failed validation. java.io.EOFException.
```
jfallows, line 27: Please remove exception class name from message.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtEventFormatter.java
```diff
@@ -49,7 +47,7 @@ public String format(
         case AUTHORIZATION_FAILED:
         {
             JwtAuthorizationFailedExFW ex = extension.authorizationFailed();
-            result = String.format(AUTHORIZATION_FAILED_FORMAT, identity(ex.identity()));
+            result = String.format("No active session found for token identity (%s).", identity(ex.identity()));
```
jfallows, line 50: Please see related comment below in tests.

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceEventFormatter.java
```diff
@@ -58,33 +51,44 @@ public String format(
...
+                    asString(ex.subject()),
+                    asString(ex.version())
+            );
             break;
         }
         case RETRIEVED_SCHEMA_ID:
         {
-            KarapaceRetrievableSchemaIdExFW ex = extension.retrievableSchemaId();
-            result = String.format(RETRIEVED_SCHEMA_ID, ex.schemaId());
+            KarapaceRetrievableSchemaIdExFW ex = extension.retrievedSchemaId();
```
jfallows, line 90: Looks like we need to rename `KarapaceRetrievableSchemaIdExFW` to `KarapaceRetrievedSchemaIdExFW` to align with naming conventions, agree?

runtime/catalog-karapace/src/main/java/io/aklivity/zilla/runtime/catalog/karapace/internal/KarapaceEventFormatter.java
```diff
@@ -58,33 +51,44 @@ public String format(
...
             KarapaceUnretrievableSchemaIdExFW ex = extension.unretrievableSchemaId();
-            result = String.format(UNRETRIEVABLE_SCHEMA_ID, ex.schemaId());
+            result = String.format("Unable to fetch schema id %d.", ex.schemaId());
             break;
         }
         case RETRIEVED_SCHEMA_SUBJECT_VERSION:
         {
-            KarapaceRetrievableSchemaSubjectVersionExFW ex = extension.retrievableSchemaSubjectVersion();
-            result = String.format(RETRIEVED_SCHEMA_SUBJECT_VERSION, asString(ex.subject()), asString(ex.version()));
+            KarapaceRetrievableSchemaSubjectVersionExFW ex = extension.retrievedSchemaSubjectVersion();
```
jfallows, line 81: Looks like we need to rename `KarapaceRetrievableSchemaSubjectVersionExFW ` to `KarapaceRetrievedSchemaSubjectVersionExFW ` to align with naming conventions, agree?

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/TlsEventFormatter.java
```diff
@@ -51,27 +45,27 @@ public String format(
...
         case TLS_KEY_REJECTED:
         {
-            result = KEY_REJECTED_FORMAT;
+            result = "Bad SSL key due to misconfiguration of the server or client SSL certificate and private key.";
             break;
         }
         case TLS_PEER_NOT_VERIFIED:
         {
-            result = PEER_NOT_VERIFIED_FORMAT;
+            result = "The peer's identity was not able to identify itself.";
```
jfallows, line 63:
```suggestion
            result = "The peer's identity could not be verified.";
```

jfallows (changes requested):

specs/model-avro.spec/src/main/scripts/io/aklivity/zilla/specs/model/avro/config/event.yaml
vordimous: `"java.io.EOFException"` is one of the exception messages returned from the `AvroRuntimeException` class. The `error` string passed in the event is set from various exception messages. 

I think this is just a bad example of what kind of errors might come through and I don't know how the tests makes this specific exception happen or how to change it to something else.

origional comment here: https://github.com/aklivity/zilla/pull/1029#discussion_r1624780061

specs/guard-jwt.spec/src/main/scripts/io/aklivity/zilla/specs/guard/jwt/config/event.yaml
vordimous: Updated please review new wording with added `reason` field.

specs/binding-mqtt-kafka.spec/src/main/resources/META-INF/zilla/mqtt_kafka.idl
vordimous: Updated, please review the new wording with the added `topic` field.

specs/model-avro.spec/src/main/scripts/io/aklivity/zilla/specs/model/avro/config/event.yaml
vordimous: I also took fixing this as a not in some follow on modifications that are needed: https://github.com/aklivity/zilla/pull/1029#issuecomment-2145701938

jfallows (approved):

jfallows (approved):

PR #1033, opened 2024-05-16
Fix: http-kafka will fetch messages that have been deleted by a reten…

### LINKED ISSUE #897: `http-kafka` will `fetch` messages that have been deleted by a retention policy
**Describe the bug**
Zilla is returning messages in a GET request using `http-kafka` binding `fetch` that have been deleted from the topic via a retention policy.

**To Reproduce**
1. use any `http-kafka` `fetch` config (bug was confirmed using the quickstart and taxi-demo)
2. configure the topic with `--config cleanup.policy=compact,delete --config delete.retention.ms=6000 --config retention.ms=6000`
3. produce messages to the topic
4. wait for messages to be deleted
5. See all messages in the GET response, but response should not have deleted messages
6. restart Zilla
7. GET response now shows the correct data

**Expected behavior**
Zilla doesn't return delete messages.

aDaemonThread:
…tion policy

Fixes #897

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheServerFetchFactory.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheServerFetchFactory.java
@@ -848,7 +848,7 @@ public final class KafkaCacheServerFetchFactory implements BindingHandler
 
                 final KafkaCachePartition.Node head = partition.head();
                 final KafkaCachePartition.Node nextHead =
-                        partition.newHeadIfNecessary(partitionOffset, key, valueLength, headersSizeMax, retentionMillisMax);
+                        partition.newHeadIfNecessary(partitionOffset, key, valueLength, headersSizeMax);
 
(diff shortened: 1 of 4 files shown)


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheServerFetchFactory.java
```diff
@@ -1187,15 +1187,16 @@ private void onServerFanoutInitialSignalSegmentDelete(
             final long now = currentTimeMillis();
 
             Node segmentNode = partition.sentinel().next();
-            while (segmentNode != partition.head() &&
-                    partition.deleteAt(segmentNode.segment(), retentionMillisMax) <= now)
+            while (segmentNode.segment() != null &&
```
jfallows, line 1190: In general, the `Node` constructors enforce either that the `segment` is non-null, or it is the `sentinel` for the partition.

If we agree on that, then let's use `!node.sentinel()` instead of `node.segment() != null`.

jfallows (approved):

PR #1035, opened 2024-05-17
Add zilla context to MQTT consumer groups

bmaidics:


runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaProxyFactory.java
```diff
@@ -75,7 +75,7 @@ public void attach(
         MqttKafkaBindingConfig kafkaBinding = new MqttKafkaBindingConfig(binding);
         bindings.put(binding.id, kafkaBinding);
 
-        factories.values().forEach(streamFactory -> streamFactory.onAttached(binding.id));
+        factories.values().forEach(streamFactory -> streamFactory.onAttached(binding.id, binding.name));
```
jfallows, line 78: Is this change necessary?

Expecting that `binding.id` is sufficient to extract both namespace and local name of binding via context, agree?

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaStreamFactory.java
```diff
@@ -30,7 +30,8 @@ default void detach(
     }
 
     default void onAttached(
-        long binding)
+        long bindingId,
+        String bindingName)
```
jfallows, line 34: Anticipating this change can be reverted.

jfallows (changes requested):

jfallows (approved):

PR #1044, opened 2024-05-22
Add detection of non-compacted session topic

bmaidics:


jfallows (approved):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
```diff
@@ -3415,6 +3423,61 @@ private KafkaSessionStateProxy(
...
+                    final KafkaConfigFW cleanupPolicyConfig =
+                        kafkaMergedBeginEx.configs().matchFirst(c -> c.name().equals(CONFIG_NAME_CLEANUP_POLICY));
+                    final String16FW cleanupPolicy = cleanupPolicyConfig != null ? cleanupPolicyConfig.value() : null;
+
+                    if (cleanupPolicy == null || !cleanupPolicy.equals(COMPACT_CLEANUP_POLICY))
+                    {
+                        Flyweight mqttResetEx = mqttSessionResetExRW.wrap(sessionExtBuffer, 0, sessionExtBuffer.capacity())
+                            .typeId(mqttTypeId)
+                            .reasonCode(MQTT_IMPLEMENTATION_SPECIFIC_ERROR)
+                            .reason(MQTT_NON_COMPACT_SESSIONS_TOPIC)
```
jfallows, line 3467: This reason seems out of scope for an MQTT client, as it is Kafka-specific.

This is more like what we need in the event log where we can describe what went wrong so that the devops can fix it, not the mqtt client.

specs/binding-mqtt-kafka.spec/src/main/resources/META-INF/zilla/mqtt_kafka.idl
```diff
@@ -30,4 +30,22 @@ scope mqtt_kafka
          int8 length;
          int16[length] packetIds = null;
     }
+
+    scope event
+    {
+        enum MqttKafkaEventType (uint8)
+        {
+            RESET_MQTT_CONNECTION (1)
```
jfallows, line 38: In general, you should think about these events as a message to devops so they can correct the problem.

For example, if the sessions topic is not configured correctly, i.e. cleanup policy is not `compact`, then that should be expressed here, rather than saying that the mqtt connection was reset, as the devops does not know how to read this event log and correct the problem in Kafka.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
```diff
@@ -1659,6 +1663,8 @@ private Consumer<KafkaMergedBeginExFW.Builder> beginExToKafkaMergedFetchOnly()
                             .latestOffset(v)
                             .metadata(metadataRW.length() > 0 ? metadataRW.toString() : null));
                 });
+                builder.configsItem(c -> c.name(CONFIG_NAME_CLEANUP_POLICY)
+                    .value(configs.get(CONFIG_NAME_CLEANUP_POLICY.asString())));
```
jfallows, line 1667: What if `cleanup.policy` is not present in `configs`?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
```diff
@@ -1952,6 +1971,7 @@ private void onTopicConfigChanged(
             long traceId,
             ArrayFW<KafkaConfigFW> configs)
         {
+            configs.forEach(c -> this.configs.put(c.name().asString(), c.value().asString()));
```
jfallows, line 1974: Do we need to store all of them?

Lets change `configs` to be keyed by `StringFW` and just keep `CONFIG_NAME_CLEANUP_POLICY` config value in the map, then we'll also avoid multiple unnecessary calls to `.asString()`.

jfallows (changes requested):

jfallows (approved):

jfallows (approved):

PR #1045, opened 2024-05-22
Set decoder to ignoreAll after session is taken over by other MQTT client

bmaidics:


runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -6061,7 +6061,7 @@ private void doPublishReset(
             private void doPublishAppEnd(
                 long traceId)
             {
-                if (!MqttState.initialClosed(state))
+                if (!MqttState.initialClosed(state) && publishPayloadBytes == 0)
```
jfallows, line 6064: What goes wrong without this change?

jfallows (changes requested):

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
bmaidics: When we receive session end, we remove the publish stream from the publishes map, and will eventually send a disconnect. But if there was already a publish going on, since we removed the stream, we'll hit NPE.

jfallows: This seems to be a difference between `close` vs `close-on-flush` meaning close after data sent.

In other binding, we often track this via the `closing` state, indicating that we are intending to close, but not yet. Then when the data is sent, if we are `closing` then we transition to `closed`.

Does that same approach work here?

jfallows (changes requested):

jfallows (approved):

PR #1048, opened 2024-05-23
Make metrics attributes configurable

attilakreiner:
## Description

Fixes # (issue)

### DIFF
--- a/specs/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/config/metrics.with.metrics.attributes.yaml
+++ b/specs/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/config/metrics.with.metrics.attributes.yaml
@@ -1,0 +1,46 @@
+#
+# Copyright 2021-2023 Aklivity Inc
+#
+# Licensed under the Aklivity Community License (the "License"); you may not use
+# this file except in compliance with the License.  You may obtain a copy of the
+# License at
+#
+#   https://www.aklivity.io/aklivity-community-license/
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations under the License.
+#
+
+---
+name: test
+telemetry:
+    attributes:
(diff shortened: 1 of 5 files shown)


specs/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/config/metrics.with.metrics.attributes.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+---
+name: test
+telemetry:
+    attributes:
+        service.namespace: example
+    metricsAttributes:
+        namespace: $NAMESPACE
+        binding: $BINDING
+        service.name: zilla
```
jfallows, line 24: Let's not introduce this separate `metricsAttributes` property and instead define `service.name` in `attributes`.

Then in `otlp` exporter, metrics can filter global telemetry attributes for presence of `service.name` and include it if present, otherwise omit.

This will let us specify the `service.name` and have it included in the exported metrics attributes as needed to fix the bug in the otlp demo.

Separately, @vordimous will define a story about enhancing our support for telemetry exporter attributes.

jfallows (changes requested):

specs/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/config/metrics.with.metrics.attributes.yaml
vordimous: @jfallows @attilakreiner, here is the ticket to track the additional changes to the telemetry attributes. We haven't decided if a global `telemetry.attributes` section is needed or if just an individual `options.attributes` section on each exporter will be better. The ticket includes both so that a decision can be made after some more discovery around implementation. 

- https://github.com/aklivity/zilla/issues/1049

vordimous (approved):

specs/exporter-otlp.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/otlp/config/metrics.with.metrics.attributes.yaml
attilakreiner: Got it, done, pls chk @jfallows

PR #1055, opened 2024-05-27
Update to handle catalog IT validation(resolve schema from subject)

aDaemonThread:


runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -275,6 +275,18 @@ private void onInitialBegin(
...
+                                String schema = handler.resolve(id);
+                                if (assertion.schema == null && schema != null)
+                                {
+                                    doInitialReset(traceId);
+                                }
+                                else if (assertion.schema != null && !assertion.schema.equals(schema))
+                                {
+                                    doInitialReset(traceId);
+                                }
+                            }
```
jfallows, line 289: Do we need to add the explicit concept of `resolve` as a boolean, or is it sufficient to derive that intent from `assertion.schema != null` instead ?

This perhaps seems more intuitive since it would read as asserting the id and schema, agree?

jfallows (changes requested):

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -275,7 +278,7 @@ private void onInitialBegin(
                             {
                                 doInitialReset(traceId);
                             }
-                            if (assertion.resolve)
+                            if (!Objects.equals(DEFAULT_ASSERTION_SCHEMA, assertion.schema))
```
jfallows, line 281:
```suggestion
                            if (assertion.schema != DEFAULT_ASSERTION_SCHEMA)
```

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -275,7 +278,7 @@ private void onInitialBegin(
...
                             {
                                 String schema = handler.resolve(id);
                                 if (assertion.schema == null && schema != null)
```
jfallows, line 284:
```
                                if (!Objects.equals(assertion.schema, schema))
                                {
                                     doInitialReset(traceId);
                                }
```

jfallows (changes requested):

jfallows (approved):

PR #1065, opened 2024-05-28
Add publish qos max options for mqtt-kafka binding

bmaidics:


cloud/docker-image/src/main/docker/incubator/zilla.pcap
jfallows, line 1: Please remove `zilla.pcap` from PR, it is also breaking the build due to license header check.

jfallows (changes requested):

jfallows (approved):

PR #1070, opened 2024-05-31
feat: replace port 8080 with 12345

vordimous:
also update any 808x ports

## Description

This change uses a dummy `12345` port for tests such that it is unlikely to conflict with any other services running locally.


jfallows (commented):
@vordimous Let's use a port outside the ephemeral port range `49152-65535`, such as `12345`, instead of `54321`.
https://en.wikipedia.org/wiki/Ephemeral_port

jfallows (approved):

PR #1092, opened 2024-06-12
Add sse payload validation

bmaidics:


specs/binding-sse.spec/src/main/scripts/io/aklivity/zilla/specs/binding/sse/schema/sse.schema.patch.json
```diff
@@ -40,6 +40,36 @@
                             {
                                 "type": "integer",
                                 "default": 2000
+                            },
+                            "paths":
```
jfallows, line 44: Not sure about `paths` ... `path`, maybe `requests` ... `paths`, or `types` ... `path`?
Let's discuss the naming.

runtime/binding-sse/src/test/java/io/aklivity/zilla/runtime/binding/sse/internal/streams/server/DataIT.java
```diff
@@ -78,6 +78,16 @@ public void shouldReceiveMultipleMessages() throws Exception
...
+    @Test
+    @Configuration("server.validator.yaml")
+    @Specification({
+        "${net}/validation/request",
+        "${app}/validation/server" })
+    public void shouldValidateMessages() throws Exception
+    {
+        k3po.finish();
+    }
+
```
jfallows, line 90: We need a `shouldRejectMessages` too to prove validation is not just always allowing.

runtime/binding-sse/src/main/java/io/aklivity/zilla/runtime/binding/sse/config/SsePathConfig.java
```diff
@@ -0,0 +1,40 @@
...
+
+import static java.util.function.Function.identity;
+
+import io.aklivity.zilla.runtime.engine.config.ModelConfig;
+
+public class SsePathConfig
+{
+    public final String path;
+    public final ModelConfig content;
+
```
jfallows, line 26: Remove extra blank line.

jfallows (changes requested):

runtime/binding-sse/src/test/java/io/aklivity/zilla/runtime/binding/sse/internal/streams/server/DataIT.java
bmaidics: Inside this test we have an invalid data as well. I just wanted to include it in the same test, as if the payload is invalid we don't close the stream. Or maybe put it in a seperate and assert for the invalid event?

specs/binding-sse.spec/src/main/scripts/io/aklivity/zilla/specs/binding/sse/schema/sse.schema.patch.json
bmaidics: Yeah, I was thinking about this as well a lot, but couldn't come up with a better one. I think the `requests` ... `path` makes more sense.

runtime/binding-sse/src/test/java/io/aklivity/zilla/runtime/binding/sse/internal/streams/server/DataIT.java
jfallows: Yes, agreed. Please separate the positive validation and negative validation scenarios.

specs/binding-sse.spec/src/main/scripts/io/aklivity/zilla/specs/binding/sse/schema/sse.schema.patch.json
jfallows: Agreed, let's go with `requests` then, I was leaning towards that one too. 😄 

The idea of `types` is too overloaded already, it's why we went with `model` for the validation.

jfallows (approved):

PR #1093, opened 2024-06-12
filtering by structured value field(s)

### LINKED ISSUE #1062: Support filtering by kafka structured value field(s)
As an appdev, I want to configure zilla to filter kafka messages by one or more fields in the structured message value.

**Acceptance criteria**
 - `http-kafka` can filter by kafka message structured value field
 - `sse-kafka` can filter by kafka message structured value field

**Additional context**
One approach is to enhance `kafka` `cache_server` to promote fields of structured messages with `model` to _virtual_ message headers, so they are indexed in the same way as actual message headers.

aDaemonThread:
Fixes #1062


runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/model/TestConverterHandler.java
```diff
@@ -46,6 +56,19 @@ public TestConverterHandler(
...
         this.handler = cataloged != null ? supplyCatalog.apply(cataloged.id) : null;
+        this.pattern = Pattern.compile(PATH);
+        this.extracted = new HashMap<>();
+    }
+
+    @Override
+    public void extract(
+        String path)
+    {
+        Matcher matcher = pattern.matcher(path);
```
jfallows, line 67: Let's reuse the matcher, via precreating with `pattern.matcher("")` and then `matcher.reset(path)` here.

Same matcher reuse for extractedLength below.

runtime/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
```diff
@@ -34,19 +39,26 @@
...
     protected final JsonModelEventContext event;
+    protected final Map<String, OctetsFW> extracted;
 
     private final Int2ObjectCache<JsonSchema> schemas;
     private final Int2ObjectCache<JsonProvider> providers;
     private final JsonProvider schemaProvider;
     private final JsonValidationService service;
     private final JsonParserFactory factory;
+
+    protected JsonParser parser;
```
jfallows, line 61: Is this change needed? Prefer to keep it `private` if possible.

runtime/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
```diff
@@ -133,4 +185,6 @@ private JsonProvider createProvider(
         }
         return provider;
     }
+
+
```
jfallows, line 189:
```suggestion
```

runtime/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
```diff
@@ -83,8 +96,47 @@ protected final boolean validate(
...
+                        case KEY_NAME:
+                            String key = parser.getString();
+                            valueBytes = extracted.get(key);
+                            break;
+                        case VALUE_STRING:
+                            if (valueBytes != null)
+                            {
+                                int offset = (int) parser.getLocation().getStreamOffset() - DOUBLE_QUOTE_LENGTH;
+                                offset += index;
+                                int valLength = parser.getString().getBytes(UTF_8).length;
```
jfallows, line 122: We should be able to calculate the UTF-8 bytes length of this String without allocating the bytes themselves.

runtime/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonModelHandler.java
```diff
@@ -83,8 +96,47 @@ protected final boolean validate(
...
+                                valueBytes.wrap(in.buffer(), offset - valLength, offset);
+                                valueBytes = null;
+                            }
+                            break;
+                        case VALUE_NUMBER:
+                            if (valueBytes != null)
+                            {
+                                int offset = (int) parser.getLocation().getStreamOffset();
+                                offset += index;
+                                int valLength = parser.getString().getBytes().length;
```
jfallows, line 132: We should be able to calculate the UTF-8 bytes length of this String without allocating the bytes themselves.

runtime/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonReadConverterHandler.java
```diff
@@ -41,6 +50,17 @@ public int padding(
         return handler.decodePadding(data, index, length);
     }
 
+    @Override
+    public void extract(
+        String path)
+    {
+        Matcher matcher = pattern.matcher(path);
```
jfallows, line 57: Reuse matcher.

runtime/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonReadConverterHandler.java
```diff
@@ -53,6 +73,35 @@ public int convert(
...
+        }
+        return value != null ? value.sizeof() : 0;
+    }
+
+    @Override
+    public void extracted(
+        String path,
+        FieldVisitor visitor)
+    {
+        Matcher matcher = pattern.matcher(path);
```
jfallows, line 94: Reuse matcher.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/config/cache.options.extract.headers.yaml
```diff
@@ -0,0 +1,67 @@
...
+        when:
+          - topic: test
+  cache0:
+    type: kafka
+    kind: cache_server
+    options:
+      topics:
+        - name: test
+          headers:
+            - correlation-id: $.correlationId
```
jfallows, line 55: I realized this syntax for headers is inconsistent with the other fields in the same object, like `key` and `value`.

In those case, each one provides a model, constraining the type, so if we would want to add that capability later for headers we would naturally have an object for headers, with header names as keys, and `model` in the value.

Then we would also be able to add `extract` or `value` as another property, indicating an expression, with support for short form in the meantime.

Also, even though we know we are using JSON path internally, it is inconsistent with our other expressions in `zilla.yaml`, so perhaps we want to use the following instead?

```suggestion
          headers:
            correlation-id: ${value.correlationId}
```
and internally treat this as `$.correlationId` but without exposing it directly in the `zilla.yaml` syntax.

jfallows (changes requested):

jfallows (approved):

PR #1100, opened 2024-06-19
Fix NegativeArraySizeException when receiving mqttFlush

bmaidics:


jfallows (commented):
Please add a test scenario that reads state using by version 1 syntax then updates the state to add a new packetId and writes the new state using version 2 syntax. 

This will confirm we can deploy a new version of zilla with updated syntax for packets ids without any negative impact caused by the old state in kafka.

jfallows (approved):

PR #1108, opened 2024-06-27
Verify public-private tls key pair

### LINKED ISSUE #1073: Verify public-private key pair obtained from vault used for TLS handshake
As devops, I want to detect when zilla is misconfigured such that TLS private key does not match public key.

**Acceptance criteria**
 - Verify public-private key pair
 - Log event if public-private key pair is invalid

**Additional context**
By performing this check in the `tls` binding where the key pair is used, then each vault does not need to reimplement.

Note: there are at least two ways to approach this.
https://stackoverflow.com/questions/49426844/how-to-validate-a-public-and-private-key-pair-in-java

attilakreiner:
## Description

This change verifies the public-private key pair obtained from vault. Logs an event if public-private key pair is invalid or missing.

Fixes #1073

### DIFF
--- a/runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsBindingConfig.java
+++ b/runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsBindingConfig.java
@@ -91,14 +91,15 @@ public final class TlsBindingConfig
 
     public void init(
         TlsConfiguration config,
+        TlsEventContext events,
         VaultHandler vault,
-        SecureRandom random,
-        TlsEventContext event)
+        SecureRandom random)
     {
-        this.verifier = new TlsKeyPairVerifier(event);
+        this.verifier = new TlsKeyPairVerifier(events);
         char[] keysPass = "generated".toCharArray();
         KeyStore keys = newKeys(config, vault, keysPass, options.keys, options.signers);
         KeyStore trust = newTrust(config, vault, options.trust, options.trustcacerts && kind == KindConfig.CLIENT);
+
         try
         {
             KeyManager[] keyManagers = null;
--- a/runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsBindingConfig.java
+++ b/runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsBindingConfig.java
@@ -76,7 +76,6 @@ public final class TlsBindingConfig
     public final List<TlsRouteConfig> routes;
 
     private SSLContext context;
-    private TlsKeyPairVerifier verifier;
 
     public TlsBindingConfig(
         BindingConfig binding)
@@ -95,9 +94,9 @@ public final class TlsBindingConfig
         VaultHandler vault,
         SecureRandom random)
     {
-        this.verifier = new TlsKeyPairVerifier(events);
+        TlsKeyPairVerifier verifier = new TlsKeyPairVerifier(events);
         char[] keysPass = "generated".toCharArray();
-        KeyStore keys = newKeys(config, vault, keysPass, options.keys, options.signers);
+        KeyStore keys = newKeys(config, vault, keysPass, verifier, options.keys, options.signers);
         KeyStore trust = newTrust(config, vault, options.trust, options.trustcacerts && kind == KindConfig.CLIENT);
 
         try
@@ -411,6 +410,7 @@ public final class TlsBindingConfig
         TlsConfiguration config,
         VaultHandler vault,
         char[] password,
+        TlsKeyPairVerifier verifier,
         List<String> keyNames,
         List<String> signerNames)
     {
--- a/runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsBindingConfig.java
+++ b/runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsBindingConfig.java
@@ -20,7 +20,6 @@ import static io.aklivity.zilla.runtime.binding.tls.internal.types.ProxyInfoType
 import static io.aklivity.zilla.runtime.binding.tls.internal.types.ProxyInfoType.AUTHORITY;
 import static io.aklivity.zilla.runtime.binding.tls.internal.types.ProxyInfoType.SECURE;
 import static io.aklivity.zilla.runtime.binding.tls.internal.types.ProxySecureInfoType.NAME;
-import static java.lang.System.currentTimeMillis;
 import static java.util.Collections.singletonList;
 import static java.util.stream.Collectors.toList;
 import static javax.net.ssl.StandardConstants.SNI_HOST_NAME;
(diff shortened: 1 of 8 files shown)


specs/binding-tls.spec/src/main/resources/META-INF/zilla/tls.idl
```diff
@@ -23,7 +23,20 @@ scope tls
             TLS_PROTOCOL_REJECTED (2),
             TLS_KEY_REJECTED (3),
             TLS_PEER_NOT_VERIFIED (4),
-            TLS_HANDSHAKE_FAILED (5)
+            TLS_HANDSHAKE_FAILED (5),
+            TLS_KEY_PAIR_VERIFICATION_FAILED (6)
```
jfallows, line 27: Even though the implementation detail is a key pair, the syntax in `zilla.yaml` is just `keys`, so let's name the event using key instead of key pair too.
```suggestion
            TLS_KEY_VERIFICATION_FAILED (6)
```

Note: this feedback applies to all naming in this PR.

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsBindingConfig.java
```diff
@@ -91,12 +92,13 @@ public TlsBindingConfig(
     public void init(
         TlsConfiguration config,
         VaultHandler vault,
-        SecureRandom random)
+        SecureRandom random,
+        TlsEventContext event)
```
jfallows, line 96: Please move this to the second parameter, also rename parameter to `events`.

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsBindingConfig.java
```diff
@@ -91,12 +92,13 @@ public TlsBindingConfig(
...
     {
+        this.verifier = new TlsKeyPairVerifier(event);
         char[] keysPass = "generated".toCharArray();
         KeyStore keys = newKeys(config, vault, keysPass, options.keys, options.signers);
         KeyStore trust = newTrust(config, vault, options.trust, options.trustcacerts && kind == KindConfig.CLIENT);
-
```
jfallows, line 99: Please revert blank line removal.

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsBindingConfig.java
```diff
@@ -91,12 +92,13 @@ public TlsBindingConfig(
     public void init(
         TlsConfiguration config,
         VaultHandler vault,
-        SecureRandom random)
+        SecureRandom random,
+        TlsEventContext event)
     {
+        this.verifier = new TlsKeyPairVerifier(event);
```
jfallows, line 98: If `verifier` is only used locally in this method, then we don't need to store it in a field.
If it is needed by a method called by this method, such as `newKeys` then please pass it as a parameter, instead of introducing more global state as a field.

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsKeyPairVerifier.java
```diff
@@ -0,0 +1,91 @@
...
+    public boolean verify(
+        KeyStore.PrivateKeyEntry entry,
+        String keyName,
+        long bindingId)
+    {
+        boolean valid = false;
+        if (entry == null)
+        {
+            event.tlsKeyPairMissing(bindingId, keyName);
+        }
```
jfallows, line 52: Let's handle this case outside the `verify` method, at the caller.

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsKeyPairVerifier.java
```diff
@@ -0,0 +1,91 @@
...
+
+                // verify signature using the public key
+                sig.initVerify(publicKey);
+                sig.update(challenge);
+                valid = sig.verify(signature);
+
+                if (!valid)
+                {
+                    event.tlsKeyPairInvalid(bindingId, keyName);
+                }
```
jfallows, line 78: Let's handle this case outside the `verifier`, when the return is `false`.

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsKeyPairVerifier.java
```diff
@@ -0,0 +1,91 @@
...
+            }
+            catch (InvalidKeyException | SignatureException ex)
+            {
+                event.tlsKeyPairInvalid(bindingId, keyName);
```
jfallows, line 82: This can be replaced as a comment, such as `// key invalid`

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsKeyPairVerifier.java
```diff
@@ -0,0 +1,91 @@
...
+    public TlsKeyPairVerifier(
+        TlsEventContext event)
+    {
+        this.event = event;
+    }
+
+    public boolean verify(
+        KeyStore.PrivateKeyEntry entry,
+        String keyName,
+        long bindingId)
```
jfallows, line 46: After applying feedback below, this method can reduce to just the `entry` parameter.

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsKeyPairVerifier.java
```diff
@@ -0,0 +1,91 @@
...
+{
+    private static final String ALGORITHM = "SHA256withRSA";
+
+    private final TlsEventContext event;
+
+    public TlsKeyPairVerifier(
+        TlsEventContext event)
+    {
+        this.event = event;
+    }
```
jfallows, line 41: This constructor is no longer needed after handling event reporting from the caller. (see other feedback below)

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsKeyPairVerifier.java
```diff
@@ -0,0 +1,91 @@
...
+import java.security.PublicKey;
+import java.security.Signature;
+import java.security.SignatureException;
+import java.util.concurrent.ThreadLocalRandom;
+
+import io.aklivity.zilla.runtime.binding.tls.internal.TlsEventContext;
+
+public class TlsKeyPairVerifier
+{
+    private static final String ALGORITHM = "SHA256withRSA";
```
jfallows, line 33: We discussed other key types like Elliptic Curve that would require a different algorithm, agree?

runtime/binding-tls/src/test/java/io/aklivity/zilla/runtime/binding/tls/internal/config/TlsKeyPairVerifierTest.java
```diff
@@ -0,0 +1,202 @@
...
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.Base64;
+
+import org.junit.Test;
+
+import io.aklivity.zilla.runtime.binding.tls.internal.TlsEventContext;
+
+public class TlsKeyPairVerifierTest
+{
+    private static final String PRIVATE_KEY = """
```
jfallows, line 37: I think this is an `RSA` key. We'll need `ECDSA` too.

jfallows (changes requested):

jfallows (approved):

PR #1145, opened 2024-07-12
Support http authorization in asyncapi generation

akrambek:


runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/AsyncapiHttpProtocol.java
```diff
@@ -167,7 +172,7 @@ protected boolean isSecure()
     private <C> HttpOptionsConfigBuilder<C> injectHttpServerOptions(
         HttpOptionsConfigBuilder<C> options)
     {
-        if (isJwtEnabled)
+        if (isOauthEnabled)
         {
             options.authorization(authorization).build();
         }
```
jfallows, line 178: Do we need to call `.build()` here?

runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/AsyncapiServerNamespaceGenerator.java
```diff
@@ -52,7 +52,7 @@ public NamespaceConfig generate(
                 .inject(n -> injectTcpServer(n, servers, options, metricRefs))
                 .inject(n -> injectTlsServer(n, servers, options))
                 .inject(n -> injectProtocolRelatedBindings(n, servers, metricRefs))
-                .inject(n -> injectProtocolServers(n, servers, metricRefs))
+                .inject(n -> injectProtocolServers(n, binding.namespace, options, servers, metricRefs))
```
jfallows, line 55: This is correct to pass `binding.namespace`, but parameter is called `qname`, which is not consistent.

Suggest we add a transient `qname` field to `HttpAuthorizationConfig` and set during `AsyncapiNamespaceGenerator.init(BindingConfig)`, to specify the qualified name of the guard being used for authorization in the `binding` namespace.

Then this parameter can be removed from the call stack, and the `HttpAuthorizationConfig.qname` can be used where we originally attempted to use `HttpAuthorizationConfig.name`, but now `qname` will just work.

I'm planning to implement this generically in the engine so you wouldn't need to explicitly set `qname`, but it will help tidy the code if we do it this way for now and I can remove the line where you set `qname` when I'm adding the general approach later.

runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/AsyncapiHttpProtocol.java
```diff
@@ -264,24 +269,23 @@ private <C> HttpRequestConfigBuilder<C> injectPathParams(
...
-                            .guarded()
-                                .name(guardName)
-                                .inject(guarded -> injectGuardedRoles(guarded, securityItem.get(securityItemLabel)))
-                                .build();
-                        break;
-                    }
-                }
+                route
+                    .guarded()
+                    .name(String.format("%s:%s", qname, options.authorization.name))
```
jfallows, line 286: Based on other feedback above, this becomes:
```suggestion
                    .name(options.authorization.qname)
```

jfallows (changes requested):

jfallows (approved):

PR #1183, opened 2024-08-06
Support extract-key kafka message transform 

### LINKED ISSUE #1176: Support `extract-key` kafka message transform
Suggested syntax.
```
  kafka_cache_server:
    type: kafka
    kind: cache_server
    options:
        topics:
          - name: my-transformed-topic
            transforms:
              - extract-key: ${message.key.id}
...
```

akrambek:
## Description

Support extract-key kafka message transform.

Fixes #1176

### DIFF
--- a/runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiCompositeGenerator.java
+++ b/runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiCompositeGenerator.java
@@ -360,7 +360,7 @@ public abstract class AsyncapiCompositeGenerator
                         {
                             if (m.payload != null)
                             {
-                                final String subject = "%s-%s-payload".formatted(m.channel.name, m.name);
+                                final String subject = "%s-%s-value".formatted(m.channel.name, m.name);
 
                                 options.schema()
                                     .subject(subject)
--- a/runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiCompositeGenerator.java
+++ b/runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiCompositeGenerator.java
@@ -233,7 +233,7 @@ public abstract class AsyncapiCompositeGenerator
             return value;
         }
 
-        protected final class CatalogsHelper
+        protected class CatalogsHelper
         {
             private final AsyncapiSchemaConfig schema;
 
@@ -356,36 +356,7 @@ public abstract class AsyncapiCompositeGenerator
                         .map(o -> o.channel)
                         .filter(c -> c.messages != null)
                         .flatMap(c -> c.messages.stream())
-                        .forEach(m ->
-                        {
-                            if (m.payload != null)
-                            {
-                                final String subject = "%s-value".formatted(m.channel.address);
-
-                                options.schema()
-                                    .subject(subject)
-                                    .version("latest")
-                                    .schema(toSchemaJson(jsonb, m.payload.model))
-                                    .build();
-                            }
-
-                            if (m.headers != null && m.headers.properties != null)
-                            {
-                                for (Map.Entry<String, AsyncapiSchemaView> header : m.headers.properties.entrySet())
-                                {
-                                    final String name = header.getKey();
-                                    final AsyncapiSchemaItemView schema = header.getValue();
-
-                                    final String subject = "%s-header-%s".formatted(m.channel.address, name);
-
-                                    options.schema()
-                                        .subject(subject)
-                                        .version("latest")
-                                        .schema(toSchemaJson(jsonb, schema.model))
-                                        .build();
-                                }
-                            }
-                        });
+                        .forEach(m -> injectInlineSubject(jsonb, options, m));
 
                     Stream.of(schema)
                         .map(s -> s.asyncapi)
@@ -413,7 +384,41 @@ public abstract class AsyncapiCompositeGenerator
                 return options;
             }
 
-            private static String toSchemaJson(
+            protected <C> void injectInlineSubject(
+                Jsonb jsonb,
+                InlineOptionsConfigBuilder<C> options,
+                AsyncapiMessageView message)
+            {
+                if (message.payload != null)
+                {
+                    final String subject = "%s-value".formatted(message.channel.address);
+
+                    options.schema()
+                        .subject(subject)
+                        .version("latest")
+                        .schema(toSchemaJson(jsonb, message.payload.model))
+                        .build();
+                }
+
+                if (message.headers != null && message.headers.properties != null)
+                {
+                    for (Map.Entry<String, AsyncapiSchemaView> header : message.headers.properties.entrySet())
+                    {
+                        final String name = header.getKey();
+                        final AsyncapiSchemaItemView schema = header.getValue();
+
+                        final String subject = "%s-header-%s".formatted(message.channel.address, name);
+
+                        options.schema()
+                            .subject(subject)
+                            .version("latest")
+                            .schema(toSchemaJson(jsonb, schema.model))
+                            .build();
+                    }
+                }
+            }
+
+            protected static String toSchemaJson(
                 Jsonb jsonb,
                 AsyncapiSchemaItem schema)
             {
--- a/runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiCompositeGenerator.java
+++ b/runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiCompositeGenerator.java
@@ -353,9 +353,8 @@ public abstract class AsyncapiCompositeGenerator
                     Stream.of(schema)
                         .map(s -> s.asyncapi)
                         .flatMap(v -> v.operations.values().stream())
-                        .map(o -> o.channel)
-                        .filter(c -> c.messages != null)
-                        .flatMap(c -> c.messages.stream())
+                        .filter(o -> o.messages != null)
+                        .flatMap(o -> o.messages.stream())
                         .forEach(m -> injectInlineSubject(jsonb, options, m));
 
                     Stream.of(schema)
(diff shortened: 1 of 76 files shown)


runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiCompositeGenerator.java
```diff
@@ -495,7 +495,7 @@ protected final void injectPayloadModel(
                                 .name("catalog0")
                                 .schema()
                                     .version("latest")
-                                    .subject(subject)
+                                    .subject(message.payload.name)
```
jfallows, line 498: This should be consistent with the others, either `subject` everywhere or `message.payload.name` everywhere.

The value of `subject` is passed as part of the URL to access schema registry.
```
/subjects/{0}/versions/{1}
```
So this should probably be based on the channel address topic with `-value` suffix, rather than relying on any naming convention alignment between message payload name and channel address (kafka topic).

Note: given that channels (kafka topics) can have more than one message type in the general case, seems like we actually want to potentially use topic-record name strategy instead of topic name strategy to differentiate the schemas for different channel (topic) messages.

In that case, it  would seem to make sense to base the record part of the topic-record name strategy on the message name within the channel messages.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaTopicTransformsConfigAdapter.java
```diff
@@ -0,0 +1,84 @@
...
+        if (headers != null)
+        {
+            for (Map.Entry<String, JsonValue> entry : headers.entrySet())
+            {
+                JsonString jsonString = (JsonString) entry.getValue();
+                topicBuilder.extractHeader(entry.getKey(), jsonString.getString());
+            }
+        }
+
+        return topicBuilder.build();
```
jfallows, line 82:
```suggestion
        KafkaTopicTransformsConfigBuilder<KafkaTopicTransformsConfig> transforms = KafkaTopicTransformsConfig.builder();

        if (object.containsKey(EXTRACT_KEY_NAME))
        {
            String extractKey = object.getString(EXTRACT_KEY_NAME);
            transforms.extractKey(extractKey);
        }

        if (object.containsKey(EXTRACT_HEADERS_NAME))
        {
            JsonObject headers = object.getJsonObject(EXTRACT_HEADERS_NAME);
            for (Map.Entry<String, JsonValue> entry : headers.entrySet())
            {
                String headerName = entry.getKey();
                JsonString headerValue = (JsonString) entry.getValue();
                transforms.extractHeader(headerName, headerValue.getString());
            }
        }

        return transforms.build();
```

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaOptionsConfigAdapterTest.java
```diff
@@ -226,10 +226,13 @@ public void shouldReadHeadersOptions()
...
-                    "{" +
-                    "\"correlation-id\": \"${message.value.correlationId}\"" +
-                    "}" +
+                    "\"transforms\":" +
+                    "[" +
+                        "\"extract-headers\":" +
+                        "{" +
+                            "\"correlation-id\": \"${message.value.correlationId}\"" +
+                        "}" +
+                    "]" +
```
jfallows, line 235: Include `extract-key` here please.

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaOptionsConfigAdapterTest.java
```diff
@@ -249,15 +252,17 @@ public void shouldWriteHeadersOptions()
             .bootstrap("test")
             .topic()
                 .name("test")
-                .header("correlation-id", "${message.value.correlationId}")
+                .transforms()
+                    .extractHeader("correlation-id", "${message.value.correlationId}")
+                    .build()
```
jfallows, line 257: Include `extract-key` here please.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -339,18 +339,37 @@
                             "enum": [ "none", "json_patch" ],
                             "deprecated": true
                         },
-                        "headers":
+                        "transforms":
```
jfallows, line 342: Make sure you have caught up to latest develop, I think this change is already present there.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/fetch/filter.extracted.header/client.rpt
```diff
@@ -75,7 +75,7 @@ read zilla:data.ext ${kafka:matchDataEx()
                              .fetch()
                                  .partition(0, 2, 8)
                                  .header("header1", "value1")
-                                 .header("correlation-id", "12345")
+                                 .header("correlation-id", "1234")
```
jfallows, line 78: Just curious, what prompted this change?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaTopicConfigAdapter.java
```diff
@@ -41,9 +37,10 @@ public final class KafkaTopicConfigAdapter implements JsonbAdapter<KafkaTopicCon
     private static final String EVENT_KEY = "key";
     private static final String EVENT_VALUE = "value";
     private static final String SUBJECT = "subject";
-    private static final String HEADERS_NAME = "headers";
+    private static final String TRANSFORMS_NAME = "transforms";
 
     private final ModelConfigAdapter converter = new ModelConfigAdapter();
+    private final KafkaTopicTransformsConfigAdapter transformsConverter = new KafkaTopicTransformsConfigAdapter();
```
jfallows, line 43:
```suggestion
    private final ModelConfigAdapter model = new ModelConfigAdapter();
    private final KafkaTopicTransformsConfigAdapter transforms = new KafkaTopicTransformsConfigAdapter();
```

These are not converters (means something else for us too) plus probably no need to include the type in the name.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -360,14 +364,14 @@ public void writeEntry(
         ConverterHandler convertKey,
         ConverterHandler convertValue,
         boolean verbose,
-        List<KafkaTopicHeaderType> headerTypes)
+        KafkaTopicTransformsConfig transforms)
```
jfallows, line 367: I think this is assuming we can specify either `extract-key` or `extract-headers` only once in the `transforms` array, right?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -621,6 +630,25 @@ else if (headerTypes != null && !headerTypes.isEmpty())
...
+                KafkaCachePaddedKeyFW.Builder paddedKeyBuilder = paddedKeyRW;
+                final int keySize = paddedKeyBuilder.key(k -> k.length(length).value(buffer, index, length)).sizeof();
+                paddedKeyBuilder.padding(logFile.buffer(), 0, paddedKeySize - keySize - SIZE_OF_INT);
+                KafkaCachePaddedKeyFW newPaddedKey = paddedKeyBuilder.build();
+                logFile.writeBytes(position, newPaddedKey.buffer(), newPaddedKey.offset(), newPaddedKey.sizeof());
+            };
+
+            convertKey.extracted(transforms.extractKey, writeKey);
+        }
+
```
jfallows, line 651: I think this is assuming that `extract-key` is always processed after `extract-key`, which might be fine for now if we cannot extract the key from a header yet, but probably not sufficient for the general case later where these transforms are expected to execute in the order defined, the output of the previous feeding into the input of the next.

jfallows (changes requested):

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/fetch/filter.extracted.header/client.rpt
akrambek: The script we are using has 4 bytes `key1` and the TestConverter was returning `12345` which has 5 bytes and it was causing testware issue. I just changed TestConverter to return 1234 to match number of bytes in server script.

runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/model/resolver/AsyncapiOperationResolver.java
```diff
@@ -14,16 +14,42 @@
...
+import java.util.stream.Stream;
 
 import io.aklivity.zilla.runtime.binding.asyncapi.internal.model.Asyncapi;
+import io.aklivity.zilla.runtime.binding.asyncapi.internal.model.AsyncapiMessage;
 import io.aklivity.zilla.runtime.binding.asyncapi.internal.model.AsyncapiOperation;
 
 public final class AsyncapiOperationResolver extends AbstractAsyncapiResolver<AsyncapiOperation>
 {
+    private AsyncapiChannelResolver channels;
+    private AsyncapiMessageResolver messages;
```
jfallows, line 29:
```suggestion
    private final AsyncapiChannelResolver channels;
    private final AsyncapiMessageResolver messages;
```

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #1214, opened 2024-08-27
Fix incorrect CRC combine in Kafka produce client

bmaidics:
Fixes https://github.com/aklivity/zilla/issues/1221


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientProduceFactory.java
```diff
@@ -1723,6 +1725,16 @@ private void doEncodeRecordInit(
...
+                final MutableDirectBuffer encodeBuffer = writeBuffer;
+                this.recordHeaderSize = recordHeader.sizeof();
+                encodeBuffer.putBytes(0, encodeSlotBuffer, encodeSlotLimit, this.recordHeaderSize);
+
+                final byte[] encodeByteBuf = encodeBuffer.byteArray();
+                crc32c.reset();
+                crc32c.update(encodeByteBuf, 0, this.recordHeaderSize);
+
+                recordHeaderChecksum = crc32c.getValue();
+
```
jfallows, line 1737: Can this be solved by adjusting the `crcDataLimit` to include the record header in `encodeCrc()` instead?

How would this work for record batches with multiple records?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientProduceFactory.java
```diff
@@ -1812,7 +1824,14 @@ private void doEncodeRecordFin(
 
                 if (encodeableRecordBytesDeferred > 0)
                 {
-                    doNetworkData(traceId, budgetId, EMPTY_BUFFER, 0, 0);
+                    if (flushableRequestBytes == 0)
+                    {
+                        doEncodeRequestIfNecessary(traceId, budgetId);
+                    }
```
jfallows, line 1830: This logic is proactively encoding the produce request when the incoming message is fragmented.

It seems to be defeating message batching because it doesn't give a chance for more message fragments and messages to arrive before sending the batched request. If no more message fragments or messages arrive, then the signal should encode the produce request anyway.

I think the original mental model for this assumed messages would only be fragmented due to flow control forcing it when the encode slot is full, but we are not explicitly verifying that so when small messages are fragmented we are unnecessarily (and incorrectly) proactively sending the produce request.

Please verify that batching is still working as expected and adjust this logic as needed.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientProduceFactory.java
```diff
@@ -2006,8 +2007,7 @@ private void encodeCrc()
                     crc.update(encodeSlotByteBuffer);
 
                     long checksum = crc.getValue();
-
-                    if (crcDataLimit == recordBatch.limit())
+                    if (crcDataLimit == recordHeaderLimit && valueCompleteSize != 0)
```
jfallows, line 2010: What is the message is `null` (tombstone) instead of empty?
Perhaps this test needs to be `&& valueCompleteSize > 0` instead?

Please add an IT scenario for `null` message in addition to `empty` message to confirm whether change is necessary.

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/ClientProduceIT.java
```diff
@@ -203,6 +203,26 @@ public void shouldSendMessageValue() throws Exception
...
+    @Test
+    @Configuration("client.when.topic.yaml")
+    @Specification({
+        "${app}/message.empty.crc/client",
+        "${net}/message.empty.crc/server"})
+    public void shouldSendMessageEmptyWithCrc() throws Exception
+    {
+        k3po.finish();
+    }
+
```
jfallows, line 215: Please add `shouldSendMessageNullWithCrc()` as well.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/produce/message.empty.crc/client.rpt
```diff
@@ -0,0 +1,84 @@
...
+                              .build()}
+
+
+write zilla:data.ext ${kafka:dataEx()
+                              .typeId(zilla:id("kafka"))
+                              .produce()
+                                  .timestamp(1724846895864)
+                                  .ackMode("LEADER_ONLY")
+                                  .key("bench_pub_682636139_8#migrate")
+                                  .header("sender-id", "zilla-5e105508-c47e-4e41-add1-1b214c2883f6")
```
jfallows, line 80: Please use more generic values for consistency with other scripts.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/produce/messages.fragmented.crc/client.rpt
```diff
@@ -0,0 +1,321 @@
...
+                                  .ackMode("NONE")
+                                  .key("bench/39")
+                                  .header("zilla:filter", "bench")
+                                  .header("zilla:filter", "39")
+                                  .header("zilla:local", "bench_pub_2543774266_39")
+                                  .header("zilla:qos", "0")
+                                  .build()
+                              .build()}
+write "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+write flush
```
jfallows, line 89: Please use more generic keys and header values for consistency with other scripts.
Also, the fragmented value doesn't need to be quite so long, just fragmented. 😄

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientProduceFactory.java
```diff
@@ -2006,8 +2007,7 @@ private void encodeCrc()
                     crc.update(encodeSlotByteBuffer);
 
                     long checksum = crc.getValue();
-
-                    if (crcDataLimit == recordBatch.limit())
+                    if (crcDataLimit == recordHeaderLimit && valueCompleteSize > 0)
```
jfallows, line 2010: I realized that `valueCompleteSize` is always `>= 0` so this should actually be
```suggestion
                    if (crcDataLimit == recordHeaderLimit && valueCompleteSize != 0)
```
due to
```
        final int valueSize = payload != null ? payload.sizeof() : 0;
```
in `flushRecordInit`.

jfallows (changes requested):

jfallows (approved):

PR #1217, opened 2024-08-27
fix: fix schema syntax and naming

vordimous:
## Description

When comparing the zilla schema to the docs there were some syntax and naming inconsistencies. These changes address them.

Refer to the live reference docs update for implementation of docs types and defaults: https://vordimous.github.io/zilla-docs/next/reference/config/overview.html


specs/binding-grpc.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/schema/grpc.schema.patch.json
```diff
@@ -32,6 +32,7 @@
                         "enum": [ "server", "client"]
                     },
                     "vault": false,
+                    "options": false,
```
vordimous, line 35: now that the options services is deprecated there aren't any other options for the binding.

specs/binding-grpc.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/schema/grpc.schema.patch.json
```diff
@@ -124,6 +125,35 @@
                                     }
                                 },
                                 "additionalProperties": false
+                            },
+                            "catalog":
```
vordimous, line 129: This didn't exist and I don't think is being added anywhere else.

vordimous, line 129: This is the same structure and schema that is found in the `asyncapi` and `openapi` bindings and matches the `GrpcCatalogSchema` class

specs/binding-grpc-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/kafka/schema/grpc.kafka.schema.patch.json
```diff
@@ -233,7 +240,8 @@
                                                 {
                                                     "title": "Acks",
                                                     "type": "string",
-                                                    "enum": [ "none", "leader_only", "in_sync_replicas" ]
+                                                    "enum": [ "none", "leader_only", "in_sync_replicas" ],
+                                                    "default": "in_sync_replicas"
```
vordimous, line 244: The added defaults were communicated in the documentation but not in the schema.

specs/binding-grpc.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/schema/grpc.schema.patch.json
```diff
@@ -124,6 +131,40 @@
...
+                                                {
+                                                    "type": "string",
+                                                    "default": "latest"
+                                                }
+                                            },
+                                            "required":
+                                            [
+                                                "subject"
+                                            ],
+                                            "additionalProperties": false
```
jfallows, line 163:
```suggestion
                                            "$ref": "#/$defs/cataloged"
```
?

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
```diff
@@ -57,7 +59,8 @@
                                     "policy":
                                     {
                                         "title": "Policy",
-                                        "type": "string"
+                                        "type": "string",
+                                        "enum": [ "same-origin" , "cross-origin" ]
```
jfallows, line 63:
```suggestion
                                        "enum": [ "same-origin" , "cross-origin" ],
                                        "default": "same-origin"
```

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
```diff
@@ -434,7 +443,7 @@
                 {
                     "^[a-zA-Z]+[a-zA-Z0-9\\._\\-]*$":
                     {
-                        "title": "Authorization",
+                        "title": "Guard",
```
jfallows, line 446: Not sure about this one. I think the structure of what follows might be a definition of an authorization, not a guard.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -343,81 +350,35 @@
                         {
                             "title": "Transforms",
                             "type": "array",
-                            "oneOf":
+                            "items":
```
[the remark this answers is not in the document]
jfallows: Let's keep the original as this is also locking down the supported order to match the implementation.
Later when we add more general support for transforms, we'll loosen this a bit more.

specs/binding-tcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tcp/schema/tcp.schema.patch.json
```diff
@@ -163,6 +171,8 @@
         "path": "/$defs/options/binding/tcp",
         "value":
         {
+            "title": "tcp Options",
```
jfallows, line 174: I don't think we've been adding `title` for other options, and if we do then it's just `Options` right, since it is contextual within the current scope.

specs/binding-tls.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tls/config/client.alpn.default.yaml
```diff
@@ -34,5 +34,4 @@ bindings:
                 - serverca
             alpn:
                 - protocol2
-                - null
```
jfallows, line 37: I need to double check this again, I think it is trying to say what if no `alpn` protocol is negotiated, saying that it is still fair to proceed in the generic case.

specs/binding-tls.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tls/config/server.alpn.default.yaml
```diff
@@ -34,5 +34,4 @@ bindings:
                 - localhost
             alpn:
                 - protocol2
-                - null
```
jfallows, line 37: Same here as client comment.

specs/binding-tls.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tls/schema/tls.schema.patch.json
```diff
@@ -229,14 +239,15 @@
                         "type": "array",
                         "items":
                         {
-                            "type": [ "string", "null" ]
+                            "type": "string"
```
jfallows, line 242: Same here.

specs/binding-tls.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tls/schema/tls.schema.patch.json
```diff
@@ -229,14 +239,15 @@
...
                         }
                     },
                     "mutual":
                     {
                         "title": "Mutual Authentication",
                         "type": "string",
-                        "enum": [ "required", "requested", "none" ]
+                        "enum": [ "required", "requested", "none" ],
+                        "default": "none"
```
jfallows, line 250: The default is dependent on the existence of `trust` peer field. If `trust` is _not_ specified, then default is `none`, but if `trust` _is_ specified, then default is `required`.

https://github.com/aklivity/zilla/blob/f9222ff6b710c9873fa8ee7445934c99bf0b74fe/runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/config/TlsOptionsConfigBuilder.java#L109-L114

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/engine.schema.json
```diff
@@ -437,35 +407,6 @@
...
-                        {
-                            "not":
-                            {
-                                "const": "remote_server"
-                            }
-                        },
-                        "entry": false
-                    }
-                }
-            ],
```
jfallows, line 468: Why did we remove this one?
I think this is trying to enforce generic intent of `remote_server` at binding level to make sure any bindings using `remote_server` do so in a consistent manner instead of treating them as independent.

jfallows (changes requested):

specs/binding-grpc.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/schema/grpc.schema.patch.json
vordimous: This change was intentional since the only binding that uses the global catalog definition is the `grpc` server and it doesn't support all of the options defined in the `#/$defs/cataloged` object. 

Unless it does support those, in which case, I will add them to the docs.

specs/binding-grpc.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/schema/grpc.schema.patch.json
jfallows: Btw, i think by the same token we don't currently do routing by `grpc` service method on `grpc` `client` binding.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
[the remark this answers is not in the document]
vordimous: The yaml validator doesn't like the redundant `oneOf` properties, because it can't tell the single prop and multiple prop options apart. If either of the single prop options are used the validator doesn't know which of the two options it could be. The order of the options didn't seem to change the outcome. 
image

Using an "anyOf" works but is the same thing as having two optional properties. An I am guessing the same issue might occur where the wrong options is used to validate if there isn't enough info for the validator to distinguish a difference.

specs/binding-tls.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tls/config/client.alpn.default.yaml
vordimous: I am good with either, I will update the docs to include this option if it needs to remain.

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/engine.schema.json
vordimous: I was adding `"catalog": false` to most of the bindings so I also added `"entry": false` moving the logic for this into the individual schemas. This specific section is not needed.

I can revert this and implement the catalog logic differently if you want. I went with this since it got rid of some of the complexity of include or exclude global properties.

specs/binding-grpc.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/schema/grpc.schema.patch.json
vordimous: This change broke some tests so I have reverted it. I will capture an issue to make these changes.

specs/binding-grpc.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/schema/grpc.schema.patch.json
jfallows: I think it is supposed to defer cataloged parsing to a generic parser instead of redoing it locally.
I'll double check with @akrambek .

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
jfallows: IMHO, a guard is a guard, this is a reference to a guard used to provide context for an authorization. 😄

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/engine.schema.json
jfallows: I think the issue is that we want a guarantee that it's not possible to diverge accidentally regarding `entry`.
The `entry` property only has meaning when `kind` is `remote_server`, so if possible, I'd like to keep that logic defined once by engine to ensure consistency.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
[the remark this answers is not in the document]
jfallows: The concern is that by making the change we now validate variations that are not supported by the implementation. Can you confirm if the ordering requirement is retained or not by these proposed changes?
Btw, I don't think we get an issue with validation in `zilla` itself.

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
vordimous: Well whichever one you want I will put in both the MQTT schema and HTTP schema

jfallows (changes requested):

specs/binding-kafka-grpc.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/grpc/schema/kafka.grpc.schema.patch.json
```diff
@@ -184,10 +192,17 @@
...
                         }
                     },
-                    "exit": false
+                    "exit": false,
+                    "entry":
+                    {
+                        "title": "Entry",
+                        "type": "string",
+                        "pattern": "^[a-zA-Z]+[a-zA-Z0-9\\._\\-]*$"
+                    }
```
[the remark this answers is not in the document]
vordimous: This should have been removed from a revert commit. It is defined back the way it was.

jfallows (approved):

PR #1248, opened 2024-09-16
create `function` support in `risingwave` binding

ankitk-me:


incubator/binding-risingwave.spec/src/main/scripts/io/aklivity/zilla/specs/binding/risingwave/config/proxy.function.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+---
+name: test
+bindings:
+  app0:
+    type: risingwave
+    kind: proxy
+    options:
+      udf:
+        url: http://localhost:8815
```
jfallows, line 24: We can predict that we'll need separate UDF servers for different languages, such as `java` and `python` over time, even though we are starting with `java` first.

Please make sure we design a syntax that can remain backwards compatible and feels natural when adding `python` UDF server later.

```suggestion
    options:
      udf:
        - server: http://localhost:8815
          language: java # default
```

Initially this array can have max length 1 similar to catalogs.

Later, when we add support for `python`, we can host different user-defined functions on different servers as needed.

```suggestion
    options:
      udf:
        - server: http://udf.zillabase.dev:8815
          language: java # default
        - server: http://udf-python.zillabase.dev:8815
          language: python
```

If needed (even later!), we can further enhance this to host different UDFs on different servers, even for the same language, by adding a pattern for matching UDF names on a specific server.  Then the injection of the `USING LINK` value would be dependent on the UDF name and match a bit like routing.

incubator/binding-risingwave/src/main/java/io/aklivity/zilla/runtime/binding/risingwave/internal/config/RisingwaveOptionsConfigAdapter.java
```diff
@@ -69,6 +76,11 @@ public OptionsConfig adaptFromJson(
             options.kafka(kafka.adaptFromJson(object.getJsonObject(KAFKA_NAME)));
         }
 
+        if (object.containsKey(UDF_NAME))
+        {
+            options.udf(udf.adaptFromJson(object.getJsonObject(UDF_NAME)));
```
jfallows, line 81: See specification feedback about options syntax.

jfallows (changes requested):

incubator/binding-risingwave.spec/src/main/scripts/io/aklivity/zilla/specs/binding/risingwave/schema/risingwave.schema.patch.json
```diff
@@ -83,7 +98,11 @@
                                                 "items":
                                                 {
                                                     "type": "string",
-                                                    "enum": [ "CREATE TOPIC" ]
+                                                    "enum":
+                                                    [
+                                                        "CREATE TOPIC",
+                                                        "CREATE FUNCTION"
+                                                    ]
```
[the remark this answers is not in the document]
ankitk-me: done

incubator/binding-risingwave.spec/src/main/scripts/io/aklivity/zilla/specs/binding/risingwave/config/proxy.function.yaml
```diff
@@ -0,0 +1,30 @@
...
+    routes:
+      - exit: app1
+        when:
+          - commands:
+              - "CREATE FUNCTION"
```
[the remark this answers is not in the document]
ankitk-me: done

incubator/binding-risingwave/src/test/java/io/aklivity/zilla/runtime/binding/risingwave/internal/config/RisingwaveOptionsConfigAdapterTest.java
```diff
@@ -82,7 +82,17 @@ public void shouldReadOptions()
...
+                  "udf": [
+                    {
+                      "server": "http://udf.zillabase.dev:8815",
+                      "language": "java"
+                    },
+                    {
+                      "server": "http://udf-python.zillabase.dev:8815",
+                      "language": "python"
+                    }
+                    ]
```
jfallows, line 95: Please include `udf` in `shouldWriteOptions` test method implementation as well.

incubator/binding-risingwave.spec/src/main/scripts/io/aklivity/zilla/specs/binding/risingwave/schema/risingwave.schema.patch.json
```diff
@@ -83,7 +83,9 @@
                                         }
                                     },
                                     "additionalProperties": false
-                                }
+                                },
```
jfallows, line 86: The `language` above should be `"const": "java"` too, right?
Since we only support the value "java" if present for now.

jfallows (changes requested):

jfallows (approved):

PR #1252, opened 2024-09-18
fix: use integer for challenge type

vordimous:
## Description

updates the type for the challenge options on the jwt guard.


jfallows (approved):

ankitk-me (approved):
LGTM

PR #1259, opened 2024-09-20
Support risingwave include keyword

akrambek:


jfallows (commented):
Suggest renaming to remove `.headers` from scenario name, since from pgsql perspective these are not headers.

Also, do we need scenarios for `create.table.with.primary.key.and.includes` ?

jfallows (approved):

PR #1277, opened 2024-10-03
Support jwt guarded identity via custom token claim

### LINKED ISSUE #1276: Support `jwt` guarded identity via custom token claim
As an app developer, I need to configure my application's concept of user identifier with a stable value that is not dynamically generated.

**Acceptance criteria**
 - Support configuration of a custom `jwt` token claim to return as guarded `identity` value
 - Continue to use `sub` claim by default, for backwards compatibility

**Additional context**
Note that `keycloak` creates a dynamic internal identifier used as `sub` claim in `jwt` access token for authorized users, whereas `preferred_username` claim is required to be unique and is also stable.

akrambek:
## Description

Support jwt guarded identity via custom token claim.

Fixes #1276


runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardContext.java
```diff
@@ -29,11 +29,13 @@ final class JwtGuardContext implements GuardContext
     private final Long2ObjectHashMap<JwtGuardHandler> handlersById;
     private final LongSupplier supplyAuthorizedId;
     private final EngineContext context;
+    private final Configuration config;
```
jfallows, line 32: Can we remove this?

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -72,14 +73,15 @@ public JwtGuardHandler(
...
 
         List<JwtKeyConfig> keysConfig = options.keys;
         if ((keysConfig == null || keysConfig.isEmpty()) && options.keysURL.isPresent())
         {
-            JsonbConfig config = new JsonbConfig()
+            JsonbConfig keyConfig = new JsonbConfig()
                     .withAdapters(new JwtKeySetConfigAdapter());
             Jsonb jsonb = JsonbBuilder.newBuilder()
-                    .withConfig(config)
+                    .withConfig(keyConfig)
```
jfallows, line 84: Let's inline the `JsonbConfig` into the parameter of the `.withConfig(...)` method.

jfallows (changes requested):

jfallows (approved):

PR #1278, opened 2024-10-04
external udf - python support

ankitk-me:


incubator/binding-risingwave.spec/src/main/scripts/io/aklivity/zilla/specs/binding/risingwave/streams/pgsql/create.function.python/client.rpt
```diff
@@ -0,0 +1,54 @@
...
+connected
+
+write zilla:data.ext ${pgsql:dataEx()
+                              .typeId(zilla:id("pgsql"))
+                              .query()
+                                .build()
+                              .build()}
+write "CREATE FUNCTION gcd(int, int) RETURNS int "
+      "LANGUAGE python AS gcd;"
+      [0x00]
```
jfallows, line 37: Risingwave also supports embedded python functions (not just external), where implementation is here with the `CREATE FUNCTION` statement.

See https://docs.risingwave.com/docs/current/udf-python-embedded/

Please add an IT to confirm the embedded one passes through successfully.

jfallows (changes requested):

incubator/binding-risingwave.spec/src/main/scripts/io/aklivity/zilla/specs/binding/risingwave/streams/pgsql/create.function.python/client.rpt
ankitk-me: Added ITs for embedded python functions.

Also, adding e2e testing result to validate the same:

```
dev=> CREATE FUNCTION gcd(a int, b int) RETURNS int LANGUAGE python AS $$
def gcd(a, b):
    while b != 0:
        a, b = b, a % b
    return a
$$;
CREATE_FUNCTION
dev=> SELECT gcd(15, 25);
 gcd
-----
   5
(1 row)
```

jfallows (approved):

PR #1293, opened 2024-10-11
Support cancel request

akrambek:


incubator/binding-risingwave/src/main/java/io/aklivity/zilla/runtime/binding/risingwave/internal/stream/RisingwaveProxyFactory.java
```diff
@@ -412,6 +416,12 @@ private void onAppData(
             }
         }
 
+        private void onAppFlush(
+            final FlushFW flush)
+        {
+            //NOOP
```
jfallows, line 422: Do we need to do anything about flow control here?

jfallows (approved):

PR #1300, opened 2024-10-17
support file write in binding-filesystem & binding-http-filesystem

ankitk-me:


specs/binding-filesystem.spec/src/main/resources/META-INF/zilla/filesystem.idl
```diff
@@ -18,7 +18,10 @@ scope filesystem
     {
         READ_PAYLOAD,
         READ_EXTENSION,
-        READ_CHANGES
+        READ_CHANGES,
+        WRITE_PAYLOAD,
+        WRITE_EXTENSION,
+        WRITE_CHANGES
```
jfallows, line 24: Not sure if we need a separate `WRITE_EXTENSION`, is there a use case for that from `http-filesystem` binding for example?

specs/binding-filesystem.spec/src/main/scripts/io/aklivity/zilla/specs/binding/filesystem/streams/application/write.file.payload/client.rpt
```diff
@@ -0,0 +1,44 @@
...
+    option zilla:window 8192
+    option zilla:transmission "half-duplex"
+
+write zilla:begin.ext ${filesystem:beginEx()
+                                  .typeId(zilla:id("filesystem"))
+                                  .capabilities("WRITE_PAYLOAD")
+                                  .path("index.html")
+                                  .type("text/html")
+                                  .payloadSize(77)
+                                  .tag("f2c912a30f38a124c3249f8e802e0d90")
```
jfallows, line 26: What is the purpose of `tag` when writing the new payload?
Perhaps it is equivalent to `if-match` in `http` `PUT` ?

runtime/binding-http-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/http/filesystem/internal/stream/HttpFileSystemProxyFactory.java
```diff
@@ -54,13 +54,16 @@ public final class HttpFileSystemProxyFactory implements HttpFileSystemStreamFac
...
     private static final String16FW HEADER_STATUS_VALUE_304 = new String16FW("304");
     private static final String8FW HEADER_ETAG_NAME = new String8FW("etag");
     private static final String8FW HEADER_CONTENT_TYPE_NAME = new String8FW("content-type");
     private static final String8FW HEADER_CONTENT_LENGTH_NAME = new String8FW("content-length");
     private static final int READ_PAYLOAD_MASK = 1 << FileSystemCapabilities.READ_PAYLOAD.ordinal();
+    private static final int WRITE_PAYLOAD_MASK = 1 << FileSystemCapabilities.WRITE_PAYLOAD.ordinal();
+    private static final int CREATE_PAYLOAD_MASK = 1 << FileSystemCapabilities.CREATE_PAYLOAD.ordinal();
 
-    private static final Predicate<HttpHeaderFW> HEADER_METHOD_GET_OR_HEAD;
+    private static final Predicate<HttpHeaderFW> ALLOWED_HTTP_METHOD;
```
jfallows, line 66:
```suggestion
    private static final Predicate<HttpHeaderFW> SUPPORTED_HTTP_METHOD;
```

jfallows (changes requested):

runtime/binding-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/filesystem/internal/stream/FileSystemServerFactory.java
```diff
@@ -81,7 +81,13 @@ public final class FileSystemServerFactory implements FileSystemStreamFactory
     private static final int READ_PAYLOAD_MASK = 1 << FileSystemCapabilities.READ_PAYLOAD.ordinal();
     private static final int WRITE_PAYLOAD_MASK = 1 << FileSystemCapabilities.WRITE_PAYLOAD.ordinal();
     private static final int CREATE_PAYLOAD_MASK = 1 << FileSystemCapabilities.CREATE_PAYLOAD.ordinal();
+    private static final int DELETE_PAYLOAD_MASK = 1 << FileSystemCapabilities.DELETE_PAYLOAD.ordinal();
+    private static final String DEFAULT_TAG = "";
```
jfallows, line 85: Can we remove this?
It seems to be a placeholder value for `null` whereas we should just be able to actually use `null` instead, agree?

runtime/binding-filesystem/src/main/java/io/aklivity/zilla/runtime/binding/filesystem/internal/stream/FileSystemServerFactory.java
```diff
@@ -1029,23 +1072,37 @@ private void doAppAbort(
...
+            String error)
         {
             if (FileSystemState.initialOpening(state) && !FileSystemState.initialClosed(state))
             {
                 state = FileSystemState.closeInitial(state);
 
-                doReset(app, originId, routedId, initialId, initialSeq, initialAck, initialMax, traceId, 0L);
+                FileSystemResetExFW.Builder extension = resetExRW
+                    .wrap(extBuffer, 0, extBuffer.capacity())
+                    .typeId(fileSystemTypeId);
```
jfallows, line 1084: Suggest we precreate these and pass them in, instead of the `String` `error`.

specs/binding-filesystem.spec/src/main/resources/META-INF/zilla/filesystem.idl
```diff
@@ -34,5 +35,10 @@ scope filesystem
             string16 tag = null;
             int64 timeout = 0;
         }
+
+        struct FileSystemResetEx extends core::stream::Extension
+        {
+            string16 error;
```
jfallows, line 41: This should be specific error codes (via enum) rather than freeform strings, so that other bindings can know what to expect in terms of different possible errors.

jfallows (changes requested):

jfallows (approved):

Issue #1173, opened 2024-08-01
Replace Ivy APIs for embedded Maven during `zpmw` install


vordimous:
**Is your feature request related to a problem? Please describe.**
The current `zpmw` install takes a long time and affects local homebrew installs and version updates.

**Describe the solution you'd like**
Improve the spread and output observability of the zilla install progress.

PR #1334, opened 2024-11-21
Replace Ivy APIs with embedded Maven during zpmw install

bmaidics:
Fixes https://github.com/aklivity/zilla/issues/1173


manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
```diff
@@ -162,21 +166,24 @@ public void invoke()
                 repositories.removeIf(r -> !r.location.startsWith("file:"));
             }
 
-            ZpmCache cache = new ZpmCache(repositories, cacheDir);
+            List<RemoteRepository> remoteRepositories = getRemoteRepositories(settingsDir, repositories);
```
jfallows, line 169: Are these actually remote repositories, or do they include the file:// local repository as well?

Suggest renaming to `asRemoteRepositories(...)` instead of `getRemoteRepositories(...)`.

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
```diff
@@ -212,26 +219,21 @@ public void invoke()
...
-                logger.sumupProblems();
-            }
-        }
     }
 
-    private void readSettings(
-        Path settingsDir) throws IOException, PlexusCipherException
+    private List<RemoteRepository> getRemoteRepositories(
+        Path settingsDir,
+        List<ZpmRepository> repositories) throws IOException, PlexusCipherException
```
jfallows, line 226: The logic in this method will change with the switch to use standard maven credentials instead.

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmRepository.java
```diff
@@ -25,6 +27,7 @@
 public final class ZpmRepository
 {
     public String location;
+    public String host;
```
jfallows, line 30: Assuming this is temporary until we are using standard maven credentials approach instead of alternative added for ivy credentials.

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmArtifact.java
```diff
@@ -35,7 +33,7 @@ public ZpmArtifact(
     {
         this.id = id;
         this.path = path;
-        this.depends = unmodifiableSet(depends);
+        this.depends = depends;
```
jfallows, line 36: Can this be reverted or do we now need depends to be mutable?

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
```diff
@@ -15,238 +15,451 @@
...
+        Path dir)
     {
-        Map<ZpmDependency, String> imported = new LinkedHashMap<>();
+        return new SessionBuilderSupplier(system)
+            .get()
+                .withLocalRepositoryBaseDirectories(dir)
+                .setRepositoryListener(new ZpmConsoleRepositoryListener())
+                .setTransferListener(new ZpmConsoleTransferListener())
+                .setConfigProperty("aether.generator.gpg.enabled", Boolean.TRUE.toString())
+                .setConfigProperty("aether.syncContext.named.factory", "noop")
```
jfallows, line 189: Are these string property names defined as constants in maven api somewhere so we could use those constants instead?

jfallows (changes requested):

cloud/docker-image/src/main/docker/incubator/debug_krishna.pcap
jfallows, line 1: Please remove. 😉

cloud/docker-image/src/main/docker/incubator/debug.pcap
jfallows, line 1: Please remove. 😉

jfallows (dismissed):

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
```diff
@@ -548,7 +549,7 @@ private void generateDelegate(
                                 String serviceName = servicePath.toString();
                                 String serviceImpl = new String(input.readAllBytes(), UTF_8);
                                 String existing = services.getOrDefault(serviceName, "");
-                                services.put(serviceName, existing.concat(serviceImpl));
+                                services.put(serviceName, existing.concat("\n").concat(serviceImpl));
```
jfallows, line 552: Let's change this so it produces the same output regardless of whether or not a newline is present in the input.

For example, we can collect the implementations as a list of implementation class names (without whitespace) instead of a concatenated string here, and when writing out, we can join on newline via
```
String.join("\n", serviceImpls)
```

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
```diff
@@ -15,238 +15,388 @@
...
+                    new ZpmArtifactId(artifact.getGroupId(), artifact.getArtifactId(), artifact.getVersion());
+                final Set<ZpmArtifactId> depends = new LinkedHashSet<>();
+                children.forEach(c ->
+                {
+                    final Artifact cArtifact = c.getArtifact();
+                    final ZpmArtifactId cid =
+                        new ZpmArtifactId(cArtifact.getGroupId(), cArtifact.getArtifactId(), cArtifact.getVersion());
+                    depends.add(cid);
+                });
+                artifacts.add(new ZpmArtifact(id, artifact.getFile().toPath(), depends));
```
jfallows, line 159: Confirming we no longer need the `noneMatch` check and `version` comparison?

jfallows (changes requested):

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
bmaidics: That's right, we don't need it anymore

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
```diff
@@ -548,8 +548,7 @@ private void generateDelegate(
                                 assert servicePath.getNameCount() == 1;
                                 String serviceName = servicePath.toString();
                                 String serviceImpl = new String(input.readAllBytes(), UTF_8);
-                                String existing = services.getOrDefault(serviceName, "");
-                                services.put(serviceName, existing.concat("\n").concat(serviceImpl));
+                                services.computeIfAbsent(serviceName, s -> new ArrayList<>()).add(serviceImpl);
```
jfallows, line 551: The service implementation can be multi-valued separated by newlines, so we would need to split it here to handle that case properly.

jfallows (changes requested):

jfallows (approved):

PR #1348, opened 2024-12-10
enabling error reporting using `supplyReporter`

ankitk-me:


runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
```diff
@@ -142,18 +142,19 @@ public void run()
...
+        final Consumer<Throwable> reporter = exceptions
             ? e -> e.printStackTrace(System.err)
             : e -> System.err.println(e.getMessage());
 
         final ErrorHandler onError = ex ->
         {
-            report.accept(ex);
+            reporter.accept(ex);
             stop.countDown();
         };
```
jfallows, line 153: There is still code duplication between `zilla start` here and the default `reporter` in `EngineBuilder`.

If we omit the `reporter` then it will default.

Perhaps it should default to reporting the exception and then triggering the error handler by default when onError would otherwise be called, so that we are not forced to explicitly create the reported here just to chain with the error handler.

We should also move the defaulting logic to `EngineConfiguration`. For that to work, we'll need to create an inner functional interface in `EngineConfiguration`, something like `ErrorReporter` which has `void report(Throwable ex)` or similar and default method that returns a different value depending on `exceptions` configuraiton setting in `EngineConfiguration`.

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineContext.java
```diff
@@ -74,6 +75,8 @@ MessageConsumer supplyReceiver(
 
     EventFormatter supplyEventFormatter();
 
+    Consumer<Throwable> supplyReporter();
+
```
jfallows, line 79:
```suggestion
    void report(
        Throwable ex);

```
Perhaps this is sufficient, since the reporter would be created once per `EngineWorker` and reused for the same core anyway.

jfallows (changes requested):

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/identity/TlsClientX509ExtendedKeyManager.java
```diff
@@ -119,6 +124,10 @@ else if (keyTypes != null)
                     }
                 }
             }
+            if (debug)
+            {
+                System.out.printf("No match found for Subject CN and Key Types");
```
jfallows, line 129: We need the actual CN, key types and issuers here please.

runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
```diff
@@ -119,10 +119,8 @@ public void run()
             props.setProperty(ENGINE_WORKERS.name(), Integer.toString(workers));
         }
 
-        if (verbose)
-        {
-            props.setProperty(ENGINE_VERBOSE.name(), Boolean.toString(verbose));
-        }
+        props.setProperty(ENGINE_VERBOSE.name(), Boolean.toString(verbose));
```
jfallows, line 122: This is a change of behavior, because it will explicitly override `verbose` to false if not specified instead of picking up the value from `zilla.properties` or the `-P` value.

Please restore the `if` check and add a similar check for `exceptions` too.

jfallows (changes requested):

jfallows (approved):

PR #1352, opened 2024-12-16
Ztable support and convert transformation logic to state machine 

akrambek:


incubator/binding-risingwave/src/main/java/io/aklivity/zilla/runtime/binding/risingwave/internal/macro/RisingwaveCreateStreamMacro.java
```diff
@@ -175,18 +206,7 @@ public RisingwaveMacroState onReady(
...
-                state = new GrantResourceState();
-                state.onStarted(traceId, authorization);
-            }
-            else
-            {
-                handler.doReady(traceId, authorization, sql.length());
-            }
-
-            return state;
+            return onReadyHandler.onReady(traceId, authorization);
```
akrambek, line 209: onReadyHandler.ready( I think it will be better

incubator/binding-risingwave/src/main/java/io/aklivity/zilla/runtime/binding/risingwave/internal/macro/RisingwaveCreateStreamMacro.java
```diff
@@ -124,6 +143,8 @@ public RisingwaveMacroState onError(
             PgsqlFlushExFW flushEx)
         {
             handler.doFlushProxy(traceId, authorization, flushEx);
+            onReadyHandler = terminateState;
+
             return this;
```
jfallows, line 148: Shouldn't this return an error state instead, which calls `handler.doReady(...)` for `onReady(...)`?
Then I think we don't need to introduce a separate ready handler interface as the transition is captured by moving to a different state.

jfallows (approved):

PR #1359, opened 2024-12-24
Fix failing test when JVM default locale is not US

epieffe:
## Description
The test [PrometheusMetricsPrinterTest.shouldWorkWithMilliseconds()](https://github.com/aklivity/zilla/blob/b3f686c8d6646a3affeba4c2668da0a7801dd3cd/runtime/exporter-prometheus/src/test/java/io/aklivity/zilla/runtime/exporter/prometheus/internal/printer/PrometheusMetricsPrinterTest.java#L104) fails when the default JVM locale does not use  dot (`.`) as the decimal separator, because floating point numbers are formatted differently and the string comparison does not match.

**Solution**: explicitly set the locale to `Locale.US` when formatting decimal numbers. This makes the test pass regardless of the JVM locale settings and generally ensures more consistent behaviour because the output is independent from the JVM local configuration.

### DIFF
--- a/runtime/exporter-prometheus/src/main/java/io/aklivity/zilla/runtime/exporter/prometheus/internal/printer/PrometheusMetricsPrinter.java
+++ b/runtime/exporter-prometheus/src/main/java/io/aklivity/zilla/runtime/exporter/prometheus/internal/printer/PrometheusMetricsPrinter.java
@@ -84,7 +84,7 @@ public class PrometheusMetricsPrinter
         String msFormat =
             "# HELP %s %s\n" +
             "# TYPE %s %s\n" +
-            "%s{namespace=\"%s\",binding=\"%s\"} %f";
+            "%s{namespace=\"%s\",binding=\"%s\"} %s";
         return milliseconds ?
             String.format(msFormat, extName, description, extName, kind, extName,
                 record.namespace(), record.binding(), record.millisecondsValueReader().getAsDouble()) :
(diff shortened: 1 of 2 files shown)


runtime/exporter-prometheus/src/main/java/io/aklivity/zilla/runtime/exporter/prometheus/internal/printer/PrometheusMetricsPrinter.java
```diff
@@ -86,7 +87,7 @@ private String formatScalar(
             "# TYPE %s %s\n" +
             "%s{namespace=\"%s\",binding=\"%s\"} %f";
```
jfallows, line 88:
```suggestion
            "%s{namespace=\"%s\",binding=\"%s\"} %s";
```

If we use `%s` formatter specifier instead, then `double` will box to `Double` and call `Double.toString()` which is locale-independent, so no need to specify locale variant of `String.format(...)`, agree?

jfallows (changes requested):

runtime/exporter-prometheus/src/main/java/io/aklivity/zilla/runtime/exporter/prometheus/internal/printer/PrometheusMetricsPrinter.java
epieffe: Yes, agreed! Thanks!

jfallows (approved):

PR #1362, opened 2025-01-03
fix: kafka header extraction expression in composite zilla.yaml

### LINKED ISSUE #1138: Asyncapi kafka header extraction expression don’t match zilla yaml expressions
Currently, it is printing

```
        headers:
          :status: $.status
          zilla:correlation-id: $.correlationid
```

instead of 

```
        headers:
          :status: ${message.value.status}
          zilla:correlation-id: ${message.value.correlationId}
```

ankitk-me:
Fixes #1138

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/config/KafkaTopicTransformsConfigBuilder.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/config/KafkaTopicTransformsConfigBuilder.java
@@ -18,27 +18,20 @@ package io.aklivity.zilla.runtime.binding.kafka.config;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import io.aklivity.zilla.runtime.engine.config.ConfigBuilder;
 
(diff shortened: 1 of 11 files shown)


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/config/KafkaTopicTransformsConfigBuilder.java
```diff
@@ -89,11 +97,11 @@ public KafkaTopicTransformsConfigBuilder<T> extractHeader(
         if (matcher.reset(path).matches())
         {
             this.extractHeaders.add(new KafkaTopicHeaderType(name,
-                String.format(INTERNAL_VALUE, matcher.group(2))));
+                String.format(INTERNAL_VALUE, matcher.group(2)), externalPath));
         }
         else if (internalMatcher.reset(path).matches())
         {
-            this.extractHeaders.add(new KafkaTopicHeaderType(name, path));
+            this.extractHeaders.add(new KafkaTopicHeaderType(name, path, externalPath));
```
jfallows, line 104: When we build the `KafkaTopicTransformsConfig` there should be no matcher used, just capture the parsed values directly and nothing else.

In constructor of `KafkaTopicType`, where we prepare the `KafkaTopicConfig` for use in the binding for each core, that's where the parsed config transforms should be converted as needed to something like `KafkaTopicTransformType` using matcher etc, instead of using parsed transforms config directly.

jfallows (changes requested):

jfallows (approved):

PR #1384, opened 2025-01-28
Support gRPC client stream/unary oneway

### LINKED ISSUE #642: Support gRPC client streaming to Kafka directly
When streaming `grpc` requests to `kafka`, I'd like the option to disregard replies, in more of a "fire and forget" mode.

**Acceptance criteria**
 - `grpc` client streaming to `kafka` topic with `acks` `none`, no `reply-to` topic
 - no acknowledgement message required back to `grpc` client
 - no idempotency required

**Additional context**
Later in this slack thread https://aklivitycommunity.slack.com/archives/C027SEH2KRC/p1702479321662669

**Tasks**
  - [ ] `grpc-kafka` support for optional `reply-to`, oneway client streaming method
  - [ ] `grpc-kafka` support for optional `reply-to`, unary method produce with empty response

ankitk-me:
Fixes #642

### DIFF
--- a/runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/config/GrpcKafkaWithProduceResult.java
+++ b/runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/config/GrpcKafkaWithProduceResult.java
@@ -94,7 +94,11 @@ public class GrpcKafkaWithProduceResult
         hash.updateHash(correlation.method.value());
         hash.updateHash(method.value());
         hash.updateHash(correlation.replyTo.value());
-        hash.updateHash(replyTo.value());
+
+        if (replyTo != null && replyTo.value() != null)
+        {
+            hash.updateHash(replyTo.value());
+        }
 
         if (overrides != null)
         {
@@ -155,8 +159,13 @@ public class GrpcKafkaWithProduceResult
 
         builder.item(this::service);
         builder.item(this::method);
-        builder.item(this::replyTo);
-        builder.item(this::correlationId);
+
+        if (replyTo != null && replyTo.value() != null)
+        {
+            builder.item(this::replyTo);
+            builder.item(this::correlationId);
+        }
+
         metadata.forEach(m -> builder.item(i -> metadata(i, m)));
     }
 
--- a/runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/config/GrpcKafkaWithProduceResult.java
+++ b/runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/config/GrpcKafkaWithProduceResult.java
@@ -89,14 +89,13 @@ public class GrpcKafkaWithProduceResult
         this.nameBuffer = new ExpandableDirectByteBuffer();
         this.nameBuffer.putStringWithoutLengthAscii(0, META_PREFIX);
 
-        hash.updateHash(correlation.service.value());
-        hash.updateHash(service.value());
-        hash.updateHash(correlation.method.value());
-        hash.updateHash(method.value());
-        hash.updateHash(correlation.replyTo.value());
-
-        if (replyTo != null && replyTo.value() != null)
+        if (hasReplyTo())
         {
+            hash.updateHash(correlation.service.value());
+            hash.updateHash(service.value());
+            hash.updateHash(correlation.method.value());
+            hash.updateHash(method.value());
+            hash.updateHash(correlation.replyTo.value());
             hash.updateHash(replyTo.value());
         }
 
@@ -157,11 +156,10 @@ public class GrpcKafkaWithProduceResult
             overrides.forEach(o -> builder.item(o::header));
         }
 
-        builder.item(this::service);
-        builder.item(this::method);
-
-        if (replyTo != null && replyTo.value() != null)
+        if (hasReplyTo())
         {
+            builder.item(this::service);
+            builder.item(this::method);
             builder.item(this::replyTo);
             builder.item(this::correlationId);
         }
@@ -249,5 +247,10 @@ public class GrpcKafkaWithProduceResult
                     .name(correlation.correlationId.value(), 0, correlation.correlationId.length())
                     .valueLen(hashCorrelationId.sizeof())
                     .value(hashCorrelationId.value(), 0, hashCorrelationId.sizeof()))));
+    }
+
+    public boolean hasReplyTo()
+    {
+        return replyTo != null && replyTo.value() != null;
     }
 }
--- a/runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/stream/GrpcKafkaProxyFactory.java
+++ b/runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/stream/GrpcKafkaProxyFactory.java
@@ -242,13 +242,26 @@ public final class GrpcKafkaProxyFactory implements GrpcKafkaStreamFactory
                 final GrpcKafkaWithProduceResult result =
                     route.with.resolveProduce(authorization, grpcBeginEx);
 
-                newStream = new GrpcProduceProxy(
-                    grpc,
-                    originId,
-                    routedId,
-                    initialId,
-                    resolvedId,
-                    result)::onGrpcMessage;
+                if (result.replyTo() != null && result.replyTo().value() != null)
+                {
+                    newStream = new GrpcProduceProxy(
+                        grpc,
+                        originId,
+                        routedId,
+                        initialId,
+                        resolvedId,
+                        result)::onGrpcMessage;
+                }
+                else
+                {
+                    newStream = new GrpcProduceNoReplyProxy(
+                        grpc,
+                        originId,
+                        routedId,
+                        initialId,
+                        resolvedId,
+                        result)::onGrpcMessage;
+                }
 
                 break;
             }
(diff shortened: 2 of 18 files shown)


specs/binding-grpc-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/grpc/kafka/streams/GrpcProduceIT.java
```diff
@@ -63,6 +63,15 @@ public void shouldExchangeMessageInClientStream() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${grpc}/client.stream.rpc.oneway/client",
+        "${grpc}/client.stream.rpc.oneway/server"})
+    public void shouldExchangeMessageInClientStreamOneway() throws Exception
```
jfallows, line 70:
```suggestion
    public void shouldSendMessagesViaClientStreamOneway() throws Exception
```
Note: exchange implies request-response like echo.

specs/binding-grpc-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/grpc/kafka/streams/GrpcProduceIT.java
```diff
@@ -99,6 +108,15 @@ public void shouldExchangeMessageInUnary() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${grpc}/unary.rpc.oneway/client",
+        "${grpc}/unary.rpc.oneway/server"})
+    public void shouldExchangeMessageInUnaryOneway() throws Exception
```
jfallows, line 115:
```suggestion
    public void shouldSendMessageViaUnaryOneway() throws Exception
```

specs/binding-grpc-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/grpc/kafka/streams/KafkaProduceIT.java
```diff
@@ -63,6 +63,15 @@ public void shouldExchangeMessageInClientStream() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${kafka}/client.stream.rpc.oneway/client",
+        "${kafka}/client.stream.rpc.oneway/server"})
+    public void shouldExchangeMessageInClientStreamOneway() throws Exception
```
jfallows, line 70:
```suggestion
    public void shouldSendMessagesViaClientStreamOneway() throws Exception
```

specs/binding-grpc-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/grpc/kafka/streams/KafkaProduceIT.java
```diff
@@ -99,6 +108,15 @@ public void shouldExchangeMessageInUnary() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${kafka}/unary.rpc.oneway/client",
+        "${kafka}/unary.rpc.oneway/server"})
+    public void shouldExchangeMessageInUnaryOneway() throws Exception
```
jfallows, line 115:
```suggestion
    public void shouldSendMessageViaUnaryOneway() throws Exception
```

jfallows (changes requested):

specs/binding-grpc-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/kafka/streams/kafka/produce/client.stream.rpc.oneway/client.rpt
```diff
@@ -0,0 +1,82 @@
...
+write zilla:data.ext ${kafka:dataEx()
+                                 .typeId(zilla:id("kafka"))
+                                 .merged()
+                                   .produce()
+                                     .deferred(0)
+                                     .partition(-1, -1)
+                                     .key("test")
+                                     .header("zilla:identity", "test")
+                                     .header("zilla:service", "example.OnewayService")
+                                     .header("zilla:method", "OnewayClientStream")
```
jfallows, line 41: I think we discussed these `zilla:service` and `zilla:method` headers not being needed by oneway mapping.
What do you think?

runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/config/GrpcKafkaWithProduceResult.java
```diff
@@ -94,7 +94,11 @@ public class GrpcKafkaWithProduceResult
         hash.updateHash(correlation.method.value());
         hash.updateHash(method.value());
         hash.updateHash(correlation.replyTo.value());
```
jfallows, line 96: Does this also belong in the condition below to verify replyTo is non-null?
Seems like there should be symmetry with builder.item(...) calls further below in `headers(...)` method.

runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/stream/GrpcKafkaProxyFactory.java
```diff
@@ -242,13 +242,26 @@ public MessageConsumer newStream(
...
                     route.with.resolveProduce(authorization, grpcBeginEx);
 
-                newStream = new GrpcProduceProxy(
-                    grpc,
-                    originId,
-                    routedId,
-                    initialId,
-                    resolvedId,
-                    result)::onGrpcMessage;
+                if (result.replyTo() != null && result.replyTo().value() != null)
```
jfallows, line 245: Perhaps add a convenience method `result.hasReplyTo()` with this implementation, so we can avoid repeating the logic in each place that needs to be conditional about presence of replyTo.

runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/stream/GrpcKafkaProxyFactory.java
```diff
@@ -1636,6 +1649,437 @@ private void doGrpcReset(
...
+
+        @Override
+        protected void onKafkaData(
+            long traceId,
+            long authorization,
+            long budgetId,
+            int reserved,
+            int flags,
+            OctetsFW payload,
+            KafkaDataExFW kafkaDataEx)
```
jfallows, line 1914: For produce oneway, there is no data flow from Kafka to grpc client, agreed?

specs/binding-grpc-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/grpc/kafka/streams/kafka/produce/client.stream.rpc.oneway/server.rpt
```diff
@@ -0,0 +1,81 @@
...
+                             .produce()
+                               .deferred(0)
+                               .partition(-1, -1)
+                               .key("test")
+                               .header("zilla:identity", "test")
+                               .header("zilla:service", "example.OnewayService")
+                               .header("zilla:method", "OnewayClientStream")
+                               .build()
+                           .build()}
+read zilla:data.null
```
jfallows, line 78: We are not needing to send a `null` message for oneway, agreed?

jfallows (commented):
Where are we implicitly sending the empty message in the grpc reply to the client?
Naturally that is not coming from Kafka because we are doing produce only oneway.

runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/stream/GrpcKafkaProxyFactory.java
```diff
@@ -1636,6 +1649,402 @@ private void doGrpcReset(
...
+        {
+        }
```
jfallows, line 1916: Please follow the same design as `HttpKafkaProduceNoReply`, where the `onKafkaData` method in super type is non-abstract and empty implementation. Then subclasses are not forced to override the method where it is not needed.

Note: AFAIK, `KafkaProduceProxy` does not actually call `onKafkaData` on `HttpProxy` delegate.

runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/stream/GrpcKafkaProxyFactory.java
```diff
@@ -1707,7 +2116,8 @@ private void doKafkaData(
 
         private void doKafkaEnd(
             long traceId,
-            long authorization)
+            long authorization,
+            boolean tombstone)
```
jfallows, line 2120: Let's avoid changing the signature of `doKafkaEnd` to handle two different use cases that are not needed by the same implementation instance.

Instead, our overall design approach here has been to use strategy objects dedicated to each use case, adjusting the behavior as needed in each implementation class.

Suggest introducing a `KafkaProduceNoReplyProxy` that is a subclass of `KafkaProduceProxy` where `doKafkaEnd` is protected and overridden in the subclass to get the desired behavior.

Then the type of the delegate in the Grpc stream remains `KafkaProduceProxy` but instantiates `KafkaProduceProxy` for `GrpcProduceProxy` and `KafkaProduceNoReplyProxy` for `GrpcProduceNoReplyProxy`. In each case calling `delegate.doKafkaEnd(traceId, authorization)` causes the desired behavior due to the method override.

jfallows (changes requested):

jfallows (approved):

PR #1388, opened 2025-01-30
Use `OpenapiView` to generate composite namespaces

jfallows:
Simplifies `openapi` codebase for composite namespace generation.


runtime/binding-openapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/internal/model/OpenapiSchemaItem.java
```diff
@@ -0,0 +1,42 @@
...
+
+import jakarta.json.bind.annotation.JsonbProperty;
+
+public class OpenapiSchemaItem
+{
+    public String type;
+    public String title;
+    public Integer multipleOf;
+    public Integer maximum;
+    public Boolean exclusiveMaximum;
```
ankitk-me, line 27: Support for `exclusiveMaximum` & `exclusiveMinimum` might be tricky as openapi uses Json Schema Draft 4 but the Zilla Json Model uses Draft 7. And in Draft 7, `exclusiveMaximum` & `exclusiveMinimum` parameters are expected to be of type `Integer`.

https://json-schema.org/understanding-json-schema/reference/numeric#range

ankitk-me (changes requested):

PR #1392, opened 2025-02-03
Do onClientFlush only when stream is not closed

bmaidics:


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientProduceFactory.java
```diff
@@ -1384,7 +1384,7 @@ private void onClientInitialFlush(
                     doClientReplyAbortIfNecessary(traceId);
                     fan.onClientFanMemberClosed(traceId, this);
                 }
-                else
+                else if (!KafkaState.initialClosed(state))
```
jfallows, line 1387: IIUC, it is not possible to receive `onClientInitialFlush` if `KafkaState.initialClosed(state)` is true, agree?

PR #1396, opened 2025-02-04
Use `OpenapiView` and `AsyncapiView` to generate composite namespaces

jfallows:
Simplifies `openapi-asyncapi` codebase for composite namespace generation.


runtime/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/config/composite/OpenapiAsyncapiCompositeGenerator.java
```diff
@@ -0,0 +1,236 @@
...
+                final AsyncapiView asyncapi = AsyncapiView.of(tagIndex++, label, asyncapiParser.parse(payload), configs);
+
+                asyncapiSchemas.add(new AsyncapiSchemaConfig(label, schemaId, asyncapi));
+            }
+        }
+
+        return generate(binding, openapiSchemas, asyncapiSchemas);
+    }
+
+    protected abstract OpenapiAsyncapiCompositeConfig generate(
```
akrambek, line 93: Perhaps it would be better methods are sorted by access level. LIke all public goes first then protected.

jfallows: Yes, we are doing that, first for methods, then for inner classes / interfaces.

akrambek (approved):

PR #1409, opened 2025-02-19
support boolean model

### LINKED ISSUE #1408: Type boolean possible in schema
We are not able to define a boolean parameter in a schema like this
schema:
            type: boolean
            default: false
For now, we use a string with "false/true" but for the future it would be great if we could just use the type boolean to define a schema in zilla.

ankitk-me:
Fixes #1408

### DIFF
--- a/runtime/model-core/src/test/java/io/aklivity/zilla/runtime/model/core/internal/BooleanValidatorTest.java
+++ b/runtime/model-core/src/test/java/io/aklivity/zilla/runtime/model/core/internal/BooleanValidatorTest.java
@@ -14,14 +14,19 @@
  */
 package io.aklivity.zilla.runtime.model.core.internal;
 
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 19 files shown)


runtime/model-core/src/test/java/io/aklivity/zilla/runtime/model/core/internal/BooleanValidatorTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.agrona.DirectBuffer;
+import org.agrona.concurrent.UnsafeBuffer;
+import org.junit.Test;
+
+import io.aklivity.zilla.runtime.engine.EngineContext;
+import io.aklivity.zilla.runtime.engine.model.ValidatorHandler;
+import io.aklivity.zilla.runtime.engine.model.function.ValueConsumer;
+
+public class BooleanValidatorTest
```
jfallows, line 28: Please add negative test, `shouldRejectBooleanMessage()` too.

jfallows (approved):

PR #1430, opened 2025-03-11
Handle reserved budget in HTTP end frame

akrambek:


runtime/binding-grpc-kafka/src/main/java/io/aklivity/zilla/runtime/binding/grpc/kafka/internal/stream/GrpcKafkaProxyFactory.java
```diff
@@ -288,8 +288,8 @@ private abstract class GrpcProxy
         protected long replySeq;
         protected long replyAck;
         protected int replyMax;
-        protected long replyBud;
-        protected int replyPad;
+        protected long replyBudgetId;
+        protected int replyPadding;
```
jfallows: Please revert this rename.

runtime/binding-grpc/src/main/java/io/aklivity/zilla/runtime/binding/grpc/internal/stream/GrpcServerFactory.java
```diff
@@ -404,10 +407,13 @@ private final class GrpcServer
         private long initialSeq;
         private long initialAck;
         private int initialMax;
+
         private long replySeq;
         private long replyAck;
-        private int replyPadding;
+        private long replyBudgetId;
         private int replyMax;
+        private int replyPadding;
```
jfallows, line 415:
```suggestion
        private long replyBud;
        private int replyPad;
        private int replyMax;
```

runtime/binding-grpc/src/main/java/io/aklivity/zilla/runtime/binding/grpc/internal/stream/GrpcServerFactory.java
```diff
@@ -171,7 +172,7 @@ public void doNetAbort(
                 long authorization,
                 String16FW status)
             {
-                server.doGrpcNetStatus(traceId, authorization, status);
+                server.doGrpcNetStatus(traceId, authorization, server.replyPadding, status);
```
jfallows, line 175: Is this correct? (padding for reserved as parameter)

jfallows (changes requested):

runtime/binding-grpc/src/main/java/io/aklivity/zilla/runtime/binding/grpc/internal/stream/GrpcServerFactory.java
akrambek: No, it is not correct it should be 0 as we are handeling abort which was not claimed anywhere. Thanks for the catch.

jfallows (approved):

PR #1438, opened 2025-03-20
Auto discover available host resources and dynamically set internal limits

### LINKED ISSUE #984: Auto discover available host resources and dynamically set internal limits
As a DevOps maintainer optimizing Zilla deployments, I would expect Zilla to run as fast as the host environment allows. In cases where specific performance improvements are needed, I would expect to be able to tune resource limits manually.

**Acceptance criteria**
 - Zilla discovers available resources
 - Zilla appropriately sets the internal resource limits automatically based on available resources
 - There is a way to discover what limits have been set
 - There are documented parameters to manually tune resource limits inside of Zilla

akrambek:
Fixes #984


jfallows (approved):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
```diff
@@ -359,7 +365,7 @@ private static int defaultStreamsBufferCapacity(
     private static int defaultEventsBufferCapacity(
         Configuration config)
     {
-        return ENGINE_BUFFER_SLOT_CAPACITY.get(config) * ENGINE_WORKER_CAPACITY.getAsInt(config);
+        return 4 * ENGINE_BUFFER_SLOT_CAPACITY.get(config);
```
jfallows, line 368: Let's default this to 4 * the default value of buffer slot capacity instead of changing it when buffer slot capacity changes.

I think making this change will fix the build breaking on the PR which is configuring the buffer slot small to test flow control and then running into issues with the events buffer being too small to log the event.

jfallows (changes requested):

jfallows (approved):

PR #1448, opened 2025-04-03
Enforce worker capacity limit across both client and server tcp connections

akrambek:


runtime/binding-tcp/src/main/java/io/aklivity/zilla/runtime/binding/tcp/internal/TcpBindingContext.java
```diff
@@ -33,14 +35,16 @@
 final class TcpBindingContext implements BindingContext
 {
     private final Map<KindConfig, TcpStreamFactory> factories;
+    private final AtomicInteger capacity;
 
     TcpBindingContext(
         TcpConfiguration config,
         EngineContext context)
     {
+        this.capacity = new AtomicInteger(ENGINE_WORKER_CAPACITY.getAsInt(config));
```
jfallows, line 44: This is not shared across cores, so doesn't need to be atomic.
Perhaps `MutableInteger` instead?

runtime/binding-tcp/src/main/java/io/aklivity/zilla/runtime/binding/tcp/internal/TcpBindingContext.java
```diff
@@ -33,14 +35,16 @@
 final class TcpBindingContext implements BindingContext
 {
     private final Map<KindConfig, TcpStreamFactory> factories;
+    private final AtomicInteger capacity;
```
jfallows, line 38: This can just be a local variable, no need to store as a field, right?

runtime/binding-tcp/src/main/java/io/aklivity/zilla/runtime/binding/tcp/internal/stream/TcpClientRouter.java
```diff
@@ -85,7 +88,11 @@ public InetSocketAddress resolve(
 
         try
         {
-            if (beginEx == null)
+            if (capacity.get() <= 0)
+            {
+                //No capacity
+            }
```
jfallows, line 94: Label the try `resolve:` and then `break resolve;` if no capacity (instead of empty if-block).

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #1454, opened 2025-04-10
support `oneOf` `allOf` & `anyOf` schema in `binding-openapi` & `binding-asyncapi`

### LINKED ISSUE #1418: Problem with oneOf ref in schema
Image

We are not able to define this small example with a oneOf Ref. We think that oneOf and allOf is not correctly validated. We see this in the log when we start the application

Image

ankitk-me:
Fixes #1418

### DIFF
--- a/runtime/binding-openapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/internal/view/OpenapiSchemaView.java
+++ b/runtime/binding-openapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/internal/view/OpenapiSchemaView.java
@@ -41,6 +41,7 @@ public final class OpenapiSchemaView
     public final Integer maximum;
     public final List<String> values;
     public final OpenapiSchemaView schema;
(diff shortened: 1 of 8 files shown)


runtime/binding-openapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/internal/view/OpenapiSchemaView.java
```diff
@@ -41,6 +41,7 @@ public final class OpenapiSchemaView
     public final Integer maximum;
     public final List<String> values;
     public final OpenapiSchemaView schema;
+    public List<OpenapiSchemaView> oneOf;
```
jfallows, line 44:
```suggestion
    public final List<OpenapiSchemaView> oneOf;
```

jfallows (changes requested):

jfallows (approved):

PR #1477, opened 2025-05-16
support `guarded` `routes` in TestBindingFactory

ankitk-me:


runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -188,11 +189,32 @@ public MessageConsumer newStream(
...
+        else
+        {
+            RouteConfig route = binding.routes.stream()
+                .filter(r -> r.when.isEmpty())
+                .findFirst()
+                .orElse(null);
+            resolvedId = route != null ? route.id : 0L;
+        }
+
+        if (resolvedId != 0L)
```
jfallows, line 217: I don't think the guard for credentials is related to the need to check authorized on the routes, so only the `guard.reauthorize(...)` call is needed in the `if (guard != null)` block.

Recommend creating a `TestBindingConfig` and attaching that instead of generic `BindingConfig`, then have a `resolve` method on `TestBindingConfig` (used here) that always checks `route.authorized(...)` and `when.isEmpty()`.

That should work for all scenarios and also avoid the code repetition above.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/config/TestBindingRouteConfig.java
```diff
@@ -0,0 +1,40 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.aklivity.zilla.runtime.engine.test.internal.binding.config;
+
+import java.util.function.LongPredicate;
+
+import io.aklivity.zilla.runtime.engine.config.RouteConfig;
+
+public final class TestBindingRouteConfig
```
jfallows, line 22: Following our existing naming conventions, this would be `TestRouteConfig` similar to `HttpRouteConfig`.

jfallows (approved):

PR #1482, opened 2025-05-23
Overriding `correlation-id` via AsyncAPI Specification

### LINKED ISSUE #1471: Support Overriding `correlation-id` via AsyncAPI Specification
Overriding the `correlation-id` header isn't supported in `binding-asyncapi`. 

Support overriding the `zilla:correlation-id` header directly via the AsyncAPI specification. This would allow users to define a custom correlation ID field within their AsyncAPI contract.

**Additional context**
Image

https://www.asyncapi.com/docs/reference/specification/v3.0.0

ankitk-me:
Fixes #1471

### DIFF
--- a/runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiProxyGenerator.java
+++ b/runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiProxyGenerator.java
@@ -545,9 +545,11 @@ public final class AsyncapiProxyGenerator extends AsyncapiCompositeGenerator
 
                 private static final Pattern HEADER_LOCATION_PATTERN = Pattern.compile("([^/]+)$");
                 private static final Pattern PARAMETER_PATTERN = Pattern.compile("\\{([^}]+)\\}");
+                private static final Pattern CORRELATION_HEADERS_NAME = Pattern.compile("\\$message\\.header#\\/(.+)");
 
                 private final Matcher headerLocation = HEADER_LOCATION_PATTERN.matcher("");
                 private final Matcher parameters = PARAMETER_PATTERN.matcher("");
+                private final Matcher correlation = CORRELATION_HEADERS_NAME.matcher("");
 
                 private final List<ProxyRouteHelper> httpKafkaRoutes;
 
@@ -821,6 +823,16 @@ public final class AsyncapiProxyGenerator extends AsyncapiCompositeGenerator
                     if (kafkaOperation.reply != null)
                     {
                         produce.replyTo(kafkaOperation.reply.channel.address);
+                    }
+
+                    AsyncapiMessageView messageView = httpOperation.messages.get(0);
+                    if (messageView.correlationId != null && messageView.correlationId.location != null)
+                    {
+                        String correlationId = messageView.correlationId.location;
+                        if (correlation.reset(correlationId).matches())
+                        {
+                            produce.correlationId(correlation.group(1));
+                        }
                     }
 
                     AsyncapiHttpKafkaOperationBinding httpKafkaBinding = httpOperation.bindings.httpKafka;
--- a/runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/config/HttpKafkaOptionsConfigBuilder.java
+++ b/runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/config/HttpKafkaOptionsConfigBuilder.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2021-2024 Aklivity Inc
+ *
+ * Licensed under the Aklivity Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ *   https://www.aklivity.io/aklivity-community-license/
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package io.aklivity.zilla.runtime.binding.http.kafka.config;
+
+import java.util.function.Function;
+
+import io.aklivity.zilla.runtime.engine.config.ConfigBuilder;
+import io.aklivity.zilla.runtime.engine.config.OptionsConfig;
+
+public final class HttpKafkaOptionsConfigBuilder<T> extends ConfigBuilder<T, HttpKafkaOptionsConfigBuilder<T>>
+{
+    private final Function<OptionsConfig, T> mapper;
+
+    private HttpKafkaCorrelationConfig correlation;
+    private HttpKafkaIdempotencyConfig idempotency;
+
+    HttpKafkaOptionsConfigBuilder(
+        Function<OptionsConfig, T> mapper)
+    {
+        this.mapper = mapper;
+    }
+
+    public HttpKafkaOptionsConfigBuilder<T> correlation(
+        HttpKafkaCorrelationConfig correlation)
+    {
+        this.correlation = correlation;
+        return this;
+    }
+
+    public HttpKafkaOptionsConfigBuilder<T> idempotency(
+        HttpKafkaIdempotencyConfig idempotency)
+    {
+        this.idempotency = idempotency;
+        return this;
+    }
+
+    @Override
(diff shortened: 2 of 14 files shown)


runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaWithProduceResult.java
```diff
@@ -83,7 +84,8 @@ public class HttpKafkaWithProduceResult
         String16FW idempotencyKey,
         List<HttpKafkaWithProduceAsyncHeaderResult> async,
         HttpKafkaWithProduceHash hash,
-        long timeout)
+        long timeout,
+        String16FW correlationId)
```
jfallows, line 88: It seems to make sense to move this up near `replyTo` as they are logically related parameters.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaWithProduceResult.java
```diff
@@ -244,9 +247,19 @@ private void correlationId(
...
+        String16FW name;
+        if (this.correlationId != null)
+        {
+            name = this.correlationId;
+        }
+        else
+        {
+            name = correlation.correlationId;
+        }
+
```
jfallows, line 259:
```suggestion
        final String16FW name = this.correlationId != null ? this.correlationId : correlation.correlationId;

```

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaWithProduceConfigAdapter.java
```diff
@@ -154,6 +166,7 @@ public HttpKafkaWithConfig adaptFromJson(
                 .overrides(newOverrides)
                 .replyTo(newReplyTo)
                 .async(newAsync)
+                .correlationId(correlationId)
```
jfallows, line 169: Move up one, right after `replyTo(...)`.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaWithProduceConfigAdapter.java
```diff
@@ -146,6 +152,12 @@ public HttpKafkaWithConfig adaptFromJson(
             }
         }
 
+        String correlationId = null;
+        if (object.containsKey(CORRELATION_HEADERS_CORRELATION_ID_NAME))
+        {
+            correlationId = object.getString(CORRELATION_HEADERS_CORRELATION_ID_NAME);
+        }
+
```
jfallows, line 160: Move processing code directly after `replyTo`.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaWithProduceConfigAdapter.java
```diff
@@ -91,6 +92,11 @@ public JsonObject adaptToJson(
             object.add(ASYNC_NAME, newAsync);
         }
 
+        if (produce.correlationId != null)
+        {
+            object.add(CORRELATION_HEADERS_CORRELATION_ID_NAME, produce.correlationId.asString());
+        }
+
```
jfallows, line 99: Move processing code directly after `replyTo`.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaOptionsConfigAdapter.java
```diff
@@ -111,45 +113,40 @@ public JsonObject adaptToJson(
...
-
-                newCorrelation = new HttpKafkaCorrelationConfig(newReplyTo, newCorrelationId);
             }
         }
 
-        return new HttpKafkaOptionsConfig(newIdempotency, newCorrelation);
+        return HttpKafkaOptionsConfig.builder()
+            .idempotency(idempotencyConfig.build())
+            .correlation(correlationConfig.build())
+            .build();
```
jfallows, line 150: Move the `options` builder up and add methods like `.idempotency()` that return a `HttpKafkaIdempotencyConfig` with a mapper that sets the `idempotency(...)` automatically when `.build()` is called.

Similar for `.correlation()` to return a `HttpKafkaCorrelationConfig` with a mapper that sets the `correlation(...)` automatically when `.build()` is called.

We shouldn't need to create separate sub builders in the adapter.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/config/HttpKafkaWithProduceConfig.java
```diff
@@ -43,7 +44,8 @@ public final class HttpKafkaWithProduceConfig
         String key,
         List<HttpKafkaWithProduceOverrideConfig> overrides,
         String replyTo,
-        List<HttpKafkaWithProduceAsyncHeaderConfig> async)
+        List<HttpKafkaWithProduceAsyncHeaderConfig> async,
+        String16FW correlationId)
```
jfallows, line 48: Move up, directly after `replyTo`.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/config/HttpKafkaOptionsConfigBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+        this.correlation = correlation;
+        return this;
+    }
+
+    public HttpKafkaOptionsConfigBuilder<T> idempotency(
+        HttpKafkaIdempotencyConfig idempotency)
+    {
+        this.idempotency = idempotency;
+        return this;
+    }
```
jfallows, line 47: For structured fields, use a nested builder that calls this method (private) as the mapper, triggered when `.build()` is called on the nested builder.

Some of the config objects predate this approach, but it is the approach going forward so we can phase it in.

runtime/binding-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/asyncapi/internal/config/composite/AsyncapiProxyGenerator.java
```diff
@@ -823,6 +825,16 @@ private <C> HttpKafkaWithProduceConfigBuilder<C> injectHttpKafkaRouteProduceWith
...
+                    AsyncapiMessageView messageView = httpOperation.messages.get(0);
+                    if (messageView.correlationId != null && messageView.correlationId.location != null)
+                    {
+                        String correlationId = messageView.correlationId.location;
+                        if (correlation.reset(correlationId).matches())
+                        {
+                            produce.correlationId(correlation.group(1));
+                        }
+                    }
+
```
jfallows, line 837: Given that HTTP is implicitly request-response, it doesn't really need a correlation id, but Kafka does to correlate across request and reply topics, so recommend we look for the correlation id in the kafka operation instead of the http operation, agree?

jfallows (changes requested):

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaOptionsConfigAdapter.java
```diff
@@ -134,19 +133,24 @@ public OptionsConfig adaptFromJson(
...
-            .idempotency(idempotencyConfig.build())
-            .correlation(correlationConfig.build())
+            .idempotency()
+                .header(header)
+                .build()
+            .correlation()
+                .replyTo(replyTo)
+                .correlationId(correlationId)
+                .build()
             .build();
```
jfallows, line 154: Suggest we create the builder at top of method.

Then, instead of capturing individual temporary variables that might be null, instead call the builder to inject those parts as we discover them in the `if` blocks above.

Then return `builder.build()` here.

jfallows (changes requested):

jfallows (approved):

PR #1492, opened 2025-06-08
Support custom role claim

### LINKED ISSUE #1476: Custom Role Claim Support in Zilla JWT Validation
### Feature Request
Zilla’s JWT authorization only validates the scope claim (e.g., scope: "openid profile email") for permissions, ignoring custom RBAC claims like:

```
{
  "role": "Admin",         // Ignored
  "groups": ["Viewer"],    // Ignored
  "scope": "read:data"     // Only this is used
}
```
### Requested Change
Extend Zilla’s JWT validator to:

-   Support custom RBAC claims (e.g., role, groups, realm_access.roles) 
-   Configurable claim names (to support standards like Keycloak, Auth0, or custom IdPs).
-   Override scope-only validation  
-   Allow role/groups to work alongside or replace scope checks.

### Backward Compatibility

- Retain scope support for legacy use cases.
- Make custom claims opt-in (no breaking changes).

akrambek:
Fixes #1476


runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/config/JwtOptionsConfig.java
```diff
@@ -46,13 +47,15 @@ public static <T> JwtOptionsConfigBuilder<T> builder(
     JwtOptionsConfig(
         String issuer,
         String audience,
+        String guarded,
```
jfallows: Let's go with `roles` instead of `guarded`.

jfallows (changes requested):

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -179,11 +184,23 @@ public long reauthorize(
                 break authorize;
             }
 
-            List<String> roles = Optional.ofNullable(claims.getClaimValue("scope"))
-                .map(s -> s.toString().intern())
-                .map(s -> s.split("\\s+"))
-                .map(Arrays::asList)
-                .orElse(null);
+            List<String> roles = null;
+            String path = (this.roles != null && !this.roles.isEmpty()) ? this.roles : "scope";
```
jfallows: Schema should prevent empty string value if not permitted here and this code can then be simplified to remove the check.
The value of `roles` should be defaulted in the builder during parse to avoid implementing the defaulting logic here where options are being used.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -179,11 +184,23 @@ public long reauthorize(
...
+                roles = listClaim.stream()
+                                 .map(Object::toString)
+                                 .map(String::intern)
+                                 .collect(Collectors.toList());
+            }
+            else if (claimObj != null)
+            {
+                roles = Arrays.asList(claimObj.toString().split(SCOPE_VALUE_PATTERN));
+                roles.replaceAll(String::intern);
+            }
```
jfallows:
```suggestion
            Object rolesValue = claimValue(claims, this.roles);
            @SuppressWarnings("unchecked")
            List<String> rolesValueAsList = (rolesValue instanceof List)
                    ? (List<String>) rolesValue
                    : Optional.ofNullable(rolesValue)
                        .map(Object::toString)
                        .map(s -> s.split("\\s+"))
                        .map(Arrays::asList)
                        .orElse(null);

            List<String> roles = rolesValueAsList != null
                ? rolesValueAsList.stream()
                    .map(Object::toString)
                    .map(String::intern)
                    .toList()
                : null;
```

runtime/guard-jwt/src/test/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandlerTest.java
```diff
@@ -650,6 +652,43 @@ public void shouldDeauthorize() throws Exception
...
+        Instant now = Instant.now();
+
+        JwtClaims claims = new JwtClaims();
+        claims.setClaim("iss", "test issuer");
+        claims.setClaim("aud", "testAudience");
+        claims.setClaim("sub", "testSubject");
+        claims.setClaim("exp", now.getEpochSecond() + 10L);
+        Map<String, Object> realmAccess = new HashMap<>();
+        realmAccess.put("roles", asList("default-roles-backend", "offline_access", "uma_authorization"));
+        claims.setClaim("realm_access", realmAccess);
```
jfallows:
```suggestion
        claims.setClaim("realm_access",
            Map.of("roles", List.of("default-roles-backend", "offline_access", "uma_authorization")));
```

runtime/guard-jwt/src/test/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandlerTest.java
```diff
@@ -65,6 +65,7 @@ public void shouldAuthorize() throws Exception
             .inject(identity())
             .issuer("test issuer")
             .audience("testAudience")
+            .roles("scope")
```
jfallows: This should already default `roles` to `scope` in the builder `.build()` method, so it can be omitted here and still have the value `scope` in the returned `JwtOptionsConfig` object.

runtime/guard-jwt/src/test/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardTest.java
```diff
@@ -150,6 +150,7 @@ public void shouldNotVerifyRolesWhenInsufficient() throws Exception
                 .inject(identity())
                 .issuer("test issuer")
                 .audience("testAudience")
+                .roles("scope")
```
jfallows: Same feedback here.

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -53,10 +53,14 @@
 
 public class JwtGuardHandler implements GuardHandler
 {
+    private static final String SCOPE_VALUE_PATTERN = "\\s+";
+    private static final String SCOPE_PATH_PATTERN = "\\.";
```
jfallows:
```suggestion
    private static final String SPLIT_VALUE_PATTERN = "\\s+";
    private static final String SPLIT_PATH_PATTERN = "\\.";
```

runtime/guard-jwt/src/main/java/io/aklivity/zilla/runtime/guard/jwt/internal/JwtGuardHandler.java
```diff
@@ -179,11 +184,22 @@ public long reauthorize(
...
+                    ? (List<String>) rolesValue
+                    : Optional.ofNullable(rolesValue)
+                        .map(Object::toString)
+                        .map(s -> s.split(SCOPE_VALUE_PATTERN))
+                        .map(Arrays::asList)
+                        .orElse(null);
+
+            List<String> roles = rolesValueAsList != null
+                ? rolesValueAsList.stream()
+                    .map(Object::toString)
```
jfallows: Can this line be removed since the type of `rolesValueAsList` is already known to be `List<String>`, so individual `role` members are already known to be of type `String`?

jfallows (changes requested):

jfallows (approved):

PR #1496, opened 2025-06-13
Use GaugesLayout for engine worker capacity metric

jfallows:
## Description

Engine metric for worker utilization is an absolute value, so should be recorded via Gauge semantics rather than specifying delta to accumulating Counter semantics.


akrambek (dismissed):
Approved

PR #1499, opened 2025-06-14
Add `engine.workers.capacity` metric and rename other `engine.worker` metrics to `engine.workers`

jfallows:


examples/tcp.echo/compose.yaml
```diff
@@ -15,7 +15,7 @@ services:
       ZILLA_INCUBATOR_ENABLED: "true"
     volumes:
       - ./etc:/etc/zilla
-    command: start -v -e
+    command: start -v -e -w 1 -Pzilla.engine.worker.capacity=32
```
akrambek: is this intendant change?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineWorker.java
```diff
@@ -1031,6 +1054,12 @@ public Metric resolveMetric(
         return metricGroupsByName.get(metricGroupName).supply(metricName);
     }
 
+    public LongConsumer supplyGaugeWriter(
+        long metricId)
```
akrambek, line 1058: so gaguge doesn't need to have namespace and bindingId?

jfallows: It does, this is to helpfully look up the gauge engine metric which as no namespace or binding.

examples/tcp.echo/compose.yaml
jfallows: No, I resolved it in the follow up PR. 😄

akrambek (dismissed):

PR #1509, opened 2025-06-23
Support load balancer health check even at full capacity

### LINKED ISSUE #1495: Zilla can appear unhealthy to tcp health check mechanisms at full engine worker utilization
**Describe the bug**
Zilla unbinds server ports at full engine worker capacity, which defends against exceeding capacity, but also defeats tcp health check mechanisms for those ports, and only rebinds when a tcp server connection is closed (not for tcp client connections).

**To Reproduce**
1. Start any zilla example with kafka cache and one or more bootstrap topics
2. Verify number of tcp connections to and from zilla (baseline connections)
3. Restart zilla example with `-Pzilla.engine.worker.capacity=N` where `N` is the number of baseline connections **_plus 1_**
4. Check metrics for `engine.workers.utilization` (verify almost full utilization)
5. Make connection via `nc` to zilla tcp server port
6. Check metrics for `engine.workers.utilization` (verify now at full utilization)
7. Verify local zilla server port no longer bound (due to full capacity)
8. Stop Kafka container
9. Check metrics for `engine.workers.utilization` (verify no longer at full utilization)
10. Verify local zilla server port no longer bound (even though no longer at full capacity)

**Expected behavior**
Local zilla server port should not unbind at full capacity.
Instead, when zilla is at full capacity it should accept and immediately clean close new tcp connections.
This lets zilla interact cleanly with tcp health check mechanisms, without exceeding capacity.
Zilla would then need the ability to hand off the new connection to a different worker, distributing the load across engine workers.

akrambek:
Fixes #1495


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineBoss.java
```diff
@@ -0,0 +1,181 @@
...
+    {
+        for (BindingConfig binding : namespace.bindings)
+        {
+            BindingController controller = controllersByType.get(binding.type);
+            if (controller != null && binding.kind == KindConfig.SERVER)
+            {
+                controller.attach(binding);
+            }
+        }
+    }
```
jfallows: Please move this private method down.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineBoss.java
```diff
@@ -0,0 +1,181 @@
...
+        NamespaceConfig namespace)
+    {
+        for (BindingConfig binding : namespace.bindings)
+        {
+            BindingController controller = controllersByType.get(binding.type);
+            if (controller != null)
+            {
+                controller.detach(binding);
+            }
+        }
```
jfallows, line 207: I think we need to offer this as a `NamespaceTask` too, to make sure it executes on the boss thread, not the caller thread.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
```diff
@@ -441,6 +442,8 @@ private void unwatch(
     private void register(
         NamespaceConfig namespace)
     {
+        this.boss.attach(namespace).join();
```
```suggestion
        boss.attach(namespace).join();
```

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
```diff
@@ -452,6 +455,8 @@ private void unregister(
     {
         if (namespace != null)
         {
+            boss.detach(namespace);
```
```suggestion
            boss.detach(namespace).join();
```

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineBoss.java
```diff
@@ -0,0 +1,181 @@
...
+    public void doStart()
+    {
+        thread = startOnThread(runner, Thread::new);
+    }
+
+    public void doClose()
+    {
+        try
+        {
+            controllersByType.values().forEach(BindingController::detachAll);
```
jfallows: This `detachAll()` for `EngineBoss` and `BindingController` seems inconsistent with the design of separation between `EngineWorker` and `BindingHandler`.

Instead, we can keep track of the `NamespaceConfig`s that we have attached to the `EngineBoss` and then iterate each `NamespaceConfig` to here in `doClose` and `detach` them all.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/binding/BindingController.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    default void detach(
+        BindingConfig binding)
+    {
+    }
+
+    default void detachAll()
+    {
+    }
```
jfallows: Please remove this method.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/binding/BindingContext.java
```diff
@@ -29,4 +31,10 @@ default void detach(
         BindingConfig binding)
     {
     }
+
+    default void accept(
+        long bindingId,
+        SocketChannel channel)
+    {
+    }
```
jfallows: This is not needed on generic `BindingContext`, only on `TcpBindingContext`, right?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineContext.java
```diff
@@ -185,4 +185,8 @@ LongConsumer supplyMetricWriter(
     MessageReader supplyEventReader();
 
     Clock clock();
+
+    default void dispatch(Runnable task)
+    {
+    }
```
jfallows:
```suggestion
    default void dispatch(
        Runnable task)
    {
        task.run();
    }
```

runtime/binding-tcp/src/test/java/io/aklivity/zilla/runtime/binding/tcp/internal/streams/ServerIT.java
```diff
@@ -414,10 +412,13 @@ public void shouldUnbindRebind() throws Exception
...
         {
             client5.connect(new InetSocketAddress("127.0.0.1", 12345));
-            fail("Client should not be able to connect, engine is at capacity");
+            ByteBuffer buf = ByteBuffer.allocate(1);
+            assertTrue(client5.read(buf) == -1);
+            client5.close();
         }
         catch (IOException ioe)
         {
+            ioe.printStackTrace();
```
jfallows: Now that we don't unbind, we should be able to remove the `try` `catch` block for this part , right?

runtime/binding-tcp/src/main/java/io/aklivity/zilla/runtime/binding/tcp/internal/stream/TcpServerFactory.java
```diff
@@ -150,56 +152,28 @@ public MessageConsumer newStream(
...
         SocketChannel network,
         InetSocketAddress local)
     {
+        TcpBindingConfig binding = bindings.get(bindingId);
         final TcpRouteConfig route = binding.resolve(local);
 
         if (route != null)
         {
-            final TcpServer server = new TcpServer(binding.id, route.id, network);
+            final TcpServer server = new TcpServer(bindingId, route.id, network);
```
jfallows:
```suggestion
            final TcpServer server = new TcpServer(binding.id, route.id, network);
```
We already used `bindingId` to resolve `binding` so at this point it is more correct to use `binding.id`.

runtime/binding-tcp/src/main/java/io/aklivity/zilla/runtime/binding/tcp/internal/stream/TcpClientFactory.java
```diff
@@ -153,15 +160,15 @@ public MessageConsumer newStream(
 
         InetSocketAddress route = null;
 
-        TcpBindingConfig binding = router.lookup(routedId);
+        TcpBindingConfig binding = bindingsById.get(routedId);
         if (binding != null)
         {
-            route = router.resolve(binding, traceId, authorization, beginEx);
+            route = binding.resolve(binding, traceId, authorization, beginEx);
```
jfallows:
```suggestion
            route = binding.resolve(traceId, authorization, beginEx);
```

runtime/binding-tcp/src/main/java/io/aklivity/zilla/runtime/binding/tcp/internal/config/TcpBindingConfig.java
```diff
@@ -68,11 +102,262 @@ public TcpRouteConfig resolve(
             .orElse(null);
     }
 
+    public InetSocketAddress resolve(
+        TcpBindingConfig binding,
+        long traceId,
+        long authorization,
+        ProxyBeginExFW beginEx)
+    {
+        final TcpOptionsConfig options = binding.options;
```
jfallows:
```suggestion
    public InetSocketAddress resolve(
        long traceId,
        long authorization,
        ProxyBeginExFW beginEx)
    {
```

runtime/binding-tcp/src/main/java/io/aklivity/zilla/runtime/binding/tcp/internal/TcpUsageTracker.java
```diff
@@ -21,14 +21,14 @@
...
 import io.aklivity.zilla.runtime.engine.EngineContext;
 
-public final class TcpCapacityTracker
+public final class TcpUsageTracker
 {
     private final int capacity;
     private final LongConsumer recordUsage;
 
-    private int usage;
+    private volatile int usage;
```
jfallows: Did we change this to `volatile` to handle access from engine boss thread during tcp accept?

runtime/binding-tcp/src/main/java/io/aklivity/zilla/runtime/binding/tcp/internal/TcpBindingController.java
```diff
@@ -0,0 +1,167 @@
...
+        ServerSocketChannel[] servers = serversById.remove(binding.id);
+        quietCloseAll(servers);
+    }
+
+    @Override
+    public void detachAll()
+    {
+        serversById.values().forEach(CloseHelper::quietCloseAll);
+        serversById.clear();
+    }
```
jfallows: Please remove this method.

jfallows (changes requested):

runtime/binding-tcp/src/main/java/io/aklivity/zilla/runtime/binding/tcp/internal/TcpUsageTracker.java
akrambek: No sorry I was debugging and forgot to remove it

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineBoss.java
```diff
@@ -142,20 +152,22 @@ public CompletableFuture<Void> attach(
...
+        NamespaceConfig namespace)
+    {
+        NamespaceTask detachedTask = new NamespaceTask(namespace, this::detachNamespace);
+        taskQueue.offer(detachedTask);
+
+        return detachedTask.future();
     }
 
-    public void detach(
+    private void detachNamespace(
```
jfallows: Please move this private method down below `attachNamespace`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineBoss.java
```diff
@@ -179,4 +191,19 @@ public PollerKey supplyPollerKey(
     {
         return poller.register(channel);
     }
+
+    private void attachNamespace(
+        NamespaceConfig namespace)
+    {
+        namespaces.add(namespace);
```
jfallows, line 185: This is good.

We also need symmetrical `namespaces.remove(namespace);` in `detachNamespace` method.

Note that namespaces can be added and removed while engine is still running, so detachNamespace is not only for doClose case, that's why we need to make sure it is kept accurate.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineBoss.java
```diff
@@ -133,6 +137,12 @@ public void doClose()
         }
     }
 
+    public void attachNow(
+        NamespaceConfig namespace)
+    {
+        attach(namespace).join();
+    }
+
```
jfallows, line 145: Do we still need `attachNow` and `detachNow`? If not, let's remove them.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineBoss.java
```diff
@@ -0,0 +1,211 @@
...
+import io.aklivity.zilla.runtime.engine.binding.BindingController;
+import io.aklivity.zilla.runtime.engine.config.BindingConfig;
+import io.aklivity.zilla.runtime.engine.config.KindConfig;
+import io.aklivity.zilla.runtime.engine.config.NamespaceConfig;
+import io.aklivity.zilla.runtime.engine.internal.poller.Poller;
+import io.aklivity.zilla.runtime.engine.poller.PollerKey;
+
+public class EngineBoss implements EngineController, Agent
+{
+    private static final String AGENT_NAME = "EngineBoss";
```
jfallows:
```suggestion
    private static final String AGENT_NAME = "engine/boss";
```

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineWorker.java
jfallows, line 1: Change `agentName` to `this.agentName = String.format("engine/worker#%d", index);`

jfallows (changes requested):

jfallows (approved):

Issue #1508, opened 2025-06-19
Extend Expression Language to Access HTTP Headers for Message Construction and Filtering


Yoni-Weisberg:
### Is your feature request related to a problem? Please describe.
Currently, Zilla's expression language primarily focuses on path parameters (`params`) for dynamic data. However, many critical pieces of information for API Gateway functionality (like user IDs, authentication tokens, trace IDs, etc.) are commonly passed via HTTP request headers.

There is a need to reliably access and utilize HTTP headers within Zilla's expression language for:

1.  **Kafka Message Construction:** To dynamically compose Kafka message keys or headers based on incoming HTTP request headers (e.g., setting a Kafka message key or header using `X-User-Id`).
2.  **Filtering and Routing:** To implement conditional logic, routing rules, or capability filters based on the presence or value of specific HTTP request headers.

Without direct support for accessing HTTP headers in the expression language, users are forced to resort to less intuitive workarounds, or essential header-based logic cannot be implemented directly within Zilla.

### Describe the solution you'd like
I would like the expression language in Zilla to be extended to robustly support direct access to HTTP request headers. This would likely involve a syntax such as `${headers.Header-Name}` (e.g., `${headers.X-User-Id}`).

This capability should be available wherever the expression language is currently used for dynamic values, For example `with.key` and `with.headers` in produce capability, and `filters.key` and `filter.headers` in fetch capability.

This enhancement would significantly increase Zilla's flexibility and power by allowing direct utilization of a common and crucial source of dynamic data from incoming HTTP requests.

### Describe alternatives you've considered
**Forcing Data into Path/Query Parameters:** Attempting to pass data typically conveyed via headers into path or query parameters.
    * **Pros:** Might work if the data can be moved.
    * **Cons:** Violates RESTful best practices for certain types of data (e.g., authorization tokens, user IDs that aren't part of the resource path), and may not be feasible or secure for all use cases.

Issue #1507, opened 2025-06-19
Enhance Fetch Capability Filters for JSON Object Comparison


Yoni-Weisberg:
### Is your feature request related to a problem? Please describe.
Currently, when defining filters for fetching capabilities (e.g., in `http-client` or `kafka-client` bindings' `capabilities.filters` or similar routing/conditional logic fields), the filtering logic appears to be limited primarily to string-based comparisons or simple exact matches. This makes it challenging to filter or route requests based on complex, structured data, particularly when that data is embedded within a Kafka message key or header that is formatted as JSON.

For instance, if a Kafka message key is a JSON string like `{ "userId": "123", "deviceId": "abc" }`, it's not currently possible to directly compare `userId` or `deviceId` fields within a Zilla filter without resorting to inefficient string matching or passing all messages downstream for filtering. This limits Zilla's ability to perform sophisticated, data-driven routing and traffic management at the gateway layer.

### Describe the solution you'd like
I would like Zilla's expression language within **fetch capability filters and other conditional logic points** to support field level access within the key/value/header object. For example
```yaml
filters:
  fields:
    - message.key.userId: "${params.userId}"
    - message.key,deviceId: "${params.deviceId}"
```

This enhancement would enable Zilla to perform advanced, data-driven routing and filtering directly at the gateway layer based on structured information present in Kafka messages or HTTP requests, significantly improving efficiency and simplifying downstream service logic.

### Describe alternatives you've considered
**Using `extract-header` Transformation (Workaround):**
    * One could potentially use an `extract-header` transformation (or similar transformation capabilities within Zilla's Kafka bindings) to parse the JSON key/header and extract specific fields into new, simple Kafka headers. For example, parse `{ "userID": "123", "deviceId": "abc" }` into new headers like `kafka.header.userID: "123"` and `kafka.header.deviceId: "abc"`. Filters could then be applied to these newly created, simple headers.
    * **Pros:** Might achieve the filtering goal.
    * **Cons:** This is a redundant and less intuitive step. It creates unnecessary headers (which may not be needed by the final consumer) simply to facilitate filtering, adding configuration overhead and potentially affecting message size. It feels like a workaround rather than a direct solution.

### Additional context
[...]

Issue #1506, opened 2025-06-19
Enhance Expression Language for String Concatenation and JSON Object Construction in Kafka Bindings


Yoni-Weisberg:
### Is your feature request related to a problem? Please describe.
Currently, when configuring Kafka bindings in Zilla (specifically for `produce` in fields `with.key` and `with.headers`), the expression language for path parameters (`${params.name}`) only supports a full match to a single placeholder. This means that:

1.  Attempting to combine a static string with a parameter, e.g., `key: "device_${params.id}"`, results in the literal string "device_${params.id}" being used without substitution.
2.  Attempting to construct a JSON object using multiple parameters, e.g., `key: '{ "userID": "${params.userId}", "deviceId": "${params.deviceId}"}'`, also fails to substitute the parameter values, resulting in a literal string with the placeholders intact.

This limitation prevents users from constructing more descriptive, application-specific, or structured Kafka message keys/headers directly within Zilla, forcing backend consumers to re-compose keys or headers based on partial information.

The observed behavior stems from the [current implementation](https://github.com/aklivity/zilla/blob/f636e08dfce461407fec31075913dc2c328e859d/runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaWithResolver.java#L345) using `Matcher.matches()` against a `PARAMS_PATTERN` that expects the *entire string* to be a parameter placeholder:
```java
Matcher keyMatcher = paramsMatcher.reset(key0);
if (keyMatcher.matches()) // This only returns true if key0 is *exactly* "${params.id}"
{
    key0 = keyMatcher.replaceAll(replacer);
}
```

### Describe the solution you'd like
I would like the expression language in Zilla's Kafka binding configurations (e.g., for `key`, `headers`, and potentially other fields) to support:

1.  **String Interpolation/Concatenation:** Allowing placeholders within a larger string literal, such as `key: "device_${params.id}".
2.  **JSON Object Construction with Parameter Substitution:** Allowing the creation of JSON-formatted strings where parameter values are correctly interpolated, e.g., `key: '{ "userID": "${params.userId}", "deviceId": "${params.deviceId}"}'`.

This enhancement would enable Zilla to produce more semantically rich and structured Kafka keys or headers, reducing the need for additional transformation logic in downstream Kafka consumers.

### Describe alternatives you've considered
[...]

PR #1512, opened 2025-06-30
Extend Expression Language to Access HTTP Headers and to Use Multiple Parameters

Yoni-Weisberg:
## Description

This is an improvement including features #1508 #1507 #1506


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
```diff
@@ -496,10 +498,12 @@ public void writeEntryStart(
...
-                KafkaCachePaddedKeyFW.Builder paddedKeyBuilder = paddedKeyRW;
-                final int keySize = paddedKeyBuilder.key(k -> k.length(length).value(buffer, index, length)).sizeof();
-                paddedKeyBuilder.padding(logFile.buffer(), 0, paddedKeySize - keySize - SIZE_OF_INT);
-                KafkaCachePaddedKeyFW newPaddedKey = paddedKeyBuilder.build();
+                paddedKeyRW.rewrap();
+                KafkaKeyFW keyFW = keyRW.rewrap().length(length).value(buffer, index, length).build();
+                paddedKeyRW.key(keyFW);
+                final int keyAndMetadataSize = keyFW.sizeof() + KafkaCachePaddedKeyFW.FIELD_OFFSET_PADDING;
+                paddedKeyRW.padding(logFile.buffer(), 0, paddedKeySize - keyAndMetadataSize);
+                KafkaCachePaddedKeyFW newPaddedKey = paddedKeyRW.build();
```
jfallows: IIUC, the intent of this change is to fix an issue with `paddedKeyRW` reuse, which can be simplified to the following:
```suggestion
                KafkaCachePaddedKeyFW.Builder paddedKeyBuilder = paddedKeyRW.rewrap();
                final int keySize = paddedKeyBuilder.key(k -> k.length(length).value(buffer, index, length)).sizeof();
                paddedKeyBuilder.padding(logFile.buffer(), 0, paddedKeySize - keySize - SIZE_OF_INT);
                KafkaCachePaddedKeyFW newPaddedKey = paddedKeyBuilder.build();
                logFile.writeBytes(position, newPaddedKey.buffer(), newPaddedKey.offset(), newPaddedKey.sizeof());
```
Agree?

runtime/binding-sse-kafka/src/main/java/io/aklivity/zilla/runtime/binding/sse/kafka/internal/config/SseKafkaWithResolver.java
```diff
@@ -31,17 +33,20 @@
...
 import io.aklivity.zilla.runtime.binding.sse.kafka.internal.types.String16FW;
 import io.aklivity.zilla.runtime.binding.sse.kafka.internal.types.String8FW;
+import io.aklivity.zilla.runtime.binding.sse.kafka.internal.types.stream.HeaderFW;
 import io.aklivity.zilla.runtime.binding.sse.kafka.internal.types.stream.SseBeginExFW;
 import io.aklivity.zilla.runtime.engine.util.function.LongObjectBiFunction;
 
 public final class SseKafkaWithResolver
 {
     private static final Pattern PARAMS_PATTERN = Pattern.compile("\\$\\{params\\.([a-zA-Z_]+)\\}");
+    private static final Pattern HEADERS_PATTERN = Pattern.compile("\\$\\{headers\\.([a-zA-Z_\\-]+)\\}");
```
jfallows: What use case did you have in mind here for headers in `sse`?

Note: `sse` abstracts away `http` headers, leaving only `schema`, `authority`, `path` (and the `lastId` for reconnect).

jfallows (commented):
@Yoni-Weisberg  Thank you for this PR. I've left a few initial comments for you.

Please update the `examples/http.kafka.crud/etc/zilla.yaml` and `examples/sse.kafka.crud/etc/zilla.yaml` configuration to demonstrate the changes and we'll go from there.

runtime/binding-sse-kafka/src/main/java/io/aklivity/zilla/runtime/binding/sse/kafka/internal/config/SseKafkaWithResolver.java
Yoni-Weisberg: Hi. This is correct if you're using a standard SSE client. However, if you want to consume messages as a stream in an application, you can use a simple HTTP client instead.  
This is the use case we're working with, and we pass a custom header to filter the messages.
Anyway, it will not affect users who use standard SSE client.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
Yoni-Weisberg: Agree, I was gradually understanding the issue, so I guess that was a leftover from a previous attempt to solve it another way.
Anyway, I've fixed it.

runtime/binding-sse-kafka/src/main/java/io/aklivity/zilla/runtime/binding/sse/kafka/internal/config/SseKafkaWithResolver.java
jfallows: I see, so you are implementing the SSE protocol from a simple HTTP client to give you more control over headers so you can apply HTTP request header values as filter criteria?

Do you have any specific need for the SSE protocol, or your use case more generally requires a streaming HTTP response with filter criteria coming from HTTP header values?

Yoni-Weisberg: I don't have specific need for the SSE protocol. if it was possible I would rather use simple HTTP, but as far as I see there is no option for streaming over HTTP with filters

Yoni-Weisberg:
Hey @jfallows, I updated the configuration of `examples/http.kafka.crud/etc/zilla.yaml` and `examples/http.kafka.avro.json/etc/zilla.yaml`. 
However, there is no `examples/sse.kafka.crud`. If you want, I can add an example for demonstrating the use-case of using SSE via curl, but this is not a standard use-case as we discussed above.
Let me know what you want...

runtime/binding-sse-kafka/src/main/java/io/aklivity/zilla/runtime/binding/sse/kafka/internal/config/SseKafkaWithResolver.java
jfallows: Okay, that makes sense thanks.

Suggest we remove the proposed changes to `sse` from this PR, since they break the abstraction, plus you don't need `sse` protocol per se. Instead let's consider adding support for `application/json-lines` in the `http-kafka` binding which is something we've been thinking to add anyway.

https://jsonlines.org/
https://github.com/Azure-Samples/azure-search-openai-demo/pull/968#pullrequestreview-1738239621

All data inside Zilla flows in streams, so even the HTTP GET many is currently streaming the Kafka messages until end of (Kafka) stream. It's just that the end of stream is currently triggered by catching up to live messages in Kafka, rather than continuing to send further live messages.

[HttpKafkaProxyFactory.HttpFetchManyProxy.onKafkaData(...)](https://github.com/aklivity/zilla/blob/develop/runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/stream/HttpKafkaProxyFactory.java#L1169)
[HttpKafkaProxyFactory.HttpFetchManyProxy.onKafkaEnd(...)](https://github.com/aklivity/zilla/blob/develop/runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/stream/HttpKafkaProxyFactory.java#L1208)

Note: unlike the `application/json` content type for merging a stream of Kafka messages in an HTTP response, the `application/json-lines` content type doesn't need any preamble or postamble around all the messages, just a newline separator between messages.

HTTP GET many also supports `with` `filters` to filter for matching Kafka messages in the response so this should work without requiring any further changes to filter the streamed response.

[HttpKafkaProxyFactory.HttpFetchManyProxy.onHttpBegin(...)](https://github.com/aklivity/zilla/blob/develop/runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/stream/HttpKafkaProxyFactory.java#L950)

Suggest we keep this PR scoped to just the expression language enhancements, and handle the `http-kafka` `application/json-lines` changes in a separate PR.

Yoni-Weisberg: Hey,
In "streaming" I meant the response is blocking until client aborts the request (or until a given timeout). and also have an event-id to resume after disconnection. So how do you suggest to implement it with json-lines?
Also, I guess I don't need to revert the entire change in SSE (including the ability to append text to parameters, and to use identity), but only the header part, am I right?

jfallows: @Yoni-Weisberg 
> In "streaming" I meant the response is blocking until client aborts the request (or until a given timeout). and also have an event-id to resume after disconnection. So how do you suggest to implement it with json-lines?

Agreed, `event-id` per message and `last-event-id` on recovering request are important concepts to retain in any streaming HTTP response format that wants to have a delivery guarantee.

There is some overlap with the concept of pagination across a larger data set, with a cursor position to advance to the next page, whereas streaming needs a cursor position to advance to the next message.

https://swagger.io/docs/specification/v3_0/links/#when-to-use-links
```
GET /items?cursor=Q1MjAwNz&limit=100
```

We can apply a similar approach to streaming. For example:
```json
{ "metadata": { "next": "<cursor>" }, "data": {"name": "Alice", "age": 30} }
{ "metadata": { "next": "<cursor>" }, "data": {"name": "Bob", "age": 25} }
{ "metadata": { "next": "<cursor>" }, "data": {"name": "Charlie", "age": 35} }
```

Then on recovery, the next request would include a `cursor` query parameter with the opaque cursor value to pick up from where it left off.

This approach feels reasonably consistent with REST API pagination strategies, and therefore more easily consumed by HTTP clients.

Feedback welcome. 😄 

> Also, I guess I don't need to revert the entire change in SSE (including the ability to append text to parameters, and to use identity), but only the header part, am I right?

Yep, makes sense to keep the new work to support identity and a broader text pattern for `sse` parameters resolved into `with`, we just don't need the headers part as you say. 👍

Yoni-Weisberg: Sounds good. So lets close this PR with the current content. And continue the discussion on another one. I made all the necessary fixes.

Yoni-Weisberg:
Hi, @jfallows, Is there anything left that's preventing this from being merged?

jfallows (dismissed):

jfallows:
> Hi, @jfallows, Is there anything left that's preventing this from being merged?

@Yoni-Weisberg
The build is passing but the `http.kafka.avro.json` example is no longer passing the automated test. 
See https://github.com/Yoni-Weisberg/zilla/blob/develop/examples/http.kafka.avro.json/.github/test.sh for details.

PR #1547, opened 2025-08-29
Add engine authorization config

bmaidics:


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
```diff
@@ -109,6 +110,7 @@ public class EngineConfiguration extends Configuration
         ENGINE_CONFIG_POLL_INTERVAL_SECONDS = config.property("config.poll.interval.seconds", 60);
         ENGINE_NAME = config.property("name", EngineConfiguration::defaultName);
         ENGINE_DIRECTORY = config.property("directory", EngineConfiguration::defaultDirectory);
+        ENGINE_AUTHORIZATION = config.property("authorization", EngineConfiguration::defaultDirectory);
```
jfallows:
```suggestion
        ENGINE_AUTHORIZATION = config.property("authorization");
```

jfallows: If this is intended to be used for config authorization over HTTP, perhaps `config.http.authorization` instead?

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
```diff
@@ -423,6 +430,15 @@ private static int defaultWorkerCapacity(
...
 
+        long totalCalculated = bufferPoolMaxSize + budgetsMaxSize + streamsMaxSize;
+        if (totalCalculated > workerEntriesMemory)
+        {
+            double scaleFactor = (double) workerEntriesMemory / totalCalculated;
+            bufferPoolMaxSize = (long) (bufferPoolMaxSize * scaleFactor);
+            budgetsMaxSize = (long) (budgetsMaxSize * scaleFactor);
+            streamsMaxSize = workerEntriesMemory - bufferPoolMaxSize - budgetsMaxSize;
+        }
+
```
jfallows: Please move this unrelated work to a separate PR.

Btw, this feels like a workaround. Is the issue not actually that the calculation for ...MaxSize above is incorrectly bounded such that `totalCalculated > workerEntriesMemor` becomes unexpectedly possible?

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -363,7 +364,14 @@ public void evaluate() throws Throwable
...
                     final String pollInterval = String.format("PT%dS", config.configPollIntervalSeconds());
-                    fs = FileSystems.newFileSystem(configURI, Map.of("zilla.filesystem.http.poll.interval", pollInterval));
+                    Map<String, String> fileSystemConfig = new HashMap<>();
+                    fileSystemConfig.put("zilla.filesystem.http.poll.interval", pollInterval);
+                    final String configAuth = config.configHttpAuthorization();
+                    if (configAuth != null)
+                    {
+                        fileSystemConfig.put("zilla.filesystem.http.config.authorization", configAuth);
+                    }
+                    fs = FileSystems.newFileSystem(configURI, fileSystemConfig);
```
jfallows: We lost the simple syntax here, let's restore it with the following instead.
```suggestion
                    final String pollInterval = String.format("PT%dS", config.configPollIntervalSeconds());
                    final String configAuth = config.configHttpAuthorization();
                    fs = FileSystems.newFileSystem(configURI, Map.ofEntries(
                        Map.entry("zilla.filesystem.http.poll.interval", pollInterval),
                        Map.entry("zilla.filesystem.http.config.authorization", configAuth)));
```

runtime/filesystem-http/src/main/java/io/aklivity/zilla/runtime/filesystem/http/internal/HttpWatchService.java
```diff
@@ -203,6 +207,7 @@ private HttpWatchKey(
             HttpPath path)
         {
             this.watcher = watcher;
+            this.authorization = watcher.authorization;
```
jfallows: Do we need to store `authorization` in the watch key if it is stored on the watcher and we already have a reference to the watcher in the watch key?

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/streams/application/reconfigure.create.via.http.basic.auth/client.rpt
```diff
@@ -0,0 +1,61 @@
...
+write http:header "If-None-Match" "AAAAAAA"
+write http:header "Prefer" "wait=86400"
+write close
+
+
+connect "zilla://streams/app0"
+    option zilla:window 8192
+    option zilla:transmission "duplex"
+
+connected
```
jfallows: Missing EOL at EOF

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
jfallows: Actually, I think we can simplify this further.

```
Map<String, ?> env = new HashMap<>();
config.properties(env::put, env::put);
fs = FileSystems.newFileSystem(configURI, env);
```
We should probably move that implementation to `Map<String, ?> Configuration.asMap()` instead.
```
fs = FileSystems.newFileSystem(configURI, config.asMap());
```
Then we can later choose to change the implementation of `.asMap()` to return an instance of AbstractMap that delegates to Configuration instead of copying all the values into a HashMap.

This approach also lets us avoid re-specifying configuration at both engine level and http filesystem level.

We can also consider this approach for http(s) based filesystem for Engine itself, not just via EngineRule.

runtime/engine/pom.xml
```diff
@@ -25,7 +25,7 @@
 
   <properties>
     <jacoco.coverage.ratio>0.74</jacoco.coverage.ratio>
-    <jacoco.missed.count>5</jacoco.missed.count>
+    <jacoco.missed.count>6</jacoco.missed.count>
```
bmaidics: Since the `asMap` method is on `Configuration`, but we don't test http configURL in the `engine`, but only in the `http-filesystem` module, I think it's okay to increase this.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -158,6 +164,20 @@ public final class Engine implements Collector, AutoCloseable
 
         this.boss = new EngineBoss(config, errorHandler, bindings);
 
+        final URI configURI = config.configURI();
+
+        if ("http".equals(configURI.getScheme()))
```
jfallows: `https` too?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
```diff
@@ -58,6 +58,7 @@ public class EngineConfiguration extends Configuration
     public static final IntPropertyDef ENGINE_CONFIG_POLL_INTERVAL_SECONDS;
     public static final PropertyDef<String> ENGINE_NAME;
     public static final PropertyDef<String> ENGINE_DIRECTORY;
+    public static final PropertyDef<String> ENGINE_CONFIG_HTTP_AUTHORIZATION;
```
jfallows: Can we remove this now, as we can instead use `HttpFileSystemConfiguration` property?

runtime/engine/pom.xml
jfallows: Let's write a `ConfigurationTest` that verifies the `.asMap()` behavior.

runtime/filesystem-http/src/main/java/io/aklivity/zilla/runtime/filesystem/http/internal/HttpFileSystemConfiguration.java
```diff
@@ -20,7 +20,8 @@
 
 public final class HttpFileSystemConfiguration
 {
-    public static final String POLL_INTERVAL_PROPERTY_NAME = "zilla.filesystem.http.poll.interval";
+    public static final String POLL_INTERVAL_PROPERTY_NAME = "zilla.engine.config.poll.interval";
+    public static final String CONFIG_AUTHORIZATION_PROPERTY_NAME = "zilla.engine.config.http.authorization";
```
jfallows: These names need the correct prefixes.

```suggestion
    public static final String POLL_INTERVAL_PROPERTY_NAME = "zilla.filesystem.http.poll.interval";
    public static final String AUTHORIZATION_PROPERTY_NAME = "zilla.filesystem.http.authorization";
```

runtime/filesystem-http/src/main/java/io/aklivity/zilla/runtime/filesystem/http/internal/HttpFileSystem.java
```diff
@@ -49,6 +50,7 @@ public final class HttpFileSystem extends FileSystem
             .version(HTTP_2)
             .followRedirects(NORMAL)
             .build();
+        this.authorization = (String) env.get("zilla.engine.config.http.authorization");
```
jfallows: Use `FileSystemConfiguration.AUTHORIZATION_PROPERTY_NAME`.

runtime/filesystem-http/src/main/java/io/aklivity/zilla/runtime/filesystem/http/internal/HttpPath.java
```diff
@@ -43,6 +43,7 @@ public final class HttpPath implements Path
 
     private final HttpFileSystem fs;
     private final URI location;
+    private final String authorization;
```
jfallows: Do we need to store the authorization on each `HttpPath`, or just read it from `HttpFileSystem` instead since we have a reference?

runtime/filesystem-http/src/main/java/io/aklivity/zilla/runtime/filesystem/http/internal/HttpWatchService.java
```diff
@@ -50,19 +50,22 @@ public final class HttpWatchService implements WatchService
...
     private final Collection<HttpWatchKey> watchKeys;
     private final BlockingQueue<WatchKey> pendingKeys;
     private final ScheduledExecutorService executor;
 
     private volatile boolean closed;
 
     HttpWatchService(
-        HttpFileSystemConfiguration config)
+        HttpFileSystemConfiguration config,
+        String authorization)
```
jfallows: Can we call `config.authorization()` instead of passing `authorization` separately?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Configuration.java
```diff
@@ -832,6 +835,53 @@ public void properties(
...
+                }
+                for (Entry<String, Object> e : entrySet())
+                {
+                    if (e.getKey().equals(key))
+                    {
+                        return e.getValue();
+                    }
+                }
+                return null;
+            }
```
jfallows: I think we should be able to do better than a linear search here.
```suggestion
                Object value = null;

                if (key instanceof String name)
                {
                    PropertyDef property = definition.properties.get(name);
                    if (property != null)
                    {
                        value = property.get(Configuration.this);
                    }
                }
                
                return value;
            }
```

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/streams/application/reconfigure.create.via.http.basic.auth/server.rpt
```diff
@@ -0,0 +1,67 @@
+#
```
jfallows, line 1: It's good that we verified the engine is initializing the filesystem correctly.

However, we also need http authorization test in http filesystem itself.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
bmaidics: I changed it because now the engine provides the environment for the filesystem, and with that the authorization as well.

jfallows: No, the configuration is global.

We use the same configuration for all the components.

So any specific component configuration should be owned by the component, not the engine, agree?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineConfiguration.java
```diff
@@ -419,9 +426,9 @@ private static int defaultWorkerCapacity(
         long workerEntriesMemory = workerMemory - eventsBufferSize - aggregateMetadataSize;
         long workerEntriesCount = workerEntriesMemory / aggregateEntrySize;
 
-        long bufferPoolMaxSize = min(bufferPoolEntrySize * workerEntriesCount + bufferPoolMetadataSize, Integer.MAX_VALUE);
-        long budgetsMaxSize = min(budgetsEntrySize * workerEntriesCount + budgetsMetadataSize, Integer.MAX_VALUE);
-        long streamsMaxSize = min(streamsEntrySize * workerEntriesCount + streamsMetadataSize, Integer.MAX_VALUE);
+        long bufferPoolMaxSize = min(bufferPoolEntrySize * workerEntriesCount, Integer.MAX_VALUE);
+        long budgetsMaxSize = min(budgetsEntrySize * workerEntriesCount, Integer.MAX_VALUE);
+        long streamsMaxSize = min(streamsEntrySize * workerEntriesCount, Integer.MAX_VALUE);
```
jfallows: Remember to rebase to latest `develop`.

runtime/filesystem-http/src/main/java/io/aklivity/zilla/runtime/filesystem/http/internal/HttpPath.java
```diff
@@ -354,13 +359,19 @@ private HttpRequest newReadRequest()
         return request.build();
     }
 
-    HttpRequest newWatchRequest()
+    HttpRequest newWatchRequest(
+        String authorization)
```
jfallows: Can we not use `fs.authorization()` here also?

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/ReconfigureHttpIT.java
```diff
@@ -39,7 +39,8 @@
 @Ignore
 public class ReconfigureHttpIT
 {
-    public static final String ENGINE_CONFIG_POLL_INTERVAL_SECONDS = "zilla.engine.config.poll.interval.seconds";
+    public static final String HTTP_FILESYSTEM_POLL_INTERVAL_SECONDS = "zilla.filesystem.http.poll.interval";
+    public static final String HTTP_FILESYSTEM_HTTP_AUTHORIZATION = "zilla.filesystem.http.authorization";
```
jfallows: These should no longer be necessary now that we have `HttpFileSystemEnvironment` constants, agree?

runtime/filesystem-http/src/main/java/io/aklivity/zilla/runtime/filesystem/http/HttpFilesystemEnvironment.java
```diff
@@ -0,0 +1,26 @@
...
+public final class HttpFilesystemEnvironment
+{
+    private HttpFilesystemEnvironment()
+    {
+
+    }
+
+    public static final String POLL_INTERVAL_PROPERTY_NAME = "zilla.filesystem.http.poll.interval";
+    public static final String AUTHORIZATION_PROPERTY_NAME = "zilla.filesystem.http.authorization";
+}
```
jfallows: Let's keep these constants here, but change the class to an interface, then add 2 accessor methods:
```
public Duration pollInterval();
public String authorization()
```
...plus a static method.
```
public static HttpFilesystemEnvironment of(
    Map<String, ?> env)
{
    return new HttpFilesystemConfiguration(env);
}
```
This of course requires `HttpFilesystemConfiguration` to implement `HttpFilesystemEnvironment` interface.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Configuration.java
```diff
@@ -832,6 +835,42 @@ public void properties(
...
+                    },
+                    (k, v) ->
+                    {
+                        if (v != null)
+                        {
+                            entries.add(Map.entry(k, v));
+                        }
+                    }
+                );
+                return entries;
```
jfallows:
```suggestion
                Set<Entry<String, Object>> entries = new HashSet<>();
                BiConsumer<String, Object> action = (k, nv) -> Optional.ofNullable(nv).map(v -> Map.entry(k, v)).ifPresent(entries::add);
                properties(action, action);
                return entries;
```

runtime/filesystem-http/src/main/java/io/aklivity/zilla/runtime/filesystem/http/internal/HttpFileSystemConfiguration.java
```diff
@@ -37,4 +38,9 @@ public Duration pollInterval()
         String value = env != null ? (String) env.get(POLL_INTERVAL_PROPERTY_NAME) : null;
         return value != null ? Duration.parse(value) : POLL_INTERVAL_PROPERTY_DEFAULT;
```
jfallows:
```suggestion
        String value = (String) env.get(POLL_INTERVAL_PROPERTY_NAME);
        return value != null ? Duration.parse(value) : POLL_INTERVAL_PROPERTY_DEFAULT;
```

examples/mqtt.proxy.jwt/compose.yaml
```diff
@@ -6,6 +6,7 @@ services:
     hostname: zilla.examples.dev
     ports:
       - 7183:7183
+      - 4444:4444
```
jfallows:
```suggestion
```

examples/mqtt.proxy.jwt/compose.yaml
```diff
@@ -16,6 +17,7 @@ services:
       MOSQUITTO_BROKER_PORT: 1883
       KEYSTORE_PASSWORD: generated
       ZILLA_INCUBATOR_ENABLED: "true"
+      JAVA_OPTIONS: "-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=*:4444"
```
jfallows:
```suggestion
```

Issue #1491, opened 2025-06-05
Migrate engine started and stopped to telemetry events


jfallows:
Engine outputs the `zilla.yaml` configuration, `started` event, exceptions, and `stopping` / `stopped` event directly to `stdout`.

However, the `started` and `stopped` events are helpful for monitoring behavior and detecting misconfigured `zilla` instances.

Migrate `started` and `stopped` events from `stdout` to `telemetry` and provide a convenient way to enable `stdout` exporter from `zilla start` instead of requiring it in `zilla.yaml`.

Only one instance of `stdout` is needed to get the benefit, so `stdout` exporter can be updated to be `no-op` for subsequent instances. This will prevent log messages being written multiple times to `stdout`.

Once this task is completed, the engine `started` and `stopped` events should also successfully export via `otlp`.

PR #1555, opened 2025-09-18
Migrate engine started and stopped to telemetry events

bmaidics:
Fixes https://github.com/aklivity/zilla/issues/1491

### DIFF
--- a/runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporter.java
+++ b/runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporter.java
@@ -15,6 +15,9 @@
 package io.aklivity.zilla.runtime.exporter.stdout.internal;
 
 import java.net.URL;
+import java.util.Comparator;
+import java.util.SortedSet;
+import java.util.concurrent.ConcurrentSkipListSet;
 
 import io.aklivity.zilla.runtime.engine.EngineContext;
 import io.aklivity.zilla.runtime.engine.exporter.Exporter;
@@ -25,11 +28,13 @@ public class StdoutExporter implements Exporter
     public static final String NAME = "stdout";
 
     private final StdoutConfiguration config;
+    private final SortedSet<StdoutExporterHandler> handlers;
 
     public StdoutExporter(
         StdoutConfiguration config)
     {
         this.config = config;
+        this.handlers = new ConcurrentSkipListSet<>(Comparator.comparingInt(System::identityHashCode));
     }
 
     @Override
@@ -56,6 +61,6 @@ public class StdoutExporter implements Exporter
     public ExporterContext supply(
         EngineContext context)
     {
-        return new StdoutExporterContext(config, context);
+        return new StdoutExporterContext(config, context, handlers);
     }
 }
--- a/runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterContext.java
+++ b/runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterContext.java
@@ -14,9 +14,8 @@
  */
 package io.aklivity.zilla.runtime.exporter.stdout.internal;
 
-import java.util.Comparator;
 import java.util.List;
-import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.SortedSet;
 import java.util.function.LongFunction;
 
 import io.aklivity.zilla.runtime.engine.EngineContext;
@@ -34,15 +33,16 @@ public class StdoutExporterContext implements ExporterContext
 {
     private final StdoutConfiguration config;
     private final EngineContext context;
-    private final ConcurrentSkipListSet<StdoutExporterHandler> handlers;
+    private final SortedSet<StdoutExporterHandler> handlers;
 
     public StdoutExporterContext(
         StdoutConfiguration config,
-        EngineContext context)
+        EngineContext context,
+        SortedSet<StdoutExporterHandler> handlers)
     {
         this.config = config;
         this.context = context;
-        this.handlers = new ConcurrentSkipListSet<>(Comparator.comparingInt(System::identityHashCode));
+        this.handlers = handlers;
     }
 
     @Override
@@ -53,13 +53,31 @@ public class StdoutExporterContext implements ExporterContext
         LongFunction<KindConfig> resolveKind)
     {
         StdoutExporterConfig stdoutExporter = new StdoutExporterConfig(exporter);
-        return new StdoutExporterHandler(config, context, stdoutExporter, handlers);
+        return new StdoutExporterHandler(config, this, stdoutExporter);
     }
 
     @Override
     public void detach(
         long exporterId)
     {
+    }
+
+    public void started(
+        StdoutExporterHandler handler)
+    {
+        handlers.add(handler);
+    }
+
+    public boolean enabled(
+        StdoutExporterHandler handler)
+    {
+        return handlers.first() == handler;
+    }
+
+    public void stopped(
+        StdoutExporterHandler handler)
+    {
+        handlers.remove(handler);
     }
 
     public String supplyQName(
@@ -69,7 +87,7 @@ public class StdoutExporterContext implements ExporterContext
     }
 
     public String supplyEventName(
-            int eventId)
+        int eventId)
     {
         return context.supplyEventName(eventId);
     }
--- a/runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterHandler.java
+++ b/runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterHandler.java
@@ -15,9 +15,7 @@
 package io.aklivity.zilla.runtime.exporter.stdout.internal;
 
 import java.io.PrintStream;
-import java.util.SortedSet;
 
-import io.aklivity.zilla.runtime.engine.EngineContext;
 import io.aklivity.zilla.runtime.engine.exporter.ExporterHandler;
 import io.aklivity.zilla.runtime.exporter.stdout.internal.config.StdoutExporterConfig;
 import io.aklivity.zilla.runtime.exporter.stdout.internal.stream.StdoutEventsStream;
@@ -26,25 +24,23 @@ public class StdoutExporterHandler implements ExporterHandler
 {
     private final StdoutExporterContext context;
     private final PrintStream out;
-    private final SortedSet<StdoutExporterHandler> handlers;
 
     private StdoutEventsStream events;
 
     public StdoutExporterHandler(
         StdoutConfiguration config,
-        EngineContext context,
-        StdoutExporterConfig exporter,
-        SortedSet<StdoutExporterHandler> handlers)
+        StdoutExporterContext context,
+        StdoutExporterConfig exporter)
     {
-        this.context = new StdoutExporterContext(config, context);
+        this.context = context;
         this.out = config.output();
-        this.handlers = handlers;
     }
 
     @Override
     public void start()
     {
-        handlers.add(this);
+        context.started(this);
+
         events = new StdoutEventsStream(context, out);
     }
 
@@ -52,17 +48,20 @@ public class StdoutExporterHandler implements ExporterHandler
     public int export()
     {
         int processed = 0;
-        if (handlers.first() == this)
+
+        if (context.enabled(this))
         {
             processed = events.process();
         }
+
         return processed;
     }
 
     @Override
     public void stop()
     {
-        handlers.remove(this);
+        context.stopped(this);
+
         this.events = null;
     }
 }


runtime/engine/pom.xml
```diff
@@ -25,7 +25,7 @@
 
   <properties>
     <jacoco.coverage.ratio>0.74</jacoco.coverage.ratio>
-    <jacoco.missed.count>5</jacoco.missed.count>
+    <jacoco.missed.count>6</jacoco.missed.count>
```
bmaidics: This was needed for the new TypeReference<>(){} when we append to the config file.
If we want to test it in the IT, we need to add the StdoutExporter to the available exporters in the EngineRule somehow so we can read the system patch

runtime/command-start/src/main/java/io/aklivity/zilla/runtime/command/start/internal/airline/ZillaStartCommand.java
```diff
@@ -77,6 +78,10 @@ public final class ZillaStartCommand extends ZillaCommand
         description = "Show exception traces")
     public boolean exceptions;
 
+    @Option(name = {"-o", "--stdout"},
+        description = "Enable stdout telemetry exporter")
+    public boolean stdout;
```
jfallows:
```suggestion
    @Option(name = {"-l", "--logs"},
        description = "Show log events")
    public boolean events;
```

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
```diff
@@ -211,6 +230,55 @@ private void onPathChanged(
...
+                String systemYamlDoc = yamlMapper.writeValueAsString(systemMap);
+
+                try (BufferedWriter writer = Files.newBufferedWriter(
+                    configPath,
+                    StandardOpenOption.APPEND))
+                {
+                    writer.newLine();
+                    writer.write(systemYamlDoc);
+                    writer.newLine();
+                }
```
jfallows: This attempts to modify the persistent file stored on disk, or the HTTP requested config stored remotely.

Instead just need to return the yaml from this method and append in memory.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -499,6 +514,12 @@ public MessageReader supplyEventReader()
         return new EventReader();
     }
 
+    public long supplyTraceId()
+    {
+        EngineWorker worker = workers.get(0);
+        return worker.supplyTraceId();
+    }
+
```
jfallows: We don't really need a `traceId` for engine `started` and `stopped` events.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineWorker.java
```diff
@@ -981,6 +981,32 @@ public void drain()
...
+        while (!exportersDrained && System.nanoTime() - exporterStart < exporterDrainTimeout)
+        {
+            exportersDrained = true;
+
+            for (AgentRunner runner : exportersById.values())
+            {
+                ExporterAgent agent = (ExporterAgent) runner.agent();
+                ExporterHandler handler = agent.getHandler();
+
+                if (handler.export() > 0)
```
jfallows: This is not a good idea as the exporter is expecting to have export method called only on the exporter agent thread.

Perhaps move the while export() > 0 with timeout to the ExporterAgent onClose() method?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
```diff
@@ -211,6 +230,55 @@ private void onPathChanged(
...
+            JsonProvider schemaProvider = JsonProvider.provider();
+
+            for (URL patch : systemNamespacePatches)
+            {
+                InputStream schemaPatchInput = patch.openStream();
+                JsonReader schemaPatchReader = schemaProvider.createReader(schemaPatchInput);
+                JsonArray schemaPatchArray = schemaPatchReader.readArray();
+                JsonPatch schemaPatch = schemaProvider.createPatch(schemaPatchArray);
+
+                systemPatched = schemaPatch.apply(systemPatched);
```
jfallows: These `schema*` names are copied from the other schema-centric code and need to be renamed for purpose here.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
```diff
@@ -211,6 +230,55 @@ private void onPathChanged(
...
+            if (systemPatched != null && !systemPatched.equals(systemBase))
+            {
+                String systemJson = systemPatched.toString();
+                ObjectMapper jsonMapper = new ObjectMapper();
+                Map<String, Object> systemMap = jsonMapper.readValue(systemJson, new TypeReference<>()
+                {
+                });
+
+                ObjectMapper yamlMapper = new ObjectMapper(new YAMLFactory());
+                String systemYamlDoc = yamlMapper.writeValueAsString(systemMap);
```
jfallows: This is solved in `EngineConfigWriter`.
https://github.com/aklivity/zilla/blob/develop/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/config/EngineConfigWriter.java#L194-L201

As used by `EngineWorker`.
https://github.com/aklivity/zilla/blob/develop/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineWorker.java#L818-L819

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/exporter/ExporterAgent.java
```diff
@@ -55,4 +55,9 @@ public String roleName()
     {
         return agentName;
     }
+
+    public ExporterHandler getHandler()
+    {
+        return handler;
+    }
```
jfallows: Please revert this change, it should not be needed if we are keeping things single threaded.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
```diff
@@ -74,6 +89,7 @@ public class EngineManager
     private static final String CONFIG_TEXT_DEFAULT = "name: default\n";
 
     private final Collection<URL> schemaTypes;
+    private final Collection<URL> systemNamespacePatches;
```
```suggestion
    private final Collection<URL> systemConfigs;
```
Try to name things for their purpose instead of the implementation details.
Here if we follow the naming convention for `schemaTypes`, then we have `systemConfigs` instead of `systemNamespacePatches`. The word `system` is already conveying the namespace, much like `URL system()` method added to `Exporter`.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
```diff
@@ -95,6 +111,7 @@ public class EngineManager
 
     public EngineManager(
         Collection<URL> schemaTypes,
+        Collection<URL> systemNamespacePatches,
```
```suggestion
        Collection<URL> systemConfigs,
```

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
```diff
@@ -156,6 +174,7 @@ private void onPathChanged(
         reconfigure:
         try
         {
+            attachSystemNamespaceIfNecessary();
```
jfallows: Perhaps append system namespace after defaulting for missing config?

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -206,15 +209,19 @@ public final class Engine implements Collector, AutoCloseable
...
+        final Collection<URL> systemNamespacePatches = new ArrayList<>();
+        systemNamespacePatches.addAll(exporters.stream().map(Exporter::system).filter(Objects::nonNull).collect(toList()));
+
         final Map<String, Binding> bindingsByType = bindings.stream()
             .collect(Collectors.toMap(b -> b.name(), b -> b));
         final Map<String, Guard> guardsByType = guards.stream()
             .collect(Collectors.toMap(g -> g.name(), g -> g));
 
-        EngineEventContext events = new EngineEventContext(this);
+        this.events = new EngineEventContext(this);
```
jfallows: If events is local scope and passed to EngineManager, then we call `manager.start()`, why not trigger the `started` event there? Same for `stopped` event.

We use `readonly` for `metrics` command.

runtime/engine/pom.xml
jfallows: We don't need to append to the file, but we do need to generate the system namespace YAML in memory. Please see comments on that section regarding `EngineConfigWriter`.

> If we want to test it in the IT, we need to add the StdoutExporter to the available exporters in the EngineRule somehow so we can read the system patch

This would most naturally be tested in the StdoutExporter project, no?

runtime/exporter-otlp/src/test/java/io/aklivity/zilla/runtime/exporter/otlp/internal/EventIT.java
```diff
@@ -62,4 +70,45 @@ public void shouldPostEventLogToOtlpCollector() throws Exception
...
+            public long millis()
+            {
+                try
+                {
+                    Thread.sleep(1000);
+                }
+                catch (InterruptedException e)
+                {
+                    throw new RuntimeException(e);
+                }
```
jfallows, line 135: Any reason why we need to actually sleep if we have full control over the timestamps?

runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporter.java
```diff
@@ -44,6 +44,12 @@ public URL type()
         return getClass().getResource("schema/stdout.schema.patch.json");
     }
 
+    @Override
+    public URL system()
+    {
+        return config.verboseEvents() ? getClass().getResource("system/stdout.system.namespace.patch.json") : null;
```
jfallows:
```suggestion
        return config.verboseEvents() ? getClass().getResource("system/stdout.system.patch.json") : null;
```

runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterContext.java
```diff
@@ -48,6 +51,11 @@ public ExporterHandler attach(
         Collector collector,
         LongFunction<KindConfig> resolveKind)
     {
+        if (!attached.compareAndSet(false, true))
+        {
+            return NOOP_HANDLER;
+        }
+
```
jfallows: This doesn't seem like it would work for the scenario we discussed, where there are 2 namespaces, both with stdout exporter, and the first one is detached, then the second one should still produce output.

However, this approach makes the second one a no-op, so the second one doesn't produce any output even after the first namespace is detached.

To make this work, we need to short circuit in the handler itself, not in attach, so the handler can take over when needed.

runtime/exporter-stdout/pom.xml
```diff
@@ -150,7 +150,10 @@
                   </fileMappers>
                 </artifactItem>
               </artifactItems>
-              <includes>io/aklivity/zilla/specs/exporter/stdout/schema/stdout.schema.patch.json</includes>
+              <includes>
+                io/aklivity/zilla/specs/exporter/stdout/schema/stdout.schema.patch.json,
+                io/aklivity/zilla/specs/exporter/stdout/system/stdout.system.namespace.patch.json
```
jfallows:
```suggestion
                io/aklivity/zilla/specs/exporter/stdout/system/stdout.system.patch.json
```

runtime/exporter-stdout/src/test/java/io/aklivity/zilla/runtime/exporter/stdout/internal/events/EventIT.java
```diff
@@ -63,6 +63,6 @@ public class EventIT
     public void shouldLogEvents() throws Exception
     {
         k3po.finish();
-        output.expect(Pattern.compile("test.net0 \\[[^\\]]+\\] \\[[^\\]]+\\] BINDING_TEST_CONNECTED test event message"));
+        output.expect(Pattern.compile("engine:events \\[[^\\]]+\\] \\[[^\\]]+\\] ENGINE_STARTED Engine Started."));
```
jfallows: We need both right?

specs/engine.spec/src/main/resources/META-INF/zilla/core.idl
```diff
@@ -117,17 +117,31 @@ scope core
...
+
+        struct EngineStoppedEx extends core::stream::Extension
+        {
+            string16 message;
+        }
+
         union EngineEventEx switch (EngineEventType)
         {
-            case CONFIG_WATCHER_FAILED: EngineConfigWatcherFailedEx configWatcherFailed;
+            case CONFIG_WATCHER_FAILED: EngineConfigWatcherFailedEx configWatcherFailed; // TODO: metadata for started/stopped events
```
jfallows: Do we still need this TODO comment?

specs/exporter-stdout.spec/src/main/scripts/io/aklivity/zilla/specs/exporter/stdout/system/stdout.system.namespace.patch.json
```diff
@@ -0,0 +1,9 @@
+[
+  { "op": "add", "path": "/telemetry", "value": {} },
+  { "op": "add", "path": "/telemetry/exporters", "value": {} },
+  {
+    "op": "add",
+    "path": "/telemetry/exporters/stdout_logs_exporter",
```
jfallows:
```suggestion
    "path": "/telemetry/exporters/stdout",
```

runtime/exporter-otlp/src/test/java/io/aklivity/zilla/runtime/exporter/otlp/internal/EventIT.java
bmaidics: Yes, we control the timestamp. But we don't control the timing of the events. In order for both the events to arrive before we export, and have a deterministic ouput, I added the sleep.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
```diff
@@ -206,6 +208,9 @@ public final class Engine implements Collector, AutoCloseable
         schemaTypes.addAll(catalogs.stream().map(Catalog::type).filter(Objects::nonNull).collect(toList()));
         schemaTypes.addAll(models.stream().map(Model::type).filter(Objects::nonNull).collect(toList()));
 
+        final Collection<URL> systemConfigs =
+            exporters.stream().map(Exporter::system).filter(Objects::nonNull).collect(Collectors.toList());
+
```
jfallows: This formatting is a bit easier to read.
```suggestion
        final Collection<URL> systemConfigs = exporters.stream()
            .map(Exporter::system)
            .filter(Objects::nonNull)
            .toList();

```

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/event/EngineEventIT.java
```diff
@@ -61,6 +61,9 @@ public class EngineEventIT
     })
     public void shouldLogEvents() throws Exception
     {
+        k3po.start();
+        Thread.sleep(1000);
```
jfallows, line 65: Can we remove the sleep now that we drain on close for exporter events?

runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporter.java
```diff
@@ -44,6 +44,12 @@ public URL type()
         return getClass().getResource("schema/stdout.schema.patch.json");
     }
 
+    @Override
+    public URL system()
+    {
+        return config.verboseEvents() ? getClass().getResource("system/stdout.system.patch.json") : null;
```
jfallows:
```suggestion
        return config.verboseEvents()
            ? getClass().getResource("system/stdout.system.patch.json")
            : null;
```

runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterHandler.java
```diff
@@ -40,18 +44,31 @@ public StdoutExporterHandler(
     @Override
     public void start()
     {
-        events = new StdoutEventsStream(context, out);
+        active = ACTIVE.compareAndSet(false, true);
+        if (active)
+        {
+            events = new StdoutEventsStream(context, out);
+        }
```
jfallows: This seems like it works to prevent subsequent stdout from producing output, but doesn't activate already started secondary stdout exporter when the namespace with the primary stdout exporter is stopped.

Also `static` scope might be a bit of a broad hammer, since it is not scoped by the StdoutExporter instance.

In cases like this, we need to keep a reference to all handlers in the StdoutExporter, so we can promote a new handler as primary, and then only produce output in `export()` for the handler that is primary according to `StdoutExporter` state.

So one approach would be to have a `ConcurrentSkipListSet<StdoutExporterHandler>` at `StdoutExporter` scope, passing it to `StdoutExporterContext` and `StdoutExporterHandler`.

In `start()` method, add the handler to the set.
In `export()` method, conditionally do output if `set.first() == this`.
In `stop()` method, remove the handler from the set.

Let me know what you think.

runtime/exporter-stdout/src/test/java/io/aklivity/zilla/runtime/exporter/stdout/internal/events/EventIT.java
```diff
@@ -63,6 +63,7 @@ public class EventIT
     public void shouldLogEvents() throws Exception
     {
         k3po.finish();
-        output.expect(Pattern.compile("test.net0 \\[[^\\]]+\\] \\[[^\\]]+\\] BINDING_TEST_CONNECTED test event message"));
+        output.expect(Pattern.compile("engine:events \\[[^\\]]+\\] \\[[^\\]]+\\] ENGINE_STARTED Engine Started."),
+            Pattern.compile("test.net0 \\[[^\\]]+\\] \\[[^\\]]+\\] BINDING_TEST_CONNECTED test event message"));
```
jfallows: Just a bit easier to read.

```suggestion
        output.expect(
            Pattern.compile("engine:events \\[[^\\]]+\\] \\[[^\\]]+\\] ENGINE_STARTED Engine Started."),
            Pattern.compile("test.net0 \\[[^\\]]+\\] \\[[^\\]]+\\] BINDING_TEST_CONNECTED test event message"));
```

runtime/exporter-stdout/src/test/java/io/aklivity/zilla/runtime/exporter/stdout/internal/events/StdoutOutputRule.java
```diff
@@ -65,19 +66,19 @@ public void evaluate() throws Throwable
...
-
-                    final String actual = in.readLine();
-                    final Pattern expect = Objects.requireNonNull(expected);
-
-                    assertThat(actual, matchesPattern(expect));
+                    for (Pattern expected : expectedPatterns)
+                    {
+                        String actual = in.readLine();
+                        assertThat(actual, matchesPattern(expected));
+                    }
```
jfallows:
```suggestion
                    for (Pattern expected : expecteds)
                    {
                        String actual = in.readLine();
                        assertThat(actual, matchesPattern(expected));
                    }
```

jfallows (changes requested):

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/internal/event/EngineEventIT.java
bmaidics: It is not needed for that. If we don't have the sleep, we immediately close the engine, not letting the handshake complete.

runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterHandler.java
```diff
@@ -15,60 +15,61 @@
...
 import io.aklivity.zilla.runtime.exporter.stdout.internal.stream.StdoutEventsStream;
 
-public class StdoutExporterHandler implements ExporterHandler
+public class StdoutExporterHandler implements ExporterHandler, Comparable<StdoutExporterHandler>
 {
-    private static final AtomicBoolean ACTIVE = new AtomicBoolean(false);
-
     private final StdoutExporterContext context;
     private final PrintStream out;
+    private final ConcurrentSkipListSet<StdoutExporterHandler> handlers;
```
jfallows:
```suggestion
    private final SortedSet<StdoutExporterHandler> handlers;
```

Narrow the type to the interface providing `.first()` method, keeping (concurrent) implementation details separate.

runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterHandler.java
```diff
@@ -15,60 +15,61 @@
...
+        handlers.remove(this);
         this.events = null;
     }
+
+    @Override
+    public int compareTo(
+        StdoutExporterHandler other)
+    {
+        return Integer.compare(System.identityHashCode(this), System.identityHashCode(other));
+    }
```
jfallows: Remove this method and the implements in the class definition.
Instead pass `Comparator.comparingInt(System::identityHashCode)` as the comparator in the `ConcurrentSkipListSet` constructor.

runtime/exporter-stdout/src/main/java/io/aklivity/zilla/runtime/exporter/stdout/internal/StdoutExporterContext.java
jfallows:
```suggestion
    private final SortedSet<StdoutExporterHandler> handlers;
```

PR #1583, opened 2025-10-07
Append local config if configured

bmaidics:


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
```diff
@@ -181,6 +183,15 @@ private void onPathChanged(
...
 
+            if (localConfigPath != null)
+            {
+                String localYamlDoc = Files.exists(localConfigPath) ? Files.readString(localConfigPath) : null;
+                if (localYamlDoc != null)
+                {
+                    newConfigText = newConfigText + "\n" + localYamlDoc;
+                }
+            }
+
```
```suggestion
            if (localConfigPath != null &&
                Files.exists(localConfigPath))
            {
                String localYamlDoc = Files.readString(localConfigPath);
                newConfigText = localYamlDoc + "\n" + newConfigText;
            }

```

Let's simplify the null checks and also put the local config first.

PR #1596, opened 2025-10-26
Add basic auth username/password accessor to HTTP server

bmaidics:


jfallows (changes requested):
Please restructure the code with less indentation and put credentials approach as first, with basic case as else.

jfallows (approved):

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -51,10 +54,12 @@
 
 public final class HttpBindingConfig
 {
+    private static final Pattern BASIC_FORMAT_PATTERN =
+        Pattern.compile("^\\s*Basic\\s+(?<format>(:?[^\\\\{]*\\{[^}]*}[^\\\\{]*)+)$");
```
jfallows:
```suggestion
        Pattern.compile("^\\s*Basic\\s+(?<format>(:?[^\\\\{]*\\{[^}]+}[^\\\\{]*)+)$");
```

Issue #1518, opened 2025-07-16
Support populating identity override without guarded route


ankitk-me:
**Is your feature request related to a problem? Please describe.**

In the `http-kafka` binding, `zilla:identity` overrides (e.g., ${guarded['guard0'].identity}) fail to populate unless explicitly configured under a guarded route. 

**Describe the solution you'd like**

The following configuration should automatically populate zilla:identity without requiring guarded route declarations:

```
routes:
  - when:
      - method: POST
        path: /events/{id}
    exit: kafka_cache_client_b
    with:
      overrides:
        zilla:identity: ${guarded['lsauthgaurd'].identity}
```

**Additional context**
https://github.com/aklivity/zilla/issues/1475

nageshwaravijay1117:
@jfallows, could you please assign this issue to me?

jfallows:
> [@jfallows](https://github.com/jfallows), could you please assign this issue to me?

Done. ✅

PR #1600, opened 2025-10-28
Support populating identity override without guarded route

nageshwaravijay1117:
Added support for guard identity expressions without explicit guarded routes

Implemented two-phase guard processing in EngineManager to enable guard expressions like ${guarded['<guard-name>'].identity} and ${guarded['<guard-name>'].attributes.*} to work without declaring the guard in a route's guarded section.

Fixes https://github.com/aklivity/zilla/issues/1518.

### DIFF
--- a/runtime/binding-sse-kafka/src/main/java/io/aklivity/zilla/runtime/binding/sse/kafka/internal/config/SseKafkaRouteConfig.java
+++ b/runtime/binding-sse-kafka/src/main/java/io/aklivity/zilla/runtime/binding/sse/kafka/internal/config/SseKafkaRouteConfig.java
@@ -49,44 +49,26 @@ public final class SseKafkaRouteConfig
     {
         this.id = route.id;
 
-        // Build identifiers and attributors maps using EngineContext
         final Map<String, LongFunction<String>> identifiers = new HashMap<>();
         final Map<String, LongObjectBiFunction<String, String>> attributors = new HashMap<>();
 
-        // Extract guard names referenced in expressions
-        Set<String> referencedGuardNames = new HashSet<>();
+        Set<String> guardNames = new HashSet<>();
         if (route.with != null)
         {
             SseKafkaWithConfig withConfig = (SseKafkaWithConfig) route.with;
-            referencedGuardNames = SseKafkaWithResolver.extractGuardNames(withConfig);
+            guardNames = SseKafkaWithResolver.extractGuardNames(withConfig);
         }
 
-        // Only add guards that are actually referenced in expressions
-        for (String guardName : referencedGuardNames)
+        for (String guardName : guardNames)
         {
-            // Try to find guard in route.guarded first
-            route.guarded.stream()
-                .filter(guarded -> guarded.name.equals(guardName))
-                .findFirst()
-                .ifPresentOrElse(
-                    guarded ->
-                    {
-                        identifiers.put(guarded.name, guarded.identity);
-                        attributors.put(guarded.name, guarded.attributes);
-                    },
-                    () ->
-                    {
-                        // Not found in route.guarded, resolve via EngineContext
-                        long guardId = context.supplyTypeId(guardName);
-                        GuardHandler guard = context.supplyGuard(guardId);
+            long guardId = route.resolveId.applyAsLong(guardName);
+            GuardHandler guard = context.supplyGuard(guardId);
 
-                        if (guard != null)
-                        {
-                            identifiers.put(guardName, guard::identity);
-                            attributors.put(guardName, guard::attribute);
-                        }
-                    }
-                );
+            if (guard != null)
+            {
+                identifiers.put(guardName, guard::identity);
+                attributors.put(guardName, guard::attribute);
+            }
         }
 
         final LongFunction<String> defaultIdentifier = a -> null;


jfallows:
@nageshwaravijay1117  I think the fix we want is in `http-kafka` and `sse-kafka` bindings.

They each rely on `route.guarded` to pick up `.identity` and `.attributes` expression support.

However, we already have `EngineContext.supplyGuard(guardId)` and `EngineContext.supplyTypeId(guardName) => guardId`, so there's no need to depend on `route.guarded` there.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfig.java
```diff
@@ -41,12 +47,53 @@ public final class HttpKafkaRouteConfig
...
+        // Build identifiers and attributors maps using EngineContext
+        final Map<String, LongFunction<String>> identifiers = new HashMap<>();
+        final Map<String, LongObjectBiFunction<String, String>> attributors = new HashMap<>();
+
+        // First, add explicitly guarded routes (from route.guarded)
+        for (var guarded : route.guarded)
+        {
+            identifiers.put(guarded.name, guarded.identity);
+            attributors.put(guarded.name, guarded.attributes);
+        }
```
jfallows: If the guards are not referenced in the expressions, do we need them at runtime?

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfig.java
```diff
@@ -82,6 +126,56 @@ public HttpKafkaRouteConfig(
...
+                            Matcher matcher = guardPattern.matcher(header.value);
+                            while (matcher.find())
+                            {
+                                guardNames.add(matcher.group(1));
+                            }
+                        }
+                    });
+                }
+            });
+        });
```
jfallows: Please use stream and flatMap instead of these deeply nested loops.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfig.java
```diff
@@ -41,12 +47,53 @@ public final class HttpKafkaRouteConfig
...
+        if (route.with != null)
+        {
+            HttpKafkaWithConfig withConfig = (HttpKafkaWithConfig) route.with;
+            Set<String> referencedGuardNames = extractReferencedGuards(withConfig);
+
+            for (String guardName : referencedGuardNames)
+            {
+                // Skip if already added from route.guarded
+                if (!identifiers.containsKey(guardName))
+                {
```
jfallows: If we don't need the explicitly referenced guards for expressions, then this condition would always be true and can therefore be removed.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfig.java
```diff
@@ -41,12 +47,53 @@ public final class HttpKafkaRouteConfig
...
+                        {
+                            // Create identity and attribute functions that delegate to GuardHandler
+                            identifiers.put(guardName, guard::identity);
+                            attributors.put(guardName, guard::attribute);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        // Guard not found or invalid - ignore and continue
+                    }
```
jfallows: Do any of the methods actually throw exceptions, if not then we don't need the try-catch block.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfig.java
```diff
@@ -82,6 +126,56 @@ public HttpKafkaRouteConfig(
...
     }
 
+    private Set<String> extractReferencedGuards(
+        HttpKafkaWithConfig withConfig)
+    {
+        Set<String> guardNames = new HashSet<>();
+
+        // Regex pattern matches ${guarded['<name>'].identity} and ${guarded['<name>'].attributes.*}
+        // Capture group (1) extracts the guard name
+        Pattern guardPattern = Pattern.compile("\\$\\{guarded\\['([a-zA-Z]+[a-zA-Z0-9\\._\\:\\-]*)'\\]");
```
jfallows: The relevant patterns are already defined in `HttpRouteWithConfg`, so perhaps it would be better to the work of extracting the guards there instead.

runtime/binding-http-kafka/src/test/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfigTest.java
```diff
@@ -0,0 +1,257 @@
...
+import io.aklivity.zilla.runtime.binding.http.kafka.config.HttpKafkaOptionsConfig;
+import io.aklivity.zilla.runtime.binding.http.kafka.config.HttpKafkaWithConfig;
+import io.aklivity.zilla.runtime.binding.http.kafka.config.HttpKafkaWithProduceConfig;
+import io.aklivity.zilla.runtime.binding.http.kafka.config.HttpKafkaWithProduceOverrideConfig;
+import io.aklivity.zilla.runtime.engine.EngineContext;
+import io.aklivity.zilla.runtime.engine.config.GuardedConfig;
+import io.aklivity.zilla.runtime.engine.config.RouteConfig;
+import io.aklivity.zilla.runtime.engine.guard.GuardHandler;
+
+public class HttpKafkaRouteConfigTest
```
jfallows: We use loosely coupled ITs to test these rather than tightly coupled unit tests. Please see other ITs with scripts and config as examples.

The idea is that the ITs rely only on the syntax that is already officially supported as a backwards compatible syntax, so it will remain stable even if the code is completely refactored.

Unit tests like these often need to be refactored when the code is refactored, then losing the benefit of having unchanged tests when updating the implementation.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfig.java
nageshwaravijay1117: @jfallows  No, we don’t need runtime guards if they aren’t referenced in any expressions. I’ll update the changes.

runtime/binding-http-kafka/src/test/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfigTest.java
nageshwaravijay1117: understood @jfallows , will remove the unit test cases added as part of this change.

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfig.java
```diff
@@ -41,12 +45,50 @@ public final class HttpKafkaRouteConfig
...
+
+        // Extract guard names referenced in expressions
+        Set<String> referencedGuardNames = new HashSet<>();
+        if (route.with != null)
+        {
+            HttpKafkaWithConfig withConfig = (HttpKafkaWithConfig) route.with;
+            referencedGuardNames = HttpKafkaWithResolver.extractGuardNames(withConfig);
+        }
+
+        // Only add guards that are actually referenced in expressions
```
jfallows: Isn't it correct to say that `route.guarded` can be ignored and instead we only want to process all of the guards referenced by guard name in the expressions?

jfallows (changes requested):

runtime/binding-http-kafka/src/main/java/io/aklivity/zilla/runtime/binding/http/kafka/internal/config/HttpKafkaRouteConfig.java
nageshwaravijay1117: I think route.guarded entries contain pre-resolved identity and attribute functions set during config initialization. Bypassing them would require redundant guard resolution via EngineContext and lose route-specific guard configurations.

jfallows: The concepts of guarded routes and guarded expressions are distinct.

Guarded routes are trying to make sure that zero or more specific roles are active for a valid session of a specific guard, such that the route is ignored if the session is invalid or the required roles are not active for the session. This is already handled by the engine, just the binding needs to ask the route to verify passing a valid session.

Guarded expressions are references to attributes of the session for a specific guard. It does not require any specific roles to be active. In each case we need the guard handler to resolve the guarded identity or guarded attribute expression.

The original mistake in the design was requiring guarded expressions on a route to only work on a guarded route, whereas this enhancement is to break that coupling, so that routes do not need to be guarded for guarded expressions to work.

So as part of these changes, we can verify that it is viable to remove the exposure of identity resolver and attribute resolver from the route as that can only be populated for guarded routes and therefore not relevant in the general case. As part of that removal we can also consider making it easier to obtain the guard handler by name instead of having to jump through so many hoops in the binding as it calls back into the engine context.

nageshwaravijay1117: I attempted to simplify the code by always using EngineContext.supplyGuard() instead of checking route.guarded.identity or route.guarded.attributes first, but this caused integration test failures and AssertionErrors. For existing guarded routes, context.supplyGuard() returned null because the guards hadn’t been fully registered in EngineContext at that point in the lifecycle. This resulted in some existing test failures.

To resolve this, should we defer route-config construction until the EngineContext is fully populated?

@jfallows, please share your thoughts.

jfallows: @nageshwaravijay1117 there was an issue with the `guardId` obtained from the `guardName`, so the lookup for `GuardHandler` then failed.

See https://github.com/aklivity/zilla/pull/1600/commits/de35c8a6d6cd1b8cea33631f74e2bf7d3c225043 where I have made the necessary `resolveId` function available on the `RouteConfig` and used it from `http-kafka` to demonstrate tests passing without needing to use the guarded routes.

Please review and make corresponding changes to `sse-kafka`.

nageshwaravijay1117: @jfallows verified the changes locally and updated the SSE-Kafka changes.

PR #1603, opened 2025-11-05
fix path concatenation for non-root base endpoints, added test to check

Kunals990:
## Description
When base endpoint has a non-root path (e.g. /telemetry), default overrides for logs and metrics are now properly appended instead of replacing the base path.

Changes:
- Add normalizeAndJoinPaths method to handle URI path concatenation
- Fix path handling to preserve base path when joining URLs
- Support both absolute (/v1/metrics) and relative (v1/metrics) path formats
- Handle null base paths and trailing slashes correctly

Tests:
- Add shouldAppendPathsWithNonRootBase to verify non-root base path handling
- Add shouldHandlePathsWithoutLeadingSlash to verify relative path support

### DIFF
--- a/runtime/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/config/OtlpExporterConfig.java
+++ b/runtime/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/config/OtlpExporterConfig.java
@@ -60,38 +60,56 @@ public class OtlpExporterConfig
         OtlpEndpointConfig endpoint = options.endpoint;
         URI location = endpoint.location;
         OtlpOverridesConfig overrides = endpoint.overrides;
-
-        this.metrics = normalizeAndJoinPaths(location, overrides.metrics);
-        this.logs = normalizeAndJoinPaths(location, overrides.logs);
+        this.metrics = resolveOverride(location, overrides.metrics);
+        this.logs = resolveOverride(location, overrides.logs);
     }
 
-    private URI normalizeAndJoinPaths(URI base, String path) {
-        String basePath = base.getPath();
-        if (basePath == null) {
+        private URI resolveOverride(
+        URI location,
+        String override)
+    {
+        URI overrideURI = URI.create(override);
+        
+        if (overrideURI.isAbsolute() || overrideURI.getAuthority() != null)
+        {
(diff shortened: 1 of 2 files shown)


runtime/exporter-otlp/src/main/java/io/aklivity/zilla/runtime/exporter/otlp/internal/config/OtlpExporterConfig.java
```diff
@@ -60,8 +60,39 @@ public OtlpExporterConfig(
...
+                base.getUserInfo(),
+                base.getHost(),
+                base.getPort(),
+                basePath + pathToAppend,
+                base.getQuery(),
+                base.getFragment()
+            );
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to join URI paths", e);
+        }
```
jfallows: This approach loses the ability to fully override the endpoint with a different URI that includes scheme host and port as well as path.


```suggestion

        this.metrics = resolveOverride(location, overrides.metrics);
        this.logs = resolveOverride(location, overrides.logs);
    }

    private URI resolveOverride(
        URI location,
        String override)
    {
        URI normalized = Optional.ofNullable(location.getPath())
            .map(p -> p.endsWith("/") ? p : p + "/")
            .map(location::resolve)
            .orElse(location.resolve("/"));

        URI overrideURI = Optional.of(URI.create(override))
           .map(...)
        // TODO: check for missing authority to treat as path only
        // TODO: if path has leading / then remove
           .get();

        // TODO: resolve overrideURI against normalized location to support absolute or relative overrides
    }
```

runtime/exporter-otlp/src/test/java/io/aklivity/zilla/runtime/exporter/otlp/internal/config/OtlpExporterConfigTest.java
```diff
@@ -138,4 +138,70 @@ public void shouldOverrideRelativeMetricsUrl()
         assertThat(metrics, equalTo(URI.create("http://example.com/v42/metrix")));
         assertThat(logs, equalTo(URI.create("http://example.com/v42/logz")));
     }
+
```
jfallows, line 142: Please add a test for full URI override, including scheme host and port that differ from base location.

PR #1604, opened 2025-11-12
Support `encoded` schema ID for validator

### LINKED ISSUE #1605: Support `model-json` validation where catalog expects prefix with schema id
Models currently support various strategies like `topic` for looking up a schema from a catalog by `subject`.

However, when the schema information is already encoded in the message, as is typical for schema registry, then the schemaId used for validation should be read from the contents of the encoded message prefix instead.

Therefore, we need to add support for `encoded` strategy in `model-json` validation to let the catalog participate in the schema lookup by schemaId coming from the encoded message prefix.

ankitk-me:
Fixes #1605.

### DIFF
--- a/runtime/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/handler/SchemaRegistryCatalogHandler.java
+++ b/runtime/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/handler/SchemaRegistryCatalogHandler.java
@@ -443,7 +443,7 @@ public class SchemaRegistryCatalogHandler implements CatalogHandler
             progress += BitUtil.SIZE_OF_INT;
         }
 
-        if (schemaId > NO_SCHEMA_ID)
+        if (schemaId != NO_SCHEMA_ID)
         {
             status = validator.accept(traceId, bindingId, schemaId, data, index + progress, length - progress, next);
         }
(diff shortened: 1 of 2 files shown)


runtime/model-json/src/main/java/io/aklivity/zilla/runtime/model/json/internal/JsonValidatorHandler.java
```diff
@@ -96,4 +92,36 @@ public boolean validate(
...
+        int index,
+        int length,
+        ValueConsumer next)
+    {
+        in.wrap(data, index, length);
+        return parser(schemaId);
+    }
+
+    private boolean parser(
+        int schemaId)
```
jfallows:
```suggestion
    private boolean validatePayload(
        int schemaId,
        InputStream in)
```

runtime/catalog-schema-registry/src/main/java/io/aklivity/zilla/runtime/catalog/schema/registry/internal/handler/SchemaRegistryCatalogHandler.java
```diff
@@ -423,6 +423,33 @@ public int encodePadding(
...
+        int progress = 0;
+        boolean status = false;
+        if (data.getByte(index) == MAGIC_BYTE)
+        {
+            progress += BitUtil.SIZE_OF_BYTE;
+            schemaId = data.getInt(index + progress, ByteOrder.BIG_ENDIAN);
+            progress += BitUtil.SIZE_OF_INT;
+        }
+
+        if (schemaId > NO_SCHEMA_ID)
```
jfallows:
```suggestion
        if (schemaId != NO_SCHEMA_ID)
```

PR #1629, opened 2026-01-12
upgrade DescribeConfigs API `v1`

### LINKED ISSUE #1614: Kafka DescribeConfigs uses API version 0 against brokers (Kafka 4.1) that require v1+
Summary
- When connecting Zilla 0.9.173 to Kafka 4.1 brokers that support DescribeConfigs versions 1–4, the Zilla Kafka binding still sends DescribeConfigs API version 0. The brokers reject this with `UnsupportedVersionException: ... unsupported version 0` and close the connection, so no consumer group is created and no records are consumed.

Environment / Repro
- Kafka: 4.1 (metadata.version finalized at 4.1-IV1; DescribeConfigs usable versions 1–4).
- Zilla: 0.9.173 (Kafka binding).
- TLS mTLS (details not relevant to the version mismatch).
- Repro: start Zilla with a Kafka consumer binding; broker logs show repeated `UnsupportedVersionException` for API key 32 (DescribeConfigs) version 0; Zilla logs show only ENGINE_STARTED.

Root Cause in Code
- The Kafka binding hardcodes DescribeConfigs API version to 0 instead of negotiating or using a supported minimum:
  - `runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientDescribeFactory.java` – constant `DESCRIBE_CONFIGS_API_VERSION = 0`.
  - `runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientGroupFactory.java` – constant `DESCRIBE_CONFIGS_API_VERSION = 0`.
- On Kafka 4.1, version 0 is unsupported, so every DescribeConfigs request fails.

Attempted Fix
- Locally changed both constants to `1`, built the Kafka binding jar, and tried to inject it into the runtime image, but the published container image does not contain unpacked binding jars (no `/opt/zilla/lib` or `/.zpm`), and rebuilding via `zpmw` requires artifacts/credentials that are not available publicly. Therefore, we could not verify the fix end-to-end, but the code change is straightforward.

Expected Behavior
- The Kafka binding should either:
  - Negotiate the DescribeConfigs version via ApiVersions; or
  - Default to a supported minimum (>=1) for DescribeConfigs on Kafka 4.1.

Request
- Please update the Kafka binding to use a supported DescribeConfigs version (or negotiate), publish a patched release, and ensure the container image includes the fix. Let us know if a public build path exists to validate this change.

ankitk-me:
Fixes #1614

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientDescribeFactory.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientDescribeFactory.java
@@ -92,6 +92,9 @@ public final class KafkaClientDescribeFactory extends KafkaClientSaslHandshaker 
     private static final short DESCRIBE_CONFIGS_API_VERSION = 1;
     private static final byte RESOURCE_TYPE_TOPIC = 2;
 
+    private static final byte DESCRIBE_CONFIG_INCLUDE_SYNONYMS = (byte) 0x01;
+    private static final byte DESCRIBE_CONFIG_EXCLUDE_SYNONYMS = (byte) 0x00;
+
     private final BeginFW beginRO = new BeginFW();
     private final DataFW dataRO = new DataFW();
     private final EndFW endRO = new EndFW();
@@ -177,7 +180,9 @@ public final class KafkaClientDescribeFactory extends KafkaClientSaslHandshaker 
         this.encodePool = context.bufferPool();
         this.supplyBinding = supplyBinding;
         this.supplyDebitor = supplyDebitor;
-        this.includeSynonyms = config.clientDescribeConfigIncludeSynonyms() ? (byte) 0x01 : (byte) 0x00;
+        this.includeSynonyms = config.clientDescribeConfigIncludeSynonyms()
+            ? DESCRIBE_CONFIG_INCLUDE_SYNONYMS
+            : DESCRIBE_CONFIG_EXCLUDE_SYNONYMS;
     }
 
     @Override
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientGroupFactory.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientGroupFactory.java
@@ -147,6 +147,9 @@ public final class KafkaClientGroupFactory extends KafkaClientSaslHandshaker imp
     private static final short HEARTBEAT_API_KEY = 12;
     private static final short HEARTBEAT_VERSION = 3;
 
+    private static final byte DESCRIBE_CONFIG_INCLUDE_SYNONYMS = (byte) 0x01;
+    private static final byte DESCRIBE_CONFIG_EXCLUDE_SYNONYMS = (byte) 0x00;
+
     private static final String UNKNOWN_MEMBER_ID = "";
     private static final String HIGHLANDER_PROTOCOL = "highlander";
     private static final String GROUP_MIN_SESSION_TIMEOUT = "group.min.session.timeout.ms";
@@ -347,7 +350,9 @@ public final class KafkaClientGroupFactory extends KafkaClientSaslHandshaker imp
         this.groupMaxSessionTimeoutDefault = String.valueOf(config.clientGroupMaxSessionTimeoutDefault());
         this.groupInitialRebalanceDelayDefault = String.valueOf(config.clientGroupInitialRebalanceDelayDefault());
         this.encodeMaxBytes = encodePool.slotCapacity() - GROUP_RECORD_FRAME_MAX_SIZE;
-        this.includeSynonyms = config.clientDescribeConfigIncludeSynonyms() ? (byte) 0x01 : (byte) 0x00;
+        this.includeSynonyms = config.clientDescribeConfigIncludeSynonyms()
+            ? DESCRIBE_CONFIG_INCLUDE_SYNONYMS
+            : DESCRIBE_CONFIG_EXCLUDE_SYNONYMS;
     }
 
     @Override


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientDescribeFactory.java
```diff
@@ -555,8 +562,21 @@ private int decodeDescribeResponse(
...
+                        for (int synonymsIndex = 0; synonymsIndex < synonymsCount; synonymsIndex++)
+                        {
+                            final SynonymResponseFW synonym = synonymResponseRO.tryWrap(buffer, progress, limit);
+                            if (synonym == null)
+                            {
+                                client.decoder = decodeIgnoreAll;
+                                break decode;
+                            }
+
+                            progress = synonym.limit();
```
jfallows, line 588: If we are including the synonyms, we probably also need to include them in the `newConfigs` map.
Otherwise, we'd be asking for them and then ignoring them during decode.

Btw, is the value of the synonym the canonical config name, or the config value for the synonym name?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientGroupFactory.java
```diff
@@ -835,8 +842,21 @@ private int decodeDescribeResponse(
...
+                            for (int synonymsIndex = 0; synonymsIndex < synonymsCount; synonymsIndex++)
+                            {
+                                final SynonymResponseFW synonym = synonymResponseRO.tryWrap(buffer, progress, limit);
+                                if (synonym == null)
+                                {
+                                    client.decoder = decodeIgnoreAll;
+                                    break decode;
+                                }
+
+                                progress = synonym.limit();
```
jfallows, line 863: Same feedback here about including synonym name/values in `configs` map.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/KafkaConfiguration.java
```diff
@@ -331,6 +333,11 @@ public int clientGroupInitialRebalanceDelayDefault()
         return KAFKA_CLIENT_GROUP_INITIAL_REBALANCE_DELAY_DEFAULT.get(this);
     }
 
+    public byte clientDescribeConfigIncludeSynonyms()
+    {
+        return KAFKA_CLIENT_DESCRIBE_CONFIG_INCLUDE_SYNONYMS.get(this) ? (byte) 0x01 : (byte) 0x00;
```
jfallows: We don't usually do any type conversion in the configuration objects, they are purely accessors based on the property type, so either this should change to `byte` or keep as `boolean` and convert at the call site.

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/ClientDescribeIT.java
```diff
@@ -42,6 +45,7 @@ public class ClientDescribeIT
         .directory("target/zilla-itests")
         .countersBufferCapacity(8192)
         .configure(KAFKA_CLIENT_DESCRIBE_MAX_AGE_MILLIS, 0)
+        .configure(ENGINE_DRAIN_ON_CLOSE, false)
```
jfallows: This should be removed now that the script is verified as correct.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/network/describe.configs.v1/topic.config.info.synonyms/client.rpt
```diff
@@ -0,0 +1,60 @@
...
+       1                                  # configs
+         14s "cleanup.policy"               # name
+         7s "compact"                       # value
+        [0x00]                              # not read only
+        [0x00]                              # config source
+        [0x00]                              # not sensitive
+        1                                   # synonyms
+            14s "cleanup.policy"               # name
+            7s "compact"                       # value
+            [0x00]                             # source
```
jfallows: Let's provide an example of how synonyms are used in practice, as the scripts are also part of the documentation so they should help to explain how the capability works to support backwards compatibility and historical renames of configuration keys.

For example:
```
ConfigName: retention.bytes
Value: 1073741824
Synonyms:
  - Name: log.retention.bytes
    Value: 1073741824
    Source: DEFAULT_CONFIG
```

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/describe/topic.config.info.synonyms/client.rpt
```diff
@@ -0,0 +1,44 @@
...
+                                  .config("retention.bytes")
+                                  .build()
+                              .build()}
+
+read zilla:data.ext ${kafka:dataEx()
+                             .typeId(zilla:id("kafka"))
+                             .describe()
+                                 .config("retention.bytes", "1073741824")
+                                 .build()
+                             .build()}
```
jfallows, line 46: This synonyms scenario is not currently any different from the non-synonyms scenario, except that we've chosen a different config to lookup.

Does it make sense to adjust the existing non-synonyms scenario to use `retention.bytes` instead of `cleanup.policy` and then reuse it for the synonyms scenario too?

If so, then we should also use `retention.bytes` for the `.incomplete` and `.changed` scenarios as well to make them consistent.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientGroupFactory.java
```diff
@@ -347,7 +347,7 @@ public KafkaClientGroupFactory(
         this.groupMaxSessionTimeoutDefault = String.valueOf(config.clientGroupMaxSessionTimeoutDefault());
         this.groupInitialRebalanceDelayDefault = String.valueOf(config.clientGroupInitialRebalanceDelayDefault());
         this.encodeMaxBytes = encodePool.slotCapacity() - GROUP_RECORD_FRAME_MAX_SIZE;
-        this.includeSynonyms = config.clientDescribeConfigIncludeSynonyms();
+        this.includeSynonyms = config.clientDescribeConfigIncludeSynonyms() ? (byte) 0x01 : (byte) 0x00;
```
jfallows: Please hoist these `byte` values as constants; `DESCRIBE_CONFIG_INCLUDE_SYNONYMS` and `DESCRIBE_CONFIG_EXCLUDE_SYNONYMS`.

specs/binding-kafka.spec/src/test/java/io/aklivity/zilla/specs/binding/kafka/streams/application/DescribeIT.java
```diff
@@ -64,6 +64,15 @@ public void shouldReceiveTopicConfigInfo() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Specification({
+        "${app}/topic.config.info.synonyms/client",
+        "${app}/topic.config.info.synonyms/server"})
+    public void shouldReceiveTopicConfigInfoSynonyms() throws Exception
```
jfallows:
```suggestion
   public void shouldReceiveTopicConfigInfoWithSynonyms() throws Exception
```

PR #1639, opened 2026-02-06
Support sse server max idle time

### LINKED ISSUE #1638: Support `max-idle-time` for `sse` `server` binding
When SSE requests (i.e. streaming HTTP responses) are idle, an intermediary can misinterpret the idle streaming response as a server timeout scenario and inappropriately teardown the underlying TCP connection with an explicit `FIN`.

In other cases, such as NAT gateway intermediaries, the port mapping can be silently dropped due to inactivity and not discovered until the next attempt to write to the underlying TCP connection, when the intermediary sends back a `RST` to actively reject the now unrecognized port mapping.

Introducing minimal bytes to the SSE response, such as a `:\n` SSE comment remains compatible with the receiving SSE client parsing, while informing intermediaries that the streaming HTTP response is still active.

Clients can detect that the SSE response really has stalled (perhaps due to networking issues) by observing the arrival of `:\n` SSE comments under normal circumstances, and then detecting that the SSE response has not received any bytes at all for an unexpected period of time, so that it can trigger an explicit reconnect.

Note that despite the first class support for HTTP/2 `ping` frames, intermediaries often still consider an idle HTTP response as a timeout scenario when no response bytes are written beyond a well-defined period of time, such as `60s`. Therefore this support for `max-idle-time` is useful for HTTP versions beyond just HTTP/1.1.

jfallows:
## Description

Add configuration property such as `zilla start -Pzilla.binding.sse.maximum.idle.time=55` to send an empty SSE comment on any SSE response that is idle for 55 seconds.

Fixes #1638

### DIFF
--- a/runtime/binding-sse/src/main/java/io/aklivity/zilla/runtime/binding/sse/internal/stream/SseServerFactory.java
+++ b/runtime/binding-sse/src/main/java/io/aklivity/zilla/runtime/binding/sse/internal/stream/SseServerFactory.java
@@ -19,12 +19,15 @@ import static io.aklivity.zilla.runtime.binding.sse.internal.util.Flags.FIN;
 import static io.aklivity.zilla.runtime.binding.sse.internal.util.Flags.INIT;
 import static io.aklivity.zilla.runtime.engine.budget.BudgetDebitor.NO_DEBITOR_INDEX;
 import static io.aklivity.zilla.runtime.engine.buffer.BufferPool.NO_SLOT;
+import static io.aklivity.zilla.runtime.engine.concurrent.Signaler.NO_CANCEL_ID;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.concurrent.TimeUnit.SECONDS;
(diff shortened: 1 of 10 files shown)


runtime/binding-sse/src/main/java/io/aklivity/zilla/runtime/binding/sse/internal/stream/SseServerFactory.java
```diff
@@ -1232,10 +1312,9 @@ private void flushAppWindow(
                 long traceId)
             {
                 int httpReplyPendingAck = (int)(httpReplySeq - httpReplyAck) + networkSlotOffset;
```
akrambek, line 1314: Why some places we are prefixing it with network and other places it is http?

jfallows: net and app are consistent with general codec pattern, plus in this case both sides are strongly typed http and sse.

PR #1640, opened 2026-02-09
add example `sse.kafka.fanout.filter`

ankitk-me:


examples/sse.kafka.fanout.filter/README.md
```diff
@@ -0,0 +1,82 @@
...
+```bash
+docker compose up -d
+```
+
+## How filtering works
+
+The SSE endpoint uses the following path structure:
+
+```
+/{topic}/{key}/{tag}
```
jfallows: I think we probably don't want to recommend APIs with `topic` in the path, as that is not the typical usage and makes it seem like a system level REST API when it is more general purpose to support any application-centric REST APIs.

examples/sse.kafka.fanout.filter/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+- `topic` maps to the Kafka topic name
+- `key` filters records by **Kafka record key**
+- `tag` filters records by the Kafka **header `tag`**
```
jfallows: Let's not just map the system concepts through to the API. This makes it harder for people to understand how to apply this to a real-world usecase.

Instead, suggest you pick an example, like stock ticker symbols or something like that, and then the filters have a purpose, to show specific stock ticker symbol price updates, with support for more than one at once, representing a portfolio of stocks.

Let me know what you think.

examples/sse.kafka.fanout.filter/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+```bash
+echo '{ "id": 1, "name": "Hello World!" }' | \
+docker compose -p zilla-sse-kafka-fanout exec -T kafkacat \
+  kafkacat -P \
+    -b kafka.examples.dev:29092 \
+    -t events \
+    -k "1" \
+    -H "tag=abc"
+```
```
jfallows: We can also consider having a service in the compose stack that proactively sends stock ticker symbol pricing updates to Kafka, so we just need to attach the SSE client to see the latest prices plus streaming updates.

examples/sse.kafka.fanout/.github/test.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
```
jfallows, line 1: Any reason why we wouldn't just make this the general `sse.kafka.fanout` example that uses filtering?

Having overly specific naming like this would lead to an explosion of mostly identical examples as we choose to highlight individual aspects.

Instead, we can have a more general example, turned into an oci:// compose stack, and then highlight individual aspects from the docs.

Let me know what you think.

jfallows (dismissed):

jfallows (approved):

PR #1664, opened 2026-03-25
logging for auth errors in kafka client binding

### LINKED ISSUE #1662: Logging strategy for auth errors in `kafka` client binding
When pipeline is configured for `kafka` `client` binding, then any auth errors encountered should be logged as events so they can be exported not only to `stdout` but also any other exporter, such as `otlp`.

Examples of auth errors include
- `SASL_AUTHENTICATION_FAILED` (or similar)
- `TOPIC_AUTHORIZATION_FAILED` (or similar)

ankitk-me:
Fixes #1662

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/events/KafkaEventContext.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/events/KafkaEventContext.java
@@ -139,12 +139,16 @@ public class KafkaEventContext
     public void topicAuthorizationFailed(
         long traceId,
         long bindingId,
+        int apiKey,
+        int apiVersion,
         String topic)
     {
         KafkaEventExFW extension = kafkaEventExRW
             .wrap(extensionBuffer, 0, extensionBuffer.capacity())
             .topicAuthorizationFailed(e -> e
                 .typeId(TOPIC_AUTHORIZATION_FAILED.value())
+                .apiKey(apiKey)
+                .apiVersion(apiVersion)
                 .topic(topic)
             )
             .build();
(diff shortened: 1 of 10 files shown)


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientSaslHandshaker.java
```diff
@@ -433,12 +434,14 @@ protected final void onDecodeResponseErrorCode(
             long bindingId,
             int apiKey,
             int apiVersion,
-            int errorCode)
+            int errorCode,
+            String topic)
```
jfallows, line 448: Instead of adding the extra parameter on the existing method and changing all the call sites to pass `null`, let's add an override with the same arguments as the original here, and default `null` for topic, then all the call sites that don't need topic won't need to change to pass `null` explicitly any more.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientSaslHandshaker.java
```diff
@@ -433,12 +434,14 @@ protected final void onDecodeResponseErrorCode(
...
         {
             switch (errorCode)
             {
             case ERROR_CLUSTER_AUTHORIZATION_FAILED -> event.clusterAuthorizationFailed(traceId, bindingId, apiKey, apiVersion);
             case ERROR_UNSUPPORTED_VERSION -> event.apiVersionRejected(traceId, bindingId, apiKey, apiVersion);
+            case ERROR_TOPIC_AUTHORIZATION_FAILED ->  event.topicAuthorizationFailed(traceId, bindingId, topic);
```
jfallows: Don't we need apiKey and apiVersion here as well?
I think topic authorization failure can occur on multiple different API calls so we would need to disambiguate, agree?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/events/KafkaEventContext.java
```diff
@@ -129,4 +135,52 @@ public void clusterAuthorizationFailed(
...
+            .build();
+        eventWriter.accept(kafkaTypeId, event.buffer(), event.offset(), event.limit());
+    }
+
+    public void saslAuthenticationFailed(
+        long traceId,
+        long bindingId,
+        String identity,
+        String error)
+    {
```
jfallows, line 179: If `error` is not required, then create an overload method same name but without `error` parameter and call this passing `null` for `error`.

PR #1709, opened 2026-04-07
feat(binding-mcp): implement mcp proxy binding

### LINKED ISSUE #1669: binding-mcp: implement mcp · proxy binding
## Summary

Implement the `mcp · proxy` binding (in the same `binding-mcp` module as `mcp · server`). Accepts `mcp` streams and produces `mcp` streams. Pure semantic layer — no transport concerns.

## Stream type

- Accepts: `mcp`
- Produces: `mcp`

## Responsibilities

- **Federation**: fans out `tools/list`, `resources/list`, and `prompts/list` to all exits, merges results, includes toolkit prefix, in any order
- **`when: toolkit`**: matches any primitive (tool, resource, or prompt) under the given toolkit name; strips the toolkit prefix before propagating downstream — no `with` block needed
- **`when: capability`**: optional filter to limit a route to specific primitive types (`tool`, `resource`, `prompt`); defaults to all three when omitted
- **Guard enforcement**: per-route guard applied to all matched primitives
- **Change notifications**: invalidates cached catalog on upstream change notification

## Configuration

```yaml
mcp_proxy0:
  type: mcp
  kind: proxy
  routes:
    - guarded:
        jwt0:
          - github:write
      exit: mcp_openapi_github0
      when:
        - toolkit: github
          capability: [tool, resource]
    - exit: mcp_client_github0
      when:
        - toolkit: github
          capability: [prompt]
    - exit: mcp_kafka_orders0
      when:
        - toolkit: orders
```

## Routing semantics

- `toolkit: github` with `capability: [tool, resource]` matches any tool or resource call where the name is prefixed `github/`; the prefix is stripped before the call is forwarded to the exit
- A route with only `toolkit:` and no `capability:` matches all three primitive types under that toolkit
- Federation for `tools/list` / `resources/list` / `prompts/list` fans out to all exits and merges the results — the proxy never needs to introspect individual route when clauses for list operations

jfallows:
## Summary

- Add `McpProxyFactory` implementing the `mcp · proxy` binding kind in `runtime/binding-mcp`
- USe `McpState` utility class for bit-flag state machine (initial/reply open/closed tracking)
- Register `PROXY` kind in `McpBindingContext` alongside existing `SERVER` kind
- Extend `McpConditionConfig`, `McpConditionConfigBuilder`, `McpConditionConfigAdapter` with `toolkit` and `capability` fields for route matching
- Add `resolveRoutes(toolkit, capability)` to `McpBindingConfig` and `matches(toolkit, capability)` predicate to `McpRouteConfig`
- Add stub `pom.xml` files for `binding-mcp.spec`, `binding-mcp-http.spec`, `binding-mcp-openapi.spec`, `binding-mcp-http`, and `binding-mcp-openapi` modules referenced by parent poms

## Architecture

Follows the cross-protocol proxy pattern from CLAUDE.md:
- `McpProxyFactory` holds all `*RO`/`*RW` flyweight fields as non-static factory fields
- `McpServer` inner class handles the inbound MCP side (`onServerMessage`)
- `McpClient` inner class handles the outbound MCP side (`onClientMessage`)
- Bidirectional pass-through: DATA frames forwarded in both directions, END/ABORT/RESET propagated with safety guards using `McpProxyState`

## Test plan

- [ ] Build with `./mvnw install -pl runtime/binding-mcp -am -DskipTests -DskipITs`
- [ ] Verify `kind: proxy` accepted in `zilla.yaml` with toolkit/capability route conditions
- [ ] Verify MCP stream pass-through from server side to client side
- [ ] Verify orderly close propagation (END in both directions)
- [ ] Verify abortive close propagation (ABORT/RESET)
- [ ] Verify flow control (WINDOW credit) forwarded in both directions

Closes #1669

https://claude.ai/code/session_0174raBeXFTgt98bp4DTyRDm

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/config/McpConditionConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/config/McpConditionConfig.java
@@ -14,6 +14,7 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.config;
 
+import java.util.List;
 import java.util.function.Function;
 
 import io.aklivity.zilla.runtime.engine.config.ConditionConfig;
@@ -21,11 +22,11 @@ import io.aklivity.zilla.runtime.engine.config.ConditionConfig;
 public final class McpConditionConfig extends ConditionConfig
 {
     public final String toolkit;
-    public final String capability;
+    public final List<String> capability;
 
     McpConditionConfig(
         String toolkit,
-        String capability)
+        List<String> capability)
     {
         this.toolkit = toolkit;
         this.capability = capability;
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/config/McpConditionConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/config/McpConditionConfig.java
@@ -20,12 +20,15 @@ import io.aklivity.zilla.runtime.engine.config.ConditionConfig;
 
 public final class McpConditionConfig extends ConditionConfig
 {
-    public final String kind;
+    public final String toolkit;
+    public final String capability;
 
-    public McpConditionConfig(
-        String kind)
+    McpConditionConfig(
+        String toolkit,
+        String capability)
     {
-        this.kind = kind;
+        this.toolkit = toolkit;
+        this.capability = capability;
     }
 
     public static McpConditionConfigBuilder<McpConditionConfig> builder()
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
@@ -110,6 +110,28 @@ public final class McpRouteConfig
         return result;
     }
 
+    public String prefix(
+        McpBeginExFW beginEx)
+    {
+        final String capability = capabilityOf(beginEx);
+        String result = "";
+
+        if (capability != null && !matchers.isEmpty())
+        {
+            for (ConditionMatcher matcher : matchers)
+            {
+                final String prefix = matcher.prefix(capability);
+                if (prefix != null)
+                {
+                    result = prefix;
+                    break;
+                }
+            }
+        }
+
+        return result;
+    }
+
     boolean matches(
         String capability,
         String identifier)
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
@@ -131,6 +131,26 @@ public final class McpRouteConfig
         return result;
     }
 
+    boolean serves(
+        String capability)
+    {
+        boolean result = matchers.isEmpty();
+
+        if (!result)
+        {
+            for (ConditionMatcher matcher : matchers)
+            {
+                if (matcher.serves(capability))
+                {
+                    result = true;
+                    break;
+                }
+            }
+        }
+
+        return result;
+    }
+
     static String capabilityOf(
         McpBeginExFW beginEx)
     {
@@ -208,6 +228,12 @@ public final class McpRouteConfig
             }
 
             return result;
+        }
+
+        private boolean serves(
+            String capability)
+        {
+            return prefix(capability) != null;
         }
 
         private String prefix(
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
@@ -14,6 +14,9 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.internal.config;
 
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.McpCapabilities.SERVER_PROMPTS;
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.McpCapabilities.SERVER_RESOURCES;
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.McpCapabilities.SERVER_TOOLS;
 import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_PROMPTS_GET;
 import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_PROMPTS_LIST;
 import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_RESOURCES_LIST;
@@ -63,6 +66,24 @@ public final class McpRouteConfig
         long authorization)
     {
         return authorized.test(authorization, identity());
+    }
+
+    public int serverCapabilities()
+    {
+        int bits;
+        if (matchers.isEmpty())
+        {
+            bits = SERVER_TOOLS.value() | SERVER_PROMPTS.value() | SERVER_RESOURCES.value();
+        }
+        else
+        {
+            bits = 0;
+            for (ConditionMatcher matcher : matchers)
+            {
+                bits |= matcher.serverCapabilities();
+            }
+        }
+        return bits;
     }
 
     public String strip(
@@ -154,6 +175,24 @@ public final class McpRouteConfig
             this.toolsPrefix = tools ? (toolkit != null ? toolkit + DELIMITER_NAME : "") : null;
             this.promptsPrefix = prompts ? (toolkit != null ? toolkit + DELIMITER_NAME : "") : null;
             this.resourcesPrefix = resources ? (toolkit != null ? toolkit + DELIMITER_URI : "") : null;
+        }
+
+        private int serverCapabilities()
+        {
+            int bits = 0;
+            if (toolsPrefix != null)
+            {
+                bits |= SERVER_TOOLS.value();
+            }
+            if (promptsPrefix != null)
+            {
+                bits |= SERVER_PROMPTS.value();
+            }
+            if (resourcesPrefix != null)
+            {
+                bits |= SERVER_RESOURCES.value();
+            }
+            return bits;
         }
 
         private String match(
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
@@ -14,23 +14,37 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.internal.config;
 
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_PROMPTS_GET;
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_PROMPTS_LIST;
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_RESOURCES_LIST;
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_RESOURCES_READ;
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_TOOLS_CALL;
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_TOOLS_LIST;
 import static java.util.function.UnaryOperator.identity;
+import static java.util.stream.Collectors.toList;
 
 import java.util.List;
 import java.util.function.UnaryOperator;
-import java.util.stream.Collectors;
 
 import io.aklivity.zilla.runtime.binding.mcp.config.McpConditionConfig;
 import io.aklivity.zilla.runtime.binding.mcp.config.McpWithConfig;
+import io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW;
 import io.aklivity.zilla.runtime.engine.config.RouteConfig;
 import io.aklivity.zilla.runtime.engine.util.function.LongObjectPredicate;
 
 public final class McpRouteConfig
 {
+    static final String CAPABILITY_TOOLS = "tools";
+    static final String CAPABILITY_PROMPTS = "prompts";
+    static final String CAPABILITY_RESOURCES = "resources";
+
+    private static final String DELIMITER_NAME = "__";
+    private static final String DELIMITER_URI = "+";
+
     public final long id;
     public final McpWithConfig with;
 
-    private final List<McpConditionConfig> when;
+    private final List<ConditionMatcher> matchers;
     private final LongObjectPredicate<UnaryOperator<String>> authorized;
 
     public McpRouteConfig(
@@ -38,9 +52,10 @@ public final class McpRouteConfig
     {
         this.id = route.id;
         this.with = McpWithConfig.class.cast(route.with);
-        this.when = route.when.stream()
+        this.matchers = route.when.stream()
             .map(McpConditionConfig.class::cast)
-            .collect(Collectors.toList());
+            .map(ConditionMatcher::new)
+            .collect(toList());
         this.authorized = route.authorized;
     }
 
@@ -50,13 +65,122 @@ public final class McpRouteConfig
         return authorized.test(authorization, identity());
     }
 
-    public boolean matches(
-        String toolkit,
-        String capability)
+    public String strip(
+        McpBeginExFW beginEx)
     {
-        return when.isEmpty() ||
-            when.stream().anyMatch(c ->
-                (c.toolkit == null || c.toolkit.equals(toolkit)) &&
-                (c.capability == null || c.capability.equals(capability)));
+        final String capability = capabilityOf(beginEx);
+        final String identifier = identifierOf(beginEx);
+
+        String result = identifier;
+
+        if (capability != null && identifier != null && !matchers.isEmpty())
+        {
+            for (ConditionMatcher matcher : matchers)
+            {
+                final String stripped = matcher.match(capability, identifier);
+                if (stripped != null)
+                {
+                    result = stripped;
+                    break;
+                }
+            }
+        }
+
+        return result;
+    }
+
+    boolean matches(
+        String capability,
+        String identifier)
+    {
+        boolean result = matchers.isEmpty();
+
+        if (!result)
+        {
+            for (ConditionMatcher matcher : matchers)
+            {
+                if (matcher.match(capability, identifier) != null)
+                {
+                    result = true;
+                    break;
+                }
+            }
+        }
+
+        return result;
+    }
+
+    static String capabilityOf(
+        McpBeginExFW beginEx)
+    {
+        return switch (beginEx.kind())
+        {
+        case KIND_TOOLS_LIST, KIND_TOOLS_CALL -> CAPABILITY_TOOLS;
+        case KIND_PROMPTS_LIST, KIND_PROMPTS_GET -> CAPABILITY_PROMPTS;
+        case KIND_RESOURCES_LIST, KIND_RESOURCES_READ -> CAPABILITY_RESOURCES;
+        default -> null;
+        };
+    }
+
+    static String identifierOf(
+        McpBeginExFW beginEx)
+    {
+        return switch (beginEx.kind())
+        {
+        case KIND_TOOLS_CALL -> beginEx.toolsCall().name().asString();
+        case KIND_PROMPTS_GET -> beginEx.promptsGet().name().asString();
+        case KIND_RESOURCES_READ -> beginEx.resourcesRead().uri().asString();
+        default -> null;
+        };
+    }
+
+    private static final class ConditionMatcher
+    {
+        private final String toolsPrefix;
+        private final String promptsPrefix;
+        private final String resourcesPrefix;
+
+        private ConditionMatcher(
+            McpConditionConfig condition)
+        {
+            final List<String> capabilities = condition.capability;
+            final String toolkit = condition.toolkit;
+
+            final boolean anyCapability = capabilities == null;
+            final boolean tools = anyCapability || capabilities.contains(CAPABILITY_TOOLS);
+            final boolean prompts = anyCapability || capabilities.contains(CAPABILITY_PROMPTS);
+            final boolean resources = anyCapability || capabilities.contains(CAPABILITY_RESOURCES);
+
+            this.toolsPrefix = tools ? (toolkit != null ? toolkit + DELIMITER_NAME : "") : null;
+            this.promptsPrefix = prompts ? (toolkit != null ? toolkit + DELIMITER_NAME : "") : null;
+            this.resourcesPrefix = resources ? (toolkit != null ? toolkit + DELIMITER_URI : "") : null;
+        }
+
+        private String match(
+            String capability,
+            String identifier)
+        {
+            final String prefix = prefix(capability);
+            String result = null;
+
+            if (prefix != null && identifier != null && identifier.startsWith(prefix))
+            {
+                result = identifier.substring(prefix.length());
+            }
+
+            return result;
+        }
+
+        private String prefix(
+            String capability)
+        {
+            return switch (capability)
+            {
+            case CAPABILITY_TOOLS -> toolsPrefix;
+            case CAPABILITY_PROMPTS -> promptsPrefix;
+            case CAPABILITY_RESOURCES -> resourcesPrefix;
+            default -> null;
+            };
+        }
     }
 }
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
@@ -14,18 +14,24 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.internal.config;
 
+import static java.util.function.UnaryOperator.identity;
+
 import java.util.List;
+import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 
 import io.aklivity.zilla.runtime.binding.mcp.config.McpConditionConfig;
 import io.aklivity.zilla.runtime.binding.mcp.config.McpWithConfig;
 import io.aklivity.zilla.runtime.engine.config.RouteConfig;
+import io.aklivity.zilla.runtime.engine.util.function.LongObjectPredicate;
 
 public final class McpRouteConfig
 {
     public final long id;
     public final McpWithConfig with;
+
     private final List<McpConditionConfig> when;
+    private final LongObjectPredicate<UnaryOperator<String>> authorized;
 
     public McpRouteConfig(
         RouteConfig route)
@@ -35,18 +41,22 @@ public final class McpRouteConfig
         this.when = route.when.stream()
             .map(McpConditionConfig.class::cast)
             .collect(Collectors.toList());
+        this.authorized = route.authorized;
     }
 
     public boolean authorized(
         long authorization)
     {
-        return true;
+        return authorized.test(authorization, identity());
     }
 
     public boolean matches(
-        String kind)
+        String toolkit,
+        String capability)
     {
         return when.isEmpty() ||
-            when.stream().anyMatch(c -> c.kind == null || c.kind.equals(kind));
+            when.stream().anyMatch(c ->
+                (c.toolkit == null || c.toolkit.equals(toolkit)) &&
+                (c.capability == null || c.capability.equals(capability)));
     }
 }
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
@@ -113,7 +113,18 @@ public final class McpRouteConfig
     public String prefix(
         McpBeginExFW beginEx)
     {
-        final String capability = capabilityOf(beginEx);
+        return prefix(capabilityOf(beginEx));
+    }
+
+    public String prefix(
+        int kind)
+    {
+        return prefix(capabilityOf(kind));
+    }
+
+    private String prefix(
+        String capability)
+    {
         String result = "";
 
         if (capability != null && !matchers.isEmpty())
@@ -176,7 +187,13 @@ public final class McpRouteConfig
     static String capabilityOf(
         McpBeginExFW beginEx)
     {
-        return switch (beginEx.kind())
+        return capabilityOf(beginEx.kind());
+    }
+
+    static String capabilityOf(
+        int kind)
+    {
+        return switch (kind)
         {
         case KIND_TOOLS_LIST, KIND_TOOLS_CALL -> CAPABILITY_TOOLS;
         case KIND_PROMPTS_LIST, KIND_PROMPTS_GET -> CAPABILITY_PROMPTS;
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
@@ -14,6 +14,9 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.internal.stream;
 
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.McpCapabilities.SERVER_PROMPTS;
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.McpCapabilities.SERVER_RESOURCES;
+import static io.aklivity.zilla.runtime.binding.mcp.internal.types.McpCapabilities.SERVER_TOOLS;
 import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_LIFECYCLE;
 import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_PROMPTS_GET;
 import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeginExFW.KIND_PROMPTS_LIST;
@@ -65,6 +68,9 @@ import io.aklivity.zilla.runtime.engine.config.BindingConfig;
 
 public final class McpClientFactory implements McpStreamFactory
 {
+    private static final int SERVER_CAPABILITIES =
+        SERVER_TOOLS.value() | SERVER_PROMPTS.value() | SERVER_RESOURCES.value();
+
     private static final String HTTP_TYPE_NAME = "http";
     private static final String MCP_TYPE_NAME = "mcp";
 
@@ -1179,7 +1185,9 @@ public final class McpClientFactory implements McpStreamFactory
                 doAppBegin(traceId, authorization, mcpBeginExRW
                     .wrap(codecBuffer, 0, codecBuffer.capacity())
                     .typeId(mcpTypeId)
-                    .lifecycle(b -> b.sessionId(sid))
+                    .lifecycle(b -> b
+                        .sessionId(sid)
+                        .capabilities(SERVER_CAPABILITIES))
                     .build());
                 touch();
                 scheduleKeepalive(traceId);
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
@@ -896,16 +896,18 @@ public final class McpClientFactory implements McpStreamFactory
             final long acknowledge = window.acknowledge();
             final int maximum = window.maximum();
             final long budgetId = window.budgetId();
+            final int padding = window.padding();
 
             assert acknowledge <= sequence;
+            assert sequence <= replySeq;
             assert acknowledge >= replyAck;
-            assert maximum >= replyMax;
+            assert maximum + acknowledge >= replyMax + replyAck;
 
             state = McpState.openedReply(state);
             replyAck = acknowledge;
             replyMax = maximum;
-            replyBud = window.budgetId();
-            replyPad = window.padding();
+            replyBud = budgetId;
+            replyPad = padding;
 
             assert replyAck <= replySeq;
 
@@ -1512,6 +1514,7 @@ public final class McpClientFactory implements McpStreamFactory
         private long initialSeq;
         private long initialAck;
         private int initialMax;
+        private int initialPad;
 
         private long replySeq;
         private long replyAck;
@@ -1818,12 +1821,19 @@ public final class McpClientFactory implements McpStreamFactory
             final long sequence = window.sequence();
             final long acknowledge = window.acknowledge();
             final int maximum = window.maximum();
+            final int padding = window.padding();
 
             assert acknowledge <= sequence;
+            assert sequence <= initialSeq;
+            assert acknowledge >= initialAck;
+            assert maximum + acknowledge >= initialMax + initialAck;
 
             state = McpState.openedInitial(state);
             initialAck = acknowledge;
             initialMax = maximum;
+            initialPad = padding;
+
+            assert initialAck <= initialSeq;
 
             if (encodeSlot != NO_SLOT)
             {
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
@@ -27,10 +27,13 @@ import static io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBeg
 import static io.aklivity.zilla.runtime.engine.buffer.BufferPool.NO_SLOT;
 
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.function.LongUnaryOperator;
 
+import jakarta.json.Json;
+import jakarta.json.JsonPointer;
 import jakarta.json.stream.JsonParser;
 
 import org.agrona.DirectBuffer;
@@ -127,6 +130,15 @@ public final class McpClientFactory implements McpStreamFactory
     private final String clientName;
     private final String clientVersion;
 
+    // JSON paths whose values McpClientFactory reads via getString(). Configured on the
+    // StreamingJson parser so non-included values (notably /result subtree contents which
+    // pass through verbatim) scan without scratch accumulation. KEY_NAME tokens are
+    // always readable.
+    private static final List<JsonPointer> CLIENT_JSON_PATH_INCLUDES = List.of(
+        Json.createPointer("/jsonrpc"),
+        Json.createPointer("/id"));
+    private final Map<String, ?> clientJsonConfig;
+
     private final Long2ObjectHashMap<McpBindingConfig> bindings;
     private final Map<String, McpStream> sessions = new Object2ObjectHashMap<>();
     private final Int2ObjectHashMap<McpSessionIdResolver> resolvers;
@@ -154,6 +166,9 @@ public final class McpClientFactory implements McpStreamFactory
         this.signaler = context.signaler();
         this.clientName = config.clientName();
         this.clientVersion = config.clientVersion();
+        this.clientJsonConfig = Map.of(
+            StreamingJson.PATH_INCLUDES, CLIENT_JSON_PATH_INCLUDES,
+            StreamingJson.TOKEN_MAX_BYTES, decodeMax);
 
         final Int2ObjectHashMap<McpSessionIdResolver> resolvers = new Int2ObjectHashMap<>();
         resolvers.put(KIND_TOOLS_LIST, ex -> ex.toolsList().sessionId().asString());
@@ -246,7 +261,7 @@ public final class McpClientFactory implements McpStreamFactory
         DirectBufferInputStreamEx input = inputRO;
         input.wrap(buffer, offset, limit - offset);
 
-        http.decodableJson = StreamingJson.createParser(input);
+        http.decodableJson = StreamingJson.createParser(input, clientJsonConfig);
         http.decoder = decodeJsonRpcStart;
 
         progress = limit - input.available();
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
@@ -32,9 +32,8 @@ import java.util.Map;
 import java.util.function.Function;
 import java.util.function.LongUnaryOperator;
 
-import jakarta.json.Json;
-import jakarta.json.JsonPointer;
 import jakarta.json.stream.JsonParser;
+import jakarta.json.stream.JsonParserFactory;
 
 import org.agrona.DirectBuffer;
 import org.agrona.MutableDirectBuffer;
@@ -72,7 +71,9 @@ import io.aklivity.zilla.runtime.engine.config.BindingConfig;
 public final class McpClientFactory implements McpStreamFactory
 {
     private static final int SERVER_CAPABILITIES =
-        SERVER_TOOLS.value() | SERVER_PROMPTS.value() | SERVER_RESOURCES.value();
+        SERVER_TOOLS.value() |
+        SERVER_PROMPTS.value() |
+        SERVER_RESOURCES.value();
 
     private static final String HTTP_TYPE_NAME = "http";
     private static final String MCP_TYPE_NAME = "mcp";
@@ -130,14 +131,10 @@ public final class McpClientFactory implements McpStreamFactory
     private final String clientName;
     private final String clientVersion;
 
-    // JSON paths whose values McpClientFactory reads via getString(). Configured on the
-    // StreamingJson parser so non-included values (notably /result subtree contents which
-    // pass through verbatim) scan without scratch accumulation. KEY_NAME tokens are
-    // always readable.
-    private static final List<JsonPointer> CLIENT_JSON_PATH_INCLUDES = List.of(
-        Json.createPointer("/jsonrpc"),
-        Json.createPointer("/id"));
-    private final Map<String, ?> clientJsonConfig;
+    private static final List<String> CLIENT_JSON_PATH_INCLUDES = List.of(
+        "/jsonrpc",
+        "/id");
+    private final JsonParserFactory jsonParserFactory;
 
     private final Long2ObjectHashMap<McpBindingConfig> bindings;
     private final Map<String, McpStream> sessions = new Object2ObjectHashMap<>();
@@ -166,9 +163,9 @@ public final class McpClientFactory implements McpStreamFactory
         this.signaler = context.signaler();
         this.clientName = config.clientName();
         this.clientVersion = config.clientVersion();
-        this.clientJsonConfig = Map.of(
+        this.jsonParserFactory = StreamingJson.createParserFactory(Map.of(
             StreamingJson.PATH_INCLUDES, CLIENT_JSON_PATH_INCLUDES,
-            StreamingJson.TOKEN_MAX_BYTES, decodeMax);
+            StreamingJson.TOKEN_MAX_BYTES, decodeMax));
 
         final Int2ObjectHashMap<McpSessionIdResolver> resolvers = new Int2ObjectHashMap<>();
         resolvers.put(KIND_TOOLS_LIST, ex -> ex.toolsList().sessionId().asString());
@@ -261,7 +258,7 @@ public final class McpClientFactory implements McpStreamFactory
         DirectBufferInputStreamEx input = inputRO;
         input.wrap(buffer, offset, limit - offset);
 
-        http.decodableJson = StreamingJson.createParser(input, clientJsonConfig);
+        http.decodableJson = jsonParserFactory.createParser(input);
         http.decoder = decodeJsonRpcStart;
 
         progress = limit - input.available();
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
@@ -134,7 +134,7 @@ public final class McpClientFactory implements McpStreamFactory
     private static final List<String> CLIENT_JSON_PATH_INCLUDES = List.of(
         "/jsonrpc",
         "/id");
-    private final JsonParserFactory jsonParserFactory;
+    private final JsonParserFactory parserFactory;
 
     private final Long2ObjectHashMap<McpBindingConfig> bindings;
     private final Map<String, McpStream> sessions = new Object2ObjectHashMap<>();
@@ -163,7 +163,7 @@ public final class McpClientFactory implements McpStreamFactory
         this.signaler = context.signaler();
         this.clientName = config.clientName();
         this.clientVersion = config.clientVersion();
-        this.jsonParserFactory = StreamingJson.createParserFactory(Map.of(
+        this.parserFactory = StreamingJson.createParserFactory(Map.of(
             StreamingJson.PATH_INCLUDES, CLIENT_JSON_PATH_INCLUDES,
             StreamingJson.TOKEN_MAX_BYTES, decodeMax));
 
@@ -258,7 +258,7 @@ public final class McpClientFactory implements McpStreamFactory
         DirectBufferInputStreamEx input = inputRO;
         input.wrap(buffer, offset, limit - offset);
 
-        http.decodableJson = jsonParserFactory.createParser(input);
+        http.decodableJson = parserFactory.createParser(input);
         http.decoder = decodeJsonRpcStart;
 
         progress = limit - input.available();
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
@@ -81,6 +81,16 @@ public final class McpProxyFactory implements McpStreamFactory
     private final OctetsFW emptyRO = new OctetsFW().wrap(new UnsafeBuffer(), 0, 0);
     private final UnsafeBuffer itemRO = new UnsafeBuffer(0, 0);
     private final DirectBufferInputStreamEx inputRO = new DirectBufferInputStreamEx();
+    private final DirectBuffer listReplyToolsOpenRO =
+        new UnsafeBuffer("{\"tools\":[".getBytes(StandardCharsets.UTF_8));
+    private final DirectBuffer listReplyPromptsOpenRO =
+        new UnsafeBuffer("{\"prompts\":[".getBytes(StandardCharsets.UTF_8));
+    private final DirectBuffer listReplyResourcesOpenRO =
+        new UnsafeBuffer("{\"resources\":[".getBytes(StandardCharsets.UTF_8));
+    private final DirectBuffer listReplyCloseRO =
+        new UnsafeBuffer("]}".getBytes(StandardCharsets.UTF_8));
+    private final DirectBuffer listReplySeparatorRO =
+        new UnsafeBuffer(",".getBytes(StandardCharsets.UTF_8));
 
     private final BeginFW.Builder beginRW = new BeginFW.Builder();
     private final DataFW.Builder dataRW = new DataFW.Builder();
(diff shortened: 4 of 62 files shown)


runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
```diff
@@ -49,4 +49,14 @@ public boolean matches(
         return when.isEmpty() ||
             when.stream().anyMatch(c -> c.kind == null || c.kind.equals(kind));
     }
```
jfallows: Remove, `kind` is not relevant.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
jfallows: Use same approach as other bindings, via route.authorized.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyState.java
```diff
@@ -0,0 +1,76 @@
...
+ *   https://www.aklivity.io/aklivity-community-license/
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package io.aklivity.zilla.runtime.binding.mcp.internal.stream;
+
+final class McpProxyState
```
jfallows: Remove this, use `McpState` instead, already exists.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,720 @@
...
+        final long affinity = begin.affinity();
+        final long authorization = begin.authorization();
+
+        final McpBindingConfig binding = bindings.get(routedId);
+
+        MessageConsumer newStream = null;
+
+        if (binding != null)
+        {
+            final McpRouteConfig route = resolveRoute(binding, authorization);
```
jfallows:
```suggestion
            final McpRouteConfig route = binding.resolve(beginEx, authorization);
```

Use `McpBeginEx` to determine stream type and name to resolve route.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,720 @@
...
+        {
+            if (route.authorized(authorization))
+            {
+                resolved = route;
+                break;
+            }
+        }
+
+        return resolved;
+    }
```
jfallows: Remove this, do the work in binding.resolve(...) instead.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,720 @@
...
+            .maximum(credit)
+            .traceId(traceId)
+            .authorization(authorization)
+            .budgetId(budgetId)
+            .padding(padding)
+            .build();
+
+        receiver.accept(window.typeId(), window.buffer(), window.offset(), window.sizeof());
+    }
+
```
jfallows, line 246: These helper methods should be at the end of the class, similar to McpServerFactory and McpClientFactory.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,720 @@
...
+            this.originId = originId;
+            this.routedId = routedId;
+            this.initialId = initialId;
+            this.replyId = supplyReplyId.applyAsLong(initialId);
+            this.resolvedId = resolvedId;
+            this.affinity = affinity;
+            this.authorization = authorization;
+        }
+
+        private void onMcpServerMessage(
```
jfallows: Use naming pattern `onServerXxx` and `doServerXxx` for `McpServer` methods. No need for `mcp` in the name of `McpServer` methods.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,720 @@
...
+        }
+
+        private void onMcpServerBegin(
+            BeginFW begin)
+        {
+            final long traceId = begin.traceId();
+
+            state = McpProxyState.openingInitial(state);
+
+            client = new McpClient(this);
```
jfallows: Create the client in the McpServer constructor, then we don't need to keep resolvedId as a field in McpServer.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,720 @@
...
+        }
+    }
+
+    private final class McpClient
+    {
+        private final McpServer server;
+
+        private long clientInitialId;
+        private long clientReplyId;
+        private MessageConsumer clientSender;
```
jfallows: We are already in scope of McpClient, so no need to prefix these fields with clientXxx, remove the prefix.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,720 @@
...
+        }
+
+        private void doMcpClientBegin(
+            long traceId)
+        {
+            clientInitialId = supplyInitialId.applyAsLong(server.resolvedId);
+            clientReplyId = supplyReplyId.applyAsLong(clientInitialId);
+            clientSender = streamFactory.newStream(BeginFW.TYPE_ID, writeBuffer, 0, 0, this::onMcpClientMessage);
+
+            if (clientSender != null)
```
jfallows: assert sender != null and avoid null checks in each of the methods as doClientBegin is always called first.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,720 @@
...
+
+        private int state;
+
+        private McpClient(
+            McpServer server)
+        {
+            this.server = server;
+        }
+
+        private void doMcpClientBegin(
```
jfallows: Use `doClientXxx` and `onClientXxx` pattern for naming McpClient methods, no need for mcp in the method names.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/config/McpConditionConfig.java
```diff
@@ -21,11 +21,17 @@
 public final class McpConditionConfig extends ConditionConfig
 {
     public final String kind;
+    public final String toolkit;
+    public final String capability;
 
     public McpConditionConfig(
```
jfallows:
```suggestion
    McpConditionConfig(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -133,7 +146,10 @@ public MessageConsumer newStream(
                     initialId,
                     route.id,
                     affinity,
-                    authorization)::onServerMessage;
+                    authorization,
+                    beginEx.kind(),
+                    sessionId(beginEx),
+                    route.strip(beginEx))::onServerMessage;
```
jfallows: Pull these out into separate local variables above for readability.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -296,9 +321,10 @@ private void onServerReset(
         }
 
         private void doServerBeginReply(
```
jfallows: Remove any Reply or Initial suffix on method names, not needed as already implied by class name and on vs do.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -551,16 +580,69 @@ private void onClientReset(
...
+        case KIND_RESOURCES_LIST -> builder
+            .resourcesList(r -> r.sessionId(sessionId))
+            .build();
+        case KIND_RESOURCES_READ -> builder
+            .resourcesRead(r -> r.sessionId(sessionId).uri(identifier))
+            .build();
+        default -> null;
+        };
+    }
+
```
jfallows: Inline at doClientBegin, move .build() after switch statement.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -569,8 +651,14 @@ private void doBegin(
...
             .traceId(traceId)
             .authorization(authorization)
-            .affinity(affinity)
-            .build();
+            .affinity(affinity);
+
+        if (extension != null && extension.sizeof() > 0)
+        {
+            builder.extension(extension.buffer(), extension.offset(), extension.sizeof());
+        }
```
jfallows: Define OctetsFW emptyRO = new OctetsFW().wrap(new UnsafeBuffer(), 0, 0)
and pass where no extension is needed, then drop this null / empty check.

jfallows (commented):
LGTM

jfallows (commented):
LGTM

jfallows (commented):
LGTM

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -559,7 +557,7 @@ private void doClientBegin(
             }
             final McpBeginExFW beginEx = builder.build();
 
-            doBegin(sender, server.originId, resolvedId, initialId,
+            sender = openStream(this::onClientMessage, server.originId, resolvedId, initialId,
```
jfallows: This should be using `newStream` helper method, placed near end of factory, just before doBegin, doData, etc.

Same pattern as McpServerFactory and McpClientFactory.

specs/binding-mcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mcp/config/proxy.toolkit.multi.yaml
```diff
@@ -0,0 +1,28 @@
...
+  app0:
+    type: mcp
+    kind: proxy
+    routes:
+      - exit: app1
+        when:
+          - toolkit: bluesky
+      - exit: app2
+        when:
+          - toolkit: openai
```
jfallows: Use `quartz` as the other toolkit, and get_current_time as the tool in `quartz` if another tool is needed.

specs/binding-mcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mcp/streams/application/tools.list.toolkit/server.rpt
```diff
@@ -0,0 +1,60 @@
...
+#   https://www.aklivity.io/aklivity-community-license/
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations under the License.
+#
+
+property serverAddress "zilla://streams/app0"
+property toolsListReply "{\"tools\":[{\"name\":\"bluesky__get_weather\",\"title\":\"Weather Information Provider\"}]}"
```
jfallows: Do not parameterize tests like this, instead inline the value directly.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+            {
+                final McpRouteConfig route = binding.resolve(beginEx, authorization);
+                if (route != null)
+                {
+                    final McpSession session = sessions.computeIfAbsent(sessionId, McpSession::new);
+                    session.originId = originId;
+                    session.routedId = routedId;
+                    session.authorization = authorization;
+                    session.affinity = affinity;
+                    session.clientCapabilities = beginEx.lifecycle().capabilities();
```
jfallows: If McpSession is not a stream, why does it need to know these details?
Also if it already exists, then aren't we overwriting those details too?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+
+                    final McpLifecycleServer lifecycle = new McpLifecycleServer(
+                        sender, originId, routedId, initialId, affinity, authorization, session, binding);
```
jfallows: We don't usually pass binding to streams.
If they need it later, they can look it up by routedId in bindings map.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+                                affinity,
+                                authorization,
+                                beginKind,
+                                sessionId,
+                                identifier,
+                                prefix,
+                                0,
+                                session,
+                                exit)::onServerMessage;
+                        }
```
jfallows: We don't need to special case the atypical case with one route, since the multi-route case will still work when the size is 1.

Let's simplify.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+    private final class McpSession
+    {
+        private final String sessionId;
+        private final Long2ObjectHashMap<McpExit> exits;
+
+        private long originId;
+        private long routedId;
+        private long authorization;
+        private long affinity;
+        private int clientCapabilities;
```
jfallows: These should be on McpLifecycleServer instead of McpSession, agree?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+    {
+        private static final int UNINITIALIZED = 0;
+        private static final int OPENING = 1;
+        private static final int OPENED = 2;
+
+        private final long exitId;
+        private final Deque<McpPending> pending;
+
+        private int state;
+        private String exitSessionId;
```
jfallows:
```suggestion
        private String sessionId;
```
This is already in the context of the exit.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+            long traceId)
+        {
+            while (!exit.pending.isEmpty())
+            {
+                exit.pending.poll().doServerReset(traceId);
+            }
+        }
+    }
+
+    private final class McpListLeg implements McpPending
```
jfallows:
```suggestion
    private final class McpListClient implements McpPending
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+
+        @Override
+        public void doServerReset(
+            long traceId)
+        {
+            if (!completed)
+            {
+                completed = true;
+                listServer.legFailed(this, traceId);
+            }
```
jfallows: Stop using boolean flags, use McpState helper with state variable to track stream state properly on all stream classes.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+        }
+
+        private void doClientBegin(
+            long traceId)
+        {
+            initialId = supplyInitialId.applyAsLong(resolvedId);
+            replyId = supplyReplyId.applyAsLong(initialId);
```
jfallows: These should be created in the constructor and be final fields, same pattern as other streams.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+            final McpBeginExFW.Builder builder = mcpBeginExRW
+                .wrap(codecBuffer, 0, codecBuffer.capacity())
+                .typeId(mcpTypeId);
+            switch (listServer.beginKind)
+            {
+            case KIND_TOOLS_LIST -> builder.toolsList(t -> t.sessionId(sid));
+            case KIND_PROMPTS_LIST -> builder.promptsList(p -> p.sessionId(sid));
+            case KIND_RESOURCES_LIST -> builder.resourcesList(r -> r.sessionId(sid));
+            default -> throw new IllegalStateException("unexpected list kind: " + listServer.beginKind);
+            }
```
jfallows: We see this a lot, where the builder has to be broken up to support conditional behavior.

Perhaps it would make sense to add an `B inject(Consumer<B> injector)` to the generated flyweight builder classes, where it takes a consumer of the builder and then returns the builder afterwards, so the conditional logic can be in the inject method parameter, and the fluent style can remain unbroken?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+
+            sender = newStream(this::onClientMessage, listServer.originId, resolvedId, initialId,
+                traceId, listServer.authorization, listServer.affinity, beginEx);
+            state = McpState.openingInitial(state);
+        }
+
+        private void doClientEnd(
+            long traceId)
+        {
+            if (sender != null && !McpState.initialClosed(state))
```
jfallows: `sender` is always non-null, you can verify in the constructor or earlier if needed, but avoid this extra check throughout the code.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+        }
+
+        private void onClientData(
+            DataFW data)
+        {
+            final OctetsFW payload = data.payload();
+            if (payload != null)
+            {
+                bufferLegPayload(payload);
+            }
```
jfallows: Never allocate on the data path, see CLAUDE.md.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+
+        private void onClientEnd(
+            EndFW end)
+        {
+            final long traceId = end.traceId();
+            state = McpState.closedReply(state);
+            if (!completed)
+            {
+                completed = true;
+                final byte[] transformed = transformedLegReply();
```
jfallows: Never allocate on the data path, see CLAUDE.md.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+            while (!exit.pending.isEmpty())
+            {
+                exit.pending.poll().doServerReset(traceId);
+            }
+        }
+    }
+
+    private final class McpListLeg implements McpPending
+    {
+        private final McpListServer listServer;
```
jfallows:
```suggestion
        private final McpListServer server;
```
We are already in the context of a `list` request.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+            this.originId = originId;
+            this.routedId = routedId;
+            this.initialId = initialId;
+            this.replyId = supplyReplyId.applyAsLong(initialId);
+            this.affinity = affinity;
+            this.authorization = authorization;
+            this.beginKind = beginKind;
+            this.sessionId = sessionId;
+            this.session = session;
+            this.legs = new ArrayList<>(routes.size());
```
jfallows:
```suggestion
            this.clients = new ArrayList<>(routes.size());
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2009 @@
...
+            for (int i = 0; i < routes.size(); i++)
```
jfallows: We are iterating the routes, but don't we only need one leg/client per distinct exit?

If the concern is that multiple routes might have different toolkit names for the same exit, then the leg/client just needs to be aware of the multiple toolkit names and still only make the list request once, possibly prefixing with more than one prefix if it is actually duplicated in the routes.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -121,6 +130,15 @@ public final class McpClientFactory implements McpStreamFactory
     private final String clientName;
     private final String clientVersion;
 
+    // JSON paths whose values McpClientFactory reads via getString(). Configured on the
+    // StreamingJson parser so non-included values (notably /result subtree contents which
+    // pass through verbatim) scan without scratch accumulation. KEY_NAME tokens are
+    // always readable.
```
jfallows: Remove this comment.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -148,6 +166,9 @@ public McpClientFactory(
         this.signaler = context.signaler();
         this.clientName = config.clientName();
         this.clientVersion = config.clientVersion();
+        this.clientJsonConfig = Map.of(
+            StreamingJson.PATH_INCLUDES, CLIENT_JSON_PATH_INCLUDES,
```
jfallows, line 167: Let's change the expected type of the value for `StreamingJson.PATH_INCLUDES` from `List<JsonPointer>` to `List<String>` where the strings are JsonPointer syntax, but not required to be created by the caller, and it is up to the internal implementation of JsonParser to decide how best to represent them.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -240,7 +261,7 @@ private int decodeJsonRpc(
         DirectBufferInputStreamEx input = inputRO;
         input.wrap(buffer, offset, limit - offset);
 
-        http.decodableJson = StreamingJson.createParser(input);
+        http.decodableJson = StreamingJson.createParser(input, clientJsonConfig);
```
jfallows: Let's add `StreamingJsonParserFactory extends JsonParserFactory`, and `StreamingJsonProvider extends JsonProvider` in `common-json`.

Then add a shared static instance of StreamingJsonProvider in StreamingJson, and add a method StreamingJson.createParserFactory(Map<String, ?> config)` to mirror `Json..createParserFactory(Map<String, ?> config)`.

Then here we mirror Json approach and do StreamingJson.createParserFactory(config).createParser(input).

Note that the JsonParserFactory can be created once up front at the factory level and then used directly here to create the parser.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -65,6 +71,9 @@
 
 public final class McpClientFactory implements McpStreamFactory
 {
+    private static final int SERVER_CAPABILITIES =
+        SERVER_TOOLS.value() | SERVER_PROMPTS.value() | SERVER_RESOURCES.value();
+
```
jfallows, line 77:
```suggestion
    private static final int SERVER_CAPABILITIES =
        SERVER_TOOLS.value() |
        SERVER_PROMPTS.value() |
        SERVER_RESOURCES.value();

```

Improves readability.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
jfallows, line 1: See feedback on McpServerFactory and apply equivalent feedback here as well.

runtime/binding-mcp/src/test/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyIT.java
jfallows, line 1: Group the methods into order of lifecycle, tools, prompts, resources.

runtime/common-json/src/test/java/io/aklivity/zilla/runtime/common/json/internal/StreamingJsonTokenizerPathTest.java
```diff
@@ -0,0 +1,371 @@
...
+import jakarta.json.stream.JsonParsingException;
+
+import org.junit.jupiter.api.Test;
+
+import io.aklivity.zilla.runtime.common.json.StreamingJson;
+
+class StreamingJsonTokenizerPathTest
+{
+    @Test
+    void shouldTrackPathThroughObjectsAndArrays() throws IOException
```
jfallows: These should be public, agree?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -130,14 +131,10 @@ public final class McpClientFactory implements McpStreamFactory
...
-    // pass through verbatim) scan without scratch accumulation. KEY_NAME tokens are
-    // always readable.
-    private static final List<JsonPointer> CLIENT_JSON_PATH_INCLUDES = List.of(
-        Json.createPointer("/jsonrpc"),
-        Json.createPointer("/id"));
-    private final Map<String, ?> clientJsonConfig;
+    private static final List<String> CLIENT_JSON_PATH_INCLUDES = List.of(
+        "/jsonrpc",
+        "/id");
+    private final JsonParserFactory jsonParserFactory;
```
jfallows: Avoid naming variables same as class with lower case initial letter.
Use `parserFactory` as the name here, as JSON is implicit from the type and usage.

Same feedback for other factories.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1690,10 +1689,10 @@ private static void forEachListItem(
             return;
         }
         final String idKey = beginKind == KIND_RESOURCES_LIST ? "uri" : "name";
-        final JsonPointer idPath = Json.createPointer("/" + arrayKey + "/-/" + idKey);
-        final Map<String, ?> config = Map.of(
+        final String idPath = "/" + arrayKey + "/-/" + idKey;
+        final JsonParserFactory factory = StreamingJson.createParserFactory(Map.of(
             StreamingJson.PATH_INCLUDES, List.of(idPath),
-            StreamingJson.TOKEN_MAX_BYTES, jsonBytes.length);
+            StreamingJson.TOKEN_MAX_BYTES, jsonBytes.length));
```
jfallows: Use StreamingJson.createParserFactory at factory level, similar to McpClientFactory and McpServerFactory.

Remove StreamingJson.createParser(config, input) method.

runtime/binding-mcp/src/test/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyIT.java
```diff
@@ -76,6 +78,8 @@ public void shouldShutdownLifecycleRequests() throws Exception
         k3po.finish();
     }
 
+    // tools --------------------------------------------------------------------
+
```
jfallows: No need for grouping comments, remove them all.

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/StreamingJsonParser.java
jfallows: These helper methods should be at the end of the class.

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/StreamingJsonTokenizer.java
```diff
@@ -96,16 +95,15 @@ public StreamingJsonTokenizer(
...
 
     private static List<String[]> compilePaths(
-        List<JsonPointer> paths)
+        List<String> paths)
     {
         if (paths.isEmpty())
         {
             return List.of();
         }
         final List<String[]> compiled = new java.util.ArrayList<>(paths.size());
```
jfallows: Avoid fully qualified class names, use ArrayList not java.util.ArrayList here.

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/StreamingJsonTokenizer.java
```diff
@@ -96,16 +95,15 @@ public StreamingJsonTokenizer(
...
-        for (JsonPointer p : paths)
+        for (String s : paths)
```
jfallows: Use `path` instead of `s` for readability.

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/StreamingJson.java
```diff
@@ -68,4 +71,15 @@ public static JsonParser createParser(
     {
         return new StreamingJsonParser(in, config);
     }
```
jfallows: Remove this method.

jfallows (commented):
Some feedback not yet applied on `McpProxyFactory`.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+            return exits.computeIfAbsent(exitId, McpExit::new);
+        }
+    }
+
+    private interface McpPending
+    {
+        void proceed(long traceId);
+
+        void doServerReset(long traceId);
+    }
```
jfallows: We don't need this abstraction.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+        {
+            this.sessionId = sessionId;
+            this.exits = new Long2ObjectHashMap<>();
+        }
+
+        private McpExit supplyExit(
+            long exitId)
+        {
+            return exits.computeIfAbsent(exitId, McpExit::new);
+        }
```
jfallows: We only have one exit at a time now, should live directly on McpLifecycleServer / McpListServer, reduces session concept to `sessionId`.

We can keep a map of `McpLifecycleServer` by `sessionId` to discover it when needed.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+        private String sessionId;
+        private McpLifecycleClient lifecycle;
+
+        private McpExit(
+            long exitId)
+        {
+            this.exitId = exitId;
+            this.pending = new ArrayDeque<>();
+        }
+    }
```
jfallows: McpLifecycleServer can reference McpLifecycleClient(s) directly.
McpLifecycleClient state should be modeled directly on McpLifecycleClient, already covered by `state` field.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+
+        private McpServer(
+            MessageConsumer sender,
+            long originId,
+            long routedId,
+            long initialId,
+            long resolvedId,
+            long affinity,
+            long authorization,
+            int beginKind,
```
jfallows:
```suggestion
            int kind,
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+            McpExit exit)
+        {
+            this.sender = sender;
+            this.originId = originId;
+            this.routedId = routedId;
+            this.initialId = initialId;
+            this.replyId = supplyReplyId.applyAsLong(initialId);
+            this.affinity = affinity;
+            this.authorization = authorization;
+            this.beginKind = beginKind;
```
jfallows:
```suggestion
            this.kind = kind;
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+        }
+
+        public void proceed(
+            long traceId)
+        {
+            client.doClientBegin(traceId);
+
+            doWindow(sender, originId, routedId, initialId, traceId, authorization, 0,
+                writeBuffer.capacity(), 0);
+        }
```
jfallows: Remove and inline at call site.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+            final long traceId = data.traceId();
+            final long budgetId = data.budgetId();
+            final int flags = data.flags();
+            final int reserved = data.reserved();
+            final OctetsFW payload = data.payload();
+
+            if (!McpState.closed(state) && payload != null)
+            {
+                client.doClientData(traceId, budgetId, flags, reserved,
+                    payload.buffer(), payload.offset(), payload.sizeof());
```
jfallows: No need to check for `payload != null`.
No need to check for `!McpState.closed(state)`, cannot receive onServerData if closed.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+
+            if (!McpState.closed(state))
+            {
+                client.doClientEnd(traceId);
+            }
+
+            if (beginKind == KIND_LIFECYCLE)
+            {
+                sessions.remove(session.sessionId);
+            }
```
jfallows: This is dead code, right?
We have a dedicated McpLifecycleServer now.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+        }
+
+        private void onServerEnd(
+            EndFW end)
+        {
+            final long traceId = end.traceId();
+
+            state = McpState.closedInitial(state);
+
+            if (!McpState.closed(state))
```
jfallows: Remove the checks for `!McpState(.closed(state)`.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+        {
+            this.server = server;
+            this.resolvedId = resolvedId;
+        }
+
+        private void doClientBegin(
+            long traceId)
+        {
+            initialId = supplyInitialId.applyAsLong(resolvedId);
+            replyId = supplyReplyId.applyAsLong(initialId);
```
jfallows: Avoid reusing McpClient instead, keep these fields final.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+
+            final String identifier = server.identifier;
+            final int capabilities = server.capabilities;
+            final boolean lifecycle = server.beginKind == KIND_LIFECYCLE;
+            final String outboundSessionId = lifecycle || server.exit.sessionId == null
+                ? server.sessionId
+                : server.exit.sessionId;
+            final McpBeginExFW.Builder builder = mcpBeginExRW
+                .wrap(codecBuffer, 0, codecBuffer.capacity())
+                .typeId(mcpTypeId);
```
jfallows: What if we added to the flyweight builder to generate a method on builders.
```
xxxBuilder inject(Consumer<xxxBuilder> injector)
{
    injector.accept(this);
    return this;
}
```
Then we can inline the switch inside the .inject(...) call and keep the builder fluent instead of broken up.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+            case KIND_RESOURCES_LIST -> builder
+                .resourcesList(r -> r.sessionId(sid))
+                .build();
+            case KIND_RESOURCES_READ ->
+            {
+                final String uri = beginEx.resourcesRead().uri().asString();
+                yield builder.resourcesRead(r -> r.sessionId(sid).uri(uri)).build();
+            }
+            default -> beginEx;
+            };
```
jfallows: Move .build() call after and refactor this as `setSessionId(builder)`.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+        private void onClientData(
+            DataFW data)
+        {
+            final long traceId = data.traceId();
+            final long budgetId = data.budgetId();
+            final int flags = data.flags();
+            final int reserved = data.reserved();
+            final OctetsFW payload = data.payload();
+
+            if (payload != null)
```
jfallows: payload cannot be null

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+
+        private byte[] itemBuffer()
+        {
+            // grows as needed; reused across items
+            if (itemTransferBuffer == null || itemTransferBuffer.length < bufferPool.slotCapacity())
+            {
+                itemTransferBuffer = new byte[bufferPool.slotCapacity()];
+            }
+            return itemTransferBuffer;
+        }
```
jfallows: Remove this.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -0,0 +1,2019 @@
...
+                    break;
+                case END_ARRAY:
+                    depth--;
+                    if (inTargetArray && depth + 1 == itemDepth - 1)
+                    {
+                        inTargetArray = false;
+                        itemDepth = -1;
+                    }
+                    break;
+                case KEY_NAME:
```
jfallows: Replace this with the `decoder` approach, `decodeXxx` methods to track different decode states and avoid allocations on the data path.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -154,29 +154,28 @@ public MessageConsumer newStream(
...
             {
-                final McpSession session = sessions.get(sessionId);
-                if (session != null)
+                final McpLifecycleServer lifecycle = sessions.get(sessionId);
+                if (lifecycle != null)
                 {
-                    if (isListKind(beginKind))
+                    if (isListKind(kind))
                     {
                         final List<McpRouteConfig> routes = binding.resolveAll(beginEx, authorization);
```
jfallows:
```suggestion
                        final List<McpRoutePrefix> prefixes = binding.resolveAll(beginEx, authorization)
                            .stream()
                            .map(r -> new McpRoutePrefix(r.id, r.prefix(kind))
                            .toList();
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -186,9 +185,8 @@ public MessageConsumer newStream(
...
                             affinity,
                             authorization,
-                            beginKind,
-                            sessionId,
-                            session,
+                            kind,
+                            lifecycle,
                             routes,
                             beginEx);
                         newStream = server::onServerMessage;
```
jfallows:
```suggestion
                        newStream = new McpListServer(
                            lifecycle,
                            kind,
                            initialId,
                            affinity,
                            authorization,
                            prefixes)::onServerMessage;
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -210,13 +207,11 @@ public MessageConsumer newStream(
                                 route.id,
                                 affinity,
                                 authorization,
-                                beginKind,
-                                sessionId,
+                                kind,
                                 identifier,
                                 prefix,
                                 0,
```
jfallows: If capabilities is always 0 then let's remove it.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -210,13 +207,11 @@ public MessageConsumer newStream(
                                 route.id,
                                 affinity,
                                 authorization,
-                                beginKind,
-                                sessionId,
+                                kind,
                                 identifier,
                                 prefix,
```
jfallows:
```suggestion
                            newStream = new McpServer(
                                lifecycle,
                                kind,
                                sender,
                                originId,
                                routedId,
                                initialId,
                                route.id,
                                affinity,
                                authorization,
                                identifier,
                                prefix,
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -376,25 +322,7 @@ private void onServerBegin(
...
-                    exit.lifecycle = new McpLifecycleClient(exit, session);
-                    exit.lifecycle.doClientBegin(traceId);
-                }
-            }
-        }
-
-        public void proceed(
-            long traceId)
-        {
+            this.client = new McpClient(this, resolvedId);
```
jfallows: Initialize in constructor instead, mark client field final.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -424,14 +349,11 @@ private void onServerEnd(
...
-                client.doClientEnd(traceId);
-            }
+            client.doClientEnd(traceId);
 
-            if (beginKind == KIND_LIFECYCLE)
+            if (kind == KIND_LIFECYCLE)
             {
-                sessions.remove(session.sessionId);
+                sessions.remove(lifecycle.sessionId);
             }
```
jfallows: Can kind be lifecycle if we have separate stream for lifecycle?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -558,24 +478,27 @@ private McpClient(
...
-            replyId = supplyReplyId.applyAsLong(initialId);
+            lifecycle.doClientBegin(traceId);
 
             final String identifier = server.identifier;
             final int capabilities = server.capabilities;
-            final boolean lifecycle = server.beginKind == KIND_LIFECYCLE;
-            final String outboundSessionId = lifecycle || server.exit.sessionId == null
-                ? server.sessionId
-                : server.exit.sessionId;
+            final boolean lifecycleKind = server.kind == KIND_LIFECYCLE;
```
jfallows: I think this predates creation of separate McpLifecycleServer stream.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -947,67 +873,61 @@ private void onServerReset(
...
 
+        private MessageConsumer sender;
         private int state;
+        private String sessionId;        // upstream-provided session id, set on BEGIN reply
 
         private McpLifecycleClient(
-            McpExit exit,
-            McpSession session)
+            McpLifecycleServer server,
+            long exitId)
```
jfallows: Use routedId instead of exitId naming.
Also define local `originId` initialized to `server.routedId` for clarity.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1061,71 +981,45 @@ private void onClientBegin(
...
-            }
+            state = McpState.openedReply(state);
         }
 
         private void onClientEnd(
             EndFW end)
         {
-            final long traceId = end.traceId();
-
             state = McpState.closedReply(state);
```
jfallows, line 1318: We need to respond to this specific client with doClientEnd and also remove the client lifecycle from the server lifecycle.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1061,71 +981,45 @@ private void onClientBegin(
...
-
-            failPending(traceId);
         }
 
         private void onClientAbort(
             AbortFW abort)
         {
-            final long traceId = abort.traceId();
-
             state = McpState.closedReply(state);
```
jfallows, line 1338: We need to respond to this specific client with doClientAbort and also remove the client lifecycle from the server lifecycle.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1061,71 +981,45 @@ private void onClientBegin(
...
-
-            failPending(traceId);
         }
 
         private void onClientReset(
             ResetFW reset)
         {
-            final long traceId = reset.traceId();
-
             state = McpState.closedInitial(state);
```
jfallows, line 1378: We need to respond to this specific client with doClientReset and also remove the client lifecycle from the server lifecycle.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1284,19 +1159,16 @@ private void onClientBegin(
         private void onClientData(
             DataFW data)
         {
+            final long traceId = data.traceId();
             final OctetsFW payload = data.payload();
-            if (payload != null)
+            final boolean appended = appendToSlot(payload);
```
jfallows: appendToSlot is not following the decoder pattern, see how McpClientFactory and McpServerFactory handle onXxxData and call to decodeNet, here we would have a method just called decode instead of decodeNet.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1400,88 +1272,7 @@ private void streamItems(
...
-                    }
-                    else if (awaitingIdValue)
-                    {
-                        awaitingIdValue = false;
-                    }
-                    break;
-                default:
-                    break;
-                }
+                decoder.decode(this, ev, traceId, beforeStreamOffset, afterStreamOffset, slotBuffer);
```
jfallows: Convert to decodeNet approach, call method decode instead.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1525,6 +1304,136 @@ private void cleanupClientSlot()
...
+    @FunctionalInterface
+    private interface McpListClientDecoder
+    {
+        void decode(
+            McpListClient client,
+            JsonParser.Event event,
+            long traceId,
+            long beforeOffset,
+            long afterOffset,
+            MutableDirectBuffer slotBuffer);
```
jfallows: Method signature should align with McpClientFactory and McpServerFactory equivalent inner functional interfaces.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1525,6 +1304,136 @@ private void cleanupClientSlot()
...
+    }
+
+    private final McpListClientDecoder decodeListBeforeItems = this::decodeListBeforeItems;
+    private final McpListClientDecoder decodeListWithinItems = this::decodeListWithinItems;
```
jfallows: decodeReply - start object to reply
decodeItemsKey - find matching key -> decodeItems
decodeSkipObject - for non matching keys
decodeItems - start array for items
decodeItem - decode each item until end-array

This design is similar to one for McpServerFactory requests and McpClientFactory responses.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1563,9 +1477,8 @@ private McpListServer(
             long initialId,
             long affinity,
             long authorization,
-            int beginKind,
-            String sessionId,
-            McpSession session,
+            int kind,
+            McpLifecycleServer lifecycle,
             List<McpRouteConfig> routes,
             McpBeginExFW beginEx)
```
jfallows: Avoid passing flyweights to constuctors, prevents misuse.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1630,7 +1540,7 @@ private void onServerBegin(
 
             doServerBegin(traceId);
             emitPrelude(traceId);
```
jfallows:
```suggestion
            doEncodeBeginItems(traceId);
```

Then later `doEncodeEndItems(traceId)` too.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1639,7 +1549,7 @@ private void onServerEnd(
             final long traceId = end.traceId();
             state = McpState.closedInitial(state);
 
-            if (currentClient != null && McpState.initialOpened(currentClient.state))
+            if (currentClient != null)
```
jfallows: Rename currentClient to client.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1669,54 +1578,40 @@ private void onServerReset(
...
             long traceId)
         {
             currentClient = null;
-            startNextRoute(traceId);
+            onNextClient(traceId);
         }
 
-        private void clientFailed(
+        private void onClientError(
             McpListClient client,
```
jfallows: No need to pass client, we already have it as the current client.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyFactory.java
```diff
@@ -1669,54 +1578,40 @@ private void onServerReset(
             {
                 currentClient.doClientReset(traceId);
             }
-            // queued clients have not opened an upstream stream — drop them
             remaining.clear();
         }
 
-        private void clientDone(
+        private void onClientClosed(
             McpListClient client,
```
jfallows: No need to pass client, we already have it as the current client.

jfallows:
Pushed `733dd707` addressing this review batch.

**Addressed:**
- L180/L192 — Added `McpRoutePrefix(resolvedId, prefix)` record and `McpRouteConfig.prefix(int kind)`; factory site resolves prefixes via `binding.resolveAll(...).stream().map(r -> new McpRoutePrefix(r.id, r.prefix(kind))).toList()`. `McpListServer` no longer takes the `beginEx` flyweight; constructor is now `(lifecycle, kind, initialId, affinity, authorization, prefixes)` with `sender/originId/routedId` derived from `lifecycle`.
- L212/L213 — `McpServer` constructor reordered with `lifecycle, kind` first; always-zero `capabilities` parameter removed.
- L325 — `McpServer.client` is now `final`, created in the constructor; `resolvedId` field dropped (passed straight into `new McpClient(this, resolvedId)`).
- L357/L493 — Dead `kind == KIND_LIFECYCLE` branches removed from `McpServer.onServerEnd`/`onServerAbort` and from `McpClient.doClientBegin`/`rewriteReplyBeginEx` (lifecycle has its own server/client streams).
- L898 — `McpLifecycleClient.exitId` renamed to `routedId`; introduced local `originId = server.routedId` at each send site for clarity.
- L997/L1003/L1009 — `McpLifecycleClient.onClientEnd/Abort/Reset` now send the corresponding `doClientEnd/Abort/Reset` half-close back and remove `this` from `server.clients`. Added `doClientAbort` and `doClientReset`.
- L1483 — `McpListServer` no longer takes a flyweight in its constructor; prefixes are pre-extracted at the factory.
- L1542 — `emitPrelude` → `doEncodeBeginItems`; `closeReply` → `doEncodeEndItems`.
- L1552 — `currentClient` → `client`.
- L1585/L1593 — `onClientClosed`/`onClientError` no longer take a redundant `client` parameter.
- Allocation hygiene — `LIST_REPLY_*_OPEN/CLOSE/SEPARATOR` are now factory-level `UnsafeBuffer`-wrapped `DirectBuffer` constants; `streamItem` reuses a factory-level `itemRO` `UnsafeBuffer` instead of allocating per item.

**Deferred to a follow-up commit:** L1164/L1275/L1316/L1320 — restructuring the JSON-event decoder into the `decodeReply` → `decodeItemsKey` → `decodeSkipObject` → `decodeItems` → `decodeItem` decomposition with the `(client, traceId, authorization, budgetId, reserved, buffer, offset, progress, limit)` signature alignment is non-trivial because the current path is driven by the `JsonParser` event model rather than byte-level decoding, so I'd like to do that as its own focused commit.

All 19 `McpProxyIT` tests pass; jacoco coverage holds; checkstyle clean.

---

jfallows:
Pushed `f42446f3` with the decoder restructure for `McpListClient` (the deferred item from the previous batch).

**`McpListClientDecoder`** signature aligned with `McpServerDecoder`:
```java
int decode(
    McpListClient client,
    long traceId,
    long authorization,
    long budgetId,
    int reserved,
    DirectBuffer buffer,
    int offset,
    int progress,
    int limit);
```

**Five decode states** replace `decodeListBeforeItems` / `decodeListWithinItems`:
- `decodeReply` — consume the `START_OBJECT` of the reply envelope.
- `decodeItemsKey` — at depth 1, look for the matching key (`tools` / `prompts` / `resources`); on match → `decodeItems`, on non-match → `decodeSkipObject`; on `END_OBJECT` → `decodeIgnore`.
- `decodeSkipObject` — skip the value of a non-matching key, tracking `decodeSkipDepth` so containers and primitives both terminate correctly.
- `decodeItems` — consume the `START_ARRAY` for the items.
- `decodeItem` — emit each item between `START_OBJECT`/`END_OBJECT` pairs at the array level, splicing the configured prefix in front of the item's id (`name` / `uri`); transitions back to `decodeItemsKey` on `END_ARRAY`.
- `decodeIgnore` — terminal state, consumes remaining input.

**Driver** `McpListClient.decode(...)` mirrors `McpServerFactory.decodeNet`: accumulate to slot if needed, drive `while (previous != decoder)` loop, then compact prior to `itemStartStreamOffset` (when mid-item) or `parser.getLocation().getStreamOffset()` (when between items). `streamItems` / `appendToSlot` are gone.

**Field hygiene:** `inputRO` (`DirectBufferInputStreamEx`) hoisted to the factory and shared across all `McpListClient` instances on the worker; the per-stream `jsonInput` field removed. `parser` → `decodableJson`; `depth` / `itemDepth` / `skipDepth` → `decodeDepth` / `decodeItemDepth` / `decodeSkipDepth`.

All 19 `McpProxyIT` tests pass; jacoco coverage holds; checkstyle clean.

Note: the `testing (http.kafka.crud)` CI failure earlier appears to have been transient — it passed on the next run before this commit. The new commit's checks are running now.

---

jfallows:
Rebased onto `develop` (now picking up #1738) and pushed `c90f7d8a` simplifying the four sites that previously broke the fluent `McpBeginExFW.Builder` chain into a temporary variable so a switch could mutate it:

- `McpClient.doClientBegin` — request BEGIN ext
- `McpClient.rewriteReplyBeginEx` — reply BEGIN ext rewrite (keeps its short-circuit for unrecognised reply kinds via a single guard return at the top)
- `McpListClient.doClientBegin` — upstream list-request BEGIN ext
- `McpListServer.doServerBegin` — downstream list-reply BEGIN ext

All four now read as a single chained expression:

```java
final McpBeginExFW beginEx = mcpBeginExRW
    .wrap(codecBuffer, 0, codecBuffer.capacity())
    .typeId(mcpTypeId)
    .inject(b ->
    {
        switch (kind)
        {
        case KIND_TOOLS_LIST -> b.toolsList(...);
        case KIND_PROMPTS_LIST -> b.promptsList(...);
        case KIND_RESOURCES_LIST -> b.resourcesList(...);
        default -> throw new IllegalStateException("unexpected list kind: " + kind);
        }
    })
    .build();
```

(Block lambda is needed because `Consumer.accept` returns void, which is not a valid switch-expression target — so the switch stays a statement.)

All 19 `McpProxyIT` tests pass; jacoco coverage holds; checkstyle clean.

---

PR #1710, opened 2026-04-07
feat(store-memory): implement in-process memory store

### LINKED ISSUE #1667: store-memory: implement in-process memory store
## Summary

Implement the `memory` store type in a new `runtime/store-memory` module. This is the default store for single-instance deployments and development environments.

## Implementation

* `MemoryStoreFactorySpi` — registers the `memory` type name
* `MemoryStore` — creates per-core `MemoryStoreContext` instances
* `MemoryStoreContext` — supplies the `MemoryStoreHandler`
* `MemoryStoreHandler` — delegates all operations to the shared `MemoryStore` instance which owns the backing `ConcurrentHashMap<String, MemoryEntry<?>>`, but state is isolated per store to support multiple named stores

## Behaviour

* All `StoreHandler` operations invoke `completion.completed()` synchronously on the calling thread (inline, no thread switch)
* `putIfAbsent` uses `ConcurrentHashMap.putIfAbsent` — atomic
* `getAndDelete` uses `ConcurrentHashMap.remove` — atomic, returns previous value if present
* Entries carry an expiry timestamp; `get` returns `null` for expired entries (lazy expiry — entry not removed on miss)
* `memory` is intentionally unbounded — no eviction occurs; entries are retained until TTL expiry or explicit `delete`/`getAndDelete`. Size-bound eviction would silently corrupt store-dependent behaviour (OAuth nonces, idempotency keys, session tokens) and is not supported. For high-cardinality or multi-instance deployments, use a distributed store implementation.

## Key implementation sketches

```java
// get — inline, no allocation
@Override
public <V> void get(String key, Class<V> type, StoreCompletion<V> completion) {
    MemoryEntry<?> entry = map.get(key);
    V value = entry != null && !entry.expired() ? type.cast(entry.value) : null;
    completion.completed(value, null);  // inline on calling thread
}

// getAndDelete — atomic via ConcurrentHashMap.remove
@Override
public <V> void getAndDelete(String key, Class<V> type, StoreCompletion<V> completion) {
    MemoryEntry<?> entry = map.remove(key);
    V value = entry != null && !entry.expired() ? type.cast(entry.value) : null;
    completion.completed(value, null);
}
```

## Configuration

```yaml
stores:
  session_store0:
    type: memory
```

## Module

`runtime/store-memory`

Service loader registration:

```
META-INF/services/io.aklivity.zilla.runtime.engine.store.StoreFactorySpi
io.aklivity.zilla.runtime.store.memory.internal.MemoryStoreFactorySpi
```

## Dependencies

* `engine: add Store concept` — `Store`, `StoreContext`, `StoreHandler`, `StoreFactorySpi` interfaces

## Implementation Notes

`MemoryStore` is a single shared instance that acts as a concurrent write-through cache. It owns the backing `ConcurrentHashMap` and coordinates all atomic operations (`putIfAbsent`, `getAndDelete`) directly. Each call to `supply(context)` returns a new `MemoryStoreContext` wrapping a `MemoryStoreHandler` that delegates all operations back to the shared `MemoryStore` instance — handlers are lightweight per-worker facades, not independent stores. TTL expiry is managed by the shared instance via lazy expiry on read; no background sweep or eviction thread is used.

jfallows:
## Summary

Closes #1667

Implements the `memory` store type in a new `runtime/store-memory` module. This is the default store for single-instance deployments and development environments.

- `MemoryStoreFactorySpi` — registers the `memory` type name
- `MemoryStore` — shared `ConcurrentHashMap`-backed state, created once per store instance
- `MemoryStoreContext` — per-worker context supplying `MemoryStoreHandler`
- `MemoryStoreHandler` — delegates all operations inline to `MemoryStore` (synchronous, no thread switch)
- `specs/store-memory.spec/` — schema patch registering `memory` as a valid store type
- `MemoryStoreHandlerTest` — unit tests covering all 5 operations and TTL expiry

## Behaviour

- All `StoreHandler` operations invoke `completion` synchronously on the calling thread
- `putIfAbsent` uses `ConcurrentHashMap.putIfAbsent` — atomic
- `getAndDelete` uses `ConcurrentHashMap.remove` — atomic
- Entries carry an expiry timestamp; `get` returns `null` for expired entries (lazy expiry)
- Intentionally unbounded — no eviction; entries are retained until TTL expiry or explicit removal

## Test plan

- [x] `./mvnw install -DskipITs -pl runtime/store-memory -am` compiles without errors
- [x] `./mvnw test -pl runtime/store-memory` — all unit tests pass
- [x] A `zilla.yaml` with `stores: type: memory` parses without schema error


runtime/store-memory/src/main/java/io/aklivity/zilla/runtime/store/memory/internal/MemoryStore.java
```diff
@@ -0,0 +1,110 @@
...
+        final MemoryEntry existing = entries.putIfAbsent(key, newEntry);
+        if (existing == null || existing.expired())
+        {
+            if (existing != null && existing.expired())
+            {
+                entries.replace(key, existing, newEntry);
+            }
+            return null;
+        }
+        return existing.value;
```
jfallows:
```suggestion
        if (existing != null && existing.expired())
        {
            entries.replace(key, existing, newEntry);
            existing.value = null;
        }
        return existing != null ? existing.value : null;
```

runtime/store-memory/src/main/java/io/aklivity/zilla/runtime/store/memory/internal/MemoryEntry.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.aklivity.zilla.runtime.store.memory.internal;
+
+final class MemoryEntry
```
jfallows: Suggest using `record` instead.

runtime/store-memory/src/main/java/io/aklivity/zilla/runtime/store/memory/internal/MemoryStore.java
```diff
@@ -0,0 +1,61 @@
...
+    public String name()
+    {
+        return NAME;
+    }
+
+    @Override
+    public StoreContext supply(
+        EngineContext context)
+    {
+        return new MemoryStoreContext(this);
```
jfallows: Instead pass `this::supplyEntries` and rename `attach` to `supplyEntries`, also make it `private`.
Then `MemoryStoreContext` can have a parameter of type `LongFunction<ConcurrentMap<String, MemoryEntry>>`.

runtime/store-memory/src/main/java/io/aklivity/zilla/runtime/store/memory/internal/MemoryStore.java
```diff
@@ -0,0 +1,61 @@
...
+
+import io.aklivity.zilla.runtime.engine.EngineContext;
+import io.aklivity.zilla.runtime.engine.store.Store;
+import io.aklivity.zilla.runtime.engine.store.StoreContext;
+
+final class MemoryStore implements Store
+{
+    static final String NAME = "memory";
+
+    private final ConcurrentMap<Long, ConcurrentMap<String, MemoryEntry>> entries;
```
jfallows: Suggest renaming this as `storage` instead of `entries`.

runtime/store-memory/src/test/java/io/aklivity/zilla/runtime/store/memory/internal/MemoryStoreHandlerTest.java
```diff
@@ -0,0 +1,122 @@
...
+import org.junit.Test;
+
+public class MemoryStoreHandlerTest
+{
+    private MemoryStoreHandler handler;
+
+    @Before
+    public void setUp()
+    {
+        handler = new MemoryStoreHandler(new MemoryStore(new MemoryStoreConfiguration()).attach(0L));
```
jfallows: Use the normal pattern to create the handler, via `MemoryStore` `supplyContext(...)`, `attach(StoreConfig)`.

runtime/store-memory/src/test/java/io/aklivity/zilla/runtime/store/memory/internal/MemoryStoreHandlerTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Test
+    public void shouldReturnNullForMissingKey()
+    {
+        final AtomicReference<String> result = new AtomicReference<>();
```
jfallows: Do we need `AtomicReference<String>` in these tests, or is `MutableReference<String>` sufficient?

runtime/store-memory/src/main/java/io/aklivity/zilla/runtime/store/memory/internal/MemoryStore.java
```diff
@@ -53,9 +53,15 @@ public URL type()
...
         long storeId)
     {
-        return entries.computeIfAbsent(storeId, id -> new ConcurrentHashMap<>());
+        return storage.computeIfAbsent(storeId, id -> new ConcurrentHashMap<>());
+    }
+
+    private void removeEntries(
+        long storeId)
+    {
+        storage.remove(storeId);
```
jfallows: There can be multiple contexts per store, one context per engine worker, so removing the entries from storage on the first call to detach is not correct.

Instead we need a reference count on the storage, and only remove when it reaches zero.

Perhaps the lambdas should be called acquireEntries and releaseEntries instead?

PR #1734, opened 2026-04-23
feat(engine): add store reference, preauthorize, and roles on GuardHandler

akrambek:
## Summary

- `GuardConfig` gains an optional `store` field (plus transient `storeId`/`qstore`), mirroring how `BindingConfig` references a vault. Resolved at namespace init; guards retrieve the `StoreHandler` at runtime via `EngineContext.supplyStore(storeId)`.
- `GuardHandler` gains a new `NEEDS_PREAUTHORIZE = -1L` sentinel and two default methods:
  - `preauthorize(traceId, bindingId, contextId, callback)` — returns the URL the user must visit when an out-of-band consent step is needed. The caller passes a callback URL the guard treats as opaque and embeds in the returned authorization URL.
  - `roles(sessionId)` — read-only scope query for filter-time evaluation; never triggers credential acquisition or any other side effect.
- `engine.schema.json` accepts the new `store` property on the `guard` object.

All changes are purely additive — no existing guard runtime or caller needs to change. Default `preauthorize` returns \`null\`, default \`roles\` returns \`Set.of()\`, and `reauthorize` retains its \`0L\` / positive contract for guards that never need pre-authorization.

## Motivation

Engine-side primitives required by the binding-mcp elicitation flow ([#1739](https://github.com/aklivity/zilla/issues/1739)): the \`mcp · client\` binding needs \`NEEDS_PREAUTHORIZE\` and \`preauthorize\` to drive an interactive consent step when a backend MCP server requires it, and \`roles\` for filter-time tool listing on \`mcp · cache\`.

The new \`store\` reference on \`GuardConfig\` lets guards that maintain cross-request state — for example, cached credentials or short-lived correlation tokens — share a backing store across workers and replicas.

## Test plan

- [x] \`GuardFactoryTest\` covers both new defaults: \`shouldReturnNullPreauthorizeByDefault\`, \`shouldReturnEmptyRolesByDefault\`
- [x] \`NamespaceConfigAdapterTest\` gains read/write round-trip cases for a guard with \`store: session_store0\`
- [x] All 313 existing engine tests still pass (\`./mvnw -pl runtime/engine test\`)
- [x] \`guard-jwt\` still builds against the updated engine (\`./mvnw -pl runtime/guard-jwt -am clean install\`)

### DIFF
--- a/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
+++ b/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
@@ -1143,7 +1143,7 @@ public final class HttpServerFactory implements HttpStreamFactory
                         }
                         else
                         {
-                            error = guard != null && exchangeAuth == NOT_AUTHORIZED
+                            error = guard != null && exchangeAuth <= NOT_AUTHORIZED
                                 ? credentialsMatch != null ? response403 : response401
                                 : response404;
                         }
@@ -5102,7 +5102,7 @@ public final class HttpServerFactory implements HttpStreamFactory
                         if (route == null)
                         {
                             Array32FW<HttpHeaderFW> headers40x =
-                                guard != null && exchangeAuth == NOT_AUTHORIZED
+                                guard != null && exchangeAuth <= NOT_AUTHORIZED
                                     ? credentialsMatch != null ? headers403 : headers401
                                     : headers404;
                             doEncodeHeaders(traceId, authorization, streamId, headers40x, true);
@@ -6107,7 +6107,7 @@ public final class HttpServerFactory implements HttpStreamFactory
 
             private void deauthorizeIfNecessary()
             {
-                if (HttpState.closed(state) && sessionId != 0L && guard != null)
+                if (HttpState.closed(state) && sessionId > NOT_AUTHORIZED && guard != null)
                 {
                     guard.deauthorize(sessionId);
                 }
@@ -7233,7 +7233,7 @@ public final class HttpServerFactory implements HttpStreamFactory
         int contextId)
     {
         long expiringId = NO_CANCEL_ID;
-        if (sessionId != NOT_AUTHORIZED)
+        if (sessionId > NOT_AUTHORIZED)
         {
             final long expiringAt = guard.expiringAt(sessionId);
             if (expiringAt != EXPIRES_NEVER)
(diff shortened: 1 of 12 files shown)


jfallows (commented):
Now that the return value of reauthorize could be non-zero and not a valid sessionId, please double check all the call sites of GuardHandler.reauthorize to ensure any check of the sessionId afterwards is > 0 vs != 0 to verify a valid sessionId.

This is needed for _all_ guard reauthorize callers.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -1143,7 +1143,7 @@ else if (!isCorsRequestAllowed(server.binding, headers))
                         }
                         else
                         {
-                            error = guard != null && exchangeAuth == NOT_AUTHORIZED
+                            error = guard != null && exchangeAuth <= NOT_AUTHORIZED
```
jfallows: I think we want to treat this as a mask instead, as it's counter intuitive to say `<= NOT_AUTHORIZED`.

How about we define `MASK_AUTHORIZED = 0x7fff_ffff_ffff_ffffL` and `NEED_PREAUTHORIZE = 0x8000_0000_0000_0000L` then say `(sessionId & MASK_AUTHORIZED) != 0` to detect valid sessionId and `(sessionId & MASK_AUTHORIZED) ==0` to detect invalid sessionId.

I think this probably reads better in the code and also keeps it closer to the original logic with `!= 0` (good) and `== 0` (bad).

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/guard/GuardHandler.java
```diff
@@ -36,11 +36,24 @@ public interface GuardHandler
...
+     * out-of-band pre-authorization step before a session can be created. The caller
+     * should then invoke {@link #preauthorize} to obtain the URL the user must visit.
+     * <p>
+     * The value is encoded with the high bit set and {@link #MASK_AUTHORIZED} bits zero,
+     * so callers can use {@code (sessionId & MASK_AUTHORIZED) != 0} to test for a valid
+     * session and {@code (sessionId & MASK_AUTHORIZED) == 0} to test for any
+     * unauthorized result without enumerating the individual sentinel values.
+     * </p>
+     */
+    long NEED_PREAUTHORIZE = 0x8000_0000_0000_0000L;
```
jfallows: `NEEDS_PREAUTHORIZE`

jfallows (changes requested):

jfallows (approved):

PR #1738, opened 2026-04-27
feat(flyweight-maven-plugin): add inject(Consumer&lt;Builder&gt;) on generated struct + union builders

jfallows:
Adds `inject(Consumer<Builder>)` to generated struct and union flyweight builders so conditional mutations can stay inside the fluent chain instead of forcing it to be broken into a temporary variable.

Picks up on the suggestion in #1709 ([discussion thread](https://github.com/aklivity/zilla/pull/1709#discussion_r3141297926)).

## Before

```java
XxxFW.Builder b = builderRW.wrap(buf, 0, capacity).typeId(typeId);
if (cond)
{
    b.someField(value);
}
XxxFW msg = b.build();
```

## After

```java
XxxFW msg = builderRW.wrap(buf, 0, capacity)
    .typeId(typeId)
    .inject(b -> { if (cond) b.someField(value); })
    .build();
```

## Implementation notes

- Wired in via `StructFlyweightGenerator.BuilderClassGenerator.generate` and `UnionFlyweightGenerator.BuilderClassGenerator.generate`.
- Generated method signature is per-subclass: `public Builder inject(Consumer<Builder> injector)`. The lambda parameter receives the concrete builder type without an explicit type witness — Java's inference picks the receiver type directly.
- I prototyped placing this on `Flyweight.Builder<T>` as a generic method `<B extends Builder<T>> B inject(Consumer<B>)` first; that approach forced callers to write `.<XxxFW.Builder>inject(...)` because `B` isn't constrained by the receiver's runtime type. Per-subclass generation gives clean ergonomics without invasive F-bounded changes to the base class (which would have rippled into all hand-written `Flyweight.Builder<T>` subclasses across the bindings).
- Other flyweight shapes (list, map, array, string, bounded-octets, variant) can adopt the same pattern in follow-up commits if needed; this PR covers the cases that unblock the proxy / extension-builder simplification in #1709.

## Test plan

- [x] `./mvnw test -pl build/flyweight-maven-plugin` — all flyweight-maven-plugin tests pass, including new ones below.
- [x] `FlatFWTest.shouldChainViaInjectPreservingBuilderType` — struct case, chain continues after `inject(...)` with subclass methods.
- [x] `UnionChildFWTest.shouldChainViaInjectPreservingBuilderType` — union case, ditto.
- [x] `./mvnw clean install -pl runtime/engine,runtime/binding-mcp -am -DskipTests -DskipITs` — sanity build of a binding consuming the regenerated flyweights succeeds.

---
_Generated by [Claude Code](https://claude.ai/code/session_0196ranuZwz1vjCEQ9mGHhfW)_

### DIFF
--- a/build/flyweight-maven-plugin/src/main/java/io/aklivity/zilla/build/maven/plugins/flyweight/internal/generate/StructFlyweightGenerator.java
+++ b/build/flyweight-maven-plugin/src/main/java/io/aklivity/zilla/build/maven/plugins/flyweight/internal/generate/StructFlyweightGenerator.java
@@ -1575,6 +1575,7 @@ public final class StructFlyweightGenerator extends ClassSpecGenerator
                           .addMethod(wrapMethodWithArray.generate())
                           .addMethod(rewrapMethod())
                           .addMethod(setMethod())
+                          .addMethod(injectMethod())
                           .addMethod(buildMethod())
                           .build();
         }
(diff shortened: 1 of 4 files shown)


jfallows (commented):
LGTM

PR #1741, opened 2026-04-29
feat(engine): add async reauthorize overload on GuardHandler

akrambek:


runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/guard/GuardHandler.java
```diff
@@ -84,6 +86,36 @@ long reauthorize(
...
+     */
+    default void reauthorize(
+        long traceId,
+        long bindingId,
+        long contextId,
+        String credentials,
+        LongConsumer onResult)
+    {
+        onResult.accept(reauthorize(traceId, bindingId, contextId, credentials));
+    }
```
jfallows, line 130: I was thinking maybe we want something more general purpose for Guards, inspired by [JDK CompletionHandler](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/CompletionHandler.html) for example.

```
interface GuardHandler
{
...
    long reauthorize(
        long traceId,
        long bindingId,
        long contextId,
        String credentials);

    default void reauthorize(
        long traceId,
        long bindingId,
        long contextId,
        String credentials,
        LongCompletionCallback completion)
    {
        try
        {
            long result = reauthorize(traceId, bindingId, contextId, credentials);
            completion.completed(contextId, result);
        }
        catch (Throwable ex)
        {
            completion.failed(contextId, ex);
        }
    }
...
    interface CompletionCallback<V>
    {
        void completed(
            long contextId,
            V result);

        void failed(
            long contextId,
            Throwable ex);
    }

    interface LongCompletionCallback extends CompletionCallback<Long>
    {
        default void completed(
            long contextId,
            Long result);
        {
            completed(contextId, result.longValue());
        }

        void completed(
            long contextId,
            long result);
    }
}
```

I suspect that bindings calling this async method with completion handler will want to react to the callback by sending a signal to the originating stream, just to make sure we get back to single threaded data structures, so we may want to fill out some more fields in the callback completed and failed signatures to make that easier to get back to the right stream.

jfallows (approved):

PR #1743, opened 2026-05-01
feat(engine): drive guard preauthorize from test binding; resolveId on GuardConfig

akrambek:
## Summary

Two related additions, both about letting guards participate in the config layer the way bindings already do.

### 1. `GuardConfig.resolveId`

Mirror of `BindingConfig.resolveId`. `EngineManager` populates it from the same `NameResolver` used for bindings, so a guard implementation can resolve another guard's name to its namespaced id at attach time.

This unblocks patterns where a guard references another guard — e.g. a `via` reference to an upstream guard that holds an inbound subject credential, so a delegated guard can call `engine.supplyGuard(viaId).credentials(contextId)` to fetch the user's bearer for an upstream exchange.

### 2. Test binding async authorization + preauthorize orchestration

Today the engine's `TestBindingFactory` only invokes the synchronous `reauthorize` overload. Any guard that returns `NEEDS_PREAUTHORIZE` on the first call cannot be driven through the test binding alone. With these additions any guard that overrides `preauthorize` can be exercised end-to-end via `EngineRule` + `@Configuration`.

Per the existing [`GuardHandler.preauthorize` contract](https://github.com/aklivity/zilla/blob/develop/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/guard/GuardHandler.java), the redirect-callback URL (with whatever query params the guard embedded) is passed back to `reauthorize` as the credentials. The test binding stays **guard-agnostic** — it forwards the preauthorize URL's query string verbatim onto the configured callback URL and appends any additional `callbackParams` declared in the test config; it never interprets any specific param name itself.

- **Test binding async authorization** — `TestBindingFactory.newStream` invokes the async `reauthorize` overload on every stream begin and defers route resolution until the callback fires. Stream begin is buffered until the deferred mode resolves; on success the post-auth body runs as before, on failure a reset is sent. Synchronous guards continue to work via the default async-overload bridge that delegates to the synchronous `reauthorize`.

- **Preauthorize orchestration** — when async `reauthorize` resolves with `NEEDS_PREAUTHORIZE` and the authorization config supplies a `callback` URL, the test binding calls `preauthorize`, copies the preauthorize URL's query string onto the callback URL, appends any configured `callbackParams`, and calls `reauthorize` again with the resulting URL as the credentials.

- **`TestAuthorizationConfig` + builder + adapter + schema patch:**
  - `callback` — optional callback URL the guard should embed in its preauthorize URL.
  - `callbackParams` — optional string-to-string map of additional query params to append to the callback response (e.g. an authorization code returned by a simulated upstream).
  - `expectIdentity`, `expectCredentials` — optional assertions; after authorization resolves, the test binding resets the stream if `guard.identity(sessionId)` or `guard.credentials(sessionId)` doesn't match. Mirrors the existing `attributes:` assertion pattern.

### 3. `NamespaceRegistry.attach` order fix

Stores now attach **before** guards. Previously guards attached first, so a guard with a configured `store: …` could not resolve it via `context.supplyStore` at attach time.

## Example

```yaml
stores:
  store0:
    type: test
guards:
  jwt0:
    type: jwt
    options: { ... }
  guard0:
    type: my-pre-authz-guard
    store: store0
    options:
      via: "${guarded['jwt0'].credentials}"
      callback: /cb
      ...
bindings:
  net0:
    type: test
    kind: server
    options:
      authorization:
        guard0:
          credentials: subject-token
          callback: /cb
          callbackParams:
            code: stub-code
          expectCredentials: ACCESS-FROM-IDP
    exit: app0
```

The K3PO connect script triggers `newStream`; the test binding drives `reauthorize → preauthorize → reauthorize(callbackURL)` end-to-end against the guard. After authorization resolves, `expectCredentials` asserts the resulting session's bearer matches the upstream-issued token. The guard implementation, in turn, can use `guard.resolveId.applyAsLong("jwt0")` at attach time to thread the inbound subject credential through to the upstream exchange.

## Test plan

- [ ] `./mvnw clean verify -pl runtime/engine` — engine unit + IT tests pass.
- [ ] Synchronous guards (jwt, azure-ad, identity, etc.) continue to work via the default async-overload bridge.
- [ ] A pre-authorization-capable guard (e.g. an OAuth-style guard in a downstream module) can be exercised declaratively via the new `callback` / `callbackParams` authorization fields and asserted via `expectIdentity` / `expectCredentials`. The guard's `via` reference is resolved through `GuardConfig.resolveId`.

### DIFF
--- a/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
+++ b/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
@@ -15,9 +15,13 @@
  */
 package io.aklivity.zilla.runtime.engine.test.internal.binding;
 
+import static io.aklivity.zilla.runtime.engine.guard.GuardHandler.NEEDS_PREAUTHORIZE;
+import static io.aklivity.zilla.runtime.engine.guard.GuardHandler.NOT_AUTHORIZED;
 import static io.aklivity.zilla.runtime.engine.test.internal.binding.config.TestBindingOptionsConfigAdapter.DEFAULT_ASSERTION_SCHEMA;
 import static java.util.Collections.emptyList;
 
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.security.KeyStore;
 import java.util.LinkedList;
 import java.util.List;
@@ -38,6 +42,7 @@ import io.aklivity.zilla.runtime.engine.config.BindingConfig;
 import io.aklivity.zilla.runtime.engine.config.CatalogedConfig;
 import io.aklivity.zilla.runtime.engine.config.SchemaConfig;
 import io.aklivity.zilla.runtime.engine.guard.GuardHandler;
+import io.aklivity.zilla.runtime.engine.guard.GuardHandler.LongCompletionCallback;
 import io.aklivity.zilla.runtime.engine.metrics.Metric;
 import io.aklivity.zilla.runtime.engine.model.ConverterHandler;
 import io.aklivity.zilla.runtime.engine.model.function.ValueConsumer;
@@ -101,6 +106,10 @@ final class TestBindingFactory implements BindingHandler
     private List<CatalogAssertion> catalogAssertions;
     private GuardHandler guard;
     private String credentials;
+    private String callbackUri;
+    private Map<String, String> callbackParams;
+    private String expectIdentity;
+    private String expectCredentials;
     private Map<String, String> attributes;
     private List<Event> events;
     private int eventIndex;
@@ -156,6 +165,10 @@ final class TestBindingFactory implements BindingHandler
                 int guardId = context.supplyTypeId(options.authorization.name);
                 this.guard = context.supplyGuard(NamespacedId.id(namespaceId, guardId));
                 this.credentials = options.authorization.credentials;
+                this.callbackUri = options.authorization.callback;
+                this.callbackParams = options.authorization.callbackParams;
+                this.expectIdentity = options.authorization.expectIdentity;
+                this.expectCredentials = options.authorization.expectCredentials;
                 this.attributes = options.authorization.attributes;
             }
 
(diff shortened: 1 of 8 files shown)


runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -219,22 +230,149 @@ public MessageConsumer newStream(
...
+        String code)
+    {
+        StringBuilder builder = new StringBuilder(callbackUri);
+        builder.append(callbackUri.indexOf('?') >= 0 ? '&' : '?');
+        builder.append("state=").append(java.net.URLEncoder.encode(state, java.nio.charset.StandardCharsets.UTF_8));
+        if (code != null)
+        {
+            builder.append("&code=").append(java.net.URLEncoder.encode(code, java.nio.charset.StandardCharsets.UTF_8));
+        }
+        return builder.toString();
```
jfallows, line 346: These params are oauth-specific, it should just be generic.

Preauthorize url will return something like https://auth.example.com/path?state=xyz, and callback URL is known to the oauth guard via config, also to the test binding via config, so this can just transfer the params from the preauthorize url onto the callback url verbatim (no need for "state" param awareness) and then also append any additional query params (specified in test binding) to work correctly with oauth guard, i.e. "code=abc".

This prevents the test binding from becoming oauth-specific but configured to work correctly for oauth-specific tests, and could reasonably be configured to support any guard with this approach, agree?

jfallows (changes requested):

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -219,22 +232,145 @@ public MessageConsumer newStream(
...
+            if (route != null)
+            {
+                newStream = new TestSource(source, originId, routedId, initialId, replyId, route.id)::onMessage;
+            }
+        }
+        else
+        {
+            // Always go through the async overload — sync guards run inline via the default
+            // {@link GuardHandler#reauthorize(long,long,long,String,LongCompletionCallback)}
+            // bridge; only OAuth-style guards actually defer the callback.
```
jfallows: Remove this comment.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -219,22 +232,145 @@ public MessageConsumer newStream(
...
-            final long resolvedId = route.id;
-            newStream = new TestSource(source, originId, routedId, initialId, replyId, resolvedId)::onMessage;
+            onAuthorized(traceId, routedId, deferred, binding, NOT_AUTHORIZED);
+            return;
         }
+        // Per the GuardHandler.preauthorize contract, the redirect-callback URL is passed
+        // back to reauthorize as the credentials. Forward whatever query params the guard
+        // embedded in the preauthorize URL verbatim, then append any additional params the
+        // test configured. The test binding stays guard-agnostic — it doesn't interpret any
+        // specific param name.
```
jfallows: Remove this comment.

jfallows (changes requested):

jfallows (approved):

Issue #1744, opened 2026-05-02
`binding-http`: per-route affinity extraction


jfallows:
## Overview

Add a generic mechanism for the `binding-http` `server` kind to extract a routing key from an inbound request (header or query parameter, with optional regex extraction) and use it to dispatch the resulting application stream to a consistent worker thread within the gateway instance.

## Motivation

Today the http-server binding propagates whatever affinity flows up from the network layer (TCP/TLS connection-level). Multiple HTTP requests for the same logical session can land on different worker threads within an instance, breaking per-worker session state assumptions held by upstream bindings.

A fresh HTTP request for an existing session-affined flow may not carry the original session header, but does carry an opaque token (e.g., a `state` query parameter) from which the session identifier can be derived. Routing it correctly requires extracting the key from the request content rather than connection metadata.

## Configuration schema

Per-route `with: affinity:` block:

```yaml
bindings:
  http_server0:
    type: http
    kind: server
    routes:
      - when:
          - path: /example/auth/callback
        with:
          affinity:
            query: state
            match: &#34;[^.]+&#34;
        exit: example_server0
      - when:
          - path: /example
          - path: /example/*
        with:
          affinity:
            header: x-session-id
        exit: example_server0
```

### `with: affinity:`

- Exactly one source: `header: ` xor `query: `.
- `match:` (optional) — regex applied to the source value. If the regex has capture groups, group 1 is used; otherwise the full match. If the regex doesn&#39;t match, the rule yields no key and the route falls through.
- `match:` omitted — the full source value is used as the key.
- Source value absent (header / query parameter not present) — rule yields no key, route falls through.
- Route without `with: affinity:` — no extraction, network-level affinity propagates as today.

## Behavior

For each inbound request, after route matching:

1. Resolve `with: affinity:` for the matched route; extract the key string per the rules above. If no key is extracted, fall through to existing behavior.
2. Hash the key to a `long` and emit it as the affinity on the outgoing application BEGIN. The engine routes the application stream to the matching worker.

## Implementation notes

- `HttpServerFactory` already builds the outgoing application BEGIN with the inbound network affinity. The new logic compiles the per-route `affinity` rules at config load and applies them when the route matches, overriding the affinity field on the outgoing BEGIN with the hashed extracted key.

## Spec coverage

New scenarios under `specs/binding-http.spec/src/main/scripts/.../streams/network/`:

- `request.with.header.affinity` — header-based extraction, no regex.
[...]

PR #1745, opened 2026-05-02
feat(engine, binding-http): per-route affinity extraction with worker pinning (#1744)

jfallows:
## Description

Closes (partial) #1744 — adds the **affinity primitive** to `binding-http`: extract a routing key from a request header or query parameter, hash it, and pin the application stream to a deterministic worker on this Zilla instance via a new engine `supplyInitialId(bindingId, hash)` overload.

This gives any upstream binding (e.g. `binding-mcp`, future `binding-oauth`) **per-instance worker affinity** for session-scoped state, eliminating one source of state-loss within a single node. Same session-id always lands on the same worker on the same instance.

The PR also lands the **Store SPI strict-async contract** as preparatory hardening — formalised Javadoc, test-binding contract assertions, and signaler-deferred `MemoryStore`/`TestStore` callbacks. No production code in this PR consumes the store; the hardening clarifies the SPI before any consumer arrives.

### Engine

- `EngineContext.supplyInitialId(bindingId, hash)` overload + `EngineWorker` impl using `Math.floorMod(hash, mask.cardinality())`. No-op overrides on `TlsWorker`/`EchoWorker` test stubs.
- k3po zilla transport adds accept-side `OPTION_AFFINITY` assertion + `NO_AFFINITY = 0L` sentinel — required by spec authoring for asserting affinity on inbound application BEGINs.
- `Store` / `StoreHandler` / `StoreContext` Javadoc formalises the strict-async contract: callbacks fire **strictly later** than the call returns, on the caller's I/O thread; state-sharing scope is implementation-defined.
- `TestBindingFactory` extends its store-assertion loop and the attach-time init `putIfAbsent` to verify on every callback that (a) it fires on the dispatch thread, (b) it has not already fired (single-fire), and (c) it did not fire synchronously. A violation trips `doInitialReset` so any non-conforming store fails the spec scripts.
- `TestStoreHandler` defers each completion callback via the engine `Signaler` to satisfy the contract.

### HTTP binding

- `with: affinity:` per-route clause: extracts the key from a request header or query parameter, with optional regex extraction.
- Config: `HttpAffinityConfig`, `HttpAffinityConfigBuilder`, `HttpAffinitySource`; extension on `HttpWithConfig`; adapter changes; `HttpAffinityResolver`; `HttpRouteConfig.affinity()` and `route.supplyInitialId(hash)`.
- Hot-path local-dispatch wiring at the resolve site in `HttpServer` (HTTP/1.1) and `Http2Server` (HTTP/2). When a route resolves and an `affinity` clause is present, the resolver computes the hash and the outbound application stream id is allocated via `route.supplyInitialId(hash)`.

### Store-memory

- `MemoryStoreHandler` defers each completion callback via the engine `Signaler` (same shape as `TestStoreHandler`), threaded through `MemoryStoreContext.attach`.
- Existing isolated unit test replaced by `MemoryStoreHandlerIT`, which drives `MemoryStore` through `TestBinding` so the strict-async contract assertions cover it end-to-end.
- New `engine.spec/server-memory-store.yaml` wires the test binding to a memory store for the IT.

### Specs + ITs

- 32 peer-to-peer affinity `.rpt` scripts across `rfc7230` and `rfc7540`, application and network sides, header- and query-extracted keys.
- `server.affinity.yaml` config under both `v1.1` and `v2`, with `SchemaTest` validation entries.
- 16 `AffinityIT` methods exercising the primitive end-to-end through the live engine.

## Verification

```bash
./mvnw install -pl runtime/engine,runtime/binding-tls,runtime/binding-echo,runtime/binding-http,runtime/store-memory,specs/binding-http.spec -am
./mvnw verify -pl specs/binding-http.spec -Dit.test='AffinityIT'
./mvnw verify -pl runtime/binding-http -Dit.test='*AffinityIT'
./mvnw verify -pl runtime/store-memory -Dit.test='MemoryStoreHandlerIT'
./mvnw test -pl runtime/binding-http,runtime/engine
```

Acceptance bar (all green locally):

- 32 spec `AffinityIT`s pass peer-to-peer (rfc7230 + rfc7540, app + net, header + query)
- 16 binding `AffinityIT`s pass through the live engine
- `MemoryStoreHandlerIT` passes (strict-async contract verified end-to-end)
- `EngineIT` passes (`TestStore` + `TestBinding` contract assertions exercised)
- `HttpOptionsConfigAdapterTest`, `HttpConfigurationTest`, `HttpWithConfigAdapterTest` pass
- `EngineConfigurationTest` passes
- `SchemaTest` validates `server.affinity.yaml` (v1.1 + v2)
- `mvn checkstyle:check license:check` clean across all touched modules

## Test plan

- [ ] CI green on the branch (Build + smoke tests)
- [ ] Reviewer confirms the affinity primitive shape — `with: affinity:` clause, header xor query, optional regex, `Math.floorMod(hash, mask)` pinning
- [ ] Reviewer confirms the Store SPI strict-async contract Javadoc + test-binding assertions are appropriately scoped (no production consumer in this PR)
- [ ] Reviewer confirms `binding-http` carries no cluster-aware code paths


runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/k3po/ext/behavior/ZillaStreamFactory.java
```diff
@@ -177,6 +176,20 @@ private void onBegin(
             final long acknowledge = begin.acknowledge();
             final OctetsFW beginExt = begin.extension();
 
+            final ZillaChannelConfig config = channel.getConfig();
+            if (channel instanceof ZillaChildChannel && config.hasAffinity())
```
jfallows: Let's treat `0L` as `NO_AFFINITY`, and default affinity to that, so this hasAffinity() is not needed.

jfallows: Done in 6fdca0345 — added `NO_AFFINITY = 0L` constant on `ZillaChannelConfig`, removed `hasAffinity()` and the `affinityExplicit` field, `onBegin` now compares `config.getAffinity() != NO_AFFINITY`.

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
```diff
@@ -163,6 +163,25 @@
...
+                                "description": "Reference to a store binding. Presence enables cross-instance affinity lookups.",
+                                "type": "string"
+                            },
+                            "alt-svc-max-age":
+                            {
+                                "title": "Alt-Svc Max Age (seconds)",
+                                "type": "integer",
+                                "default": 86400,
+                                "minimum": 0
+                            },
```
jfallows: Let's move this to an HttpConfiguration property instead, still defaulting to 86400.

`HTTP_ALT_SVC_MAX_AGE`

jfallows: Done in ebeef071e — schema entry removed; `HTTP_ALT_SVC_MAX_AGE` (default 86400) will land on `HttpConfiguration` with the binding implementation in Phase 3.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineContext.java
```diff
@@ -417,6 +433,26 @@ GuardHandler supplyGuard(
...
 
+    /**
+     * Returns the externally-reachable hostname for this engine instance, configured via
+     * {@code zilla.engine.service.hostname}. Used by bindings (e.g. http) to compose a
+     * per-instance authority for cross-instance redirect responses. The port portion of
+     * the authority is composed from each network connection's destination port at use site.
+     *
+     * @return the configured service hostname, or {@code null} if unset
+     */
+    String serviceHostname();
```
jfallows: This method is not needed, can be obtained via EngineConfiguration.ENGINE_SERVICE_HOSTNAME property instead.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/EngineContext.java
```diff
@@ -417,6 +433,26 @@ GuardHandler supplyGuard(
...
+
+    /**
+     * Returns the engine-level shared store handle, resolved from the qualified name
+     * configured via {@code zilla.engine.store.name} (default {@code sys:state}). When
+     * {@code zilla.engine.store.type} is unset and the user has not declared a store
+     * binding under the configured name, this returns {@code null}.
+     *
+     * @return the engine store handle, or {@code null} if no store is configured
+     */
```
jfallows:
```suggestion
    /**
     * Returns the engine-level shared store handle.
     *
     * @return the engine store handle, or {@code null} if no store is configured
     */
```

runtime/store-memory/src/main/java/io/aklivity/zilla/runtime/store/memory/internal/MemoryStore.java
```diff
@@ -54,6 +57,36 @@ public URL type()
         return getClass().getResource("schema/memory.schema.patch.json");
     }
 
+    @Override
+    public Optional<NamespaceConfig> contribute(
+        EngineConfiguration config)
+    {
```
jfallows: Remove this, use this approach instead, see StdoutExporter as an example.
```java
@Override
public URL system()
{
   ...
}
```

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/schema/http.schema.patch.json
```diff
@@ -346,7 +345,11 @@
                                 }
                             }
                         },
-                        "additionalProperties": false
+                        "additionalProperties": false,
+                        "dependentRequired":
+                        {
+                            "store": [ "self" ]
+                        }
```
jfallows: We don't need this now, agree?

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/streams/application/rfc7230/affinity/request.with.header.affinity.match/client.rpt
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+# header value is "prefix.sess-1.suffix"; the binding's affinity rule
+# uses match: "[^.]+\.([^.]+)\.[^.]+" which captures group 1 = "sess-1"
+# affinity = "sess-1".hashCode() & 0xffff_ffffL = 0xca029156
```
jfallows: Remove these kinds of comments from scripts.

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/streams/application/rfc7230/affinity/request.with.header.affinity.match/server.rpt
```diff
@@ -0,0 +1,45 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+# header value is "prefix.sess-1.suffix"; the binding's affinity rule
+# uses match: "[^.]+\.([^.]+)\.[^.]+" which captures group 1 = "sess-1"
+# affinity = "sess-1".hashCode() & 0xffff_ffffL = 0xca029156
+
```
jfallows: Remove

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/streams/application/rfc7230/affinity/request.with.header.affinity.missing/client.rpt
```diff
@@ -0,0 +1,41 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+# affinity source (session-id header) absent from request, so the
+# route's affinity rule yields no key and the application BEGIN carries
+# the network-level affinity (here defaulted to 0L)
+
```
jfallows: Remove, see all similar comments and remove them too.

jfallows:
PR feedback addressed in 650ea09bc:
- Dropped `EngineContext.serviceHostname()` — bindings now read `ENGINE_SERVICE_HOSTNAME` directly from `EngineConfiguration`.
- Simplified `EngineContext.store()` Javadoc per the suggestion.
- Replaced `Store.contribute(EngineConfiguration)` SPI with `URL system()` matching the `StdoutExporter` pattern. `MemoryStore` returns the system patch URL when `ENGINE_STORE_TYPE` equals `memory`, and the engine wires `Store::system` URLs alongside `Exporter::system` ones into `buildSystemNamespaceIfNecessary()`.
- Dropped leftover `dependentRequired: { store: [self] }` from the http schema patch.
- Stripped explanatory comments from the affinity spec scripts.
- Adjusted PR title to reflect the broader scope of the work landed here.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -7208,6 +7328,83 @@ private URI createTargetURI(
...
+        migrateBuffer.putByte(altSvcOffset, (byte) '"');
+        altSvcOffset += 1;
+        final int altSvcLength = altSvcOffset;
+
+        return headersRW
+            .wrap(extBuffer, 0, extBuffer.capacity())
+            .item(h -> h.name(HEADER_STATUS).value(STATUS_429))
+            .item(h -> h.name(HEADER_CONTENT_LENGTH).value("0"))
+            .item(h -> h.name(HEADER_RETRY_AFTER).value(VALUE_RETRY_AFTER_NOW))
+            .item(h -> h.name(HEADER_ALT_SVC).value(migrateBuffer, 0, altSvcLength))
```
jfallows: Use the `.value(Consumer<String16FW.Builder> mutator)` overload to avoid the need for a separate migrateBuffer, then move to callsite for readability.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -645,6 +683,7 @@ public MessageConsumer newStream(
 
             HttpVersion version = null;
             boolean secure = false;
+            String serviceHost = null;
```
jfallows:
```suggestion
            String service = null;
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -1630,6 +1706,7 @@ private final class HttpServer
         private final long replyId;
         private final long affinity;
         private final boolean upgrade;
+        private final String serviceHost;
```
jfallows:
```suggestion
        private final String service;
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -1669,7 +1746,8 @@ private HttpServer(
             long initialId,
             long affinity,
             boolean secure,
-            boolean upgrade)
+            boolean upgrade,
+            String serviceHost)
```
jfallows:
```suggestion
            String service)
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -588,6 +620,12 @@ public HttpServerFactory(
         this.encodeMax = bufferPool.slotCapacity();
         this.bindings = new Long2ObjectHashMap<>();
         this.event = new HttpEventContext(context);
+        this.serviceHostname = ENGINE_SERVICE_HOSTNAME.get(config);
```
jfallows:
```suggestion
        this.serviceHost = ENGINE_SERVICE_HOSTNAME.get(config);
```

specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/config/server-memory-store.yaml
jfallows, line 1: This cannot live here, memory store depends on engine.

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/config/v1.1/server.affinity.migrate.yaml
jfallows, line 1: This is no longer needed, agree?

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/config/v2/server.affinity.migrate.yaml
jfallows, line 1: This is no longer needed, agree?

runtime/binding-http/src/test/java/io/aklivity/zilla/runtime/binding/http/internal/streams/rfc7230/server/AffinityMigrateIT.java
jfallows, line 1: This is no longer needed, agree?

jfallows (commented):
LGTM

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/store/StoreHandler.java
```diff
@@ -24,11 +24,22 @@
...
+ * local backend this means deferring via the engine signaler to the next event-loop tick of
+ * the same worker; for a networked backend it means hopping the network-completion event
+ * back onto the calling worker's signaler before invoking the callback. In either case the
+ * caller observes "callback runs on my thread, later".
  * </p>
  * <p>
  * Completion callbacks are pre-allocated per-stream and passed into each operation to avoid
- * heap allocation on the data path.
+ * heap allocation on the data path. The result of an operation is only valid inside the
+ * callback.
```
jfallows: Remove this implementation detail, caller may be something other than a binding, so referring to stream is too much detail.

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -188,7 +188,21 @@ public void attach(
                     options.storeAssertions.get(0).assertions : null;
                 if (this.store != null && this.storeAssertions == null)
                 {
-                    this.store.putIfAbsent("init", "", Long.MAX_VALUE, v -> {});
+                    final Thread dispatchThread = Thread.currentThread();
+                    final boolean[] callbackFired = { false };
```
jfallows: Use MutableBoolean instead of boolean[1];

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/store/TestStoreHandler.java
```diff
@@ -69,14 +80,30 @@ public void delete(
...
+        // contract: callback fires strictly later than the call, on the caller's I/O thread
+        if (signaler != null)
+        {
+            signaler.signalAt(System.currentTimeMillis(), 0, ignored -> task.run());
+        }
+        else
+        {
+            // unit-test fallback (no engine context); run immediately
+            task.run();
+        }
```
jfallows: We only test from ITs for this now, so unit test fallback is no longer required, instead signaler should always be non-null and can be verified in the constructor via `Objects.requireNonNull(...)`, agree?

runtime/store-memory/pom.xml
```diff
@@ -41,6 +41,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>engine.spec</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
```
jfallows: This should not be required directly from store-memory, instead it should be a transitive dependency from store-memory.spec project.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -1133,9 +1138,27 @@ else if (!isCorsRequestAllowed(server.binding, headers))
...
+                                final int hash = affinityKey.hashCode();
+                                requestId = route.supplyInitialId(hash);
+                                requestAffinity = hash & 0xffff_ffffL;
+                            }
+                            else
+                            {
+                                requestId = supplyInitialId.applyAsLong(route.id);
+                                requestAffinity = server.affinity;
+                            }
+
```
jfallows: Instead of the complex if-else logic, how about we move this to the `HttpRouteConfig.resolveAffinity(headers::get)` but return a record `HttpRouteAffinity` with `initialId` and `affinity` fields. Then for the case where there is no internal resolver, we would use the regular supplyInitialId and return an affinity of 0.

jfallows: Also we need to use a hash function that provides more predictable distribution, not just based on internal JVM String hash code.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -567,6 +571,7 @@ public HttpServerFactory(
         this.streamFactory = context.streamFactory();
         this.supplyDebitor = context::supplyDebitor;
         this.supplyInitialId = context::supplyInitialId;
+        this.supplyInitialIdByHash = (routedId, hash) -> context.supplyInitialId(routedId, (int) hash);
```
jfallows: Would the mechanics be easier if we changed the type of hash to long in EngineContext.supplyInitialId ?

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpBindingConfig.java
```diff
@@ -78,13 +80,28 @@ public final class HttpBindingConfig
...
         this.options = HttpOptionsConfig.class.cast(binding.options);
         this.routes = binding.routes.stream().map(route ->
-            new HttpRouteConfig(route, options != null ? options.overrides : null)).collect(toList());
+            new HttpRouteConfig(route, options != null ? options.overrides : null,
+                supplyInitialId != null
+                    ? () -> supplyInitialId.applyAsLong(route.id)
+                    : null,
+                supplyInitialIdByHash != null
+                    ? hash -> supplyInitialIdByHash.applyAsLong(route.id, hash)
+                    : null)).collect(toList());
```
jfallows: There is a lot of conditional logic here that could be moved to the HttpRouteConfig constructor if we pass context as the first parameter.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpAffinityResolver.java
```diff
@@ -0,0 +1,93 @@
...
+            break;
+        case QUERY:
+            if (path != null)
+            {
+                queryMatcher.reset(path);
+                if (queryMatcher.find())
+                {
+                    value = queryMatcher.group(1);
+                }
+            }
```
jfallows: Do we need to pass `path` separately? Isn't it just the `:path` header value?

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpAffinityResolver.java
```diff
@@ -0,0 +1,93 @@
...
+            return value;
+        }
+
+        matcher.reset(value);
+        if (!matcher.find())
+        {
+            return null;
+        }
+
+        return matcher.groupCount() > 0 ? matcher.group(1) : matcher.group(0);
```
jfallows:
```suggestion

        String key = null;

        if (value != null &&
            matcher != null &&
            matcher.reset(value).find())
        {
            value = matcher.groupCount() > 0
                ? matcher.group(1)
                : matcher.group(0);
        }

        return value;
```
Prefer single return.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpAffinityResolver.java
```diff
@@ -0,0 +1,93 @@
...
+            // matches `?<name>=<value>` or `&<name>=<value>` in :path; group 1 is the value
+            this.queryPattern = Pattern.compile("[?&]" + Pattern.quote(name) + "=([^&]*)");
+            this.queryMatcher = queryPattern.matcher("");
+        }
+        else
+        {
+            this.queryPattern = null;
+            this.queryMatcher = null;
+        }
+    }
```
jfallows, line 53: Instead of trying to handle all combinations in a single implementation, perhaps make this class abstract, change the constructor to a static .of method and have different targeted private inner class implementations that handle just QUERY or just HEADER. Then we can also have a NONE static final constant that is used by default, avoiding null checks at the call site but getting the correct behavior.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -1133,9 +1133,11 @@ else if (!isCorsRequestAllowed(server.binding, headers))
                             HttpPolicyConfig policy = binding.access().effectivePolicy(headers);
                             final String origin = policy == CROSS_ORIGIN ? headers.get(HEADER_NAME_ORIGIN) : null;
 
+                            final HttpRouteAffinity routeAffinity = route.resolveAffinity(headers::get);
+
```
jfallows:
```suggestion
                            final HttpRouteAffinity resolved = route.resolve(headers::get);

```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -5130,8 +5139,11 @@ else if (!isCorsRequestAllowed(binding, headers))
 
                             HttpRequestType requestType = binding.resolveRequestType(beginEx);
 
-                            final Http2Exchange exchange = new Http2Exchange(originId, routedId, NO_REQUEST_ID, streamId,
-                                exchangeAuth, traceId, policy, origin, contentLength, requestType);
+                            final HttpRouteAffinity routeAffinity = route.resolveAffinity(headers::get);
+
```
jfallows:
```suggestion
                            final HttpRouteAffinity resolved = route.resolve(headers::get);

```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRouteConfig.java
```diff
@@ -68,6 +74,33 @@ public long compositeId()
         return with != null ? with.compositeId() : NO_COMPOSITE_ID;
     }
 
+    public HttpAffinityResolver affinity()
+    {
+        return with != null ? with.affinity() : HttpAffinityResolver.NONE;
+    }
```
jfallows: If we no longer need this public method, then inline it into resolveAffinity instead.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpAffinityResolver.java
```diff
@@ -0,0 +1,125 @@
...
+
+        @Override
+        public String resolveKey(
+            Function<String, String> headerByName)
+        {
+            return applyMatcher(headerByName.apply(name), matcher);
+        }
+    }
+
+    private static final class QueryAffinityResolver extends HttpAffinityResolver
```
jfallows:
```suggestion
    private static final class ByQueryParameter extends HttpAffinityResolver
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpAffinityResolver.java
```diff
@@ -0,0 +1,125 @@
...
+        {
+            key = matcher.reset(value).find()
+                ? matcher.groupCount() > 0 ? matcher.group(1) : matcher.group(0)
+                : null;
+        }
+
+        return key;
+    }
+
+    private static final class HeaderAffinityResolver extends HttpAffinityResolver
```
jfallows:
```suggestion
    private static final class ByHeader extends HttpAffinityResolver
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
```diff
@@ -408,7 +408,7 @@ public int originTypeId()
     public void attach(
         BindingConfig binding)
     {
-        HttpBindingConfig httpBinding = new HttpBindingConfig(binding, supplyValidator);
+        HttpBindingConfig httpBinding = new HttpBindingConfig(context, binding, supplyValidator);
```
jfallows: We can omit `supplyValidator` parameter now that we are passing `context` and assign it in the constructor using `context::supplyValidator` instead.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRouteConfig.java
```diff
@@ -24,32 +25,39 @@
...
         RouteConfig route,
         Map<String8FW, String16FW> overrides)
     {
+        this.context = context;
         this.id = route.id;
         this.with = Optional.ofNullable(route.with)
             .map(HttpWithConfig.class::cast)
             .map(HttpWithResolver::new)
             .orElse(null);
+        this.affinity = with != null ? with.affinity() : HttpAffinityResolver.NONE;
```
jfallows:
```suggestion
        this.affinity = Optional.ofNullable(with)
            .map(HttpWithConfig::affinity)
            .orElse(HttpAffinityResolver.NONE);
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRouteConfig.java
```diff
@@ -68,6 +76,28 @@ public long compositeId()
...
+        }
+        else
+        {
+            final CRC32C crc = new CRC32C();
+            crc.update(key.getBytes(UTF_8));
+            final int hash = (int) crc.getValue();
+            initialId = context.supplyInitialId(id, hash);
+            affinity = hash & 0xffff_ffffL;
+        }
+        return new HttpRouteAffinity(initialId, affinity);
```
jfallows:
```suggestion
        final HttpRouteAffinity resolved;
        if (key == null)
        {
            final long initialId = context.supplyInitialId(id);
            resolved = new HttpRouteAffinity(initialId, 0L);
        }
        else
        {
            final CRC32C crc = new CRC32C();
            crc.update(key.getBytes(UTF_8));
            final int hash = (int) crc.getValue();
            final long initialId = context.supplyInitialId(id, hash);
            final long affinity = hash & 0xffff_ffffL;
            resolved = new HttpRouteAffinity(initialId, affinity);
        }

        return resolved;
```

jfallows: Move the CRC32C instance to a field, and use .reset() to reuse.

jfallows: Would still prefer to avoid allocation of byte array here if possible. Might require a different hashing approach.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRouteConfig.java
```diff
@@ -24,32 +24,43 @@
...
 
 public final class HttpRouteConfig
 {
     public final long id;
 
+    private final EngineContext context;
     private final List<HttpConditionMatcher> when;
     private final HttpWithResolver with;
+    private final HttpAffinityResolver affinity;
+    private final CRC32C affinityHasher;
```
jfallows:
```suggestion
    private final CRC32C crc32c;
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRouteConfig.java
```diff
@@ -68,6 +79,34 @@ public long compositeId()
...
+            affinityHasher.reset();
+            for (int i = 0; i < key.length(); i++)
+            {
+                final char c = key.charAt(i);
+                affinityHasher.update(c & 0xff);
+                affinityHasher.update((c >>> 8) & 0xff);
+            }
+            final int hash = (int) affinityHasher.getValue();
+            final long initialId = context.supplyInitialId(id, hash);
+            final long affinity = hash & 0xffff_ffffL;
```
jfallows: Let's drop the CRC32C completely and use .hashCode() instead.

```suggestion
            final long affinity = key.hashCode();
```

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/config/HttpRouteConfig.java
```diff
@@ -68,6 +79,34 @@ public long compositeId()
...
+            resolved = new HttpRouteAffinity(initialId, affinity);
+        }
+
+        return resolved;
```
jfallows: With that simplification in place we can reduce this to the following:
```
return key != null
    ? new HttpRouteAffinity(
        context.supplyInitialId(id, key.hashCode())
        key.hashCode())
    : new HttpRouteAffinity(
        context.supplyInitialId(id),
        NO_AFFINITY);
```

jfallows (commented):
LGTM

PR #1752, opened 2026-05-07
feat(binding-mcp): MCP elicitation across server, proxy, and client kinds (#1739)

### LINKED ISSUE #1739: `binding-mcp` support elicitation
# MCP Elicitation Support
 
## Overview
 
Add support for the MCP elicitation protocol to enable MCP clients to authorize access to backend services requiring OAuth credentials. This covers the `mcp` binding (server, proxy, client) and `http` server binding changes needed to support the elicitation flow end-to-end.
 
When an MCP client binding requires credentials for a backend MCP server and none are available, it triggers an elicitation flow — sending an `elicitation/create` notification upstream to the remote MCP client with a URL for the user to authorize access. Once the user completes authorization in their browser, the OAuth callback is received by the MCP server binding, routed back to the correct MCP client binding, and the original suspended request is resumed.
 
## Pipeline Shape
 
```
[Remote MCP Client]
       ↕  SSE (mcp-session-id: abc123xyz)
[HTTP Server Binding]
       ↕
[MCP Server Binding]
       ↕
[MCP Proxy Binding]
       ↕
[MCP Client Binding]     ← triggers elicitation when credentials unavailable
       ↕
[Target MCP Server]
```
 
### Phase 1 — Tool Call Arrives, Elicitation Triggered
 
The remote MCP client makes a tool call. The MCP client binding determines credentials are unavailable and triggers elicitation:
 
```
Remote MCP Client
  POST /mcp  HTTP/1.1
  mcp-session-id: abc123xyz
  { "method": "tools/call", "params": { "name": "github__search_repos", ... } }
  ▼
HTTP Server Binding
  routes to correct worker thread via mcp-session-id header
  ▼
MCP Server Binding
  forwards tools/call downstream
  ▼
MCP Proxy Binding
  routes "github__search_repos" → GitHub MCP client binding
  ▼
MCP Client Binding
  credentials unavailable → NEEDS_PREAUTHORIZE
  suspends tools/call stream
  obtains authorization URL from guard
  extracts state param from URL: "7f3a9b1c"
  stores thread-local: stateStore["7f3a9b1c"] → streamId
  sends elicitation/create upstream
  ▼
MCP Proxy Binding
  intercepts elicitation/create flowing upstream
  prepends toolkit prefix to state: "github__7f3a9b1c"
  forwards upstream
  ▼
MCP Server Binding
  intercepts elicitation/create flowing upstream
  prepends mcp-session-id to state: "abc123xyz.github__7f3a9b1c"
  replaces CALLBACK_PLACEHOLDER with real callback URI in URL
  forwards upstream
  ▼
Remote MCP Client
  receives elicitation/create { mode:url, elicitationId, url }
  responds: { "action": "accept" }
  user clicks URL in browser
```
 
### Phase 2 — Out of Band: Browser OAuth Consent
 
```
User's browser
  → authorization URL with state=abc123xyz.github__7f3a9b1c
  → user consents
  → provider 302: GET /mcp/auth/callback?code=xyz&state=abc123xyz.github__7f3a9b1c
  → browser follows redirect → hits Zilla
```
 
### Phase 3 — Callback Received, Routed, Stream Resumed
 
```
GET /mcp/auth/callback?code=xyz&state=abc123xyz.github__7f3a9b1c
  ▼
HTTP Server Binding
  extracts [^\.]+ from state → "abc123xyz"
  consistent hash → correct engine worker
  ▼
MCP Server Binding
  recognizes elicitation.callback path suffix
  strips sessionId prefix from state → "github__7f3a9b1c"
  derives real callback URI from Host header + config path
  emits internal: elicitation/callback { url: "https://zilla.corp/mcp/auth/callback?code=xyz&state=github__7f3a9b1c" }
  returns 200 "Authorization complete" to browser
  ▼
MCP Proxy Binding
  extracts toolkit "github" from before "__"
  strips toolkit prefix → state: "7f3a9b1c"
  routes to GitHub MCP client binding
  strips toolkit prefix from state param in URL
  forwards: elicitation/callback { url: "https://zilla.corp/mcp/auth/callback?code=xyz&state=7f3a9b1c" }
  ▼
MCP Client Binding
  looks up stateStore["7f3a9b1c"] → streamId
  cleans up stateStore entry
  passes url directly to guard: handler.reauthorize(traceId, bindingId, 0L, url)
  on success: SIGNALs streamId to resume
  sends elicitation/complete upstream
  ▼
Remote MCP Client
  receives elicitation/complete
  retries tools/call
```
 
### Phase 4 — Tool Call Resumed, Completes
 
```
MCP Client Binding
  receives SIGNAL on suspended streamId
  calls handler.reauthorize() → HIT, positive sessionId
  calls handler.credentials(sessionId) → access token
  injects Authorization: Bearer <token>
  forwards tools/call to target MCP server
  ▼
Target MCP Server
  validates token ✓
  returns tool result
  ▼
MCP Client Binding → MCP Proxy Binding → MCP Server Binding → Remote MCP Client ✓
```
 
## State Parameter Lifecycle
 
The `state` parameter is used to carry routing context through the OAuth redirect. Each binding layer adds its own prefix on the way up and strips it on the way back down:
 
```
Guard generates:            "7f3a9b1c"
MCP proxy prepends:         "github__7f3a9b1c"
MCP server prepends:        "abc123xyz.github__7f3a9b1c"   ← provider sees and returns this
HTTP server extracts:       "abc123xyz"                    → correct worker thread
MCP server strips:          "github__7f3a9b1c"             → downstream
MCP proxy strips:           "7f3a9b1c"                     → correct client binding
MCP client resolves:        stateStore["7f3a9b1c"]         → suspended streamId
```
 
The `CALLBACK_PLACEHOLDER` (`https://replace.me/callback`) is used by the MCP client binding when calling the guard's `preauthorize()`. The MCP server binding replaces it with the real derived callback URI before forwarding `elicitation/create` upstream. This avoids passing the callback URI through the pipeline on every lifecycle stream.
 
### HTTP Server Binding
 
- Support dual affinity key extraction for worker thread routing:
  - `mcp-session-id` header → normal MCP requests
  - `[^\.]+` regex on `state` query param → callback requests matching `elicitation.callback` path
- Apply extracted key for consistent worker thread routing

### MCP Server Binding
 
- **Config** — add `elicitation.callback` option with default `auth/callback`:
  ```yaml
  options:
    elicitation:
      callback: auth/callback    # default
  ```
 
- **Callback recognition** — on incoming request matching `elicitation.callback` path suffix:
  - Extract `code`, `state` query params
  - Strip `sessionId.` prefix from `state`
  - Derive real callback URI from scheme + Host header + mounted path + `elicitation.callback`
  - Emit internal `elicitation/callback { code, state, callback }` downstream
  - Return `200 "Authorization complete, you may close this tab"` to browser
- **elicitation/create upstream** — intercept outgoing `elicitation/create`:
  - Extract `state` param from authorization URL
  - Prepend `mcp-session-id + "."` to state
  - Replace `CALLBACK_PLACEHOLDER` with real derived callback URI
  - Forward modified URL upstream
- **elicitation/complete upstream** — forward as-is

### MCP Proxy Binding
 
- **elicitation/create upstream** — intercept outgoing `elicitation/create`:
  - Extract `state` param from authorization URL
  - Prepend toolkit prefix: `toolkitName + "__"`
  - Rewrite URL state param, forward upstream
- **elicitation/callback downstream** — intercept `elicitation/callback { code, state, callback }`:
  - Extract toolkit name from before `__` in state
  - Strip toolkit prefix from state
  - Route to correct downstream MCP client binding based on toolkit name
  - Forward with stripped state downstream
- **elicitation/complete upstream** — forward as-is

### MCP Client Binding
 
- **Shared constant** — define internally within mcp binding project:
  ```java
  static final String CALLBACK_PLACEHOLDER = "https://replace.me/callback";
  ```
 
- **Credential miss on tool call / initialize:**
  - On `NEEDS_PREAUTHORIZE` returned from guard:
    - Call `handler.preauthorize(traceId, bindingId, contextId, CALLBACK_PLACEHOLDER)`
    - Extract `state` param from returned authorization URL
    - Store thread-local: `stateStore[state] → streamId`
    - Suspend stream, do not forward to target MCP server
    - Send `elicitation/create { mode:url, elicitationId:contextId, url }` upstream
- **elicitation/callback handling:**
  - Receive `elicitation/callback { url }`
  - Look up thread-local: `stateStore[state] → streamId`
  - Clean up `stateStore` entry
  - Pass url directly: handler.reauthorize(traceId, bindingId, 0L, url)
  - On positive result: SIGNAL `streamId` to resume
- **Resumed stream:**
  - Guard now has credentials → positive `sessionId`
  - Inject `Authorization: Bearer <token>` on outbound call
  - Forward tool call to target MCP server, return result upstream
- **elicitation/complete:**
  - After SIGNAL sent, forward `elicitation/complete { elicitationId, status:completed }` upstream via SSE session
- **Initialize suspension:**
  - Same pattern as tool call — suspend initialize on `NEEDS_PREAUTHORIZE`, resume after SIGNAL

## Config Example
 
```yaml
bindings:
  http_server0:
    type: http
    kind: server
    routes:
      - when:
          - path: /mcp
          - path: /mcp/*
        exit: mcp_server0
 
  mcp_server0:
    type: mcp
    kind: server
    options:
      elicitation:
        callback: auth/callback
    exit: mcp_proxy0
 
  mcp_proxy0:
    type: mcp
    kind: proxy
    routes:
      - when:
          - toolkit: github
        exit: mcp_client_github0
 
  mcp_client_github0:
    type: mcp
    kind: client
    options:
      authorization:
        oauth_github0:
          credentials:
            headers:
              authorization: Bearer {credentials}
      exit: sys:http_client
```
 
## Acceptance Criteria
 
- Remote MCP client making a tool call with no stored credentials receives `elicitation/create` with a valid authorization URL
- State parameter is correctly assembled (`sessionId.toolkit__opaqueToken`) as it flows upstream through server and proxy bindings
- Guard callback routes to the correct Zilla instance (via load balancer) and correct worker thread (via HTTP server binding) using `sessionId` extracted from `state`
- Callback is recognized by MCP server binding, emits `elicitation/callback` downstream with full callback `url` including `state` and `code`
- MCP proxy binding correctly strips toolkit prefix from `state` parameter and routes to the correct MCP client binding
- MCP client binding resumes the suspended stream after SIGNAL and the tool call completes successfully
- `elicitation/complete` is sent upstream to the remote MCP client after successful token storage
- Initialize is suspended and resumed by the same mechanism as tool calls
- Multiple toolkit routes behind the same MCP proxy binding are correctly demultiplexed by toolkit prefix in state

jfallows:
## Summary

MCP elicitation support across all three `mcp` binding kinds (`server`, `proxy`, `client`). Lets an upstream tool implementation suspend a `tools/call` request to ask the human user to authorize via OAuth, then resume the request once the user completes (or declines, or the auth times out). On the `client` kind, the binding can also act as the OAuth initiator itself — pre-authorising the upstream `tools/call` via a configured guard before any request leaves the gateway.

Closes #1739.

### Protocol shape

The same app-side wire shape applies to all three binding kinds: every elicitation is `CHALLENGE mcp:elicitCreate(id, url)` from the upstream → `FLUSH mcp:elicitCallback(url)` from the binding once the human has navigated the OAuth flow → `FLUSH mcp:elicitComplete(COMPLETED | DECLINED | CANCELLED)` from the upstream, all carried on the same app-stream initial direction.

- **Unified contract:** `END` is never forwarded while an elicitation is in flight. An app stream that emits `END` mid-elicit is treated as a protocol error and the reply is `ABORT`ed. This invariant holds in `server`, `proxy`, and `client` kinds.
- **`server` kind:** the MCP SSE response stream that delivered the `tools/call` is upgraded to surface JSON-RPC elicitation events — `id: :1 + data: { method: "elicitation/create", params: { url, elicitationId } }`, then `…:2 + elicitation/complete`, then either `…:3 + result` (completed) or `…:3 + error` (declined / cancelled). The configurable auth-callback endpoint (`options.elicitation.callback`, default `auth/callback`) looks up the elicitation by session and emits the binding-side `mcp:elicitCallback` flush.
- **`proxy` kind:** initial-direction FLUSHes are relayed unchanged between server and client sides so a `mcp:elicitCallback` from a downstream server traverses the proxy to the upstream.
- **`client` kind:** two upstream-elicitation modes.
  - **Upstream-driven** — the upstream `mcp` server emits SSE `elicitation/create` / `elicitation/complete` events on the response stream; the client decodes them on the network side and surfaces `CHALLENGE mcp:elicitCreate` / `FLUSH mcp:elicitComplete` on the app side. To avoid the cycle "app holds initial END until elicit completes, but upstream needs the request body END to respond", the client also proactively ENDs the network HTTP request as soon as the JSON-RPC params body is complete (driven by a brace-depth tracker on the request body).
  - **Binding-driven** — when the configured guard returns `NEEDS_PREAUTHORIZE` on `tools/call`, the binding holds the upstream POST, emits `CHALLENGE mcp:elicitCreate(id="1", url=preauthorizeURL)` to the app, and defers the POST until the app replies with `FLUSH mcp:elicitCallback(callbackURL)`. On COMPLETED it opens the POST with `Authorization: Bearer <credentials>` and emits `FLUSH mcp:elicitComplete(COMPLETED)`; on DECLINED or timeout it emits `FLUSH mcp:elicitComplete(DECLINED | CANCELLED)`.

## What's in this branch

**`binding-mcp` runtime**

- New extension shapes (`mcp:elicitCreate`, `mcp:elicitCallback`, `mcp:elicitComplete`) on CHALLENGE / FLUSH frames in the binding IDL; `McpFunctions` builders/matchers for each, exercised by `McpFunctionsTest`.
- `McpElicitationConfig` under `options` (`callback` path, default `auth/callback`); new `binding.options.authorization` block on `client` kind, resolved against an engine guard.
- **Server kind** (`McpServerFactory`): callback path recognition on `GET`, per-session `elicitations` map keyed by sessionId-prefixed state token, `McpRequestStream.onAppChallenge` / `onAppFlush(elicitComplete)` to drive the SSE event encoder; JSON-RPC error event on declined/cancelled; `McpBindingConfig.resolveRedirectUri` synthesises the binding-side redirect URI from request headers (with a flyweight-reuse fix in `Array32FW.matchFirst`); deferred network END forwarding while in `elicitMode`.
- **Proxy kind** (`McpProxyFactory`): `doClientFlush` on the client side and an `onServerFlush → client.doClientFlush` relay so initial-direction FLUSHes (`elicitCallback`) traverse the proxy.
- **Client kind** (`McpClientFactory`):
  - `McpToolsCallStream` request state machine: tracks `pendingAuth`, `requestSent`, `bufferedBody`, elicit-timeout; gates `doEncodeRequestEnd` on `requestSent` to prevent encoder buffer leaks on the decline/timeout paths.
  - `onAppChallenge` overrides `doAppChallenge` to use `initialId` (client-kind throttle direction), distinct from the server-kind default.
  - JSON brace-depth body tracker drives proactive net END on JSON-RPC params completion.
  - SSE decoder hooks `elicitation/create` and `elicitation/complete` (with `elicitationId`, `url`, `mode`, `status` extracted by `JsonReader`) → `mcp.onDecodeElicitCreate` / `onDecodeElicitComplete`.
- **Shared** (`McpStream`): `proceedWithRequest`, `bufferAppData`, and `deferAppEnd` hooks made overridable so the three kinds can compose the elicit state machine without duplicating the BEGIN/DATA/END plumbing; null-net guards added so handlers don't NPE on streams torn down mid-elicit.

**`engine`**

- Generic engine concept: test guard mode `NEEDS_PREAUTHORIZE` via a `preauthorizeURL` option on the test guard (drives the client-kind binding-driven path in ITs).
- k3po test transport: drain the engine task queue (`channel.engine.drainTasks()`) before sending the eager `WINDOW` on accept-side child streams. This preserves the wire ordering trick scripts rely on (`read advise zilla:challenge` placed *before* `connected` requires the CHALLENGE frame to land on the streams buffer ahead of the eager WINDOW) without the regressions seen with the original "defer WINDOW via task" approach.
- Picks up #1760 (worker/boss `onStart` ordering NPE fix), #1755/#1758/#1759 (REDIRECT control frame, binding-http REDIRECT, command-dump REDIRECT support) via `develop` merge.

**`binding-mcp.spec` scripts and ITs**

- App-side and net-side `tools.call.elicit.{completed,declined,timeout}` scenarios for all three kinds:
  - `…/server` flow (already on develop, retained)
  - `…/.proxied/{client,server}` for `mcp-proxy` and the `mcp-client` upstream-driven path
  - `…/.guarded/{client,server}` for `mcp-client` binding-driven path (new `client.guarded.yaml` config)
- `reject.auth.callback.unknown.elicitation` scenario.
- Existing `toolkit.elicit` scenario adapted to the unified contract.

**`incubator/command-dump`**

- Kafka pcap fixtures (`KafkaProduce_shouldSendMessageValueWithProducerId.txt`, `KafkaFetch_shouldRequestPartitionOffset.txt`) refreshed for the new accept-side WINDOW order from the `drainTasks` change.

## Test plan

- [x] `./mvnw install -pl runtime/binding-mcp --also-make -Djacoco.skip=true`
- [x] `McpServerIT` — `shouldCallToolElicit{Completed,Declined,Timeout}` + `shouldCallToolToolkitElicit`
- [x] `McpClientIT` — `shouldCallToolElicit{Completed,Declined,Timeout}{Proxied,Guarded}` (6 new scenarios, no `@Ignore`)
- [x] `McpProxyIT` — FLUSH relay exercised by the proxied elicit ITs
- [x] `binding-mcp.spec` peer-to-peer: `ApplicationIT` + `NetworkIT` covering all elicit scenarios
- [x] `engine` driver tests with the wire-order-trick scripts: `DuplexIT`, `HalfDuplexIT`, `SimplexIT`, `EngineIT`, `AdvisoryIT` (binding-http)
- [x] `command-dump` `KafkaProduceApplicationIT` + `KafkaFetchApplicationIT` after fixture refresh

https://claude.ai/code/session_0172bbrihatvupUXDAn5ddvC

---
_Generated by [Claude Code](https://claude.ai/code/session_0172bbrihatvupUXDAn5ddvC)_


runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
```diff
@@ -122,4 +128,33 @@ public List<McpRouteConfig> resolveAll(
...
+    {
+        String redirectUri = null;
+        if (httpBeginEx != null)
+        {
+            final HttpHeaderFW authorityHeader = httpBeginEx.headers()
+                .matchFirst(h -> HTTP_HEADER_AUTHORITY.equals(h.name().asString()));
+            // matchFirst returns a shared flyweight; copy its value out before
+            // calling matchFirst again, otherwise the next call re-wraps the
+            // shared instance and authorityHeader.value() reads the wrong header
+            final String authority = authorityHeader != null ? authorityHeader.value().asString() : null;
```
jfallows:
```suggestion
            final String authority = Optional.ofNullable(httpBeginEx.headers()
                .matchFirst(h -> HTTP_HEADER_AUTHORITY.equals(h.name().asString())))
                .map(h -> h.value().asString())
                .orElse(null);
```
Same pattern for `path`.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
```diff
@@ -122,4 +128,33 @@ public List<McpRouteConfig> resolveAll(
...
+                .matchFirst(h -> HTTP_HEADER_PATH.equals(h.name().asString()));
+            final String path = pathHeader != null ? pathHeader.value().asString() : null;
+            if (authority != null && path != null)
+            {
+                final int queryAt = path.indexOf('?');
+                final String pathOnly = queryAt >= 0 ? path.substring(0, queryAt) : path;
+                final McpElicitationConfig elicitation = options != null ? options.elicitation : null;
+                final String callback = elicitation != null
+                    ? elicitation.callback
+                    : McpElicitationConfig.DEFAULT_CALLBACK_PATH;
```
jfallows:
```suggestion
                final String callback = Optional.ofNullable(options)
                    .map(o -> o.elicitation)
                    .map(e -> e.callback)
                    .orElse(DEFAULT_CALLBACK_PATH);
```
Use static import for `McpElicitationConfig.DEFAULT_CALLBACK_PATH`.

specs/binding-mcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mcp/streams/application/tools.call.elicit.completed/client.rpt
```diff
@@ -0,0 +1,96 @@
...
+                               .sessionId("session-1")
+                               .name("get_weather")
+                               .build()
+                           .build()}
+
+write advised zilla:challenge ${mcp:matchChallengeEx()
+                                    .typeId(zilla:id("mcp"))
+                                    .elicitCreate()
+                                        .id("elicit-1")
+                                        .url("https://server.example.com/authorize?state=7f3a9b1c&redirect_uri=https%3A%2F%2Freplace.me%2Fcallback")
```
jfallows:
```suggestion
                                        .url("https://server.example.com/authorize?state=7f3a9b1c&redirect_uri=%s".formatted(http:encodeQuery("https://replace.me/callback"))
```
Does this seem more readable?
If so we can add `encodeQuery` to `HttpFunctions` and call from here to get the same final output via:
```
URLEncoder.encode(rawValue, StandardCharsets.UTF_8).replace("+", "%20");
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
jfallows: Combine this using Optional.ofNullable with .map and .orElse(null);

jfallows: Combine this using Optional.ofNullable with .map and .orElse(null);

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -312,6 +326,10 @@ public MessageConsumer newStream(
                 .matchFirst(h -> HTTP_HEADER_ACCEPT.equals(h.name().asString()));
             final String accept = acceptHeader != null ? acceptHeader.value().asString() : null;
```
jfallows: Combine this using Optional.ofNullable with .map and .orElse(null);

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -312,6 +326,10 @@ public MessageConsumer newStream(
...
 
+            final HttpHeaderFW pathHeader = httpBeginEx.headers()
+                .matchFirst(h -> HTTP_HEADER_PATH.equals(h.name().asString()));
+            final String path = pathHeader != null ? pathHeader.value().asString() : null;
```
jfallows: Combine this using Optional.ofNullable with .map and .orElse(null);

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -312,6 +326,10 @@ public MessageConsumer newStream(
...
+
             final McpLifecycleStream resolvedSession = session;
```
jfallows: We don't need `resolvedSession`, can just use `session` directly instead, agree?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -323,17 +341,36 @@ public MessageConsumer newStream(
                 }
                 else
                 {
+                    final String redirectUri = binding != null ? binding.resolveRedirectUri(httpBeginEx) : null;
```
jfallows: `binding` cannot be null here due to `route` being non-null, agree?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -323,17 +341,36 @@ public MessageConsumer newStream(
...
                     newStream = new McpServer(
                         sender,
                         originId,
                         routedId,
                         initialId,
                         resolvedId,
-                        resolvedSession)::onNetMessage;
+                        resolvedSession,
+                        redirectUri)::onNetMessage;
```
jfallows: Prefer `redirectURI` over `redirectUri` naming.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -323,17 +341,36 @@ public MessageConsumer newStream(
...
+                        .matchFirst(h -> HTTP_HEADER_AUTHORITY.equals(h.name().asString()));
+                    final String authority = authorityHeader != null ? authorityHeader.value().asString() : "";
+                    final String callbackUrl = scheme + "://" + authority + path;
+                    newStream = new McpAuthCallbackHandler(
+                        sender,
+                        originId,
+                        routedId,
+                        initialId,
+                        path,
+                        callbackUrl)::onNetMessage;
```
jfallows: Prefer `callbackURL` over `callbackUrl` naming.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -323,17 +341,36 @@ public MessageConsumer newStream(
...
+                        redirectUri)::onNetMessage;
                 }
                 break;
             case "GET":
-                if (!acceptIncludesEventStream(accept))
+                if (isAuthCallbackPath(binding, path))
+                {
+                    final HttpHeaderFW schemeHeader = httpBeginEx.headers()
+                        .matchFirst(h -> HTTP_HEADER_SCHEME.equals(h.name().asString()));
+                    final String scheme = schemeHeader != null ? schemeHeader.value().asString() : "https";
```
jfallows: Combine this using Optional.ofNullable with .map and .orElse(null);

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -323,17 +341,36 @@ public MessageConsumer newStream(
...
+                    final HttpHeaderFW authorityHeader = httpBeginEx.headers()
+                        .matchFirst(h -> HTTP_HEADER_AUTHORITY.equals(h.name().asString()));
+                    final String authority = authorityHeader != null ? authorityHeader.value().asString() : "";
```
jfallows: Combine this using Optional.ofNullable with .map and .orElse(null);

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -1201,6 +1243,7 @@ private McpServer(
             this.session = session;
             this.decoder = decodeJsonRpc;
             this.initialMax = decodeMax;
+            this.redirectUri = redirectUri;
```
```suggestion
            this.redirectURI = redirectURI;
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -1979,6 +2033,10 @@ private void doEncodeResponsePreamble(
                 int codecLimit = 0;
                 if (sseUpgrade)
                 {
+                    if (elicitMode)
+                    {
+                        codecLimit += encodeSseIdLine(codecBuffer, codecLimit, decodedId, ++elicitSeq);
```
jfallows: For the final response, there is no need to recover, so no preamble event id needed, same for error cases.
Suggest removal of elicitSeq.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -1184,13 +1221,18 @@ private final class McpServer
 
         private McpRequestStream stream;
 
+        private final String redirectUri;
+        private boolean elicitMode;
```
jfallows: Can we avoid this boolean flag? Is the state it represents captureable or captured elsewhere?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
```diff
@@ -122,4 +129,32 @@ public List<McpRouteConfig> resolveAll(
 
         return result;
     }
+
+    public String resolveRedirectUri(
+        HttpBeginExFW httpBeginEx)
+    {
+        String redirectUri = null;
```
jfallows:
```suggestion
        String redirectURI = null;
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
```diff
@@ -122,4 +129,32 @@ public List<McpRouteConfig> resolveAll(
 
         return result;
     }
+
+    public String resolveRedirectUri(
```
jfallows:
```suggestion
    public String resolveRedirectURI(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -4221,6 +4721,128 @@ private static boolean acceptIncludesEventStream(
              accept.contains("text/*"));
     }
 
+    private static String manipulateElicitUrl(
+        String originalUrl,
+        String sessionId,
+        String redirectURI)
+    {
```
jfallows, line 4723: Can we use Pattern, Matcher and replace via regex here instead of verbose parsing?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -101,6 +115,9 @@ public final class McpServerFactory implements McpStreamFactory
     private static final String LIFECYCLE_STREAM_ID_PREFIX = "";
 
     private static final String SSE_DATA_PREFIX = "data: ";
+    private static final int ELICIT_INIT_SEQ = 0;
+    private static final int ELICIT_CREATE_SEQ = 1;
+    private static final int ELICIT_COMPLETE_SEQ = 2;
```
jfallows: These should not be needed.
The `id` field from the initial challenge ex and reply flush ex frames is the SSE event id for elicitation, same approach as other similar frames unrelated to elicitation.

PR #1755, opened 2026-05-08
feat(engine): add REDIRECT control frame primitive

### LINKED ISSUE #1753: Add REDIRECT control frame to defer affinity routing to sender
## Overview

Introduce a new engine-level control frame `REDIRECT` that allows a consumer binding to instruct its parent (producer) binding to abort the current stream and reissue it through the affinity resolver with a different hash key. This pushes protocol-specific affinity extraction out of generic transport bindings (e.g. `binding-http` server) and into the protocol-aware binding (e.g. `binding-mcp` server) that actually understands the routing key.

## Motivation

Today, `HttpAffinityConfig` requires HTTP server binding configuration to express protocol-specific extraction rules — e.g. extract `mcp-session-id` from a header, run regex `[^\.]+` over a `state` query parameter. This bakes downstream protocol knowledge into the upstream transport binding, and grows linearly with the number of session-bearing protocols layered on HTTP.

The protocol-aware binding (`binding-mcp` server, WebSocket session bindings, etc.) is the natural place for this knowledge — it already parses the protocol and knows which fields constitute the affinity key. But by the time the protocol-aware binding receives a stream, the engine has already selected a worker via the upstream binding's affinity hash. Moving affinity extraction "down" requires a way to revisit the routing decision after `BEGIN` inspection.

`REDIRECT` is that primitive.

## Frame shape

`REDIRECT` flows upstream (consumer → producer), same direction as `WINDOW`, `RESET`, and `CHALLENGE`. It closes the original stream like `RESET`. Fields:

```
REDIRECT
  ├── streamId, sequence, acknowledge, maximum   # same as BEGIN
  ├── traceId, authorization                     # carry-through
  ├── affinity                                   # original BEGIN's affinity
  ├── timestamp
  ├── hash : DirectBuffer                        # the new affinity key bytes
  └── extension : OctetsFW                       # round-tripped from BEGIN
```

The round-tripped extension means the parent doesn't need to buffer `BEGIN`'s extension data while awaiting `WINDOW` — the consumer returns it via `REDIRECT` if it decides to redirect.

## Timing rule

`REDIRECT` is mutually exclusive with `WINDOW` for the first response from the consumer:

```
producer                                     consumer
  BEGIN ──────────────────────────────────────►
                                                inspects extension
                                                decides
  ◄────────────────────────────── WINDOW   (commit, normal flow)
       OR
  ◄──────────────────────────── REDIRECT   (abort, reissue)
```

Once the consumer issues `WINDOW`, the producer is permitted to send `DATA`. Allowing `REDIRECT` after that would lose in-flight bytes. Mutual exclusion makes the contract crisp.

For protocols where the consumer needs `DATA` before deciding, future extensions could relax this with explicit reissue semantics — out of scope for v1.

## Parent handling

```
parent.onRedirect(redirect):
  closeStream(redirect.streamId)            // RESET-equivalent semantics
  beginEx = redirect.extension
  hashKey = redirect.hash
  consumer = affinityResolver.resolve(routedBindingId, hashKey)
  newStreamId = supplyInitialId(...)
  consumer.accept(BEGIN { ext = beginEx, traceId, authorization, ... })
```

`AffinityResolver` is the existing/proposed pluggable SPI; for local resolution it picks a worker, for clustered routing it returns a remote consumer. `REDIRECT` does not depend on clustering, but composes with it.

## Constraints

- **Loop bound.** At most one `REDIRECT` per request lineage. Enforced engine-side via a per-trace counter, or by convention. For MCP specifically the second `BEGIN` lands on the right worker and resolves locally with no further redirect.
- **Capability discovery.** Parent bindings opt in to handling `REDIRECT`. Emitting `REDIRECT` against a non-supporting parent yields `RESET` and an operator config error — bindings that emit `REDIRECT` should declare a parent that supports it.
- **Trace continuity.** The reissued `BEGIN` preserves `traceId`. Observability sees one logical request crossing a worker boundary, not two.
- **Authorization carry-through.** `authorization` is on `REDIRECT` and forwarded onto the new `BEGIN`. No re-auth.

## Beneficiaries beyond MCP

This is a generic primitive. Any binding with protocol-specific session affinity sitting behind a generic transport can use it:

- WebSocket-over-HTTP with session-shaped routing
- MQTT `clientId`-based affinity
- gRPC with session metadata
- MCP (motivating use case)

In each case, the protocol-aware binding extracts the affinity key after seeing `BEGIN`; the upstream HTTP/TCP/etc. binding stays generic. `HttpAffinityConfig`'s protocol-shaped extraction options become optional once consumers can drive their own affinity via `REDIRECT`.

## k3po test support

The zilla engine's k3po-zilla transport needs `REDIRECT` support so test scripts can exercise the new frame. This change lives in zilla engine's test scope, not in `aklivity/k3po`.

DSL syntax mirrors the existing advisory verbs for upstream-flowing frames (e.g. `challenge` / `challenged`):

```
# consumer-side script (reading data stream, advising producer)
read advise zilla:redirect
  hash "abc123xyz"
  ext ${zilla:beginExHttp(...)}

# producer-side script (writing data stream, being advised by consumer)
write advised zilla:redirect
  hash "abc123xyz"
  ext ${zilla:beginExHttp(...)}
```

## Implementation surface

- New frame type in the engine's primitive frame ABI: `REDIRECT`.
- Flyweight builder/decoder + RW types alongside existing `BeginFW` / `WindowFW` / `ResetFW` / `ChallengeFW`.
- Parent-side handler hook on bindings that opt in (`binding-http` server is the immediate consumer for the MCP use case).
- Small addition to `AffinityResolver` callers to accept an external hash key (vs. extracting from `BEGIN`).
- k3po-zilla transport extensions: `read advise zilla:redirect` and `write advised zilla:redirect` with `hash` and `ext` fields, mirroring the `challenge` / `challenged` advisory shapes.
- No SPI changes.
- No schema changes to existing bindings.

## Relationship to other work

Emerged from cluster state management discussions around #1739 (MCP elicitation across clustered zilla deployments). The cluster work needs an `AffinityResolver` SPI with cross-node routing; `REDIRECT` is the optimization that makes that SPI usable without leaking MCP-specific config into HTTP server binding.

Treat as a follow-on optimization to whatever lands for the cluster work — non-blocking, but unlocks cleaner layering once present.

## Acceptance criteria

- [ ] `REDIRECT` frame defined in engine ABI with the documented fields, including round-tripped extension.
- [ ] `binding-http` server handles `REDIRECT`: closes original stream, reissues `BEGIN` with round-tripped extension and the supplied hash via the affinity resolver path.
- [ ] `AffinityResolver` callers accept an external hash key from the `REDIRECT` path and route accordingly.
- [ ] One `REDIRECT`-per-lineage enforcement (engine counter or documented convention).
- [ ] `traceId` preserved across redirect; observability shows one logical request.
- [ ] `authorization` preserved across redirect; no re-auth required.
- [ ] `binding-mcp` server emits `REDIRECT` for misrouted requests (sessionId mismatch on POST `/mcp`, sessionId-prefix mismatch on `state` for callback path), allowing `HttpAffinityConfig` to drop MCP-specific extraction options.
- [ ] k3po-zilla transport supports `read advise zilla:redirect` / `write advised zilla:redirect` with `hash` and `ext` fields.
- [ ] Test: consumer emits `REDIRECT` in response to `BEGIN`; original stream is closed; reissued stream lands on the affinity-resolved worker with the correct extension.
- [ ] Test: consumer emits `REDIRECT` against a non-supporting parent → `RESET` and a config-time error surfaced.

jfallows:
## Summary

Introduces a new engine-level control frame `REDIRECT` (type id `0x40000005`)
that flows upstream from consumer to producer, mirroring `RESET` /
`CHALLENGE` direction. `REDIRECT` carries a new affinity (the hash result
from a protocol-aware consumer) and the round-tripped `BEGIN` extension,
giving a consumer binding a way to abort a misrouted stream and instruct
the parent to reissue `BEGIN` with a different affinity.

This PR lands the foundational primitive only — the frame ABI plus
engine throttle dispatch and k3po-zilla test-transport support.

## Frame shape

```
struct Redirect extends Frame [0x40000005]
{
    int64 affinity;     // the new affinity to use
    octets extension;   // round-tripped from BEGIN
}
```

Design note (vs. the original issue): `REDIRECT` carries the *hash result*
as `int64 affinity`, not the bytes-to-hash. This avoids any coupling
between protocol-aware bindings and the choice of hash function — the
consumer pre-hashes and the parent's logic becomes "use this affinity
instead of the one on the original BEGIN." The frame shape mirrors
`Begin.affinity` exactly, so the parent's resolver call shape doesn't
change. The original `BEGIN`'s affinity is dropped from the frame entirely
— the parent already knows it because it dispatched the BEGIN.

## Engine internals

- `EngineWorker` throttle dispatch recognizes `RedirectFW.TYPE_ID` on both
  initial and reply paths and removes the dispatcher (REDIRECT closes the
  stream, like RESET).
- `Target` write paths recognize `RedirectFW.TYPE_ID`, remove the stream
  from the streams map, and drop any registered streamId-set entry; on
  the reply path also drop any pending correlation.

## k3po-zilla test transport

- New `ADVISORY_REDIRECT` `StructuredTypeInfo` registered as a
  read-advisory + write-advisory pair (mirroring `ADVISORY_CHALLENGE`).
- New `ZillaExtensionKind.REDIRECT` (unaligned) with overrides routing
  both `encodeBuffer` and `decodeBuffer` to `writeExtBuffer`. This keeps
  the script-supplied affinity isolated from the BEGIN ext, which lives
  in `readExtBuffer(BEGIN)` for the round-trip.
- `ZillaStreamFactory.doRedirect`: reads the script-supplied affinity via
  `redirectEx.getLong(redirectEx.readerIndex())` and pulls the BEGIN ext
  bytes from `readExtBuffer(BEGIN)` (absolute-index reads, so it survives
  any prior `read zilla:begin.ext` advancing the reader index).
- `ZillaTarget.doRedirect(...)`: builds a `RedirectFW` with `affinity` and
  round-tripped `extension`, plus carry-through `authorization`.
- `Throttle.onRedirect`: surfaces only `RedirectFW.affinity()` to
  `writeExtBuffer(REDIRECT, false).writeLong(affinity)`, unregisters the
  throttle, and fires `ADVISORY_REDIRECT`. The frame's `extension` is
  consumed but not surfaced to the script — the round-trip is silent.

## DSL form

```
read advise zilla:redirect  0x00000000000000b1L
write advised zilla:redirect 0x00000000000000b1L
```

A single long literal (`Long.decode()` parsed). Native byte order on the
channel buffer; round-trip is consistent regardless of platform.

## Test plan

- [x] `simplex/server.sent.read.advise.redirect.ext/{client,server}.rpt`
      peer-to-peer scripts exercise the full encode/decode path with a
      non-empty `zilla:begin.ext` round-trip.
- [x] `SimplexIT#shouldReceiveServerSentReadAdviseRedirectWithExtension`
      runs the scripts; 37/37 SimplexIT tests pass.
- [x] DuplexIT/HalfDuplexIT continue to pass (no regressions in CHALLENGE
      handling).
- [x] 316/316 engine unit tests pass.
- [x] `./mvnw checkstyle:check -pl runtime/engine` clean.

## Acceptance criteria addressed

- [x] `REDIRECT` frame defined in engine ABI with documented fields,
      including round-tripped extension.
- [x] `traceId` preserved across redirect (carried on `Frame`).
- [x] `authorization` preserved (carried on `Frame` and forwarded in
      k3po `doRedirect`).
- [x] k3po-zilla transport supports `read advise zilla:redirect` /
      `write advised zilla:redirect`.
- [x] Test: consumer emits `REDIRECT`; original stream is closed at the
      engine throttle level (dispatcher removal); ext is round-tripped
      onto the new frame.

Fixes #1753


jfallows (commented):
LGTM

PR #1758, opened 2026-05-08
feat(binding-http): handle REDIRECT and remove `with.affinity` syntax

### LINKED ISSUE #1756: binding-http: handle REDIRECT and remove `with.affinity` syntax
## Overview

Wire `REDIRECT` (added in #1753) into the `binding-http` server, and at
the same time remove the `with.affinity` extraction syntax that
`HttpAffinityConfig` introduced in #1744. The two changes are paired —
the latter is a stop-gap for protocol-specific session affinity over
HTTP, and `REDIRECT` lets a protocol-aware downstream binding drive
that decision itself, so the upstream HTTP server stops needing to know.

## Motivation

`HttpAffinityConfig` requires HTTP server binding configuration to
express protocol-specific extraction rules (e.g. extract a session id
from a header, run a regex over a query parameter). This bakes
downstream protocol knowledge into the upstream transport binding and
grows linearly with each session-bearing protocol layered on HTTP. The
regex match-group form is also fragile.

`REDIRECT` (already on `develop` after #1753) is the clean replacement.
The protocol-aware binding extracts the affinity key after seeing
`BEGIN`, hashes it itself, and emits `REDIRECT` upstream with the new
affinity. The HTTP server reissues the application `BEGIN` through the
affinity-selection path with the supplied affinity. No protocol-shaped
options on the HTTP server.

## Scope

This issue covers landing both halves in a single PR so users on
`develop` never see a release where `with.affinity` is gone but no
working alternative is in place. Since `with.affinity` only just landed
on develop in #1744 and presumably hasn't shipped in a release yet,
hard-remove (no soft-deprecate) is appropriate.

### Add: REDIRECT support in `binding-http` server

- Recognize incoming `REDIRECT` on the application stream throttle —
  same dispatch site as `WINDOW`, `RESET`, `CHALLENGE`.
- Close the original application stream (`RESET`-equivalent
  semantics). The HTTP request that originated it is still alive on
  the network side; we abandon only the misrouted application stream.
- Reissue `BEGIN` through the affinity-selection path with the
  supplied affinity. Use the round-tripped `extension` from the
  `REDIRECT` frame on the new `BEGIN`. Preserve `traceId` and
  `authorization` (no re-auth).
- One-redirect-per-lineage enforcement: track on the per-request
  state that this request has already redirected once; a second
  `REDIRECT` against the reissued stream becomes a `RESET` plus a
  surfaced error rather than another reissue.

The shape of the affinity-selection call is small enough today that
this can be inline in `binding-http` (no new SPI).

### Remove: `with.affinity` extraction syntax

- Remove `HttpAffinityConfig`, `HttpAffinityConfigBuilder`, and the
  associated config / schema fields under `routes[].with.affinity`.
- Remove `HttpAffinityResolver` and the call sites in `binding-http`
  that consume header / query / regex extraction.
- Update `specs/binding-http.spec/.../binding-http.schema.json` to
  drop the `affinity` property from `with`.
- Update or remove any IT scripts that exercise the old extraction
  path.

## Acceptance criteria

- [ ] `binding-http` server `onRedirect` handler closes the original
       application stream and reissues `BEGIN` with the supplied
       affinity, round-tripped extension, preserved `traceId` and
       `authorization`.
- [ ] One-redirect-per-lineage enforcement (second `REDIRECT` on the
       reissued stream surfaces an error / `RESET`).
- [ ] `HttpAffinityConfig`, `HttpAffinityConfigBuilder`,
       `HttpAffinityResolver`, and related config + schema removed.
- [ ] No `with.affinity` references remain anywhere in the repo
       (`./mvnw test` confirms nothing broke).
- [ ] `binding-http` ITs cover: REDIRECT roundtrip, REDIRECT against a
       supporting parent (success), second REDIRECT on a reissued
       stream (rejected), schema-time rejection of any
       `with.affinity` snippet for safety.

## Related

- Depends on #1753 (REDIRECT primitive).
- Replaces the `with.affinity` syntax introduced in #1744.

jfallows:
Closes #1756

## Summary

Wire `REDIRECT` (added in #1755) into the `binding-http` server, and at the
same time remove the `with.affinity` extraction syntax that
`HttpAffinityConfig` introduced in #1744. The two changes are paired —
`with.affinity` was a stop-gap for protocol-specific session affinity
over HTTP, and `REDIRECT` lets a protocol-aware downstream binding drive
that decision itself, so the upstream HTTP server stops needing to know.

## Add: REDIRECT support in `binding-http` server

- Recognize `RedirectFW.TYPE_ID` on the application stream throttle for
  both the HTTP/1.1 (`HttpExchange`) and HTTP/2 (`Http2Exchange`)
  dispatch sites, alongside `WINDOW`, `RESET`, `CHALLENGE`.
- On `REDIRECT`: close the original misrouted application stream, mint
  a fresh `requestId` via the affinity-selection path
  (`supplyInitialId(routedId, Long.hashCode(affinity))`), and reissue
  `BEGIN` carrying the round-tripped extension and the supplied
  affinity. The originating HTTP request stays alive on the network
  side; only the misrouted application stream is abandoned.
- Preserve `traceId` from the `REDIRECT` frame and `authorization`
  from the original exchange (no re-auth).
- One-redirect-per-lineage enforcement: a `redirected` flag on the
  per-request state rejects a second `REDIRECT` against the reissued
  stream, surfacing a 500 to the client and closing the exchange.

## Remove: `with.affinity` extraction syntax

- Delete `HttpAffinityConfig`, `HttpAffinityConfigBuilder`,
  `HttpAffinitySource`, and `HttpAffinityResolver`.
- Drop `affinity` from `HttpWithConfig`, `HttpWithConfigBuilder`,
  `HttpWithConfigAdapter`, `HttpWithResolver`, and `HttpRouteConfig`.
- Replace `HttpRouteConfig.resolve(headerByName)` with `resolve()`
  (initial route) and `resolve(long affinity)` (REDIRECT reissue).
- Drop the `affinity` property from `binding-http.schema.patch.json`
  `with` and rely on `additionalProperties: false` to reject any
  `with.affinity` snippet at validation time.
- Remove the `affinity` IT scripts, `AffinityIT` test classes,
  `server.affinity.yaml` config samples, and the affinity-specific
  cases from `HttpWithConfigAdapterTest` and `SchemaTest`.

## Tests

- `RedirectIT` (rfc7230 network) exercises the peer-to-peer `REDIRECT`
  frame round-trip alongside a happy-path GET and a rejected
  second-redirect.
- `SchemaTest.shouldRejectWithAffinity` validates that any
  `with.affinity` snippet is rejected at schema-load time
  (`server.with.affinity.invalid.yaml`).

## Test plan
- [x] `./mvnw verify -pl specs/binding-http.spec` — 475 tests, 0 failures
- [x] `./mvnw verify -pl runtime/binding-http` — 374 tests, 0 failures
- [x] `./mvnw checkstyle:check -pl runtime/binding-http,specs/binding-http.spec`
- [x] `./mvnw license:check -pl runtime/binding-http,specs/binding-http.spec`
- [ ] Follow-up: runtime-level `RedirectIT` driving the network →
      `binding-http` → application flow with `read advise zilla:redirect`.

https://claude.ai/code/session_0112qRPXfYtALkaPwDC3qu6N

---
_Generated by [Claude Code](https://claude.ai/code/session_0112qRPXfYtALkaPwDC3qu6N)_


jfallows (commented):
LGTM

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -3024,6 +3034,56 @@ private void onExchange(
...
+
+                requestSeq = 0;
+                requestAck = 0;
+                requestMax = 0;
+                requestPad = 0;
+                requestCaps = 0;
+                responseSeq = 0;
+                responseAck = 0;
+                responseMax = 0;
+                requestState = HttpExchangeState.PENDING;
```
jfallows: move this into a `clear()` method.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -6019,6 +6084,53 @@ private void onExchange(
...
+                responseId = supplyReplyId.applyAsLong(newRequestId);
+                affinity = newAffinity;
+
+                requestSeq = 0;
+                requestAck = 0;
+                requestMax = 0;
+                requestPad = 0;
+                requestBud = 0L;
+                requestCaps = 0;
+                state = 0;
```
jfallows, line 6153: Move to a `clear()` method.
It should clear _all_ the field, then set just the ones needed afterwards, not mixed order like this.

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/config/v2/server.redirect.yaml
```diff
@@ -14,13 +14,17 @@
...
+    type: http
+    kind: server
+    options:
+      versions:
+        - h2
+    routes:
+      - exit: app0
+        when:
+          - headers:
+              :authority: localhost:8080
```
jfallows: Do we need a different config for server redirect, or is there an existing one we can reuse instead?

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/streams/application/rfc7230/redirect/request.redirect/server.rpt
```diff
@@ -19,21 +19,36 @@ accept "zilla://streams/app0"
...
+                              .header(":method", "POST")
+                              .header(":path", "/items")
+                              .header(":authority", "localhost:8080")
+                              .header("content-length", "100")
+                              .build()}
+
 connected
 
-read closed
+read advise zilla:redirect 0x00000000000000b1L
```
jfallows, line 33:
```suggestion
read advise zilla:redirect 0x00000000000000b1L
write aborted
```
To verify the RESET is sent.
Needs a companion read abort in the peer script.

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/streams/application/rfc7230/redirect/request.redirect/server.rpt
```diff
@@ -19,21 +19,36 @@ accept "zilla://streams/app0"
...
+                              .header(":authority", "localhost:8080")
+                              .header("content-length", "100")
+                              .build()}
+
+connected
 
 write zilla:begin.ext ${http:beginEx()
                               .typeId(zilla:id("http"))
                               .header(":status", "200")
+                              .header("connection", "close")
```
jfallows: We don't want to add this, since the connection close stuff is just part of the lower level HTTP/1.1 protocol and not portable to HTTP/2, so we abstract it away.

specs/binding-http.spec/src/main/scripts/io/aklivity/zilla/specs/binding/http/streams/application/rfc7540/redirect/request.redirect/server.rpt
```diff
@@ -17,25 +17,38 @@
...
+                              .header(":authority", "localhost:8080")
+                              .header("content-length", "100")
+                              .build()}
+
+connected
 
 write zilla:begin.ext ${http:beginEx()
                               .typeId(zilla:id("http"))
                               .header(":status", "200")
+                              .header("content-length", "0")
```
jfallows: We tend not to send content-length like this, please remove.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
jfallows: also verify in `clear()` via `assert HttpState.closed(state)`.
this will make sure we send the RESET after receiving REDIRECT before clearing for reuse.

runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
```diff
@@ -2833,10 +2838,10 @@ private final class HttpExchange
...
             private final long routedId;
             private final long traceId;
-            private final long requestId;
-            private final long responseId;
+            private long requestId;
+            private long responseId;
             private final long sessionId;
-            private final long affinity;
+            private long affinity;
             private final HttpPolicyConfig policy;
```
jfallows: Move any newly non-final fields down below the final fields in their own group.

jfallows (commented):
LGTM

PR #1757, opened 2026-05-08
feat(engine): add RouterFactorySpi for pluggable stream factory composition

jfallows:
## Summary

Adds an SPI surface (`RouterFactorySpi`) that lets engine extensions contribute to the composition of `EngineContext.streamFactory()`. Mirrors the factory + context shape already used by `Store`:

- `RouterFactorySpi.create(Configuration) → Router`
- `Router.supply(RouteableContext) → RouterContext`
- `RouterContext.attach(RouterConfig) → BindingHandler` (paired with `detach(long routerId)`)

`RouteableContext` is the under-engine surface available to the router during setup: it exposes the engine `Configuration`, the engine's current default `BindingHandler` stream factory (which the router may wrap), and `attachComposite(...)` / `detachComposite(...)` for injecting synthesized namespaces alongside operator-authored ones.

When `zilla.engine.router` selects a router by name, the engine instantiates it via `ServiceLoader`, supplies a `RouteableContext`, attaches a synthesized `RouterConfig`, and installs the returned `BindingHandler` as the worker's stream factory. With no router configured, runtime behavior is unchanged.

The change is engine-only:

- New `runtime.engine.router` package: `RouterFactorySpi`, `RouterFactory`, `Router`, `RouterContext`, `RouteableContext`
- New `runtime.engine.config.RouterConfig` (with `id`) + `RouterConfigBuilder`
- `EngineConfiguration` adds `zilla.engine.router` property and `routerName()` accessor
- `EngineBuilder` / `Engine` / `EngineWorker` thread the router through; `EngineWorker.streamFactory()` returns a field initialised to `this::newStream` and replaced by the wrapped handler when a router attaches; `onClose` invokes `RouterContext.detach(routerId)`
- moditect `module-info` exports the new package and declares `uses RouterFactorySpi`
- Test impl under `test/internal/router/` (`TestRouterFactorySpi`, `TestRouter`, `TestRouterContext`) returns the engine's default handler unchanged; wired via `META-INF/services`
- `RouterFactoryTest` covers discovery, creation, and the unrecognized-name error citing `RouterFactory.names()`
- `EngineIT.shouldHandshakeWithRouter` exercises the full startup path with a router attached

Closes #1754

## Test plan
- [x] `./mvnw -pl runtime/engine clean verify` — 319 unit tests, 203 ITs pass; JaCoCo coverage check passes
- [x] `RouterFactoryTest` (3 tests): discovery, creation, unrecognized-name error
[...]

PR #1760, opened 2026-05-10
fix(engine): await worker/boss onStart before EngineManager registration

jfallows:
## Summary

Fixes an NPE introduced by #1757 that surfaces in `binding-pgsql-kafka` (`ProxyIT.shouldDropTopic` and others):

```
java.lang.NullPointerException: Cannot invoke "EngineRegistry.attach(NamespaceConfig)" because "this.registry" is null
    at EngineWorker.attach(EngineWorker.java:1086)
    at EngineManager.lambda$register$1(EngineManager.java:588)
```

After #1757, `EngineWorker` builds its `registry` (and the per-worker binding/guard/vault/catalog/store/model/metric contexts) inside `onStart`, on the worker thread, so handler `supply(this)` calls happen on the owning thread. But `EngineWorker.doStart` was fire-and-forget — it returned as soon as `startOnThread` had launched the runner, before the runner had reached `onStart`.

`Engine.start` then proceeded to start `EngineManager`, whose config watch could fire `register` and call `EngineWorker.attach(namespace)` → `registry.attach(namespace)` against a worker whose `onStart` hadn't yet run, producing the NPE. The pre-#1757 code happened to be safe only because all of that initialization ran in the constructor.

The fix makes startup ordering deterministic with a `CountDownLatch` per worker / boss:

- Counted down at the end of `onStart` in a `finally` (so init failures still release the latch — `await` returns and the runner's error handler surfaces the real exception).
- Awaited at the end of `doStart` before returning.

`EngineBoss` currently has no work in `onStart`, but adding the symmetric scaffolding keeps the worker/boss lifecycle consistent and protects against future asymmetric init.

## Test plan
- [x] `./mvnw -pl runtime/engine clean install` — 320 unit + 204 ITs pass; JaCoCo coverage check passes
- [x] `./mvnw -pl incubator/binding-pgsql-kafka clean verify` — `ProxyIT` 5/5 pass (previously 4 run with 3 errors)
- [x] `./mvnw -pl runtime/binding-http -am clean verify` — green
- [x] `./mvnw -pl runtime/engine checkstyle:check` — clean

https://claude.ai/code/session_01RXTcFJG39snE5KhLkMSALD

---
_Generated by [Claude Code](https://claude.ai/code/session_01RN8xhzqGY2rsXGdkaELa42)_

### DIFF
--- a/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineBoss.java
+++ b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineBoss.java
@@ -27,6 +27,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.CountDownLatch;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 
@@ -59,6 +60,7 @@ public class EngineBoss implements EngineController, Agent
     private final Map<String, BindingController> controllersByType;
     private final int maxSelectMillis;
     private final int maxIdleCount;
+    private final CountDownLatch started = new CountDownLatch(1);
 
     private volatile Thread thread;
     private Set<NamespaceConfig> namespaces;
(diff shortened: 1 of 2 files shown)


jfallows (commented):
LGTM

PR #1764, opened 2026-05-12
mcp: session id alignment + redirect handling and configurable attempts

### LINKED ISSUE #1761: binding-mcp server: align mcp-session-id worker affinity via REDIRECT
## Overview

The MCP server binding maintains per-worker session state in `McpServerFactory.sessions` (a per-worker `Map<String, McpLifecycleStream>` keyed by `mcp-session-id`). For correctness, every request for a given session must reach the worker that owns it — otherwise lookups fail and the request cannot be served.

This issue introduces two changes:

1. On session creation, generate an `mcp-session-id` whose hash maps to the worker creating the session.
2. On observing an inbound `mcp-session-id` (header or extracted from `state` query param), if its hash does not map to the current worker, emit `REDIRECT` to route to the right worker.

## Motivation

All requests for a session must arrive on the worker that holds it. HTTP/2 connection pinning provides this for free for the typical case: all multiplexed requests on a single connection land on the same worker, and a session created on that connection lives there. The hot path needs no extra work.

The cases that don't get connection-level pinning need an explicit alignment mechanism:

- **HTTP/1.1 clients** open a new TCP connection per request (or with limited pipelining); each connection may land on a different worker.
- **Reconnected HTTP/2 clients** (idle timeout, network blip, multi-tab, multi-process) open a new connection that lands on whichever worker the engine assigns.
- **Separate connections in the same MCP session lifecycle** — e.g. the OAuth callback path is a different connection from the original MCP client.

For these, the request reaches some worker via default affinity (connection-based), but the session lives elsewhere. The MCP server binding inspects the inbound `mcp-session-id`, finds it does not hash to the current worker, and emits `REDIRECT` carrying the session-id's bytes so the producer (HTTP server binding) reissues the stream on the correct worker.

### Session creation

On `initialize` request:

- Generate an opaque token.
- Compute the engine's affinity hash for the token.
- If `Math.floorMod(hash, workerCount) == self`, accept the token as the session-id.
- Otherwise, generate another token and retry until the hash maps to `self`.
- Bound the retry loop with a cap (e.g. 16 attempts) and fail cleanly if exceeded.

The resulting `mcp-session-id` value is guaranteed to hash to this worker. Subsequent requests carrying this session-id route here via default affinity without `REDIRECT`.

### Session-id observation on inbound streams

For each inbound stream arriving at `binding-mcp` server with a session-id — either via the `mcp-session-id` header on POST `/mcp` or via the `[^\.]+` prefix of the `state` query parameter on the elicitation callback path:

- Extract the session-id bytes.
- Compute the hash.
- If `Math.floorMod(hash, workerCount) == self`, proceed as today — the session is local.
- Otherwise, emit `REDIRECT` upstream with:
  - `hash` populated with the session-id bytes
  - `extension` round-tripped from the inbound `BEGIN` so the producer can reissue without re-buffering

The HTTP server binding closes the original stream, reissues `BEGIN` through the affinity resolver path with the supplied hash, landing on the right worker.

### Worker-locality fast path

When the inbound session-id resolves to this worker, the existing per-worker `sessions` map answers the lookup; no behavioral change from today's path.

## Outcome by client type

| Client | Outcome |
|---|---|
| HTTP/2 client, same connection as session creation | All requests pinned to the session-creating worker; no `REDIRECT` ever fires. |
| HTTP/2 client, new connection (reconnect, multi-tab) | First request per session triggers `REDIRECT`; reissued stream lands on the correct worker. Correct in all cases. |
| HTTP/1.1 client | Each new connection's first request for a session triggers `REDIRECT` if the connection's worker differs from the session's worker. Correct in all cases. |
| Separate connection for the same session (e.g. OAuth callback) | Lands on whatever worker the new connection picked; `REDIRECT` aligns with the session's worker. |

## Implementation surface

- Session-id generator in `binding-mcp` server: hash-and-retry loop bounded by attempt count, using the engine's canonical affinity hash function.
- `binding-mcp` server `BEGIN` handler: extract session-id key from `mcp-session-id` header or `state` query-param prefix, compute affinity, conditionally emit `REDIRECT` when not aligned.
- Shared canonical hash function: producer (HTTP server binding's affinity path) and consumer (`binding-mcp` server's worker check) must compute identical results for the same key bytes.
- Configuration option (optional): max retry attempts for session-id generation, with reasonable default.

## Acceptance criteria

- [ ] On `initialize`, the generated `mcp-session-id` hashes to the current worker, verified by recomputing the hash and applying `Math.floorMod(hash, workerCount)`.
- [ ] Session-id generation retry loop is bounded; exceeding the cap produces a clean error rather than an infinite loop.
- [ ] On observing an inbound `mcp-session-id` not aligned to this worker, `REDIRECT` is emitted with the session-id bytes as `hash` and the inbound `BEGIN` extension round-tripped.
- [ ] When the inbound session-id resolves locally, the existing path is taken with no added overhead.
- [ ] Test: HTTP/2 client doing `initialize` followed by N tool calls on the same connection — zero `REDIRECT` emissions.
- [ ] Test: client opens a second connection with the same session-id, lands on a different worker — `REDIRECT` emitted, reissued stream serves the request on the correct worker.
- [ ] Test: HTTP/1.1 client request carrying an existing `mcp-session-id` on a non-aligned worker — `REDIRECT` emitted, request handled on the correct worker.
- [ ] Test: OAuth callback path with `state` carrying `<sessionId>.<...>` prefix on a non-aligned worker — `REDIRECT` emitted, callback handled on the correct worker.

jfallows:
### Motivation

- Ensure MCP session IDs are aligned to the local stream index and provide a redirect when a client-supplied session ID maps to a different index.
- Allow tuning how many candidate session IDs are generated to find an aligned one via configuration.

### Description

- Add configuration property `zilla.binding.mcp.session.id.attempts` with getter `sessionIdAttempts()` and default `ENGINE_WORKERS * 2` in `McpConfiguration` and expose it as `MCP_SESSION_ID_ATTEMPTS`.
- Introduce `supplyInitialIdByHash`, `sessionIdAttempts`, and `localIndex` in `McpServerFactory`, plus a `RedirectFW` builder to support redirect messages.
- Implement session alignment logic: `newSessionId(...)` tries up to configured attempts to obtain an aligned session id, `isSessionIdAligned(...)` checks alignment using `supplyInitialIdByHash` and `streamIndex`, and `extractSessionIdFromState(...)` pulls session id from `state` URL parameter.
- When an incoming HTTP begin contains a session id that is not aligned, construct a `McpRedirectHandler` to send a `RedirectFW` pointing the client to the correct affinity, and create lifecycle sessions with the chosen aligned session id or reset when none found.

### Testing

- Updated unit test `McpConfigurationTest` to include the new property constant and ran the binding-mcp module tests with `mvn -pl runtime/binding-mcp test`, which succeeded.
- Built the module and ran unit tests locally for the modified classes, and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a02685dd9d88322abf59ab0d2117e12)

Closes #1761

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/McpConfiguration.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/McpConfiguration.java
@@ -14,6 +14,8 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.internal;
 
+import static io.aklivity.zilla.runtime.engine.EngineConfiguration.ENGINE_WORKERS;
+
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
@@ -39,6 +41,7 @@ public class McpConfiguration extends Configuration
     public static final PropertyDef<String> MCP_CLIENT_NAME;
     public static final PropertyDef<String> MCP_CLIENT_VERSION;
     public static final PropertyDef<Duration> MCP_INACTIVITY_TIMEOUT;
+    public static final IntPropertyDef MCP_SESSION_ID_ATTEMPTS;
     public static final IntPropertyDef MCP_KEEPALIVE_TOLERANCE;
     public static final PropertyDef<Duration> MCP_SSE_KEEPALIVE_INTERVAL;
 
@@ -59,6 +62,8 @@ public class McpConfiguration extends Configuration
             McpConfiguration::defaultServerVersion);
         MCP_INACTIVITY_TIMEOUT = config.property(Duration.class, "inactivity.timeout",
             (c, v) -> Duration.parse(v), "PT60S");
+        MCP_SESSION_ID_ATTEMPTS = config.property("session.id.attempts",
+            McpConfiguration::defaultSessionIdAttempts);
         MCP_KEEPALIVE_TOLERANCE = config.property("keepalive.tolerance", 2);
         MCP_SSE_KEEPALIVE_INTERVAL = config.property(Duration.class, "sse.keepalive.interval",
             (c, v) -> Duration.parse(v), "PT15S");
@@ -114,6 +119,11 @@ public class McpConfiguration extends Configuration
     public int keepaliveTolerance()
     {
         return MCP_KEEPALIVE_TOLERANCE.getAsInt(this);
+    }
+
+    public int sessionIdAttempts()
+    {
+        return MCP_SESSION_ID_ATTEMPTS.getAsInt(this);
     }
 
     public Duration sseKeepaliveInterval()
@@ -183,6 +193,12 @@ public class McpConfiguration extends Configuration
     private static String defaultSessionIdSupplier()
     {
         return UUID.randomUUID().toString();
+    }
+
+    private static int defaultSessionIdAttempts(
+        Configuration config)
+    {
+        return Math.max(1, ENGINE_WORKERS.getAsInt(config) * 2);
     }
 
     private static ElicitationIdSupplier decodeElicitationIdSupplier(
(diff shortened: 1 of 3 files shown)


jfallows (commented):
LGTM

PR #1769, opened 2026-05-13
Support routing based on TLS client certificate presence/signer for mixed-auth endpoints on shared port

### LINKED ISSUE #1697: Support routing based on TLS client certificate presence/signer for mixed-auth endpoints on shared port
## Summary

When multiple products/endpoints share the same port with different authentication mechanisms (e.g., mTLS with one CA, mTLS with a different CA, or API keys with no client cert), the current TLS server configuration cannot handle them correctly.

## Problem

The `tls` binding's `mutual` option is a connection-level decision made before any routing occurs. If all products on a port use mTLS with the same CA, `mutual: required` works fine. But if there is a mix — some products using mTLS (possibly with different CAs) and others using API keys — then:

1. `mutual: required` cannot be used because API key clients have no client certificate.
2. `mutual: requested` would need to be used instead, making the cert optional.
3. After TLS handshake, routing would need to branch based on:
   - Whether a client certificate was presented at all.
   - Which CA signed the client certificate (to distinguish between different mTLS products).

Neither of these routing conditions is currently supported.

## Impact

The zilla-plus gateway schema for multi-product configurations on a shared port currently assumes all endpoints on that port use the same auth mechanism. Mixed mTLS/API-key auth on a shared port is a gap that cannot be bridged without this capability.

## Proposed Solution

Add support for routing guards/rules based on:
- Presence of a TLS client certificate (`tls.client.cert` present/absent)
- The issuer/signer of the TLS client certificate

This would allow `mutual: requested` at the TLS layer and then policy-driven routing per product based on cert attributes.

akrambek:
Fixes #1697


specs/binding-tls.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tls/config/server.mutual.signer.routed.yaml
```diff
@@ -0,0 +1,140 @@
...
+            mutual: requested
+        routes:
+            - when:
+                - cert:
+                    signer:
+                      cn: serverca
+              exit: app0
+            - when:
+                - cert: false
+              exit: app1
```
jfallows, line 168: The TLS server has a `localhost` Server Certificate signed by the Server CA, and it also trusts Client Certificates signed by the Client CA. This Client CA has an alias `clientca` in the `server` vault.

This syntax and example is quite confusing because the route seems to be saying it matches any Client Certificate with a signer that has Common Name `serverca`, which seems to convey a Server CA.

I think we need to maintain the abstraction based on alias instead of unpacking the Common Name of the signer, so that we don't force config changes when the content of the vault changes such that the alias maps to a new CA with a different Common Name.

We can use `mutual: none` instead of the `cert` `false` to indicate no client certificate.

When `mutual:none`, then `trust` cannot be specified on the `route` `when` condition.
When `trust` is specified on the `route` `when` condition, then `mutual:required` is implied.
When condition does not support `mutual:requested` as a value to match on, that doesn't specify either outcome.

```suggestion
bindings:
    net0:
        type: tls
        kind: server
        vault: server
        options:
            keys:
                - localhost
            trust:
                - clientca
            mutual: requested
        routes:
            - when:
                - trust:
                      - clientca
                  mutual: required     # implied when trust is present, can be omitted
              exit: app1
            - when:
                - mutual: none         # implies `trust` must not be present
              exit: app0
```

jfallows (changes requested):

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/vault/VaultHandler.java
```diff
@@ -78,4 +80,24 @@ KeyManagerFactory initSigners(
...
+     *
+     * @param certRefs  list of vault entry names identifying the trusted certificates to resolve
+     * @return a map from resolved alias to the corresponding X.509 certificate, or {@code null}
+     *         if no aliases could be resolved
+     */
+    default Map<String, X509Certificate> resolveTrust(
+        List<String> certRefs)
+    {
+        return null;
+    }
```
jfallows: I'd like us to try to solve this without changing the VaultHandler abstraction if possible.
See comments below.

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/stream/TlsServerFactory.java
```diff
@@ -1683,7 +1683,27 @@ private void onDecodeHandshakeFinished(
...
+                }
+            }
+            catch (SSLPeerUnverifiedException ex)
+            {
+                // no client cert presented under mutual: requested
+            }
+
+            final TlsRouteConfig route = binding != null
+                ? binding.resolve(authorization, tlsHostname, tlsProtocol, port, clientCertPresent, clientCertTrustAliases)
+                : null;
```
jfallows: The current VaultHandler abstract maps from aliases to `TrustManagerFactory` which we use to configure the SSLEngine.

Once that handshake completes successfully, we then want to know if we trust the client certificate to follow a specific route.

I _think_ we can do that by creating a different TrustManagerFactory instance (via the VaultHandler) per route, for just the trusted aliases on that route. Then we can assess if the client certificate is trusted by using the TrustManager[] for that route, iterating over it to find the X509TrustManager instances, and checking if the client is trusted, if so follow the route.
For the `mutual:none` case, we just need to check that the client did _not_ present a peer certificate chain.

So in terms of APIs, we can probably just pass the client certificate chain to binding.resolve(...) and let it figure out which routes are valid using the TrustManagerFactory for the trusted routes (mutual:required), and verifying no client certificate chain for the mutual:none routes.

jfallows (changes requested):

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/stream/TlsServerFactory.java
```diff
@@ -1682,8 +1682,20 @@ private void onDecodeHandshakeFinished(
...
 
+            Certificate[] clientCerts = null;
+            try
+            {
+                clientCerts = tlsSession.getPeerCertificates();
+            }
+            catch (SSLPeerUnverifiedException ex)
+            {
+                // no client cert presented under mutual: requested
+            }
```
jfallows: Lets move this to a static method to reduce the noise here regarding handling of the exception.

runtime/binding-tls/src/test/java/io/aklivity/zilla/runtime/binding/tls/internal/streams/ServerIT.java
```diff
@@ -299,6 +300,26 @@ public void shouldRequestMutualAuthentication() throws Exception
...
+    @Test
+    @Configuration("server.mutual.signer.routed.yaml")
+    @Specification({
+        "${net}/server.mutual.cert.absent/client",
+        "${app}/server.mutual.cert.absent/server"})
+    public void shouldRouteWhenClientCertAbsent() throws Exception
+    {
+        k3po.finish();
+    }
+
```
jfallows, line 322: Let's also add a negative test, where the client certificate is not one of the trusted aliases, so even though the TLS handshake succeeds, we reject because there is no valid route.

jfallows (changes requested):

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/stream/TlsServerFactory.java
jfallows: Also, let's only attempt to get the peer certificates if the options mutual is _not_ `none`, i.e. if it is `required` or `requested`. This will avoid triggering a predictable exception in the common case when mutual TLS is not configured.

runtime/binding-tls/src/main/java/io/aklivity/zilla/runtime/binding/tls/internal/stream/TlsServerFactory.java
```diff
@@ -2408,6 +2402,25 @@ private static Optional<TlsServer.TlsStream> nullIfClosed(
         return TlsState.closed(state) ? NULL_STREAM : stream;
     }
 
+    private static Certificate[] peerCertificates(
```
jfallows: Rename to `clientCertificates` and move private static method after private non-static methods.

specs/binding-tls.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tls/streams/network/server.mutual.signer.untrusted/client.rpt
```diff
@@ -0,0 +1,30 @@
+#
```
jfallows, line 1: This should have a peer script, and a corresponding spec IT method.

jfallows (changes requested):

jfallows (approved):

PR #1773, opened 2026-05-16
feat(binding-mcp): emit Alt-Svc response header from server

### LINKED ISSUE #1770: binding-mcp server: emit Alt-Svc header advertising engine service hostname
## Overview

Add support for emitting the HTTP `Alt-Svc` response header from `binding-mcp` server. The advertised authority comes from the engine `Configuration` service hostname. Clients that honor Alt-Svc cache the alternative authority and direct subsequent requests for the same origin to it.

## Motivation

`Alt-Svc` (RFC 7838) is a standard HTTP mechanism for an origin to advertise an alternative service endpoint that clients may use for subsequent requests. Typical use:

- Clients reach the origin through an intermediary (load balancer, ingress, service mesh, etc.).
- Origin responds with `Alt-Svc: h2="<service-hostname>:<port>"; ma=<seconds>`.
- HTTP/2-capable clients cache the alternative and connect to it directly for subsequent requests, bypassing the intermediary.

For zilla deployments where the gateway has a stable, externally-resolvable service hostname distinct from whatever intermediary fronts it (k8s headless service per pod, dedicated DNS per gateway instance, direct DNS records, etc.), this provides a no-cost optimization for clients that honor the standard.

`Alt-Svc` is a hint — clients are free to ignore it — so emitting it has no correctness impact and no downside for clients that don't implement it. The TLS certificate, URL, and origin authority are unchanged from the client's perspective; only the underlying TCP/DNS routing shifts.

### Configuration

The advertised authority is sourced from engine `Configuration`. Approximate shape (refine to existing zilla conventions):

```properties
zilla.engine.service.hostname=<host>
zilla.engine.service.port=<port>
```

If the service hostname is unset, `Alt-Svc` emission is skipped (no header added to responses).

### Emission

On HTTP responses emitted by `binding-mcp` server, include:

```
Alt-Svc: h2="<configured hostname>:<configured port>"; ma=<configured TTL>
```

Where `ma` (max-age, in seconds) is a configurable TTL with a reasonable default (e.g. 24 hours, matching common HTTP/2 server implementations).

If the request arrived at the same authority that would be advertised (i.e. the client is already on the direct path), emission may be skipped as a small optimization.

## Implementation surface

- Read engine service hostname and port from `Configuration` at binding init.
- Construct the `Alt-Svc` value (one-time at init, or per-response if dynamic).
- Insert the header in the response builder path of `binding-mcp` server.
- Configuration option (optional): `ma` value, with a sensible default.

## Acceptance criteria

- [ ] `binding-mcp` server emits `Alt-Svc` header on responses when engine service hostname is configured.
- [ ] When engine service hostname is unset, no `Alt-Svc` header is emitted.
- [ ] `ma` value is configurable with a documented default.
- [ ] Test: response includes `Alt-Svc` with the configured hostname and port.
- [ ] Test: with engine service hostname unset, no `Alt-Svc` header in response.
- [ ] Test (optional): when request authority matches the configured service hostname, `Alt-Svc` is skipped.

jfallows:
## Summary

- Adds RFC 7838 `Alt-Svc` response header emission to `binding-mcp` server. The binding emits a placeholder authority that `binding-http` (already merged via #1775) translates to the wire-level ALPN protocol id, service host, and physical port.
- New binding-mcp config:
  - `zilla.binding.mcp.alt.svc.enabled` (boolean) — per-binding opt-in. Default is driven from the presence of engine `zilla.engine.service.hostname` so operators get sensible behaviour out of the box, and individual bindings can override.
  - `zilla.binding.mcp.alt.svc.max.age` (Duration, default `PT24H`) — the `ma` TTL.
- Adds a `headerMissing()` matcher to `binding-http.spec` `HttpFunctions` so spec scripts can actively assert header absence.

## Emitted value

When enabled, binding-mcp emits per response:

```
Alt-Svc: http=":<logical-port>"; ma=<seconds>
```

The port is taken from the request's `:authority`. Example, for `:authority = localhost:8080`:

```
Alt-Svc: http=":8080"; ma=86400
```

`http=` is a placeholder ALPN identifier; the empty host and the logical port are placeholders for the service host and physical port respectively. `binding-http` (#1775) substitutes these before the header reaches the wire — turning the placeholder into e.g. `h2="backend.example.com:8443"; ma=86400` or `http%2F1.1="..."; ma=...` based on the active request's protocol and the engine's configured service host / `ProxyBeginEx.destination.port`.

## Notes

- The 13 response-begin sites in `McpServerFactory` go through a small `applyAltSvc(builder, altSvc)` helper that no-ops when `altSvc` is `null`.
- `altSvc` is computed once per request in `newStream` (uses the request's `:authority`) and passed into each handler constructor (`McpServer`, `McpAuthCallbackHandler`, `McpShutdownHandler`, `McpEventStream`); `McpRequestStream` reaches `server.altSvc` from its enclosing `McpServer`.
- The existing `lifecycle.initialize` scripts now actively assert `alt-svc` is absent via `.headerMissing("alt-svc")` — so `shouldInitializeLifecycle` doubles as the negative-case IT. New `lifecycle.initialize.alt.svc` scripts cover the positive case with `@Configure(MCP_ALT_SVC_ENABLED_NAME, "true")`.

## Test plan

- [x] `McpConfigurationTest` extended for the two new binding-mcp property name constants.
- [x] `HttpFunctionsTest`: two new tests for `headerMissing()` (match + fail-when-present).
- [x] `binding-mcp` ITs (`McpServerIT`): `shouldInitializeLifecycle` (no Alt-Svc, asserted absent) + new `shouldInitializeLifecycleAltSvc` (Alt-Svc placeholder present). 142 tests pass.
- [x] `binding-mcp.spec` peer-to-peer ITs (`NetworkIT` / `ApplicationIT`): new `shouldInitializeLifecycleAltSvc` in both. 113 tests pass.
- [x] `binding-http.spec` coverage met after adding `headerMissing` tests.
- [x] `./mvnw checkstyle:check` passes on all touched modules.

Closes #1770. The translator side already landed in #1775.

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
@@ -98,6 +98,7 @@ public final class McpServerFactory implements McpStreamFactory
     private static final String HTTP_HEADER_SESSION = "mcp-session-id";
     private static final String HTTP_HEADER_STATUS = ":status";
     private static final String HTTP_HEADER_CONTENT_TYPE = "content-type";
+    private static final String HTTP_HEADER_ALT_SVC = "alt-svc";
     private static final String HTTP_HEADER_ACCEPT = "accept";
     private static final String HTTP_HEADER_LAST_EVENT_ID = "last-event-id";
     private static final String CONTENT_TYPE_JSON = "application/json";
@@ -187,6 +188,8 @@ public final class McpServerFactory implements McpStreamFactory
     private final Supplier<String> supplyElicitationId;
     private final String serverName;
     private final String serverVersion;
+    private final boolean altSvcEnabled;
+    private final long altSvcMaxAgeSeconds;
     private final long inactivityTimeoutMillis;
     private final long sseKeepaliveIntervalMillis;
     private final Signaler signaler;
@@ -245,6 +248,8 @@ public final class McpServerFactory implements McpStreamFactory
         this.supplyElicitationId = config.elicitationIdSupplier();
         this.serverName = config.serverName();
         this.serverVersion = config.serverVersion();
+        this.altSvcEnabled = config.altSvcEnabled();
+        this.altSvcMaxAgeSeconds = config.altSvcMaxAge().toSeconds();
         this.inactivityTimeoutMillis = config.inactivityTimeout().toMillis();
         this.sseKeepaliveIntervalMillis = config.sseKeepaliveInterval().toMillis();
         this.signaler = context.signaler();
@@ -351,6 +356,12 @@ public final class McpServerFactory implements McpStreamFactory
                 .map(h -> h.value().asString())
                 .orElseGet(() -> extractSessionIdFromState(path));
 
+            final String authority = Optional.ofNullable(httpBeginEx.headers()
+                    .matchFirst(h -> HTTP_HEADER_AUTHORITY.equals(h.name().asString())))
+                .map(h -> h.value().asString())
+                .orElse(null);
+            final String altSvc = buildAltSvc(authority);
+
             if (sessionId != null && !isSessionIdAligned(resolvedId, sessionId))
             {
                 newStream = new McpRedirectHandler(
@@ -383,7 +394,8 @@ public final class McpServerFactory implements McpStreamFactory
                         initialId,
                         resolvedId,
                         session,
-                        redirectURI)::onNetMessage;
+                        redirectURI,
+                        altSvc)::onNetMessage;
                 }
                 break;
             case "GET":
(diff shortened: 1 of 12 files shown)


runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -4467,6 +4490,30 @@ private MessageConsumer newStream(
...
+        return result;
+    }
+
+    private HttpBeginExFW.Builder applyAltSvc(
+        HttpBeginExFW.Builder builder,
+        String altSvc)
+    {
+        if (altSvc != null)
+        {
+            builder.headersItem(h -> h.name(HTTP_HEADER_ALT_SVC).value(altSvc));
```
jfallows: Use `builder.inject(...)` with method reference at call site, passing only the builder and using field state in method implementation, instead of wrapping.

specs/binding-http.spec/src/main/java/io/aklivity/zilla/specs/binding/http/internal/HttpFunctions.java
```diff
@@ -388,7 +398,8 @@ private HttpBeginExFW match(
             if (beginEx != null &&
                 matchCompositeId(beginEx) &&
                 matchTypeId(beginEx) &&
-                matchHeaders(beginEx))
+                matchHeaders(beginEx) &&
+                matchMissingHeaders(beginEx))
```
jfallows:
```suggestion
                matchHeadersMissing(beginEx))
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -1278,6 +1291,16 @@ private McpServer(
             this.decoder = decodeJsonRpc;
             this.initialMax = decodeMax;
             this.redirectURI = redirectURI;
+            this.altSvc = altSvc;
+        }
+
+        private void applyAltSvc(
```
```suggestion
        private void injectAltSvc(
```

Also move to end of (inner) class as it is a utility method.
Same feedback for `applyAltSvc` method on other inner classes.

PR #1786, opened 2026-05-21
fix(binding-kafka): immediate retry metadata refresh

ankitk-me:


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientProduceFactory.java
```diff
@@ -2260,6 +2266,23 @@ private void onDecodeProducePartition(
...
                     break;
+                case ERROR_LEADER_NOT_AVAILABLE:
+                case ERROR_NOT_LEADER_FOR_PARTITION:
+                case ERROR_KAFKA_STORAGE_ERROR:
+                {
+                    final LongConsumer metaFlushSignal = clientRoute.metaFlushSignal;
+                    if (metaFlushSignal != null)
+                    {
+                        metaFlushSignal.accept(traceId);
+                    }
```
jfallows: Perhaps it would be simpler to have clientRoute.metaFlushSignal be a no-op instead of null when not set, then just always call it unconditionally instead of testing for null each time.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/KafkaBindingContext.java
```diff
@@ -55,7 +55,7 @@ final class KafkaBindingContext implements BindingContext
         Map<KindConfig, KafkaStreamFactory> factories = new EnumMap<>(KindConfig.class);
         factories.put(CLIENT, new KafkaClientFactory(config, context, this::supplyClientRoute));
         factories.put(CACHE_SERVER, new KafkaCacheServerFactory(config, context, supplyCache,
-            this::supplyCacheRoute));
+            this::supplyCacheRoute, this::supplyClientRoute));
```
jfallows: Why does cache server need client route?

I don't think we can require this as they could be on different workers as they are different bindings in the pipeline.

In general, client-to-client coordination is okay, i.e. in the same kafka client factory, and cache-server-to-cache-server is also okay, i.e. in the same kafka cache server factory, but coordination across factories is outside the abstraction and also not thread safe.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientRoute.java
```diff
@@ -28,6 +30,7 @@ public final class KafkaClientRoute
     public final Int2ObjectHashMap<Int2IntHashMap> partitions;
 
     public volatile long metaInitialId;
+    public volatile LongConsumer metaFlushSignal;
```
jfallows: I'd prefer we avoid the term `signal` here, as it already has a defined meaning.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheServerMetaFactory.java
```diff
@@ -0,0 +1,1172 @@
...
+        {
+            if (!KafkaState.initialOpened(state))
+            {
+                this.reconnectAttempt = 0;
+
+                final long traceId = window.traceId();
+
+                state = KafkaState.openedInitial(state);
+
+                clientRoute.metaFlushSignal = this::doMetaFanoutFlushIfNecessary;
```
jfallows: This cross-pollination from client to cache should be removed, as mentioned in other feedback.

This is triggering a FLUSH back to the meta client anyway, so the reaction can instead be fully self-contained there with the same effect.

jfallows (changes requested):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientFetchFactory.java
```diff
@@ -3223,6 +3230,8 @@ private void cleanupNetwork(
                 doNetworkResetIfNecessary(traceId);
                 doNetworkAbortIfNecessary(traceId);
 
+                clientRoute.metaFlush.accept(traceId);
+
```
jfallows: Why do we need this for transaction abort?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientProduceFactory.java
```diff
@@ -898,7 +900,10 @@ private void onClientFanInitialReset(
             final int error = kafkaResetEx != null ? kafkaResetEx.error() : UNKNOWN_ERROR;
             doClientFanReplyResetIfNecessary(traceId);
 
-            if (reconnectDelay != 0 && !members.isEmpty() && error == ERROR_NOT_LEADER_FOR_PARTITION)
+            if (reconnectDelay != 0 && !members.isEmpty() &&
+                (error == ERROR_LEADER_NOT_AVAILABLE ||
+                 error == ERROR_NOT_LEADER_FOR_PARTITION ||
+                 error == ERROR_KAFKA_STORAGE_ERROR))
```
jfallows: What governs which errors trigger reconnect, i.e. for "recoverable" errors only?

If so, then perhaps we should model an enum of error codes and whether or not they are recoverable, handling unknown errors as a catchall, not recoverable?

jfallows (changes requested):
When this is working, then we need a final pass to provide exponential backoff so we get quick recovery for network blips but no rapid polling for longer distruptions.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientMetaFactory.java
jfallows, line 1: This file used to be one implementation shared by cache client and cache server, but now it is split to add reconnect logic primarily on one side.
Do we want that on the client side, or catch it earlier on the server side to minimize impact?
For example, if we get an error and it propagates meta cache server to meta cache client, then all the meta cache clients now need to reconnect, versus just the meta cache server needing to reconnect.
Are the implementations actually different across meta cache client and meta cache server?

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/ClientFetchIT.java
```diff
@@ -147,16 +147,6 @@ public void shouldRejectOffsetNotAvailable() throws Exception
...
-    @Test
-    @Configuration("client.when.topic.yaml")
-    @Specification({
-        "${app}/partition.not.leader/client",
-        "${net}/broker.not.available/server"})
-    public void shouldRejectBrokerNotAvailable() throws Exception
-    {
-        k3po.finish();
-    }
-
```
jfallows: Is this the change of behavior that fixes the underlying bug?

jfallows (changes requested):

runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/ClientFetchIT.java
ankitk-me: I added this scenario as part of this PR, but latest reviewed that this error is not `RETRIABLE` so remove it. 

```
BROKER_NOT_AVAILABLE	8	False	The broker is not available.
```

jfallows (approved):

PR #1789, opened 2026-05-21
Add support for MCP list changed notifications

### LINKED ISSUE #1776: feat(binding-mcp): notification-driven cache refresh via list_changed
## Summary

When `options.cache` is configured on the `mcp · proxy` binding (landed in #1774, closes #1737), list-method payloads (`tools` / `resources` / `prompts`) are populated at attach (hydrate) and refreshed periodically based on `options.cache.ttl.<kind>`. The MCP protocol also lets the downstream server proactively signal that a method list has changed via these notifications, sent on the lifecycle session:

- `notifications/tools/list_changed`
- `notifications/resources/list_changed`
- `notifications/prompts/list_changed`

A cache-aware proxy should consume those notifications on the hydrate lifecycle session and trigger an immediate refresh for the affected kind, rather than waiting for the next TTL boundary. This narrows the stale window when the downstream is willing to notify.

This was explicitly deferred from #1774's scope — see the "Notification-driven refresh" line in #1774's PR body.

## Proposed behaviour

- `McpHydrateSession` already owns the lifecycle stream to downstream. Extend its `onMessage` / `onData` handler to parse inbound JSON-RPC notifications and recognise the three `*/list_changed` methods.
- On receipt: cancel any pending periodic refresh timer for that kind (`signaler.cancel(cancelId)`) and start a new `McpHydrateListStream` immediately, going through the existing per-kind lease so a single worker still issues the refresh on the wire.
- After completion the regular `scheduleRefresh(kind)` re-arms the periodic timer.
- Aborted notification-driven refresh follows the same recovery as TTL-driven refresh — prior cached entry retained, next periodic refresh fires after `ttl(kind)`.

## Scope

- Runtime: `McpHydrateSession` parses notifications; existing lease, settle, and refresh paths handle the rest unchanged.
- Spec scripts: new scenarios under `specs/binding-mcp.spec/.../streams/application/`:
  - `cache.refresh.tools.notified` / `.resources.notified` / `.prompts.notified` — server sends a `*/list_changed` notification on the lifecycle session; proxy issues an immediate refresh on the wire.
  - `cache.refresh.tools.notified.before.ttl` — notification arrives before TTL elapses; verifies the pending TTL timer is cancelled so only one refresh fires.
- ITs: peer + engine-driven for each of the new scenarios, mirroring the existing `cache.refresh.*` test pairs.

## Out of scope (already landed in #1774)

- Periodic TTL-driven refresh.
- Lease coordination across workers.
- Hydrate / serve / lifecycle gating / `options.cache.authorization`.

## References

- #1737 (parent issue, original cache design).
- #1774 (proxy cache option landing — deferred this feature).
- MCP spec — `notifications/{tools,resources,prompts}/list_changed`.

jfallows:
Fixes #1776

## Description

End-to-end support for MCP `notifications/{tools,prompts,resources}/list_changed`
across all three binding kinds (`mcp · server`, `mcp · client`, `mcp · proxy`),
with cache-coordinated propagation in the cache-enabled proxy and
resumability across multi-route deployments.

## Behavior delivered

**Initialize advertise.** `mcp · server` unconditionally advertises
`capabilities.{tools,prompts,resources}.listChanged: true` in the JSON-RPC
initialize response. The MCP spec treats the flag as advisory (SHOULD), and
Zilla in fact delivers notifications in every deployment shape — passthrough
proxies forward upstream `list_changed`; cache-enabled proxies generate
notifications from their refresh path. Encoding the bits unconditionally
collapses an 8-variant byte-array selector to one constant and removes a
back-reference from `McpLifecycleStream` to `McpServer` that previously had
the initialize encode wait on the lifecycle reply.

**Passthrough proxy.** Upstream `list_changed` FlushEx flows through to the
agent unchanged. Event IDs survive across multi-route deployments via the
aggregate-event-ID scheme introduced in PR #1791.

**Cache-enabled proxy — single-fire, fresh-content-only.** When an upstream
emits `list_changed`, the per-route lifecycle client (one-per-active-agent-
route) does **not** forward it to the agent. Instead the proxy:
1. Records the upstream event id (for resume).
2. Drives a cache refresh via the hydrater.
3. At cache settle, emits exactly one `list_changed` to each connected agent
   session, but **only if the refreshed content differs from the previous
   cached value** (CRC32 diff-gate). TTL-driven refreshes that produce
   identical content are silent.

This guarantees: (a) the agent never sees `list_changed` before the cache is
fresh; (b) the agent never sees a notification for content that did not
change; (c) exactly one notification per upstream signal that produces real
content change, never two.

**Cache-driven notifications carry no event id.** The SSE encoder skips the
entire `id:` line when the FlushEx id has the absent sentinel
(`length() == -1`). Per the SSE spec, an empty `id:` line resets the
client's `lastEventId` — only the *absence* of the field preserves it. This
means the agent's prior upstream-anchored `Last-Event-ID` survives a
cache-driven notification, so reconnect can still resume from upstream.

**Cross-worker fan-out.** The cache-enabled proxy elects a single
"lifecycle owner" worker per binding via `StoreHandler.lock` (introduced in
PR #1790) and holds the lease for the lifetime of the binding via
`StoreHandler.renew` (PR #1792). The owner subscribes to upstream and writes
refreshed content into the shared store; non-owner workers subscribe to
`StoreHandler.watch` (PR #1790) on the same keys, so a settle on one worker
fires the change-detection + per-session broadcaster on every worker — every
agent gets notified regardless of which worker holds the upstream stream.

**Multi-route resumability.** When the proxy has multiple routes (e.g. one
upstream per toolkit), inbound FlushEx event IDs from each per-route
lifecycle client are recorded keyed by `routedId`. Outbound FlushEx ids
sent to the agent are minted as an aggregate using PR #1791's
shortest-unique-prefix scheme over CRC32C(toolkit). On reconnect the agent
sends `Last-Event-ID: =;=`, the proxy decodes it,
and each per-route client issues its own `McpResumeChallengeEx` against
that route's upstream id. Single-route proxies and the hydrater's
self-loop bypass aggregation as a transparent relay.

## Engine dependencies (already merged to `develop`)

- **#1790** `StoreHandler.lock` / `unlock` / `watch` — coordination
  primitives consumed by this PR.
- **#1791** Aggregate event IDs for multi-toolkit MCP proxy — encodes
  per-route resumability; this PR consumes it on the agent-facing side.
- **#1792** `StoreHandler.renew` — lets the lifecycle-owner worker hold
  its lock across the binding's lifetime with periodic lease renewals.

## Key implementation files

**Binding side (`runtime/binding-mcp/.../stream/`):**

- `McpServerFactory` — unconditional `INITIALIZE_RESPONSE_CAPABILITIES`
  constant; SSE encoder `encodeSseNotifyEvent` skips the `id:` line when
  length is -1.
- `McpClientFactory` — parses upstream `capabilities.{...}.listChanged`
  and surfaces them on lifecycle BEGIN_EX; decodes inbound SSE
  `notifications/.../list_changed` into FlushEx.
- `McpProxyLifecycleFactory` — defers upstream list_changed in
  `McpLifecycleClient.onClientFlush` when cache is enabled; mints aggregate
  event IDs on outbound FlushEx; decodes inbound `McpResumeChallengeEx`
  and dispatches per-route resumes; `doNotifyListChanged` emits a FlushEx
  with the id field left as the absent sentinel.
- `cache/McpProxyCache` — CRC32 diff-detect on `McpListCache.put`; per-kind
  lock lifecycle via `StoreHandler.lock/unlock`; `renewLifecycle()` exposes
  the lifecycle lock to the manager for periodic renewal.
- `cache/McpProxyCacheManager` — lifecycle lock held for binding lifetime;
  scheduled `renew` at `leaseTtl/3`; settle dispatch matrix
  `(pending, changed)` → emit-or-drain.
- `cache/McpProxyCacheHydrater` — intercepts upstream list_changed and
  drives `handler.onListChanged(kind)` → cache refresh.

**Config (`runtime/binding-mcp/.../config/`):**

- `McpBindingConfig` — toolkit prefix map for aggregate-id encode/decode;
  validation that multi-route deployments specify `with.toolkit` per route.

**Spec (`specs/binding-mcp.spec/.../`):**

- `McpFunctions` matcher gains null-tolerant semantics on the three
  list_changed FlushEx variants: omitting `.id(...)` matches any id
  (wildcard, unchanged); `.id("X")` matches the exact value; `.id(null)`
  asserts the id field is absent (length=-1) — distinct from wildcard.
- IDL: `McpLifecycleBeginEx.capabilities` widened to `uint16`; new
  capability bits `SERVER_TOOLS_LIST_CHANGED` (64),
  `SERVER_PROMPTS_LIST_CHANGED` (128), `SERVER_RESOURCES_LIST_CHANGED`
  (256).

## Test coverage

**Initialize advertise (no dedicated scenario needed):**

The unconditional advertise is exercised implicitly by every existing
`lifecycle.initialize*` scenario — each now asserts the new
`"prompts":{"listChanged":true},"resources":{"listChanged":true},
"tools":{"listChanged":true}` JSON in its expected response. A regression
that re-couples output to input bits would break those tests first.

**Passthrough list_changed:**

- `lifecycle.notify.{tools,prompts,resources}.list.changed` — single-route
  passthrough, upstream id flows through.
- `lifecycle.notify.tools.list.changed.toolkit.multi` and
  `.toolkit.multi.prefixed` — multi-route, aggregate event ID minting.
- `lifecycle.events.resume.{aggregate,partial,partial.prefixed,
  aggregate.prefixed}` — agent reconnect with aggregate `Last-Event-ID`;
  per-route resume challenge dispatch.

**Cache-driven list_changed:**

- `cache.notify.tools.list.changed` — upstream `list_changed` →
  cache refresh → agent sees one notification with id-absent
  (asserted via `.id(null)`).
- `cache.notify.tools.list.changed.after.tools.call` — agent first invokes
  `tools/call` (forces per-route lifecycle client to exist, exercising the
  defer-not-suppress path), then receives the cache-driven notification.
- `cache.refresh.tools.notify` — TTL-driven refresh with new content
  triggers fan-out; id-absent on the wire.

**ITs:**

- `McpProxyCacheIT` — `shouldRefreshToolsOnListChangedNotification`,
  `shouldEmitOneListChangedAfterAgentInvokesToolsCall`, plus the
  pre-existing cache contention and TTL refresh ITs continue passing.

https://claude.ai/code/session_01Gx5yC2CuFd54Fyoy7kL3qg


runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -131,14 +142,47 @@ public void register(
     void acquireLifecycle(
         Consumer<Boolean> completion)
     {
-        store.putIfAbsent(STORE_LOCK_KEY_LIFECYCLE, STORE_LOCK_VALUE, leaseTtl,
-            prior -> completion.accept(prior == null));
+        store.lock(STORE_LOCK_KEY_LIFECYCLE, leaseTtl, (k, t) ->
+        {
+            lifecycleLockToken = t;
```
jfallows:
```suggestion
            lockToken = t;
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -131,14 +142,47 @@ public void register(
     void acquireLifecycle(
```
jfallows:
```suggestion
    void acquireLock(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -131,14 +142,47 @@ public void register(
...
-        store.putIfAbsent(STORE_LOCK_KEY_LIFECYCLE, STORE_LOCK_VALUE, leaseTtl,
-            prior -> completion.accept(prior == null));
+        store.lock(STORE_LOCK_KEY_LIFECYCLE, leaseTtl, (k, t) ->
+        {
+            lifecycleLockToken = t;
+            completion.accept(t != null);
+        });
     }
 
     void releaseLifecycle(
```
jfallows:
```suggestion
    void releaseLock(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -202,28 +248,74 @@ public void put(
...
         {
-            store.put(storeKey, value, STORE_TTL_FOREVER, completion.andThen(this::checkPut));
+            crc32.reset();
+            crc32.update(value.getBytes(StandardCharsets.UTF_8));
+            final long newChecksum = crc32.getValue();
+            final boolean changed = lastChecksum != -1L && lastChecksum != newChecksum;
+            lastChecksum = newChecksum;
+            store.put(storeKey, value, STORE_TTL_FOREVER, completion.andThen(this::checkPut).andThen(k ->
+            {
+                if (onSettled != null)
```
jfallows: Let's make onSettled null case a no-op instead, so we can avoid this null check and use andThen directly.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCacheHandler.java
```diff
@@ -22,4 +22,7 @@ interface McpProxyCacheHandler
 
     void hydrate(
         int kind);
+
+    void onListChanged(
```
jfallows:
```suggestion
    void onChanged(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCacheListener.java
```diff
@@ -21,5 +21,8 @@ interface McpProxyCacheListener
     void onError(
         int kind);
 
+    void onListChanged(
```
jfallows:
```suggestion
    void onChanged(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCacheManager.java
```diff
@@ -79,6 +91,36 @@ public void stop()
...
+    private void onStoreChanged(
+        int kind)
+    {
+        if (stopped)
+        {
+            return;
+        }
+        cache.caches().get(kind).get((k, v) ->
+        {
+        });
```
jfallows: Define no-op constant and use here instead of lambda.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCacheManager.java
```diff
@@ -79,6 +91,36 @@ public void stop()
...
+    private void closeWatch(
+        int kind)
+    {
+        final Closeable handle = watchHandles[kind];
+        if (handle != null)
+        {
+            watchHandles[kind] = null;
+            try
+            {
+                handle.close();
```
jfallows: Use Agrona CloseHelper.quietClose instead.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -61,10 +63,19 @@ public final class McpProxyCache
...
+    private final CRC32 crc32 = new CRC32();
 
     boolean populated;
+    private String lifecycleLockToken;
 
     Runnable onReady;
+    public OnSettled onSettled;
+
+    @FunctionalInterface
+    public interface OnSettled
```
jfallows:
```suggestion
    public interface ListChangedListener
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -131,14 +142,47 @@ public void register(
...
+        {
+            store.unlock(STORE_LOCK_KEY_LIFECYCLE, token, completion);
+        }
+        else
+        {
+            completion.accept(null);
+        }
+    }
+
+    void renewLifecycle(
```
jfallows:
```suggestion
    void renewLock(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -181,6 +225,8 @@ public final class McpListCache
         private final String storeLockKey;
 
         boolean populated;
```
jfallows: Move populated below private fields with blank line separator.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -202,28 +248,74 @@ public void put(
...
+            else
+            {
+                changed = false;
+            }
             populated = value != null;
             checkReady();
+            if (onSettled != null)
+            {
+                onSettled.accept(kind, changed);
+            }
```
jfallows: Same here, no-op instead of null for onSettled, eliminate null check.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCacheHydrater.java
```diff
@@ -318,6 +339,41 @@ private void onLifecycleAbort(
...
+                listKind = KIND_RESOURCES_LIST;
+                break;
+            default:
+                listKind = -1;
+                break;
+            }
+            if (listKind != -1)
+            {
+                handler.onListChanged(listKind);
+            }
```
jfallows: Assume flushEx is valid.

```suggestion
            final OctetsFW extension = flush.extension();
            final McpFlushExFW flushEx = mcpFlushExRO.wrap(extension.buffer(), extension.offset(), extension.limit());

            switch (flushEx.kind())
            {
            case KIND_TOOLS_LIST_CHANGED:
                handler.onListChanged(KIND_TOOLS_LIST);
                break;
            case KIND_PROMPTS_LIST_CHANGED:
                handler.onListChanged(KIND_PROMPTS_LIST);
                break;
            case KIND_RESOURCES_LIST_CHANGED:
                handler.onListChanged(KIND_RESOURCES_LIST);
                break;
            default:
                break;
            }
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyLifecycleFactory.java
```diff
@@ -558,6 +561,52 @@ private void doServerFlush(
...
+                flushEx = mcpFlushExRW
+                    .wrap(codecBuffer, 0, codecBuffer.capacity())
+                    .typeId(mcpTypeId)
+                    .resourcesListChanged(b -> {})
+                    .build();
+                break;
+            default:
+                flushEx = null;
+                break;
+            }
```
jfallows: Use a switch expression instead of a switch statement.

specs/binding-mcp.spec/src/main/java/io/aklivity/zilla/specs/binding/mcp/internal/McpFunctions.java
```diff
@@ -1102,7 +1103,7 @@ public final class McpToolsListChangedFlushExMatcherBuilder
             public McpToolsListChangedFlushExMatcherBuilder id(
                 String id)
             {
-                this.id = new String16FW(id);
+                this.id = new String16FW(id, StandardCharsets.UTF_8);
```
jfallows: Use static import for UTF_8.

specs/binding-mcp.spec/src/main/resources/META-INF/zilla/mcp.idl
```diff
@@ -14,14 +14,17 @@
...
         SERVER_TOOLS(1),
         SERVER_PROMPTS(2),
         SERVER_RESOURCES(4),
         CLIENT_ROOTS(8),
         CLIENT_SAMPLING(16),
-        CLIENT_ELICITATION(32)
+        CLIENT_ELICITATION(32),
+        SERVER_TOOLS_LIST_CHANGED(64),
+        SERVER_PROMPTS_LIST_CHANGED(128),
+        SERVER_RESOURCES_LIST_CHANGED(256)
```
jfallows: Move the `SERVER_` capabilities together.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCacheManager.java
```diff
@@ -50,14 +57,20 @@ public final class McpProxyCacheManager implements McpProxyCacheListener
         this.signaler = signaler;
         this.hydrateBackoffMs = new long[KIND_SLOTS];
         this.hydrateRetryIds = new long[KIND_SLOTS];
+        this.watchHandles = new Closeable[KIND_SLOTS];
         Arrays.fill(this.hydrateRetryIds, NO_CANCEL_ID);
         this.refreshCancelId = NO_CANCEL_ID;
         this.reconnectCancelId = NO_CANCEL_ID;
+        this.renewCancelId = NO_CANCEL_ID;
```
jfallows:
```suggestion
        this.refreshId = NO_CANCEL_ID;
        this.reconnectId = NO_CANCEL_ID;
        this.renewId = NO_CANCEL_ID;
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCacheManager.java
```diff
@@ -141,6 +190,52 @@ private void scheduleRefresh()
...
+    {
+        if (cache.leaseTtl == null)
+        {
+            return;
+        }
+        cancelLifecycleRenew();
+        // renew at one third of the lease TTL so two consecutive renew failures still leave
+        // headroom before the lock would expire and let another worker take over
+        renewCancelId = signaler.signalAt(
+            Instant.now().plusMillis(cache.leaseTtl.toMillis() / 3L), 0, this::onLifecycleRenew);
```
jfallows: Move calculation to `cache.renewTtl` with duration derived from `cache.leastTtl`, then use direclty here via `cache.renewTtl.toMillis()`.

Also remove comment above.

specs/binding-mcp.spec/src/main/resources/META-INF/zilla/mcp.idl
```diff
@@ -14,11 +14,14 @@
...
     {
         SERVER_TOOLS(1),
         SERVER_PROMPTS(2),
         SERVER_RESOURCES(4),
+        SERVER_TOOLS_LIST_CHANGED(64),
+        SERVER_PROMPTS_LIST_CHANGED(128),
+        SERVER_RESOURCES_LIST_CHANGED(256),
         CLIENT_ROOTS(8),
         CLIENT_SAMPLING(16),
         CLIENT_ELICITATION(32)
```
jfallows: Values for bitmask should be ordered, but grouping server_ together before client_ capabilities.

jfallows (commented):
LGTM

PR #1792, opened 2026-05-23
feat(engine): add renew operation to StoreHandler

jfallows:
## Summary

- Adds `StoreHandler.renew(key, token, ttl, completion)` to the engine SPI for callers that hold a coordination lock longer than its initial TTL — they schedule renewals at an interval shorter than the lease TTL, and a failed renewal signals that ownership has been lost so the caller can surrender state and let a new owner take over.
- Brings the in-tree TestStoreHandler up to the same contract surface as the lock/unlock/watch SPI introduced in #1790, so spec-level ITs can be written against `type: test` without depending on `store-memory` for correctness. TestStoreContext now supplies per-store watcher and lock maps shared across workers (mirroring the existing `supplyEntries` pattern); the new `TestWatcher` record carries the registering worker's signaler so cross-worker notify dispatches listener invocations onto the right thread.
- `store-memory` and `TestStoreHandler` both implement `renew` with an atomic `ConcurrentMap.replace` against the previously-observed lock entry — token match under unexpired TTL returns the original token; mismatch or expiry returns null. Expired entries are evicted opportunistically, matching the unlock cleanup behaviour.

## Commits

- `02630399` `test(engine): share watchers and locks per store across workers in TestStoreHandler` — the cross-worker test infrastructure that the renew SPI tests depend on.
- `aba18122` `feat(engine): add renew operation to StoreHandler` — the SPI method plus store-memory and TestStoreHandler implementations, TestBindingFactory renew assertion, and the `store.renew.yaml` spec config.

## Consumer

This SPI is consumed by `binding-mcp`'s cache lifecycle-lock-hold change (on a separate branch, depends on this PR). That change schedules `renew` at `leaseTtl / 3` for the worker that owns a binding's upstream lifecycle stream, so a single cache owner stays uninterrupted during normal operation and a holding node's crash triggers TTL-bounded takeover by another worker.

## Test plan

- [x] `./mvnw clean verify -pl runtime/store-memory -am` — `MemoryStoreHandlerIT` 5/5 (including new `shouldRenew`).
- [x] `./mvnw clean install -DskipTests -DskipITs -pl runtime/engine -am` — engine compiles after both commits.
- [ ] CI verifies the full reactor.

https://claude.ai/code/session_01Gx5yC2CuFd54Fyoy7kL3qg

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gx5yC2CuFd54Fyoy7kL3qg)_

### DIFF
--- a/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/store/StoreHandler.java
+++ b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/store/StoreHandler.java
@@ -158,6 +158,40 @@ public interface StoreHandler
         Consumer<String> completion);
 
     /**
+     * Extends the TTL of a lock previously acquired via {@link #lock}, provided the supplied token
+     * matches the current holder.
+     * <p>
+     * Used by callers that hold a lock longer than its initial TTL — for example, a singleton
+     * worker that owns a coordination lease for the lifetime of its binding. The caller schedules
+     * renewals at an interval shorter than the lease TTL; if any renewal fails the caller has
+     * lost ownership and must surrender any externally-visible state predicated on holding the
+     * lock.
+     * </p>
+     * <p>
+     * On a successful ownership-checked renewal, the completion receives the supplied token and
+     * the lock's expiry is reset to {@code now + ttl}. On failure — the token does not match the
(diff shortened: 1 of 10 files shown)


jfallows (commented):
LGTM

PR #1794, opened 2026-05-23
feat(guard-identity): add identity guard for unvalidated bearer pass-through

jfallows:
## Summary

Adds a new `identity` guard that captures the inbound credentials string verbatim and re-emits it via `GuardHandler.credentials(sessionId)`. No validation, no options.

Useful when Zilla is chaining a bearer token to an upstream that performs its own validation:

- MCP proxies forwarding GitHub PATs to `github-mcp-server`
- Internal API gateways trusting an upstream IdP
- Sidecar / mesh patterns where the gateway is behind a verified perimeter

The current shipped guard (`guard-jwt`) validates JWT format on capture, so it cannot carry opaque tokens. `identity` fills that gap without introducing new SPI surface — it implements the existing `GuardHandler` contract (`reauthorize` / `credentials` / `deauthorize`) and integrates with any binding that supports `options.authorization.name`, including `binding-mcp` and `binding-http`.

### Behavior

- `reauthorize(credentials)` — interns `credentials` as the session's identity; returns an auto-generated authorized session id
- `credentials(sessionId)` — returns the stored identity verbatim
- `identity(sessionId)` — same string (these are the same value for this guard)
- Sessions are shared per `(context, identity)` and ref-counted; null credentials produce fresh non-shared sessions
- `expiresAt` / `expiringAt` → `EXPIRES_NEVER`; `challenge` → `false`; no `preauthorize`
- Schema rejects any `options` block — the guard is configuration-free (`type: identity` only)

### Usage example

```yaml
guards:
  passthrough:
    type: identity
bindings:
  north_http_server:
    type: http
    kind: server
    options:
      authorization:
        passthrough:
          credentials:
            headers:
              authorization: Bearer {credentials}
    routes:
      - exit: app
  app_client:
    type: mcp
    kind: client
    options:
      authorization:
        name: passthrough
    routes:
      - exit: net
        with:
          headers:
            :scheme: https
            :authority: upstream.example.com
            :path: /mcp
```

The client's `Authorization: Bearer …` header is captured at `north_http_server`, propagated through the stream's authorization id, and re-stamped on the outbound request from `app_client` to the upstream.

## Test plan

- [x] `./mvnw checkstyle:check -pl runtime/guard-identity,specs/guard-identity.spec -am` — clean
- [x] `./mvnw clean install -pl runtime/guard-identity,specs/guard-identity.spec` — passes including ITs and jacoco coverage check (0.82 ratio)
- [x] `IdentityGuardFactoryTest` — factory smoke test
- [x] `IdentityGuardTest` — handler unit tests covering capture/return/refcount, shared-vs-fresh session semantics
- [x] `IdentityIT` — k3po-driven binding integration test exercising the guard via engine spec network/application scripts
- [x] `SchemaTest` — validates a minimal `type: identity` config against the patched schema

---
_Generated by [Claude Code](https://claude.ai/code/session_0184bhWpuAK8AdyQrSFgGQtc)_


jfallows (commented):
LGTM

PR #1796, opened 2026-05-24
feat(binding-mcp): propagate upstream Bearer auth challenges and defer per-request app stream end

### LINKED ISSUE #1795: binding-mcp: propagate upstream auth challenges to inbound client
## Summary

Today `McpClientFactory.onNetBegin` (~line 4081) special-cases only `HTTP 405` (events-unsupported fallback); every other non-2xx upstream response is silently dropped at `mcp.onDecodeResumable(...)` around line 4097. The upstream's `:status` and any `WWW-Authenticate` challenge never reach the inbound client. This blocks the `examples/mcp.proxy` GitHub demo (and any equivalent), where `github-mcp-server` emits `HTTP 403 + WWW-Authenticate: Bearer realm="…", scope="…"` and the inbound client should be able to discover the scopes it needs, mint a PAT, and retry.

## Proposed change

Introduce a new `McpResetEx` extension that carries an MCP-level abstraction of the upstream auth challenge, propagates it as a single RESET-with-extension through `mcp(proxy)` → `mcp(server)`, and re-renders it as an HTTP error response at the `mcp(server) → http(server)` boundary.

### IDL (`specs/binding-mcp.spec/src/main/resources/META-INF/zilla/mcp.idl`)

```idl
union McpResetEx switch (uint8) extends core::stream::Extension
{
    case 0: mcp::stream::McpBearerResetEx bearer;
}

struct McpBearerResetEx
{
    string16 realm = null;
    string16 scope = null;
}
```

- **Union, not flat struct** — sets up `Basic`, `Digest`, custom schemes as additive cases without IDL breakage.
- **Structured, not opaque** — `realm` and `scope` are parsed at the binding-mcp ↔ binding-http boundaries; the MCP wire never sees HTTP syntax. Future non-HTTP MCP transports reuse the same struct.
- **Both fields optional** per RFC 6750 (empty `Bearer` challenge round-trips).

### Frame flow (upstream 403 + Bearer challenge)

1. Upstream emits `HTTP 403 + WWW-Authenticate: Bearer realm="github", scope="repo"`.
2. `http(client)` emits BEGIN on its app reply (existing behavior).
3. `mcp(client).onNetBegin` detects `:status` ∈ {401, 403}, parses Bearer params from `www-authenticate`, builds `McpResetEx.bearer`, emits a single RESET-with-extension on the application initial stream. Resets the net side to release the upstream.
4. `mcp(proxy)` forwards the RESET extension transparently (new extension-bearing overload).
5. `mcp(server)` decodes `bearer`, formats RFC-6750 `WWW-Authenticate: Bearer realm="…", scope="…"`, emits BEGIN-with-HttpBeginEx (`:status: 401` + `www-authenticate`) on the net reply — same shape the existing 405-reject path uses (`McpServerFactory.java:3497`).
6. `http(server)` encodes the HTTP response to the inbound client.

### Why RESET-with-extension, not CHALLENGE

1. The transport-layer event is terminal — the MCP request can no longer be satisfied on this stream. RESET models "this attempt is over"; CHALLENGE models "stay open, I need something from you" (existing uses: `resume`, `suspended`, `elicitCreate`).
2. CHALLENGE forces a teardown coordination problem — both ends would race to close after the challenge is delivered. RESET-with-extension fuses "the why" and "the close" into one atomic frame.

`HttpResetEx` already establishes the precedent at the HTTP layer.

### Why HTTP 401 outbound (not 403)

When the response carries `WWW-Authenticate`, HTTP 401 is the convention-aligned status ("authenticate and retry"). The 403 distinction (authenticated-but-denied) is information the inbound client can't act on differently. Collapsing 401/403 to 401 outbound keeps the IDL free of HTTP-status modeling.

### General failure (any other non-2xx)

Bare RESET (no extension), unchanged from today's behavior. `http(server)` renders this as a generic HTTP error via its existing fallthrough. No IDL or runtime changes needed for that path.

### Holdback invariant

This relies on the existing frame-timing asymmetry: `mcp(client)` is *proactive* about emitting END on its net initial (releases upstream eagerly), while `mcp(server)` *holds back* END on its app initial precisely so reverse signals (today: `McpChallengeEx`; with this change: `McpResetEx`) can flow back on the still-open request stream. No change to that timing.

## Scope

- IDL addition (above)
- `McpFunctions.resetEx()` / `matchResetEx()` builders & matchers (mirroring `challengeEx`)
- `McpClientFactory` — new branch in `onNetBegin` for 401/403; new `doAppReset(..., Flyweight extension)` overload
- `McpProxyFactory` — add RESET propagation (no RESET handling exists today) with extension passthrough
- `McpServerFactory` — handle `McpResetEx` on app initial, emit `:status: 401` + `www-authenticate` via the existing 405-path shape
- Three spec scenarios under `network/` and `application/`: `client.reject.upstream.bearer/`, `proxy.propagate.reset.bearer/`, `server.render.reset.bearer/`
- IT methods in `McpClientIT` / `McpProxyIT` / `McpServerIT`

## Out of scope (deferred)

- Basic, Digest, or custom auth-challenge schemes — additive union cases for a follow-up.
- An OAuth-style guard implementing `preauthorize()` — lives in `zilla-plus/examples/mcp.proxy.oauth` (planned).
- Token exchange (JWT → PAT) — zilla-plus territory.

## Test-first plan

Per `AGENTS.md`, scripts first (with the existing k3po reset-ext limitation acknowledged — see commented-out blocks in `specs/binding-mcp.spec/.../reject.method.not.allowed/{client,server}.rpt`). Where `read zilla:reset.ext` / `reject zilla:reset.ext` k3po primitives aren't yet supported, the spec scenarios will assert the abort + the observable HTTP-side bytes; IT-level Java assertions backfill the rest.

## Motivation

Required by `examples/mcp.proxy` (PR #1793) — the GitHub upstream PAT-discovery demo currently silently drops the upstream's auth challenge, so the inbound client sees nothing actionable.

jfallows:
## Summary

Closes #1795.

- Propagate upstream HTTP Bearer auth challenges (401/403 with `WWW-Authenticate: Bearer`) from `mcp(client)` back to the inbound peer as a `RESET` with `McpResetEx.bearer` (`realm`, `scopes`, `error` per RFC 6750), and render that same `McpResetEx.bearer` at `mcp(server)` as an HTTP 401/403 `WWW-Authenticate: Bearer` response — bidirectional support.
- Establish the architecture invariant that per-request MCP application streams hold initial open until the reply terminates, with end deferred (rather than sent proactively) on cancel/shutdown/normal close.

## Architecture invariants

- **mcp(client) is proactive**: HTTP request `BEGIN` is sent on app `BEGIN`; body is streamed as `DATA` arrives; HTTP request End is sent based on body-completion detection (JSON brace counting for variable-arg methods, immediate for zero-arg methods) — never waits for app `END`.
- **mcp(server) holds initial open**: app-initial stays open until upstream's reply terminates (`END`, `ABORT`, or `RESET`); doesn't close just because inbound HTTP body is done.
- **mcp(proxy) holds initial open**: fan-out routes defer `doClientEnd` until upstream's reply ends, mirroring mcp(server). For cancel: `doClientReset` precedes `doClientEnd` on each route.
- **Both directions independent**: response can flow back before request body finishes; `RESET`-with-extension can flow back at any time while initial is open. This is what allows the bearer auth-challenge `RESET` to propagate cleanly from upstream through mcp(client) / mcp(proxy) to the inbound peer.

### Bearer challenge propagation (commit 1)
- `mcp.idl`: add `McpResetEx` union with `McpBearerResetEx { realm, scopes, error }`.
- `McpFunctions`: builder/matcher for bearer reset ex.
- `McpClientFactory.McpHttpStream.onNetBegin`: on `401`/`403` with `Bearer` `WWW-Authenticate`, parse `realm`/`scope`/`error` per RFC 6750 and emit `RESET` + `McpResetEx.bearer` on the app initial; default `error` to `insufficient_scope` for 403 and `invalid_token` for 401 when upstream doesn't supply it.
- New `lifecycle.initialize.reject.bearer` scripts (network + application) and matching `NetworkIT`/`ApplicationIT`/`McpClientIT` entries.

### Hold-open / proactive refactor (commit 2)
- `McpServerFactory`: defer `stream.doAppEnd` out of `onNetEnd` / `decodeNet` end-of-buffer; close app-initial only when upstream's reply has terminated (`McpRequestStream.onAppEnd` / `onAppAbort`).
- `McpClientFactory`: lift `requestSent` / `paramsBraceDepth` / `paramsInString` / `paramsEscaped` and `tryCompleteRequestBody(data)` from `McpToolsCallStream` to `McpRequestStream`. `McpToolsListStream` / `McpPromptsListStream` / `McpResourcesListStream` override `onAppBeginImpl` to send HTTP request `End` immediately (zero-arg). `McpPromptsGetStream` / `McpResourcesReadStream` override `onAppData` to drive brace counting via `tryCompleteRequestBody`.
- `McpProxyListFactory.onNextClient`: remove the `if (McpState.initialClosed(state))` guard around `doClientEnd`.
- Move `write close` (client) to AFTER `read closed` and `read closed` (server) to AFTER `write close` for tools.call, tools.list, prompts.get, prompts.list, resources.read, resources.list (plus `.10k`, `.100k`, `.with.progress`, `.elicit.*` variants and `lifecycle.shutdown`). `.aborted` variants drop the post-abort close handshake.

### Cancel cleanup (commit 3)
- `McpServerFactory.doAppCancel`: send `doAppReset` (RESET on app-reply) BEFORE `doAppEnd` (END on app-initial). Previously `doAppCancel` only sent RESET, leaving app-initial dangling after the hold-open refactor.

### Universal invariant across all per-request scripts (commit 4)
- 102 `.rpt` files transformed. Per-request blocks identified by `.toolsList()`, `.toolsCall()`, `.promptsGet()`, `.promptsList()`, `.resourcesRead()`, `.resourcesList()` BEGIN extension. Lifecycle blocks (`.lifecycle()`) excluded.
  - `client.rpt`: `write close` moved from before reads to AFTER `read closed`/`read abort`/`read aborted`.
  - `server.rpt`: `read closed` moved from before writes to AFTER `write close`/`write aborted`.
- `McpProxyListFactory`: stop sending `doClientEnd` eagerly in `onNextClient`; send `doClientEnd` from `McpListClient.onClientEnd` (after upstream reply ends) and from `McpListServer.onServerReset` (after `doClientReset`) for cancel propagation.
- `McpProxyCacheHydrater`: stop transitioning to `closingInitial` in `doListHydrateBegin`; send `doListHydrateEnd` from `onListHydrateEnd` (after upstream reply ends).

### Server-side bearer rendering (later commits)
- `McpServerFactory`: render an inbound app `RESET` carrying `McpResetEx.bearer` as an HTTP 401/403 response with `WWW-Authenticate: Bearer` (`doNetRejectBearer` / `doNetBeginRejectedBearer`); status from `McpBearerError` (403 `insufficient_scope`, else 401). The initialize 200 is encoded from `McpLifecycleStream.onAppBegin` (app accept), and the reject challenge is rendered from `onAppReset` gated on `!McpState.replyOpening(server.state)` (HTTP reply not yet begun). New `tools.call.reject.bearer` (401/`invalid_token`) scenario; `lifecycle.initialize.reject.bearer` header now carries `error=` for symmetry; peer scripts reused across client- and server-kind ITs (`McpServerIT`, `McpClientIT`, `NetworkIT`, `ApplicationIT`).
- `McpProxyLifecycleFactory`: relay an upstream bearer `RESET` across fan-out routes — `onClientReset` forwards `reset.extension()` via `onClientBearerReset` → `doServerReset(traceId, extension)` (first bearer wins by `!McpState.replyOpened(state)`, sibling routes aborted), so a bearer challenge traverses a proxy (`5d1283d1`).
- `McpServerFactory` (`McpEventStream`): when a standalone lifecycle SSE GET resumes with `Last-Event-ID` and the upstream app `RESET`s with `McpResetEx.bearer`, render the 401/403 `WWW-Authenticate` on the pending SSE GET reply (`McpEventStream.doNetRejectBearer`, opening a window first so the client reaches `connected`) rather than the already-closed initialize stream; `onAppReset` routes to the pending `sse` when the main reply has already opened. New `lifecycle.events.resume.reject.bearer` scenario (`McpServerIT` + `NetworkIT`/`ApplicationIT`).

### Proxy list reply path (later commits)
- `McpProxyListFactory`: per-direction abort/cancel teardown (no cross-direction fan-out). The upstream client is detached only once `McpState.closed(state)`, via an inline `if (McpState.closed(state)) onClientClosed(traceId);` at each state-close site in `McpListServer` (`onServerEnd`/`onServerAbort`/`onServerReset`/`doServerEnd`/`doServerAbort`). Reply envelope is lazy — `doEncodeFraming` only stages the prelude once items exist or a graceful empty end, so an abort-before-items emits no partial envelope. The three previously-`@Ignore`d held-open `*.aborted` proxy scenarios are now deterministic; `*.aborted` client scripts use passive `read aborted`.
- Reply DATA flows through a single `BufferPool` encode slot on `McpListServer`. `doServerData` is a write-or-stash emit (mirroring `HttpServerFactory.doNetworkData`): writes what the reply window allows directly, stashes the remainder, returns the bytes accepted. `encode(traceId)` is the slot drain (mirroring `decode(traceId)` from the decode side), called from `onServerWindow`. The prelude, per-client separator, postlude, id-prefix injection, and item bytes all flow through `doServerData`, so the drain is the sole producer of reply `DATA`. The decoder pauses on slot space (retaining pending bytes via the existing `decodedItemProgress` cursor in the decode slot), preserving streaming of items larger than the slot capacity.
- Decoder→server dispatch goes through `McpListClient.onDecodedItemBegin` / `onDecodedItemChunk` / `onDecodedItemEnd` rather than reaching through `client.server.doEncodeX` pointers.

## Test plan

- [x] `specs/binding-mcp.spec` — peer-to-peer scripts consistent (`NetworkIT` 59, `ApplicationIT` 94, plus `ProxyCacheIT`, `SchemaTest`, `McpFunctionsTest`)
- [x] `runtime/binding-mcp` — all green
  - `McpServerIT` 57/57 ✓ (incl. server-side bearer rendering + lifecycle SSE resume bearer)
  - `McpClientIT` 50/50 ✓ (`shouldCallToolWithProgressResume` now passing 6/6)
  - `McpProxyIT` 44/44 ✓ (held-open abort/cancel scenarios deterministic; previously-`@Ignore`d abort tests re-enabled; encode-slot refactor behavior-preserving)
  - `McpProxyCacheIT` 24/24 ✓
  - `McpProxyLifecycleIT` 7/7 ✓

https://claude.ai/code/session_01BYNJLhjKpXGDVtuNgYysgy

---
_Generated by [Claude Code](https://claude.ai/code/session_01BYNJLhjKpXGDVtuNgYysgy)_


runtime/binding-mcp/src/test/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientIT.java
```diff
@@ -64,6 +64,16 @@ public void shouldInitializeLifecycle() throws Exception
         k3po.finish();
     }
 
+    @Test
+    @Configuration("client.yaml")
+    @Specification({
+        "${app}/lifecycle.initialize.reject.bearer/client",
+        "${net}/lifecycle.initialize.reject.bearer/server"})
+    public void shouldRejectLifecycleInitializeOnUpstreamBearerChallenge() throws Exception
```
jfallows:
```suggestion
    public void shouldRejectLifecycleInitializeWithBearerChallenge() throws Exception
```

runtime/binding-mcp/src/test/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientIT.java
```diff
@@ -97,6 +107,7 @@ public void shouldPingLifecycle() throws Exception
         k3po.finish();
     }
 
+
```
jfallows:
```suggestion
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -1424,7 +1424,6 @@ private void onNetEnd(
                 stream.elicitationId == null)
             {
                 state = McpState.closedInitial(state);
-                stream.doAppEnd(traceId, authorization);
```
jfallows, line 1427: I think we want to keep this, but add a check for `McpState.replyClosed(state)` to `doAppEnd(...)`, to handle race condition where proactive HTTP request at MCP client allows reply to complete before we observe the HTTP request end.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -1747,7 +1746,6 @@ private void decodeNet(
...
```
jfallows, line 1750: This check should be the same as in `onNetEnd`.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -2356,6 +2402,11 @@ private abstract class McpRequestStream extends McpStream
         final int requestId;
         HttpEventStream eventStream;
 
+        boolean requestSent;
```
jfallows: Can this boolean be modeled instead using McpState.initialOpening/initialOpened?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -2619,6 +2670,81 @@ void onDecodeCompletion(
...
+                {
+                    requestSent = true;
+                    http.doEncodeRequestEnd(data.traceId(), data.authorization());
+                    completed = true;
+                }
+            }
+            return completed;
+        }
+
+        boolean advanceParamsBraceDepth(
```
jfallows: Use StreamingJsonParser instead of manual payload inspection.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -3221,12 +3333,26 @@ private void onNetBegin(
...
 
             final OctetsFW ext = begin.extension();
+            String status = null;
+            String wwwAuthenticate = null;
             if (ext.sizeof() > 0)
             {
                 final HttpBeginExFW httpBeginEx = httpBeginExRO.tryWrap(
                     ext.buffer(), ext.offset(), ext.limit());
                 if (httpBeginEx != null)
                 {
```
jfallows: .tryWrap handles empty extension, returns null - always attempt tryWrap, remove redundant if statement for ext size.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -3221,12 +3333,26 @@ private void onNetBegin(
...
+                    final HttpHeaderFW statusHeader = httpBeginEx.headers()
+                        .matchFirst(h -> HTTP_HEADER_STATUS.equals(h.name().asString()));
+                    if (statusHeader != null)
+                    {
+                        status = statusHeader.value().asString();
+                    }
```
jfallows:
```suggestion
                    final String status = Optional.ofNullable(httpBeginEx.headers()
                        .matchFirst(h -> HTTP_HEADER_STATUS.equals(h.name().asString())))
                        .map(HttpHeaderFW::value)
                        .map(String16FW::asString)
                        .orElse(null);
```
same style change for other headers

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -3238,9 +3364,31 @@ private void onNetBegin(
...
+                wwwAuthenticate != null &&
+                startsWithIgnoreCase(wwwAuthenticate, BEARER_PREFIX))
+            {
+                final String realm = extractFirstGroup(BEARER_REALM_PATTERN, wwwAuthenticate);
+                final String scopes = extractFirstGroup(BEARER_SCOPE_PATTERN, wwwAuthenticate);
+                String error = extractFirstGroup(BEARER_ERROR_PATTERN, wwwAuthenticate);
+                if (error == null)
+                {
+                    error = STATUS_403.equals(status) ? BEARER_ERROR_INSUFFICIENT_SCOPE : BEARER_ERROR_INVALID_TOKEN;
+                }
```
jfallows: There are 3 different error codes supported by RFC 6750; `insufficient_scope`, `invalid_token` and `invalid_request`. Let's model all 3 explicitly via mcp.idl enum, and ignore non-standard errors, tighten up the parsing accordingly.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -106,10 +109,21 @@ public final class McpClientFactory implements McpStreamFactory
...
     private static final String HTTP_HEADER_LAST_EVENT_ID = "last-event-id";
+    private static final String STATUS_401 = "401";
+    private static final String STATUS_403 = "403";
     private static final String STATUS_405 = "405";
+    private static final Pattern BEARER_REALM_PATTERN =
+        Pattern.compile("realm\\s*=\\s*\"([^\"]*)\"");
+    private static final Pattern BEARER_SCOPE_PATTERN =
+        Pattern.compile("scope\\s*=\\s*\"([^\"]*)\"");
+    private static final Pattern BEARER_ERROR_PATTERN =
+        Pattern.compile("error\\s*=\\s*\"([^\"]*)\"");
```
jfallows: Use a single pattern, with named groups for `realm`, `scope` and `error` in any order, where all are optional.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -3238,9 +3364,31 @@ private void onNetBegin(
...
             }
 
-            mcp.onNetBegin(begin);
+            if ((STATUS_401.equals(status) || STATUS_403.equals(status)) &&
+                wwwAuthenticate != null &&
+                startsWithIgnoreCase(wwwAuthenticate, BEARER_PREFIX))
+            {
+                final String realm = extractFirstGroup(BEARER_REALM_PATTERN, wwwAuthenticate);
+                final String scopes = extractFirstGroup(BEARER_SCOPE_PATTERN, wwwAuthenticate);
+                String error = extractFirstGroup(BEARER_ERROR_PATTERN, wwwAuthenticate);
```
jfallows: Create a reusable Matcher via <pattern>.matcher("") and then reuse via `.reset(input).matches()` to verify it matches the pattern. Then access each optional named group by name, and default `error` based on status code if missing.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -441,6 +441,7 @@ private void onClientEnd(
             {
                 state = McpState.closedReply(state);
                 cleanupClientSlot();
+                doClientEnd(traceId);
```
jfallows: Do we need to be proactive on sending END here, or can we instead rely on reacting to inbound onServerEnd for the proxy case?

Similar feedback for other proxy streams. However, if proxy is truly the origin of the stream, then it would make sense for it to manage the proactive end request as there would be no proxied stream to otherwise trigger it.

jfallows: McpProxyListFactory is the origin of the per-route client streams (fan-out: one inbound server, N outbound clients), so per your own caveat the proactive `doClientEnd` here is the right shape — there is no upstream-of-the-proxy stream to drive the close via 1-1 propagation. The symmetric proactive `doClientAbort` was added in `onClientAbort`/`onServerReset` for the same reason on the unclean-close paths.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Deferred to a follow-up — incrementally feeding `StreamingJsonParser` across multiple inbound `DATA` frames on the request side needs additional plumbing (a dedicated `DirectBufferInputStreamEx` per request stream so it doesn't collide with the response decoder's shared `inputRO`, and verification that PATH_INCLUDES filtering doesn't suppress the `START_OBJECT`/`END_OBJECT` events we need for depth tracking). Happy to do it in a separate PR if you'd prefer that over leaving the manual brace counter in this one.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
jfallows: I understand the relationship is 1-many, for server-to-client(s), not 1-1, but isn't the lifetime equivalent, i.e. the inbound server END frame is deferred (by mcp invariant) same as for each of the 1-many related client streams, then when inbound server END frame arrives, each of the client outbound END frames can be sent (assuming replyClosed guard succeeds), also triggered at the codepath here, but still gated by replyClosed guard, agree?

jfallows: Agreed. Added the `replyClosed` guard to `doClientEnd` in `18bec1fe`, mirroring the McpServerFactory pattern. The proactive `onClientEnd` call site already sets `closedReply` before invoking it (guard passes trivially), and the `McpListServer.onServerEnd` 1-1 propagation path now defers each route's initial close until that route's reply has terminated — same hold-open semantics as mcp(server).

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Done in `1c7fe89a`. Each `McpRequestStream` lazily creates a `JsonParser` from a dedicated factory whose `PATH_INCLUDES = ["$.NEVER_MATCH"]` — that makes every VALUE non-readable, which is what lets the tokenizer's internal `resumeOp` carry parse state across `DATA` frame boundaries without needing the InputStream to buffer unread bytes across `wrap()` calls. Depth tracks `START_OBJECT`/`START_ARRAY`/`END_OBJECT`/`END_ARRAY`; the body is complete when it returns to 0. A shared per-worker `paramsInputRO` is rewrapped fresh on each `onAppData`, mirroring the response decoder's pattern. Test suite: 174/175 (`shouldCallToolWithProgressResume` is the pre-existing one called out in the PR description).

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -221,7 +221,9 @@ public final class McpClientFactory implements McpStreamFactory
     private static final int SSE_AFTER_COLON = 2;
     private static final int SSE_SMALL_VALUE = 3;
     private static final int SSE_IGNORE_VALUE = 4;
-    private final JsonParserFactory parserFactory;
+    private final JsonParserFactory responseParserFactory;
```
jfallows, line 225: Put a blank line between static and non-static fields for readability.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -254,9 +256,12 @@ public McpClientFactory(
         this.signaler = context.signaler();
         this.clientName = config.clientName();
         this.clientVersion = config.clientVersion();
-        this.parserFactory = StreamingJson.createParserFactory(Map.of(
+        this.responseParserFactory = StreamingJson.createParserFactory(Map.of(
             StreamingJson.PATH_INCLUDES, CLIENT_JSON_PATH_INCLUDES,
             StreamingJson.TOKEN_MAX_BYTES, decodeMax));
+        this.requestParserFactory = StreamingJson.createParserFactory(Map.of(
+            StreamingJson.PATH_INCLUDES, List.of("$.NEVER_MATCH"),
+            StreamingJson.TOKEN_MAX_BYTES, encodeMax));
```
jfallows, line 265:
```suggestion
        this.requestParserFactory = StreamingJson.createParserFactory(Map.of(
            StreamingJson.PATH_INCLUDES, List.of(),
            StreamingJson.TOKEN_MAX_BYTES, encodeMax));
```
Does this work with an empty list instead?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -2681,53 +2685,38 @@ boolean tryCompleteRequestBody(
...
-            final DirectBuffer buf = payload.buffer();
-            final int end = payload.limit();
-            int depth = paramsBraceDepth;
-            boolean inStr = paramsInString;
-            boolean esc = paramsEscaped;
+            requestInputRO.wrap(payload.buffer(), payload.offset(), payload.sizeof());
+            if (paramsParser == null)
+            {
+                paramsParser = requestParserFactory.createParser(requestInputRO);
+            }
```
jfallows: What if we get multiple DATA frames of params? How is the payload fed into the params parser?
Seems like this needs a decoder abstraction similar to response, but for request to properly model multiple DATA frames and detect end of payload.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows, line 2714: This overridden method style is not so consistent with the rest of the codebase, perhaps propose an alternative approach that is more consistent.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Fixed in `339a3045`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Done in `339a3045` — but it required distinguishing two cases in `common-json` first. The old default folded `null` and `List.of()` together to mean "include all", so an empty list couldn't express "include nothing". Now `StreamingJsonParser.pathList` lets `null` propagate, `compilePaths(null)` returns `null`, and `currentPathReadable` treats `pathIncludes == null` as include-all; an empty (or non-null) list now restricts to its members, with `List.of()` cleanly meaning "include none". `createParser(InputStream)`/`createParserFactory(Map.of())` paths continue to omit `PATH_INCLUDES`, so the existing `StreamingJsonParserTest` cases that depend on the include-all default remain unchanged (79/79 in `common-json`, 175/175 in `binding-mcp`).

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Multi-frame works without buffer-stitching because the parser is created lazily on the first DATA frame and held across calls on `paramsParser`. With `PATH_INCLUDES = List.of()` (now meaning "include none") every value is non-readable, so the tokenizer never needs to mark/reset across `wrap()` calls — its internal `resumeOp` carries parse state between frames purely from event-level token transitions. Each `onAppData` rewraps `requestInputRO` with the new payload and the parser picks up exactly where it left off. This is validated end-to-end by the `tools.call.with.10k.params` / `.100k.params` ITs, whose request bodies span many `DATA` frames and reach the closing brace correctly.

That said, an explicit decoder abstraction along the lines of `HttpResponseDecoder` would be overkill here since there's only one parse state (count depth, finish at 0). If you'd prefer the consistency, I can wrap `advanceParamsParser` in a `@FunctionalInterface RequestBodyDecoder` field set in the constructor — happy to do that, or to keep the current method-on-stream form since it's a single state.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -269,7 +269,7 @@ private void doClientBegin(
         private void doClientEnd(
             long traceId)
         {
-            if (!McpState.initialClosed(state))
+            if (McpState.replyClosed(state) && !McpState.initialClosed(state))
```
jfallows:
```suggestion
           if (!McpState.initialClosed(state) &&
               McpState.replyClosed(state))
```          
Primary condition should be first for readability.
Same feedback for other doXxxEnd methods.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -441,6 +441,7 @@ private void onClientEnd(
             {
```
jfallows: We don't need guards like this in onXxx methods, those are enforced for us by the engine, so frames cannot be sent on the same stream after that stream direction closes.

Same feedback about guards to all onXxx methods.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: No doubt we'll be adding other features that rely on the json decoding structure over time, so let's just add the standard decoder pattern here instead of inventing a different one.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Done in `325ee9e1` — removed the `replyClosed` guards in `onClientEnd`, `onClientAbort`, and `onClientReset`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Reordered in `325ee9e1`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Done in `81adb691`. Added `@FunctionalInterface HttpRequestDecoder` mirroring `HttpResponseDecoder`, with `decodeJsonRpcParamsBody` and `decodeRequestEnd` as decode method references on the factory. `McpRequestStream` gains a `requestDecoder` field; the parametrised subclasses (`McpToolsCallStream`, `McpPromptsGetStream`, `McpResourcesReadStream`) initialise it to `decodeJsonRpcParamsBody` in their constructors. The body-completion logic is unchanged — depth-track `START_OBJECT`/`END_OBJECT` events on the per-stream `JsonParser`, transition to `decodeRequestEnd` and emit HTTP request End on `depth == 0` — but now expressed via the decoder strategy so future per-request decode states can be added consistently.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Cleaned up in `3da74de5` — removed the redundant `ext.sizeof() > 0` guards before `tryWrap` in the two remaining `onAppChallenge` handlers (McpRequestStream and McpLifecycleStream) and in `HttpInitializeStream.onNetBegin` and `HttpEventStream.onNetBegin`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: This thread is marked outdated by the surrounding refactors but unresolved. The `deferAppEnd` override is still here as a three-tier polymorphic boolean: `McpStream` returns false (don't defer), `McpRequestStream` returns `McpState.initialOpened(state)` (defer while body still streaming), `McpToolsCallStream` adds the elicit-cleanup side effect when `pendingAuth`.

Could you clarify the alternative you have in mind? Candidates I see: (a) inline the state check at the `onAppEnd` call site and lift the elicit-cleanup into a separate hook on `McpToolsCallStream`, (b) replace it with a `Runnable onAppEndAction` field set per stream (mirror of the new `requestDecoder` field shape from `81adb691`), or (c) fold "request body sent" into the existing decoder-state transition (`requestDecoder == decodeRequestEnd`) and reduce `deferAppEnd` to a one-liner without overrides. Happy to take any of these — your preference?

---

jfallows: (c) seems to be cleanest and most consistent with our existing design

jfallows: Done in `f11bb5a4`. Lifted `requestDecoder` up from `McpRequestStream` to `McpStream`, and `deferAppEnd` is now a single base implementation:

```java
return requestDecoder == decodeRequestEnd;
```

Both overrides are gone. The synchronous request-end emitters (`McpLifecycleStream.onAppBeginImpl`, the zero-arg request streams' `onAppBeginImpl`, the authorised elicit path in `onElicitCompleted`) now transition the decoder alongside `doEncodeRequestEnd`. `McpToolsCallStream` now overrides `onAppEnd` for the `pendingAuth` cleanup, sets `requestDecoder = decodeRequestEnd` to advance the state, then delegates to `super.onAppEnd` which trivially defers. Elicit timeout / failed paths transition the decoder too so `super.onAppEnd`'s `deferAppEnd` doesn't fire a second request end.

152/152 non-cache ITs green.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -1394,6 +1475,7 @@ private abstract class McpStream
         protected HttpStream http;
         protected String credentials;
         protected int serverCapabilities = SERVER_CAPABILITIES;
+        protected HttpRequestDecoder requestDecoder;
```
jfallows:
```suggestion
        protected McpRequestDecoder decoder;
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -1683,7 +1766,7 @@ boolean deferAppEnd(
             long traceId,
             long authorization)
         {
-            return false;
+            return requestDecoder == decodeRequestEnd;
```
jfallows: Can this be inlined instead?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -2200,18 +2281,14 @@ void onNetBegin(
...
+                        .matchFirst(h -> HTTP_HEADER_SESSION.equals(h.name().asString()));
+                    if (sessionHeader != null)
                     {
-                        final HttpHeaderFW sessionHeader = httpBeginEx.headers()
-                            .matchFirst(h -> HTTP_HEADER_SESSION.equals(h.name().asString()));
-                        if (sessionHeader != null)
-                        {
-                            responseSessionId = sessionHeader.value().asString();
-                        }
+                        responseSessionId = sessionHeader.value().asString();
```
jfallows: Use Optional.ofNullable(...).map(...).map(...).orElse(null) approach used in other similar methods.
Same feedback for other similar header code blocks.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -2451,40 +2528,37 @@ void onAppChallenge(
...
+                        eventStream = new HttpEventStream(this, prefixedId);
+                        eventStream.doNetStart(traceId, authorization);
                     }
+                    break;
+                case McpChallengeExFW.KIND_SUSPENDED:
+                    if (http != null)
+                    {
+                        http.doNetReset(traceId, authorization);
+                    }
+                    if (eventStream != null)
```
jfallows: Rename `eventStream` to `sse` as a parallel of `http`.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -269,7 +269,7 @@ private void doClientBegin(
         private void doClientEnd(
             long traceId)
         {
-            if (McpState.replyClosed(state) && !McpState.initialClosed(state))
+            if (!McpState.initialClosed(state) && McpState.replyClosed(state))
```
jfallows: Use multiple lines, break after &&.

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/StreamingJsonTokenizer.java
jfallows, line 130: Let's implement this using Optional.ofNullable(...).map(<helper method>).orElse(null).
Then helper method can optimize for empty list and use paths.stream().map(<item helper method).toList().

runtime/common-json/src/test/java/io/aklivity/zilla/runtime/common/json/internal/StreamingJsonTokenizerPathTest.java
```diff
@@ -235,7 +235,7 @@ public void shouldNotThrowWhenExcludedValueExceedsTokenMaxBytes() throws IOExcep
     {
         final List<String> excludes = List.of("/payload");
         final StreamingJsonTokenizer tokenizer =
-            new StreamingJsonTokenizer(List.of(), excludes, 8);
+            new StreamingJsonTokenizer(null, excludes, 8);
```
jfallows: Add a constant on StreamingJsonTokenizer for INCLUDE_ALL and pass instead of `null` so test is more readable.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Applied in `0070c60d`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Inlined in `0070c60d` — the call site at `McpStream.onAppEnd` now reads `if (decoder != decodeRequestEnd) { ... }` and the helper method is gone.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Done in `0070c60d`. Verified the other tryWrap-followed-by-headers blocks (`HttpEventStream.onNetBegin` status header) already use this pattern.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Renamed in `0070c60d` — field `eventStream` → `sse` and accessor `eventStreamRef()` → `sseRef()`. The class name `HttpEventStream` stays.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
jfallows: Done in `0070c60d`.

---

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/StreamingJsonTokenizer.java
jfallows: Done in `0070c60d` — `compilePaths` now uses `Optional.ofNullable(paths).map(StreamingJsonTokenizer::compilePathList).orElse(null)`, with `compilePathList` doing the empty-list short-circuit and a `paths.stream().map(StreamingJsonTokenizer::compilePath).toList()` for the non-empty case. The per-item compile body lives in `compilePath`.

---

runtime/common-json/src/test/java/io/aklivity/zilla/runtime/common/json/internal/StreamingJsonTokenizerPathTest.java
jfallows: Added `StreamingJsonTokenizer.INCLUDE_ALL` and updated the test to use it in `0070c60d`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -1237,10 +1232,6 @@ private void onNextClient(
             }
             client = new McpListClient(this, route.resolvedId(), route.prefix());
             client.doClientBegin(traceId);
-            if (McpState.initialClosed(state))
-            {
-                client.doClientEnd(traceId);
-            }
         }
 
         private void streamItemBegin(
```
jfallows: check naming for consistency, should this be doEncodeBeginItem?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
jfallows: check naming for consistency, should this be doEncodeItemChunk?

jfallows: check naming for consistency, should this be doEncodeEndItem?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -4034,17 +4035,14 @@ private void doAppEnd(
             long traceId,
             long authorization)
         {
-            if (!McpState.initialClosed(state))
+            if (McpState.replyClosed(state) && !McpState.initialClosed(state))
```
```suggestion
            if (!McpState.initialClosed(state) &&
                McpState.replyClosed(state))
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -4449,6 +4448,8 @@ private void onAppEnd(
             final long traceId = end.traceId();
             final long authorization = end.authorization();
 
+            state = McpState.closedReply(state);
+
             if (eventStream != null)
```
```suggestion
            if (sse != null)
```

Same feedback for other classes, rename eventStream to sse.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -331,6 +354,86 @@ int decode(
...
+        int limit)
+    {
+        DirectBufferInputStreamEx input = requestInputRO;
+        input.wrap(buffer, progress, limit - progress);
+        if (stream.paramsParser == null)
+        {
+            stream.paramsParser = requestParserFactory.createParser(input);
+        }
+        final JsonParser parser = stream.paramsParser;
+        boolean done = false;
```
jfallows: Does it make sense to mark the stream as "closing" via McpState here, instead of the `done` flag?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
jfallows: Renamed in `c8296d4f` to `doEncodeBeginItem`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
jfallows: Renamed in `c8296d4f` to `doEncodeItemChunk`.

---

jfallows: Renamed in `c8296d4f` to `doEncodeEndItem`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
jfallows: Done in `c8296d4f`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
jfallows: Renamed `eventStream` → `sse` and `eventStreamRef()` → `sseRef()` throughout `McpServerFactory` in `c8296d4f`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
jfallows: Done in `c8296d4f`. Replaced the local `done` boolean with a guard on the decoder transition itself: the loop now reads `while (stream.decoder != decodeRequestEnd && parser.hasNext())`. The terminal `stream.decoder = decodeRequestEnd` assignment inside the `END_OBJECT`/`END_ARRAY` branch is what exits the loop, so the extra flag is gone.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -1259,13 +1259,29 @@ private void onNextClient(
             client.doClientBegin(traceId);
         }
 
+        private void encode(
+            long traceId)
+        {
+            if (McpState.replyClosed(state))
+            {
+                return;
+            }
```
jfallows: When we transition to replyClosed, the encoder should be set to encodeIgnore, then this guard can be removed.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -1276,6 +1292,14 @@ private int doEncodeItemChunk(
...
             long traceId)
         {
+            if (encoder != encodeItems)
+            {
+                encode(traceId);
+                if (encoder != encodeItems)
+                {
+                    return 0;
+                }
+            }
```
jfallows: This suggests we need to introduce the encoder slot, so we can stage the bytes into the slot buffer, up to encodeMax (= bufferPool.slotCapacity()) and then write out based on available client window, instead of trying to have both an encoder abstraction and also write out these items directly.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
jfallows: Done in `65a60a68`. `doServerEnd` now sets `encoder = encodeIgnore` on the reply-close transition (matching `doServerAbort` and `onServerReset`), so once `replyClosed` the `encode()` loop reaches the terminal `encodeIgnore` and emits nothing. Removed the `if (McpState.replyClosed(state)) return;` guard from `encode()`.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
jfallows: Agreed this is the right end state — fold prelude/items/separator/postlude into a single encode-slot (staged up to `bufferPool.slotCapacity()`, drained to the wire per client window) so the encoder owns all output instead of mixing the state machine with direct `doEncodeItemChunk` writes. Deferring it to a follow-up rather than bundling into this PR, since it's a sizable refactor of the decode→encode item path (the JSON decoder would write items into the slot instead of straight to the wire) and warrants its own review. The current per-direction teardown + lazy-prelude fix here is complete and the abort/cancel scenarios are deterministic; the slot refactor can land on top without changing observable behavior.

---

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
```diff
@@ -156,6 +156,20 @@ else if (identifier != null)
         return resolved;
     }
 
+    public List<Long> resolveAllLifecycle(
```
jfallows:
```suggestion
    public List<Long> resolveAll(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyLifecycleFactory.java
```diff
@@ -262,6 +266,18 @@ private void onDecodeAggregateEventId(
...
 
+        private void onResumeConfiguredRoutes(
+            long traceId)
+        {
+            for (McpAggregateRoute route : binding.aggregateRoutes)
+            {
+                if (!clients.containsKey(route.routedId()))
+                {
+                    supplyClient(route.routedId()).doClientBegin(traceId);
+                }
```
jfallows:
```suggestion

                final long routedId = route.routedId();
                if (!clients.containsKey(routedId))
                {
                    final McpLifecycleClient client = supplyClient(routedId);
                    client.doClientBegin(traceId);
                }
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyLifecycleFactory.java
```diff
@@ -262,6 +266,18 @@ private void onDecodeAggregateEventId(
             }
         }
 
+        private void onResumeConfiguredRoutes(
```
jfallows:
```suggestion
        private void onServerResumeRoutes(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyLifecycleFactory.java
```diff
@@ -359,12 +376,69 @@ private void onServerBegin(
...
+        }
+
+        private void onClientBearerReset(
+            long traceId,
+            OctetsFW extension,
+            McpLifecycleClient origin)
+        {
+            if (!bearerRelayed && !McpState.replyOpened(state))
+            {
+                bearerRelayed = true;
```
jfallows: Do we actually need this flag?
If we abort the initial stream of each client, then we cannot get a client reset on the initial stream of any client.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -1237,56 +1265,180 @@ private void onNextClient(
...
+                final MutableDirectBuffer slot = bufferPool.buffer(encodeSlot);
+                accepted = Math.min(length, slot.capacity() - encodeSlotOffset);
+                slot.putBytes(encodeSlotOffset, buffer, offset, accepted);
+                encodeSlotOffset += accepted;
+                flushEncodeSlot(traceId);
+            }
+            return accepted;
+        }
+
+        private void flushEncodeSlot(
```
jfallows:
```suggestion
        private void encode(
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -376,7 +377,7 @@ private void onClientBegin(
 
             state = McpState.openedInitial(state);
 
-            flushClientWindow(traceId, 0L, 0, 0L, 0);
+            flushClientWindow(traceId, 0L, 0, 0L, bufferPool.slotCapacity());
```
jfallows:
```suggestion
            flushClientWindow(traceId, 0L, 0, 0L, decodeMax);
```

Make `decodeMax` a field on factory.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -508,11 +504,8 @@ private void onClientReset(
             assert initialAck <= initialSeq;
 
             state = McpState.closedInitial(state);
-            if (!McpState.replyClosed(state))
-            {
-                state = McpState.closedReply(state);
-                server.onClientError(traceId);
-            }
+            state = McpState.closedReply(state);
```
jfallows: This is not correct, reply has not closed yet.
Do we need to propagate RESET to server, then when server gets onServerReset, propagate RESET to client, then client is closed both initial and reply?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -822,7 +815,7 @@ private int decodeItemStart(
             {
             case START_OBJECT:
                 client.decodedItemProgress = decodedItemProgress - 1;
-                client.server.streamItemBegin(traceId);
+                client.server.doEncodeBeginItem(traceId);
```
jfallows: Usually we don't chase pointers like this.
Instead we call an onDecodedXxx method on client, and client then calls server.doXxx in reaction.

Same for encode item chunk etc.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -1133,6 +1144,7 @@ private void onServerEnd(
             {
                 client.doClientEnd(traceId);
             }
+            detachClientIfClosed();
```
jfallows: This feels wrong.

Instead each time we update the replyClosed or initialClosed state, we check for if closed and then call onClientClosed, which can do the client detach as needed without this special case.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -1237,56 +1265,180 @@ private void onNextClient(
...
+            DirectBuffer buffer,
+            int offset,
+            int length,
+            long traceId)
+        {
+            int accepted;
+            if (encodeSlot == NO_SLOT)
+            {
+                final int replyWin = replyMax - (int) (replySeq - replyAck) - replyPad;
+                final int emit = Math.min(Math.max(replyWin, 0), length);
```
jfallows: In general, we use `length` for this typically.
Replace similar occurrences of `emit` with `length` in this and other map factory classes.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -2391,15 +2494,15 @@ void markEventsUnsupported()
...
         @Override
-        HttpEventStream eventStreamRef()
+        HttpEventStream sseRef()
         {
-            return eventStream;
+            return sse;
         }
 
         @Override
         void clearEventStream()
```
jfallows:
```suggestion
        void clearSse()
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -2354,7 +2454,10 @@ private abstract class McpRequestStream extends McpStream
     {
         final McpLifecycleStream session;
         final int requestId;
-        HttpEventStream eventStream;
+        HttpEventStream sse;
```
jfallows: Why is it both package private and have an accessor?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyLifecycleFactory.java
jfallows: Can we remove this flag, and for the success case verify that initial is still opened / reply is not closed before sending the begin?

jfallows (commented):
LGTM

PR #1806, opened 2026-05-29
[2.0] Refactor MQTT session management to use store-based persistence

### LINKED ISSUE #1798: Migrate mqtt-kafka session ownership to Store SPI
## Background

`MqttKafkaSessionFactory` currently coordinates MQTT session ownership using a custom Kafka consumer-group protocol (`"highlander"`). This issue replaces that mechanism with `StoreHandler`-based coordination and folds in the related cross-replica session-lookup, redirect-hint plumbing, and the `#migrate` compacted-topic takeover signal.

A previous release added an optional `store` field and a deprecation warning announcing this change. This issue completes the migration.

## Proposal

Coordinate session ownership via the `Store` SPI using the existing `StoreHandler` surface (`lock`, `renew`, `unlock`, `put`, `watch`). **No SPI changes are required.**

### Cooperative takeover protocol

1. New owner candidate `put`s its ownership record (replica identity, acquisition timestamp, identity strength, optional externally-reachable address) on the session key.
2. Current owner's `watch` fires with the new value.
3. Current owner detects displacement (value owner != self), stops lease renewal, drains pending state writes, calls `unlock`.
4. New candidate `lock`s the session key — succeeds.
5. New candidate writes its initial session state and begins serving.

### Strong vs weak session ownership

The ownership record includes an identity strength classification:

- **Strong identity**: this replica has an externally-addressable identity (`serverRef` configured). The address is published in the ownership record.
- **Weak identity**: no externally-addressable address is configured. The ownership record carries only an opaque replica identity.

The strength classification drives whether a CONNECT arriving on a non-owner replica should redirect the client back to the owner or initiate a takeover.

### CONNECT handling with strong/weak ownership

When CONNECT for clientId X arrives at replica B:

1. B reads the current ownership record for X from the Store.
2. If the current owner A has **strong identity** → B responds with MQTT 5 CONNACK reason `SERVER_MOVED` (0x9C) and `Server Reference` property pointing to A. **No takeover claim is put.** Client reconnects to A; session locality is preserved.
3. If the current owner A has **weak identity** (or there is no current owner) → B proceeds with the cooperative takeover protocol above.

If a CONNECT later arrives directly at A for the same clientId (either via redirect from B, or directly via the load balancer), normal MQTT session-takeover semantics apply locally at A — A kicks any stale prior connection and accepts the new one.

If A becomes unreachable or hung, the lease TTL ensures the lock is released within bounded time, after which a subsequent CONNECT at B (or any other replica) acquires the lock via Case (3) above.

### Failure recovery via TTL

The lease TTL bounds recovery when the cooperative path cannot complete:

- **Owner dead** — ceases renewal → lease expires → new candidate locks.
- **Owner drain hung** — watch handler disables renewal → lease expires → new candidate locks.

Renewal must be gated on continued ownership: the watch handler sets a flag that suppresses further renewals.

### Replacement of the `#migrate` compacted-topic takeover signal

`MqttKafkaSessionFactory` currently writes a `<clientId>#migrate` record on `mqtt-sessions` with a sender-id header during session establishment. Other replicas serving the same clientId observe this via a merged-stream filter (excluding their own sender-id) and yield ownership. This is an alternate channel for the same takeover signal carried by the highlander group, using the compacted topic instead of a Kafka consumer group.

The Store-based cooperative takeover protocol replaces this mechanism directly:

| `#migrate` mechanism | Store-based replacement |
|---|---|
| New owner writes `<clientId>#migrate` record with own sender-id | New owner `put`s ownership record on session key |
| Existing owner's merged-stream filter triggers | Existing owner's `watch` fires |
| Sender-id filter distinguishes self from others | `value.owner != self` check |

After this issue, the `#migrate` write, the merged-stream filter on `clientIdMigrate`, the associated `SENDER_ID` exclude clause, the `MIGRATE_KEY_POSTFIX` constant, and the `clientIdMigrate` field are removed.

### Configuration

```yaml
stores:
  memory0:
    type: memory

bindings:
  mqtt-kafka0:
    type: mqtt-kafka
    kind: proxy
    options:
      ...
      store: memory0
      serverRef: mqtt-node-1.example.com:1883   # optional; presence makes this replica's identity strong
```

`serverRef` continues to be the externally-reachable address advertised by this replica. Its presence classifies this replica's ownership as **strong**; its absence classifies it as **weak**. The option no longer drives any Kafka consumer-coordinator plumbing — it is published directly in the Store ownership record.

New `mqtt-kafka` options for lease duration and renewal interval, with defaults sized for fast cooperative takeover and bounded TTL fallback.

## In scope

- Schema change: `store` becomes required on `mqtt-kafka` binding; validation fails if absent.
- Implement cooperative takeover protocol in `MqttKafkaSessionFactory` using `StoreHandler`.
- Watch-gated renewal lifecycle.
- Configuration knobs for lease duration and renewal interval.
- Ownership record schema includes replica identity, acquisition timestamp, identity strength, and (for strong identity) the externally-reachable address.
- On CONNECT in `MqttServerFactory` / `MqttKafkaSessionFactory`, consult the Store for current owner; if owner has **strong identity**, respond with MQTT 5 CONNACK `SERVER_MOVED` + `Server Reference` rather than initiating takeover. If owner has **weak identity** (or no owner), proceed with cooperative takeover.
- Replace the cross-replica session lookup (currently the `mqtt-clients` Kafka consumer group with `consumerId=serverRef`) with the ownership record contents from the Store.
- Remove `KafkaResetEx.consumerId → MqttResetEx.serverRef` plumbing for session redirect; the redirect hint now flows via the Store watch event payload.
- Remove the `#migrate` compacted-topic takeover signal: stop writing `<clientId>#migrate` records; remove the merged-stream filter on `clientIdMigrate` and the associated `SENDER_ID` exclude clause; remove the `MIGRATE_KEY_POSTFIX` constant and `clientIdMigrate` field.
- Remove the highlander-group code path from `MqttKafkaSessionFactory`.
- Remove `"highlander"` protocol references from `KafkaClientGroupFactory` if no other callers remain.
- Tests: cooperative takeover, TTL-recovery failure modes, `Clean Start=true` displacement, CONNACK redirect when owner has strong identity, takeover when owner has weak identity, fallback to takeover after lease expiry when a strong-identity owner becomes unreachable.
- Update existing k3po specs that exercise session takeover, server redirect, and `#migrate` signaling.
- Documentation update.

## Acceptance criteria

- All existing `MqttKafkaSessionFactory` tests pass under the new ownership coordination.
- New tests cover: cooperative takeover, TTL-recovery failure modes, `Clean Start=true` displacement, CONNACK `SERVER_MOVED` redirect for strong-identity owners, cooperative takeover for weak-identity owners, and takeover via TTL expiry when a strong-identity owner is unreachable.
- The `"highlander"` protocol name is no longer referenced by `binding-mqtt-kafka`; the corresponding election code in `MqttKafkaSessionFactory` is removed.
- The `mqtt-clients` Kafka consumer group is no longer created by the binding.
- No `<clientId>#migrate` records are written by the binding; `MIGRATE_KEY_POSTFIX` and `clientIdMigrate` are removed from `MqttKafkaSessionFactory`.
- Client redirect hint (when owner identity is strong) is sourced from the Store ownership record, not from `KafkaResetEx.consumerId`.
- Configurations missing the `store` option fail validation with a clear error message.
- Documentation updated for the new session ownership model, strong/weak identity classification, CONNACK redirect behavior, and configuration knobs.

## References

- `runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java`
- `runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java`
- `runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientGroupFactory.java`
- `runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/store/StoreHandler.java`
- `runtime/store-memory/src/main/java/io/aklivity/zilla/runtime/store/memory/internal/MemoryStoreHandler.java`

### LINKED ISSUE #1799: Move MQTT QoS state into mqtt-sessions topic and consolidate session keys
## Background

`MqttKafkaPublishFactory` and `MqttKafkaSubscribeFactory` currently persist MQTT QoS in-flight state via Kafka offset-commit metadata on per-clientId consumer-group identities (`<groupIdPrefix>-<clientId>-session` for publish-side QoS 2; `<groupIdPrefix>-<clientId>` for subscribe-side QoS 1+). The offset-commit metadata string carries in-flight QoS 1/2 packet IDs, QoS 2 idempotent producer state (`producerId`, `producerEpoch`, `producerSequence`), and per-partition consumed offsets.

This has two issues for long-lived MQTT sessions:

1. **Kafka offset retention deletes session state.** Committed offsets are deleted after `offsets.retention.minutes` (default 7 days) once a group has no active members. Sessions with a session expiry interval longer than the retention window — or MQTT 3.1.1 `Clean Session=false` clients, which have no spec-defined expiry — can lose QoS state before the session is supposed to expire.
2. **Per-clientId Kafka consumer-group identities remain** as offset-commit namespaces, contributing Kafka resource footprint proportional to MQTT client count.

Moving this state into the compacted `mqtt-sessions` topic — which already holds subscription, will, and expiry state — eliminates both. Compacted topic retention is "keep latest value per key, indefinitely." After this change, `mqtt-kafka` maintains **zero Kafka consumer groups** (combined with the coordination move in #1798).

## Proposal

Store all per-session state, including QoS in-flight state, on the compacted `mqtt-sessions` topic under a **fixed, bounded set of keys per session**. Session state remains Kafka-native — no new store implementation and no `StoreHandler` SPI changes are required. (Ownership coordination is handled separately by #1798 via the Store SPI.)

### Fixed key set per session

| Key | Holds | Update frequency |
|---|---|---|
| `<clientId>` | Subscriptions, session metadata (expiry config, flags) | Low |
| `<clientId>#offsets` | All QoS state as one blob: per-(topic, partition) consumed offset, in-flight QoS 1/2 packet IDs (publish + subscribe), QoS 2 idempotent producer state | High |
| `<clientId>#will` | Will message and delivery signal (renamed from `#will-signal`) | Low |
| `<clientId>#expiry` | Session-expiry signal (renamed from `#expiry-signal`) | Low |

The `<clientId>#offsets` blob is a single record holding a map of `(topic, partitionId)` → `{ consumedOffset, inFlightPacketIds, producerId, producerEpoch, producerSequence }`. Each QoS 1/2 state change rewrites this single record. Each session is owned by a single worker on a single replica, so publish-side and subscribe-side updates to `#offsets` serialize naturally without concurrent-writer contention.

The blob is bounded in size: in-flight packet IDs are capped by the MQTT receive-maximum / in-flight limit, and partition count is bounded by topic configuration. QoS 0 traffic never touches it.

### Key renames

`#will-signal` → `#will` and `#expiry-signal` → `#expiry`, performed in this same change since record keys on `mqtt-sessions` are already being modified. (`#migrate` is removed separately by #1798.)

### Read/write changes

- `MqttKafkaPublishFactory.doCommitOffsetIncomplete`: rewrite to update the `<clientId>#offsets` record instead of committing a Kafka offset.
- `MqttKafkaSubscribeFactory` QoS 1+ offset persistence: rewrite to update the `<clientId>#offsets` record.
- `MqttKafkaSessionFactory` session-resume: read the `<clientId>#offsets` record to restore offsets, in-flight packet IDs, and producer state, instead of issuing `OffsetFetch` against the per-clientId group.
- Remove `newOffsetFetchStream` and `newOffsetCommitStream` paths and the per-clientId groupId construction in `MqttKafkaSessionFactory`.
- Update `#will-signal` / `#expiry-signal` producers and consumers to use the new `#will` / `#expiry` key postfixes.

### Cleanup

`mqtt-sessions` cleanup uses per-key tombstones (records with the same key and `payload == null`). With a fixed key set, cleanup is a known, bounded sequence — no enumeration or dynamic key discovery is required.

On session end (expiry signal fires, or `Clean Start=true` CONNECT arrives), tombstone the session keys in this order:

1. `<clientId>#offsets`
2. `<clientId>#will`
3. `<clientId>#expiry`
4. `<clientId>` (main record last)

This ordering ensures no replica reading `mqtt-sessions` ever observes a state where the main session record is gone while auxiliary records remain. Kafka log compaction physically removes tombstoned records after `delete.retention.ms` (default 24 hours).

### Migration

The compacted topic and the legacy mechanisms coexist for one release.

**QoS state**: the session-resume path reads from both during this window — read the `<clientId>#offsets` record first; if absent, fall back to fetching offset commits from the per-clientId Kafka group. Going forward, write only to `<clientId>#offsets`. After the deprecation window, the offset-commit read fallback is removed; stale offset commits in `__consumer_offsets` are cleaned up by Kafka retention.

**Key renames**: on resume, read both the new (`#will`, `#expiry`) and legacy (`#will-signal`, `#expiry-signal`) keys; prefer the new key when present. Going forward, write only the new keys and tombstone the corresponding legacy key. After the deprecation window, drop the legacy-key read path.

### Consequences

- **Zero Kafka consumer groups** created by `mqtt-kafka` (with #1798).
- **No Kafka offset retention exposure** for MQTT session state — sessions can have arbitrary expiry intervals.
- **MQTT 3.1.1 `Clean Session=false` persists indefinitely** until tombstoned, matching protocol intent.
- **Single durable storage mechanism** for all per-session state under a fixed key set.

## In scope

- Define the `<clientId>#offsets` blob schema (map of `(topic, partition)` → consumed offset, in-flight packet IDs, idempotent producer state).
- Rewrite `MqttKafkaPublishFactory.doCommitOffsetIncomplete` (and equivalents) to update the `#offsets` record.
- Rewrite `MqttKafkaSubscribeFactory` QoS 1+ offset persistence to update the `#offsets` record.
- Rewrite session-resume in `MqttKafkaSessionFactory` to read QoS state from the `#offsets` record.
- Rename `#will-signal` → `#will` and `#expiry-signal` → `#expiry`, with read-both/write-new/tombstone-legacy migration.
- Fixed-key-set cleanup with defined tombstone ordering (`#offsets` → `#will` → `#expiry` → `<clientId>`).
- Transitional read fallback to Kafka offset commits (for QoS state) and to legacy signal keys (for renames) for one release window.
- After the deprecation window, remove offset-fetch/commit code paths, per-clientId groupId construction, and legacy-signal-key read paths.
- Tests: QoS 1 in-flight replay on resume, QoS 2 publish idempotency across reconnect, session resume after offline duration greater than Kafka `offsets.retention.minutes`, MQTT 3.1.1 `Clean Session=false` indefinite persistence, fixed-key-set cleanup on session expiry and on `Clean Start=true`, key-rename migration (resume reads legacy keys, writes new keys, tombstones legacy).
- Documentation update.

## Acceptance criteria

- All existing QoS 1/2 publish and subscribe tests pass with the new persistence mechanism.
- A new test covers session resume after an offline duration greater than Kafka's `offsets.retention.minutes`, with full QoS state restored.
- A new test covers MQTT 3.1.1 `Clean Session=false` resumption after an extended offline period with no state loss.
- After the deprecation window, no `mqtt-kafka` code path creates, fetches, or commits to a Kafka consumer group identity for offset state.
- `kafka-consumer-groups.sh --list` shows no groups attributable to `mqtt-kafka` for any active or recently-disconnected session.
- The four session keys (`<clientId>`, `<clientId>#offsets`, `<clientId>#will`, `<clientId>#expiry`) are tombstoned in the documented order on session expiry and on `Clean Start=true`; compaction removes them after `delete.retention.ms`.
- Sessions established under the legacy `#will-signal` / `#expiry-signal` keys resume correctly after upgrade, with state rewritten under the new keys and legacy keys tombstoned.
- Documentation updated for the new persistence model, the fixed key set, and cleanup behavior.

## References

- `runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java` (offset fetch/commit paths, signal keys, cleanup)
- `runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishFactory.java` (`doCommitOffsetIncomplete`)
- `runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSubscribeFactory.java`
- `specs/binding-mqtt-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mqtt/kafka/config/` (`mqtt-sessions` topic schema/configuration)

### LINKED ISSUE #1800: Move replica identity from per-binding serverRef to engine-level service config
## Background

Replica identity in `mqtt-kafka` is currently declared via the per-binding `serverRef` option. After #1798, this option drives both:

- The externally-reachable address advertised in MQTT 5 `Server Reference` redirects (CONNACK `SERVER_MOVED`, DISCONNECT with `serverReference` property).
- Classification of this replica's identity as **strong** (when present) or **weak** (when absent) in the Store ownership record.

The same identity is conceptually engine-scoped, not binding-scoped — it describes the replica as a whole. Future protocol-level redirect or affinity behavior in other bindings would have to redeclare the same address per binding.

This issue moves identity declaration to engine-level configuration. Bindings derive identity from engine configuration rather than declaring it locally.

## Proposal

Reuse the existing engine-level service identity configuration (`zilla.engine.service.hostname`), already consumed by `binding-mcp`. Bindings derive identity from engine configuration rather than declaring it locally.

### Engine configuration

The engine exposes a service hostname via `EngineConfiguration` (`zilla.engine.service.hostname`):

```
zilla.engine.service.hostname=mqtt-node-1.example.com:1883
```

```yaml
bindings:
  mqtt-kafka0:
    type: mqtt-kafka
    kind: proxy
    options:
      store: memory0
      # serverRef no longer needed; inherited from engine service hostname
```

- `service.hostname` — primary externally-reachable address. Presence classifies this replica's identity as **strong**.
- Absence of a service hostname classifies this replica's identity as **weak**.

### Binding consumption

- `mqtt-kafka` reads the engine service hostname via `EngineConfiguration` and publishes it in Store ownership records (replacing direct read of binding-level `serverRef`).
- The strong/weak classification used in CONNECT handling (see #1798) is driven by the engine service hostname presence.
- Other bindings that introduce redirect or affinity behavior can consume the same configuration.

### Compatibility with existing per-binding `serverRef`

Per-binding `serverRef` continues to function for one release with a deprecation warning when set:

```
WARN [mqtt-kafka0] The 'serverRef' option on mqtt-kafka is deprecated.
Configure the engine 'service.hostname' property (zilla.engine.service.hostname) instead.
```

If both the engine service hostname and binding `serverRef` are set, binding `serverRef` takes precedence (preserves existing behavior for in-place upgrades). In a subsequent release, binding `serverRef` is removed; the engine service hostname becomes the sole source.

## In scope

- `mqtt-kafka` reads the engine service hostname and publishes it in Store ownership records.
- Deprecation warning on `mqtt-kafka` `serverRef` option when set.
- Precedence rule: binding-level `serverRef` overrides engine identity when both are present (transitional behavior).
- Documentation: engine-level service hostname configuration, deprecation of binding-level `serverRef`, migration guidance.
- Tests: identity resolution from engine config, override precedence when both are configured, deprecation warning emission.

## Acceptance criteria

- `mqtt-kafka` resolves the replica identity from the engine service hostname; identity is exposed to the binding via engine configuration.
- `mqtt-kafka` Store ownership records carry strong identity (with address) when the engine service hostname is configured, weak identity when absent.
- Setting `serverRef` on `mqtt-kafka` produces a deprecation warning logged once per binding instance.
- When both the engine service hostname and binding `serverRef` are configured, binding `serverRef` wins.
- Existing configurations using `serverRef` continue to function with the warning.
- Documentation updated.

## Related

- Depends on the strong/weak ownership model introduced in #1798.

## References

- Engine root configuration (`EngineConfiguration.ENGINE_SERVICE_HOSTNAME`).
- `runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/config/MqttKafkaOptionsConfig.java`
- `runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java`

jfallows:
## Description

This PR refactors MQTT session management to use a store-based persistence model instead of Kafka consumer group coordination. The changes introduce a new session storage mechanism that allows MQTT brokers to persist and retrieve session state independently, enabling better session takeover and redirect capabilities.

Fixes #1798 
Fixes #1799
Fixes #1800

### Key Changes

**Core Session Management:**
- Introduced `MqttKafkaSessionOffsetsHelper` to manage session offset persistence in a dedicated store
- Added `MqttKafkaSessionOffsetsFW` type for serializing session offset metadata
- Removed Kafka consumer group-based session coordination (`KafkaGroup`, `KafkaGroupBeginExFW`, `KafkaGroupFlushExFW`)
- Updated `MqttKafkaSessionFactory` to use the new store-based approach instead of group coordination

**MQTT Server Enhancements:**
- Added `SERVER_BUSY` reason code support for connection rejection
- Introduced `InstanceId` class for unique broker instance identification
- Added `Closeable` interface support for proper resource cleanup
- Enhanced session takeover and redirect handling with timestamp-based tracking

**Configuration & Storage:**
- Added memory store configuration support to MQTT options
- Updated `MqttConfiguration` to support configurable session stores
- Modified `MqttBindingConfig` to include store references
- Updated example configuration to include memory store definition

**Test Updates:**
- Added `MqttKafkaSessionOffsetsHelperTest` for session offset persistence logic
- Removed obsolete Kafka group-based session test scenarios
- Updated MQTT-Kafka integration tests to work with new session model
- Added new MQTT session redirect and takeover test scenarios

**Specification Changes:**
- Updated IDL to include `MqttKafkaSessionOffsets` structure
- Removed test scripts for deprecated Kafka group coordination flows
- Simplified session management test scripts by removing group-based coordination
- Added new test scenarios for store-based session management

### Motivation

The previous Kafka consumer group-based approach had limitations for session management across multiple broker instances. The new store-based model provides:
- Better isolation between session state and message publishing
- Simpler session takeover and redirect logic
- Support for external session stores (memory, persistent, etc.)
- Cleaner separation of concerns between session management and message flow

### Testing

Existing unit and integration tests have been updated to validate the new session persistence model. The test suite covers session creation, takeover, redirect, and expiry scenarios with the store-based approach.

https://claude.ai/code/session_01Gaf31VM5LZAFmUvtuBmqfz


akrambek:
### Code review

Found 1 issue:

1. `KafkaOffsetFetchStream` does not handle `FlushFW`, so a QoS2 client with no prior session-offsets record never gets a CONNACK.

The merged offset-fetch stream is opened `FETCH_ONLY` with a key filter and `KafkaEvaluation.EAGER` (`newOffsetFetchStream`). When the keyed `#offsets` record is absent — a fresh QoS2 client — the cache delivers a FLUSH after catch-up rather than DATA. But `onOffsetFetchMessage` only switches on `Begin`/`Data`/`End`/`Abort`/`Reset`; the FLUSH is silently dropped, so `onOffsetFetched` (the sole gate into `onProducerInit` → offset-commit → session establishment) is never called and the session stalls.

https://github.com/aklivity/zilla/blob/0e1140c3d1b4d47bd16857c94204ba1bd7f4ae45/runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java#L3847-L3871

The sibling stream with identical `FETCH_ONLY` + key + `EAGER` semantics, `KafkaFetchWillSignalStream`, handles exactly this no-data case by driving establishment from its flush handler — confirming the missing-key path arrives as a FLUSH:

https://github.com/aklivity/zilla/blob/0e1140c3d1b4d47bd16857c94204ba1bd7f4ae45/runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java#L3534-L3552

Note `onOffsetFetched` already anticipates empty entries (`initProducer` branch), but that branch is only reachable via `onKafkaData` — so it cannot run when delivery is a FLUSH. Existing QoS2 specs always pre-seed the `#offsets` record, so this path is untested.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishMetadata.java
```diff
@@ -172,4 +156,66 @@ public String16FW offsetMetadataToString(
...
+            sessionOffsetsRW.wrap(offsetBuffer, 0, offsetBuffer.capacity());
+            sessionOffsetsRW.version(SESSION_OFFSETS_VERSION);
+            entries.forEach(metadata -> sessionOffsetsRW.entriesItem(item -> encodeEntry(item, metadata)));
+            return sessionOffsetsRW.build();
+        }
+
+        public List<KafkaOffsetMetadata> decode(
+            MqttKafkaSessionOffsetsFW offsets)
+        {
+            final List<KafkaOffsetMetadata> entries = new ArrayList<>();
```
akrambek, line 196: Creating garbage.

runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
```diff
@@ -216,71 +167,6 @@ public void shouldSessionStreamReceiveServerSentReset() throws Exception
...
     }
 
-    @Test
-    @Configuration("proxy.yaml")
-    @Configure(name = WILL_AVAILABLE_NAME, value = "false")
-    @Configure(name = PUBLISH_MAX_QOS_NAME, value = "1")
-    @Specification({
-        "${mqtt}/session.server.sent.reset/client",
-        "${kafka}/session.group.server.sent.reset/server"})
-    public void shouldGroupStreamReceiveServerSentReset() throws Exception
```
akrambek, line 226: Just trying to understand if we are just refactoring implementation why do we need to remove spec tests?

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -3001,6 +3090,355 @@ else if (this.authField.equals(MqttConnectProperty.PASSWORD))
...
+            else if (claimed)
+            {
+                // Lock contended: the current owner either yields cooperatively (its watch fired
+                // on our record) or, if unresponsive, we force-unlock its stale token and steal.
+                stealAt = signaler.signalAt(currentTimeMillis() + sessionRenew.toMillis(),
+                    originId, routedId, initialId, traceId, SIGNAL_STEAL_SESSION_OWNERSHIP, 0);
+            }
+        }
+
+        // Shared success path for both the initial lock acquisition and a steal that succeeded
```
akrambek: feels like to much of comments.

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -3001,6 +3090,355 @@ else if (this.authField.equals(MqttConnectProperty.PASSWORD))
...
+        // and the lock is held by yet another contender). Then re-attempt the lock; the result
+        // is dispatched to onStealComplete which accepts on success or rejects with
+        // CONNACK SERVER_BUSY on failure.
+        private void onStealOwnership(
+            long traceId,
+            long authorization)
+        {
+            stealAt = NO_CANCEL_ID;
+
+            if (MqttState.initialClosed(state) || MqttState.replyClosed(state))
```
akrambek, line 3271: I think this can be simplified with below condition

runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
```diff
@@ -7198,5 +7638,86 @@ private int calculatePublishApplicationFlags(
...
+                        nonce = body.substring(nonceStart, secondSeparator);
+                        token = body.substring(secondSeparator + FIELD_SEPARATOR.length());
+                    }
+                    else
+                    {
+                        nonce = body.substring(nonceStart);
+                    }
+                }
+
+                record = new OwnershipRecord(!weak, identity, nonce, token);
```
akrambek: Looks like this is garbage

akrambek (changes requested):

runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaPublishMetadata.java
```diff
@@ -172,4 +156,66 @@ public String16FW offsetMetadataToString(
...
+        {
+            this.offsetBuffer = offsetBuffer;
+        }
+
+        public MqttKafkaSessionOffsetsFW encode(
+            Collection<KafkaOffsetMetadata> entries)
+        {
+            sessionOffsetsRW.wrap(offsetBuffer, 0, offsetBuffer.capacity());
+            sessionOffsetsRW.version(SESSION_OFFSETS_VERSION);
+            entries.forEach(metadata -> sessionOffsetsRW.entriesItem(item -> encodeEntry(item, metadata)));
```
akrambek: Following up on the "Creating garbage." note here: `encode()`/`encodeEntry()` run on the per-QoS2 offset-commit path (`MqttKafkaPublishFactory.doCommitOffset*`), so these are warm-path allocations rather than one-time setup:

- L178 `entries.forEach(metadata -> sessionOffsetsRW.entriesItem(item -> encodeEntry(item, metadata)))` — the inner `item -> encodeEntry(item, metadata)` captures `metadata`, so a fresh lambda is allocated per entry, per commit (and the outer lambda captures `this`).
- L217 `metadata.packetIds.forEach(p -> item.appendPacketIds(p.shortValue()))` — `packetIds` is an Agrona `IntArrayList`; `forEach(Consumer)` boxes every int to `Integer`, plus the capturing lambda. `IntArrayList.forEachInt(IntConsumer)` avoids both.

Both are avoidable with index loops over `entries`/`packetIds` (and `forEachInt`), keeping this on the no-alloc-on-the-hot-path convention. `decode()` just below already uses the primitive `(IntConsumer) packetIds::add` for the same reason.

runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
jfallows: We are no longer using consumer groups, so the spec scripts associated with them are no longer needed.

PR #1808, opened 2026-05-29
Separate engine initialization from startup

jfallows:
## Description

This change separates engine initialization from startup by introducing an explicit `init()` method and refactoring the startup sequence. The key improvements are:

1. **New `init()` method**: Starts workers and boss threads eagerly, with idempotent behavior (safe to call multiple times)
2. **Deferred binding attachment**: The `start()` method now attaches bindings after initialization, allowing for optional pre-startup hooks
3. **EngineRule enhancement**: Added `beforeStart()` hook to allow tests to execute custom logic (e.g., waiting for external services) on a separate thread before engine startup, without blocking the test body

### Changes

**Engine.java**
- Added `initialized` AtomicBoolean field to track initialization state
- New `init()` method that idempotently starts workers and boss
- Modified `start()` to call `init()` first, ensuring backward compatibility

**EngineRule.java**
- Added `beforeStart` field and `beforeStart(Runnable)` configuration method
- Refactored `evaluate()` to spawn a separate "engine-rule-before-start" thread that:
  - Calls the optional `beforeStart` hook
  - Calls `engine.start()` to attach bindings
  - Handles errors and interrupts the test thread if needed
- Added `starter.join()` in finally block to ensure proper cleanup

**EngineTest.java**
- Added `shouldInitThenStart()` test: verifies init() followed by start() works correctly
- Added `shouldInitIdempotent()` test: verifies calling init() multiple times is safe

### Motivation

This separation enables test scenarios where initialization (worker/boss startup) must happen before a blocking operation (e.g., waiting for an external runtime), while keeping the test body non-blocking. The idempotent init() design maintains backward compatibility with existing code that calls start() directly.

### Test Plan

Added unit tests covering the new init() behavior and idempotency. Existing tests continue to pass as start() remains backward compatible.

https://claude.ai/code/session_01XcNeWBUJAA87NSf9R4KdVT

### DIFF
--- a/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
+++ b/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
@@ -101,7 +101,6 @@ public final class EngineRule implements TestRule
         this.properties = new Properties();
         this.exceptions = m -> false;
         this.interruptible = true;
-        this.beforeStart = null;
 
         configure(ENGINE_DRAIN_ON_CLOSE, true);
         configure(ENGINE_SYNTHETIC_ABORT, true);
--- a/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
+++ b/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
@@ -93,6 +93,7 @@ public final class EngineRule implements TestRule
     private Predicate<String> exceptions;
     private boolean interruptible;
     private boolean clean;
+    private Runnable beforeStart;
 
     public EngineRule()
     {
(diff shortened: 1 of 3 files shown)


runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/EngineRule.java
```diff
@@ -93,13 +93,15 @@ public final class EngineRule implements TestRule
...
     private boolean clean;
+    private Runnable beforeStart;
 
     public EngineRule()
     {
         this.builder = Engine.builder();
         this.properties = new Properties();
         this.exceptions = m -> false;
         this.interruptible = true;
+        this.beforeStart = null;
```
jfallows: Remove assignment of implicit default value.

jfallows (commented):
LGTM

PR #1793, opened 2026-05-23
feat(examples): mcp.proxy demonstrates url-mode elicitation aggregation

jfallows:
## Description

The `examples/mcp.proxy` example, rebased onto `develop` now that all the
binding-mcp / binding-http runtime work it relied on has merged independently
(#1812, #1814, #1816, #1818, #1820). Only the example remains here.

Aggregates the `everything` reference server and a minimal `urlelicit` upstream
behind one Streamable HTTP endpoint, demonstrating:

- **Multi-toolkit aggregation** — `routes[].when.toolkit` fans `/mcp` into
  multiple upstream MCP servers, with a shared in-memory cache for
  `tools`/`prompts`/`resources` listings.
- **Protocol version negotiation** — Zilla offers MCP `2025-11-25` and echoes the
  negotiated version (#1820).
- **Form elicitation pass-through** — via the `everything` server.
- **URL-mode elicitation pass-through (SEP-1036)** — when the client advertises
  `elicitation.url`, Zilla advertises it upstream and relays the `mode:"url"`
  `elicitation/create` and `notifications/elicitation/complete` end-to-end. The
  bundled `url-elicit/` server (published `@modelcontextprotocol/sdk`, no OAuth)
  exercises this on its `authorize` tool.

## Testing

Verified end-to-end against a `develop`-built Zilla image (`docker compose up`):
`initialize` negotiates `2025-11-25`, `everything` + `urlelicit` toolkits
aggregate in `tools/list`, no worker faults. Smoke test: `./.github/test.sh`.
Interactive url-mode round-trip via MCP Inspector (documented in the README).

## Notes

- github (PAT pass-through, bootstrap credential, 401-challenge) is intentionally
  **not** in this OSS example — it will be demonstrated in a zilla-plus example
  alongside the per-toolkit `oauth` guard (#1810).
- A `time` (mcp-proxy bridge) route and a headless SDK-client driver were trialed
  and dropped: both surfaced independent Zilla interop edges (tools not
  aggregated from the mcp-proxy bridge; SDK `StreamableHTTPClientTransport` →
  Zilla `-32700`) to be investigated separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1812, opened 2026-06-02
fix(binding-http): client connection reuse per origin/authority + per-connection exchange/queue scope

### LINKED ISSUE #1811: binding-http: client connection reuse per origin/authority + per-connection exchange/queue scope
## Summary

`binding-http`'s client connection pool mishandles a single `http` client binding that fans
out to multiple `:authority` values: it reuses idle connections regardless of authority and
keys per-connection state (exchanges map, request queue, HTTP/2 stream ids) at **pool**
scope. The result is responses mis-routed/stalled across authorities, plus no distribution
of concurrent same-origin requests across connections.

## Problems

- `HttpClientPool` is keyed only by `resolvedId`; `supplyClient()` reused any idle connection
  by HTTP version + idle/reply-closed state only — **never matching `:authority`**. So a
  request to authority B could be sent over a connection established to authority A (wrong
  Host), and that authority's response comes back on the wrong exchange.
- The pool kept a **single `exchanges` map and a single request queue keyed by `streamId`**.
  Stream ids come from `HttpClient.nextStreamId()` and are **per-connection** (1, 3, 5 …), so
  two concurrent exchanges to different authorities both get stream id 1 → the later
  `exchanges.put` evicts the earlier and the pool-wide queue resolves to the wrong exchange →
  one request stalls forever. HTTP/2 likewise mis-routes responses/window via
  `exchanges.get(streamId)`.
- Concurrent same-origin requests were not distributed across available connections.
- NPE: a null exchange could be dereferenced while decoding HTTP/1.1 response headers in a
  teardown race.

## Fix

- Reuse/create client connections **per origin / `:authority`**, via a `canServe` coalescing
  seam on `HttpClient`.
- Move `exchanges` and the request queue **down into `HttpClient`** (per-connection), since
  stream ids are per-connection.
- Pin HTTP/2 stream-id scope per connection.
- Distribute concurrent same-origin requests across connections (tiered `supplyClient`:
  prefer-ready / new-under-limit / at-limit-fallback).
- Guard the null exchange when decoding HTTP/1.1 response headers.

## Tests

`rfc7230`/`rfc7540` client + server connection-management ITs, plus regression scenarios:
`requests.different.authorities`, `concurrent.requests.different.authorities`,
`concurrent.requests.same.authority` (each fails — timeout/mis-route — against the old code
and passes after the fix).

## Notes

Extracted from the `examples/mcp.proxy` branch (#1793), where this surfaced via the proxy's
shared `sys:http_client` fanning out to multiple upstream authorities. The defect and fix are
general to `binding-http` — any single `http` client binding routing to multiple authorities
is affected — so it belongs on `develop` independently of the MCP example.

jfallows:
## Description

General `binding-http` client connection-management fixes, **extracted from the
`examples/mcp.proxy` branch (#1793)** so they can land on `develop` independently. The
defect surfaced via the MCP proxy's shared `sys:http_client` fanning out to multiple
upstream authorities, but it is general to `binding-http`: any single `http` client binding
that routes to more than one `:authority` is affected.

Fixes #1811.

## Changes

- **Reuse client connections per origin / `:authority`** via a `canServe` coalescing seam on
  `HttpClient` (previously `supplyClient()` reused any idle connection by HTTP version + idle
  state only, never matching `:authority`, so a request could be sent over a connection to a
  different host).
- **Scope `exchanges` + the request queue per connection** (moved from `HttpClientPool` down
  into `HttpClient`). Stream ids are per-connection (1, 3, 5 …), so the pool-wide map/queue
  collided across concurrent authorities and stalled a request.
- **Pin HTTP/2 stream-id scope per connection.**
- **Distribute concurrent same-origin requests across connections** (tiered `supplyClient`:
  prefer-ready / new-under-limit / at-limit-fallback).
- **Guard a null exchange** when decoding HTTP/1.1 response headers in a teardown race.

## Testing

`./mvnw verify -pl runtime/binding-http` — `rfc7230`/`rfc7540` client + server
connection-management ITs green, including new regression scenarios:

- `requests.different.authorities` — sequential requests to two authorities through one
  client binding must open two connections (old: response2 never arrives).
- `concurrent.requests.different.authorities` — two concurrent cross-authority exchanges
  coexist (old: `TestTimedOut`; new: passes).
- `concurrent.requests.same.authority` — concurrent same-origin requests distributed.

### DIFF
--- a/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
+++ b/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
@@ -2148,12 +2148,8 @@ public final class HttpClientFactory implements HttpStreamFactory
             String scheme,
             String authority)
         {
-            HttpClient client = clients.stream().filter(
-                c -> c.canServe(scheme, authority) &&
-                (!HttpState.replyOpened(c.state) ||
-                c.encoder == HttpEncoder.HTTP_2 ||
-                c.exchange == null && c.encoder == HttpEncoder.HTTP_1_1 ||
-                c.encoder == HttpEncoder.H2C))
+            HttpClient client = clients.stream()
+                .filter(c -> c.canServe(scheme, authority) && c.available(versions))
                 .findFirst()
                 .orElse(null);
 
@@ -2163,6 +2159,14 @@ public final class HttpClientFactory implements HttpStreamFactory
                 client.scheme = scheme;
                 client.authority = authority;
                 onCreated(client);
+            }
+
+            if (client == null)
+            {
+                client = clients.stream()
+                    .filter(c -> c.canServe(scheme, authority) && c.reusable())
+                    .findFirst()
+                    .orElse(null);
             }
 
             return client;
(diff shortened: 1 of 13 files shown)


jfallows (commented):
LGTM

PR #1816, opened 2026-06-02
feat(binding-mcp): single session-id contract, negotiated protocol-version, partial-success hydration

jfallows:
## Description

The **single session-id contract** across the mcp server/proxy/client, with **negotiated
MCP-Protocol-Version** and **per-toolkit partial-success hydration**. Fixes the mcp.proxy
`400 "No valid session ID"` retry-storm (the cache hydrater originated a local session id
the remote never issued).

**Extracted from the `examples/mcp.proxy` branch (#1793)**; second of three sequential
binding-mcp extractions (cache refresh #1813 → **this** → proxy fixes + bootstrap credential).
Fixes #1815.

## Commits

1. **`feat(engine,binding-mcp): isLocalIndex worker affinity + worker-aligned local
   session-id generation`** — a prerequisite split from the example commit "prevent stall on
   local sessionId generation" (example edits excluded). Adds the engine SPI
   `EngineContext.isLocalIndex` + `LongIntPredicate` util (`EngineWorker` impl;
   `EchoWorker`/`TlsWorker` test EngineContext implementers updated per the engine SPI
   convention), and mints worker-aligned session ids in `McpServerFactory`.
2. **`feat(binding-mcp): single session-id contract, negotiated protocol-version,
   partial-success hydration`** — see below.

## What commit 2 does

**Single session-id contract (WS1):** lifecycle (`initialize`) requests carry no sessionId;
the reply carries the minted id; non-lifecycle requests carry it (one id per stream). The
**proxy** mints a worker-aligned unified id and is the sole minter/mapper, hash-routing each
south request by its id; the **server** maps `S↔U` and still returns `S` on the HTTP
`mcp-session-id` header; the **client** adopts the remote-minted `X` for the outbound HTTP
header; the **cache hydrater** sends no id and adopts the proxy-minted one. New engine util
`LongIntToLongFunction` for hash-routed `supplyInitialId`.

**Negotiated MCP-Protocol-Version (WS3):** default `2025-06-18`; capture
`result.protocolVersion` from the `initialize` response and send the negotiated version on
subsequent request headers.

**Per-toolkit partial success (WS2):** `hydrate.attempts.max` bounds per-kind retries; on
exhaustion the kind is marked degraded (`settled = populated || degraded`) and serves an
empty list envelope instead of aborting.

## Engine SPI note

Adds one method to `EngineContext` (`isLocalIndex`), with **all in-tree implementers
[...]

PR #1814, opened 2026-06-02
fix(binding-mcp): deterministic cache refresh startup/teardown + coalesced re-arming

jfallows:
## Description

Deterministic MCP proxy cache hydration/refresh lifecycle — **extracted from the
`examples/mcp.proxy` branch (#1793)** to land on `develop` independently. First of three
sequential binding-mcp extractions (this → single session-id contract → proxy fixes +
bootstrap credential).

Fixes #1813.

## Changes

- **Deterministic cache refresh startup/teardown** — replace the `cache.start.delay` timing
  hack with proper k3po coordination: `EngineRule.aroundStart(Supplier<Runnable>)` + k3po
  `deferStartable()` (the starter thread blocks until k3po is prepared, runs
  `engine.start()`, then releases so k3po proceeds only once bindings are attached).
- **Coalesce refresh re-arming** to one hydrate per cycle (was re-armed by both the local
  put-completion and the store-watch self-trigger).
- **`McpProxyCache.acquire`** no longer clobbers a held lock token on a failed overlapping
  acquire, so `release()` can always unlock (an orphaned per-kind lock previously stalled
  hydrate retries for the full lease TTL).
- **`McpProxyCacheHydrater.stop()`** aborts in-flight hydrate streams before ending the
  lifecycle (otherwise a stream left registered tripped the engine's synthetic-teardown
  `onServerReset` assertion).
- **`store-memory`:** synchronous dispatch hook; engine `TestStore*` dispatch support
  (test-jar) used to make the ITs deterministic.
- **`McpProxyCacheIT`** sped up via store dispatch.

## Dependency note

Includes a root `pom.xml` bump **k3po `3.4.0 → 3.4.2`** (required for `deferStartable()`).
This is repo-wide — every spec IT uses k3po — so CI exercises the full suite against 3.4.2;
the local validation below is binding-mcp-scoped.

## Testing

`./mvnw verify -pl runtime/store-memory,specs/binding-mcp.spec,runtime/binding-mcp` —
`McpProxyCacheIT` + binding-mcp ITs green (engine test-jar + store-memory built with the
changes).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1818, opened 2026-06-02
feat(binding-mcp): proxy hydration/forwarding fixes + per-upstream bootstrap credential

### LINKED ISSUE #1817: binding-mcp: proxy hydration/forwarding fixes + per-upstream bootstrap credential
## Summary

The remaining MCP proxy hydration/forwarding fixes plus the **per-upstream bootstrap
credential** — the third and final binding-mcp extraction from the `examples/mcp.proxy`
branch (#1793), after cache refresh (#1813) and the single session-id contract (#1815).

## Changes

- **Content-length framing for request bodies** — `fix`: frame client HTTP request bodies
  with `Content-Length`; `feat`: frame streamed `tools/call` / `prompts/get` /
  `resources/read` request bodies — the server stamps a guesstimate `contentLength` on
  `McpBeginEx` and the client declares a south `Content-Length` and whitespace-pads (no
  buffering / chunked).
- **Grant the proxy reply window** when the request reply begins (fixes a north
  `tools/list` hang from a reply-window race).
- **Strip the toolkit prefix** from the forwarded tool name in the proxy request body.
- **Degrade gracefully when a south route fails cache hydration** — skip a never-established
  route (`sessionId == null`) from hydration and isolate its per-route lifecycle failure
  from the shared hydrater lifecycle, so the healthy routes still aggregate.
- **Per-upstream bootstrap credential** on the mcp client binding
  (`options.authorization.credentials`) — used when there is no inbound client token, so the
  cache hydrater and token-less calls can authorize a route that requires it.
- **Widen the redirect hash unsigned** — `RedirectFW` carried `sessionId.hashCode()`
  sign-extended; binding-http folds it via `Long.hashCode`, so a negative hash routed the
  cross-worker redirect to the wrong worker → re-redirect → HTTP 500 for ~half of sessions
  under multiple workers. Widen with `Integer.toUnsignedLong` so the fold round-trips.

## Notes

General binding-mcp behavior, independent of the example (the example-config edits in two of
these commits are excluded and stay on the example branch). Final binding-mcp extraction;
after this, only the example remains on #1793.

jfallows:
## Description

The remaining MCP proxy hydration/forwarding fixes plus the **per-upstream bootstrap
credential** — the third and final binding-mcp extraction from the `examples/mcp.proxy`
branch (#1793), after cache refresh (#1814) and the single session-id contract (#1816).

Fixes #1817.

## Changes

- **`fix`: frame client HTTP request bodies with `Content-Length`.**
- **`feat`: frame streamed request bodies with `Content-Length`** — the server stamps a
  guesstimate `contentLength` on `McpBeginEx`; the client declares a south `Content-Length`
  and whitespace-pads (no buffering / chunked) for streamed `tools/call` / `prompts/get` /
  `resources/read`.
- **`fix`: grant the proxy reply window** when the request reply begins (north `tools/list`
  reply-window race).
- **`fix`: strip the toolkit prefix** from the forwarded tool name in the proxy request body.
- **`fix`: degrade gracefully when a south route fails cache hydration** — skip a
  never-established route (`sessionId == null`) and isolate its per-route lifecycle failure
  from the shared hydrater, so healthy routes still aggregate.
- **`feat`: per-upstream bootstrap credential** on the mcp client binding
  (`options.authorization.credentials`) — used when there's no inbound client token.
- **`fix`: widen the redirect hash unsigned** — `RedirectFW` carried `sessionId.hashCode()`
  sign-extended; binding-http folds it via `Long.hashCode`, so a negative hash routed the
  cross-worker redirect to the wrong worker → re-redirect → HTTP 500 for ~half of sessions
  under multiple workers. Widen with `Integer.toUnsignedLong` so the fold round-trips.

## Notes

General binding-mcp behavior, independent of the example. The example-config edits that
accompanied two of these commits (`degrade gracefully`, `bootstrap credential`) are
**excluded** and remain on the example branch. After this, only `examples/mcp.proxy` remains
on #1793.

## Testing

`./mvnw verify -pl specs/binding-mcp.spec,runtime/binding-mcp` — binding-mcp ITs
(server/client/proxy/lifecycle/cache) + spec ITs green, against develop's engine
(post #1816).

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
@@ -42,6 +42,7 @@ public final class McpBindingConfig
     public final long id;
     public final McpOptionsConfig options;
     public final GuardHandler guard;
+    public final String credentials;
     public final McpProxyCache cache;
     public final Map<String, McpProxySession> sessions;
     public final Map<String, McpRouteConfig> routeByPrefix;
@@ -85,6 +86,12 @@ public final class McpBindingConfig
             .map(a -> a.name)
             .map(binding.resolveId::applyAsLong)
             .map(context::supplyGuard)
+            .orElse(null);
+
+        this.credentials = Optional.ofNullable(options)
+            .map(o -> o.authorization)
+            .map(a -> a.credentials)
+            .filter(c -> !c.isEmpty())
             .orElse(null);
 
         this.cache = Optional.ofNullable(options)
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpOptionsConfigAdapter.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpOptionsConfigAdapter.java
@@ -24,6 +24,7 @@ import jakarta.json.JsonObjectBuilder;
 import jakarta.json.JsonString;
 import jakarta.json.bind.adapter.JsonbAdapter;
 
+import io.aklivity.zilla.runtime.binding.mcp.config.McpAuthorizationConfigBuilder;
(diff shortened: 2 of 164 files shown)


jfallows (commented):
LGTM

PR #1820, opened 2026-06-02
feat(binding-mcp): offer 2025-11-25 and negotiate elicitation url capability

jfallows:
## Description

Enable url-mode elicitation (SEP-1036) to flow through the `mcp` proxy by offering
MCP protocol **2025-11-25** and negotiating the **`elicitation.url`** client capability,
so the south client advertises it to upstreams only when the originating north client
supports it.

Closes #1819.

## Changes

Two commits:

1. **`feat`: offer MCP protocol version 2025-11-25** — bump the offered version from
   `2025-06-18` to `2025-11-25` across the two factories and the spec scripts; per-side
   down-negotiation to an older peer is preserved.
2. **`feat`: negotiate elicitation client capability**
   - `mcp.idl`: add `CLIENT_ELICITATION_FORM` and `CLIENT_ELICITATION_URL` to
     `McpCapabilities`, carried on the existing `McpLifecycleBeginEx.capabilities`.
   - **north server** parses `capabilities.elicitation.{url}` from the `initialize`
     request (bare `elicitation` implies form) and stamps the bits on the lifecycle
     begin instead of a fixed capability set; it now echoes the negotiated protocol
     version in the `initialize` response instead of always responding with the latest.
   - **south client** serializes `capabilities` back to the upstream from the bits,
     emitting `elicitation.url` only when the north client advertised it; the cache
     hydrater advertises no client capabilities.

The proxy already propagates the lifecycle capabilities north-to-south, so url-mode
elicitation is gated per-side: the south advertises `url` only when the north client
supports it.

## Notes

Independent binding-mcp enablement, off `develop` (after the example-branch extractions
#1812 / #1814 / #1816 / #1818). Precursor to wiring the `mcp.proxy` url-elicitation demo
on the rebased example branch (#1793).

## Testing

`./mvnw verify -pl specs/binding-mcp.spec,runtime/binding-mcp` — binding-mcp ITs (193)
+ spec NetworkIT/ApplicationIT peer ITs (179) green; checkstyle clean. New scenarios:
`lifecycle.initialize.elicitation.url`, `lifecycle.initialize.elicitation.form`,
`lifecycle.initialize.negotiate`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1821, opened 2026-06-03
fix(binding-mcp): parse initialize without JPMS-blocked reflection

jfallows:
## Problem

On the packaged **module-path** runtime image (the Docker image), every MCP `initialize` crashes the engine worker:

```
jakarta.json.bind.JsonbException: Error accessing field 'capabilities' declared in ... McpInitializeParams
Caused by: java.lang.IllegalAccessException: access to public member failed:
  ...McpInitializeParams.capabilities / ...McpInitializeParams$Capabilities/getField, from public Lookup
```

`McpServerFactory.onDecodeInitialize` deserializes `McpInitializeParams` with `JsonbBuilder.create().fromJson(...)`. yasson uses `MethodHandles` to build read handles; in a named module it cannot reflect into the encapsulated `internal.codec` package, so reading the `capabilities` field — whose type is the nested `Capabilities`/`Elicitation` classes added in #1820 — fails. Fields of accessible types (`String`, `JsonValue`) were unaffected, which is why this regressed only when the nested codec types were introduced.

It does not reproduce in the integration tests because they run on the **classpath** (unnamed module — every package open); it manifests only on the **module path**.

## Fix

- Read `capabilities` as a `jakarta.json.JsonObject` (an accessible jakarta.json type) and inspect `elicitation` and its `url` member via the JSON API, instead of reflecting into nested `internal.codec` POJOs.
- Add `opens io.aklivity.zilla.runtime.binding.mcp.internal.codec;` to the module descriptor as defense for any other JSON-B deserialization of codec POJOs (keeps compile-time encapsulation; permits runtime reflection only).

## Testing

- `./mvnw verify -pl runtime/binding-mcp` — 193 ITs green, checkstyle + coverage met.
- Existing `lifecycle.initialize.elicitation.url` / `.form` ITs cover client-capability detection.
- Verified against the module-path Docker image: `initialize` previously crash-looped on every request; now returns `200` with the negotiated protocol version.

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/codec/McpInitializeParams.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/codec/McpInitializeParams.java
@@ -14,20 +14,10 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.internal.codec;
 
-import jakarta.json.JsonValue;
+import jakarta.json.JsonObject;
 
 public class McpInitializeParams
 {
     public String protocolVersion;
-    public Capabilities capabilities;
-
-    public static class Capabilities
-    {
-        public Elicitation elicitation;
-    }
-
-    public static class Elicitation
-    {
-        public JsonValue url;
-    }
+    public JsonObject capabilities;
 }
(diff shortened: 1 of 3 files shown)


jfallows (commented):
LGTM

PR #1822, opened 2026-06-03
fix(binding-mcp): decode JSON-RPC requests independent of field order

jfallows:
## Problem

The official MCP TypeScript SDK (and therefore MCP Inspector, Claude, and SDK-based clients) serialize JSON-RPC as `{method, params, jsonrpc, id}` — `id` **last**. The server decoder assumed top-level keys arrive as `jsonrpc, id, method, params` and returned `HTTP 400 {"code":-32700,"Parse error"}` on `initialize`, so **no spec-compliant client could connect**. JSON object key order is insignificant per JSON-RPC.

This was never caught because the k3po scripts hand-craft `jsonrpc`-first ordering — a `reject.request.method.before.id` test even codified the wrong behavior.

## Fix

Make the envelope decode order-independent **while preserving streaming of `params`**:

- Capture `id` wherever it appears; defer id-dependent work to the end of the envelope (`END_OBJECT`): the request-needs-id validation, `session.requests` registration / request-id binding, and the `ping` response.
- Keep `method`-before-`params` required, so the BEGIN kind and route are known and `params` stream without buffering.
- New `McpServer.onDecodeRequestId(id)` binds the id to the in-flight stream and registers it; the decoder no longer reaches into stream state. `McpRequestStream` no longer takes the id in its constructor.

No `.idl`/extension change is needed: the request id is north-local (it never crosses a binding boundary), and the client/result decode path was already order-independent.

## Spec changes

Replaced `reject.request.method.before.id` (which asserted HTTP 400 for SDK ordering) with `*.id.last` accept scenarios for `initialize`, `tools.list`, and `ping`, plus matching `NetworkIT`/`McpServerIT` methods. Kept `reject.request.params.before.method` (method must still lead).

## Testing

- Test-first: the new `*.id.last` ITs fail (HTTP 400) on the old decoder and pass on the fix.
- `./mvnw verify -pl runtime/binding-mcp` — 195 ITs green, checkstyle + coverage met.
- `./mvnw verify -pl specs/binding-mcp.spec` — `NetworkIT` `*.id.last` peer tests green.

Builds on #1821 (merged).


jfallows (commented):
LGTM

Issue #1829, opened 2026-06-04
dump: general payload reassembly across DATA frames for handed-off Wireshark dissectors


jfallows:
## Context

The `dump` command's Wireshark Lua dissector (now pluggable via `ZillaDumpDissectorSpi`) tags each zilla DATA frame with its protocol type and, for the payload "next protocol", hands the DATA payload off to a Wireshark dissector. In `zilla.lua` this is:

```
-- handle_data_frame(...)
local dissector = resolve_dissector(protocol_type, slice_payload:tvb())
if dissector then
    dissector:call(slice_payload:tvb(), pinfo, tree)
end
```

i.e. `resolve_dissector` returns a Wireshark dissector handle (`Dissector.get("tls")`, `Dissector.get("http")`/`http2`, `Dissector.get("kafka")`, `Dissector.get("mqtt")`, …) and we invoke it on **a single DATA frame's payload as an isolated tvb**.

## Problem

When a protocol PDU spans **multiple** zilla DATA frames (fragmentation), the handed-off dissector sees truncated / partial PDUs and loses important context — it cannot reassemble.

Two distinct behaviors, only one of which works today:

- **Conversation state across frames — works.** The dump synthesizes a stable per-stream TCP conversation (consistent synthetic src/dst ports across all of a stream's frames), so conversation-keyed dissectors advance their state machine correctly across separate DATA frames (request vs. reply, handshake phase, etc.), as long as each PDU sits within a single DATA frame.
- **Byte-level reassembly of one PDU split across frames — does not work.** If a single PDU's bytes are split across two DATA frames, frame _N_ gets a truncated PDU and frame _N+1_ the remainder; neither dissects completely.

### Why TCP desegmentation doesn't rescue this

The synthetic TCP stream carries **zilla-framed** bytes (zilla header + payload, interleaved per frame), not contiguous protocol bytes. So Wireshark's native TCP reassembly would reassemble zilla framing, not the inner protocol's bytes. And `pinfo.desegment_len` / `desegment_offset` requests are not honored when a dissector is invoked via `Dissector:call` on a standalone payload tvb (there is no reassembly-capable parent driving it).

## Impact

[...]

PR #1830, opened 2026-06-04
feat(command-dump): pluggable Wireshark dissector via ZillaDumpDissectorSpi

jfallows:
## Summary

Makes the `dump` command's Wireshark `.lua` dissector **pluggable**, so dissector logic for additional protocols can be contributed out-of-tree instead of forking the monolithic `zilla.lua`. The core protocols are fully migrated to the new mechanism to prove the abstraction.

The `ZillaDumpCommand` Java was already protocol-agnostic (it tags frames with `CRC32C(typeLabel)`); all protocol-specific logic lived in the single checked-in `zilla.lua`. This PR reworks that file into a registry + fragments and adds an SPI.

## Changes

**1. Registry-based dispatch (behavior-preserving refactor)**
`zilla.lua` is reorganized into assembly shape — a head (version/proto/parsing helpers + a dissector registry), self-contained per-protocol blocks (id → enums → fields → handlers → `register_dissector`), and a tail. New `add_field(...)` and `register_dissector(type_id, name, ext_handler, payload_resolver)` helpers back the registry; `resolve_type`, `resolve_dissector`, and `handle_extension` become table lookups instead of hardcoded `if/elseif` chains. Pure transform — the ~30 tshark ITs pass byte-for-byte against the existing expected output.

**2. `ZillaDumpDissectorSpi` + fragment assembly**
- New SPI `io.aklivity.zilla.runtime.command.dump.ZillaDumpDissectorSpi` (`String type()`, `URL dissector()`), package exported in `module-info` with `uses`/`provides`.
- The monolith is split into a template (`zilla.lua`, carrying a `-- @dissectors@` marker) plus one self-contained Lua fragment per protocol under `internal/dissector/`, each backed by a core `ZillaDumpDissectorSpi` implementation.
- `ZillaDumpDissectors.assemble()` discovers providers via `ServiceLoader`, orders them by `type()`, and splices their fragments into the template at the marker. `ZillaDumpCommand.run()` and the `TsharkRunner` test harness both install this assembled output.

## Testing

- `ZillaDumpDissectorsTest` asserts the assembled script equals the golden monolith byte-for-byte (modulo the build `@version@` line).
- All existing command-dump tshark ITs pass byte-for-byte (`./mvnw verify -pl incubator/command-dump`, tshark image pinned at `4.2.0`).

## Known follow-up

Payload reassembly when a handed-off protocol PDU spans multiple DATA frames is not yet handled — tracked in #1829.

https://claude.ai/code/session_01VsRLeDneTbAa6sczLu1FxY

---
_Generated by [Claude Code](https://claude.ai/code/session_01VsRLeDneTbAa6sczLu1FxY)_


incubator/command-dump/pom.xml
```diff
@@ -170,6 +170,7 @@
         <configuration>
           <excludes>
             <exclude>src/test/resources/io/aklivity/zilla/runtime/command/dump/internal/*.txt</exclude>
+            <exclude>src/main/resources/io/aklivity/zilla/runtime/command/dump/internal/dissector/*.lua</exclude>
```
jfallows: These should not be excluded and instead have a license header if possible.

incubator/command-dump/pom.xml
```diff
@@ -267,7 +267,15 @@
         <directory>src/main/resources</directory>
         <filtering>true</filtering>
         <includes>
-          <include>**/zilla.lua</include>
+          <include>**/airline/zilla.lua</include>
```
jfallows, line 270: This is generated dynamically at runtime (not build time)now, so why included?

jfallows: The *aggregate* is generated at runtime, but this entry packages the *template* — the head (version/registry/parsing helpers) + the `-- @dissectors@` marker + tail — which `ZillaDumpDissectors.assemble()` reads as its input via `getResource("zilla.lua")`. Without it on the classpath, assembly has nothing to splice into. So the resource itself is still required.

It's in the *filtered* set only for the `@version@` substitution; the Lua already falls back to `"dev"` when unfiltered, so build-time filtering only matters to surface the real version on release builds.

If the intent is to drop build-time filtering entirely, I can resolve the version inside `assemble()` at runtime (from the module's implementation version) and move the template into the unfiltered resource block alongside the fragments. Want me to do that?

---

incubator/command-dump/pom.xml
jfallows: Done in `0e96a92b` — dropped the `.lua` exclude and added the Aklivity Community License header to all ten dissector fragments. To avoid the header repeating in the aggregate, `assemble()` now strips each fragment's preamble up to its `-- <name> dissector` marker, so the assembled `zilla.lua` keeps a single header from the template head (the `assembled == golden` test still passes). The remaining exclude is only the expected-output `*.txt` fixtures.

---

jfallows (commented):
LGTM

PR #1837, opened 2026-06-05
Add MCP metrics support with counters and histograms

### LINKED ISSUE #1825: metrics-mcp: instrument all MCP methods `initialize`, `tools/*`, `resources/*`, `prompts/*`
## Summary

The MCP pipeline emits no first-class metrics today. Operators can see *that* an agent connected but not *which tools it called, how often, how fast, whether the calls succeeded, or how much a cache absorbed*. This issue adds an `mcp` metrics group defined over MCP method semantics, so it applies uniformly to any binding that speaks MCP — the bindings in this gateway today and any added later. It covers every MCP method:

| Method | Stream trigger |
|---|---|
| `initialize` | session establishment |
| `tools/list` | tool discovery |
| `tools/call` | tool invocation |
| `resources/list` | resource discovery |
| `resources/read` | resource fetch by URI |
| `prompts/list` | prompt discovery |
| `prompts/get` | prompt template resolution |

`initialize` and the `*/list` methods are counted like the rest, and where a cache serves them their latency simply drops — so the benefit of a cache shows directly in the list-method latency distribution (a fast served mode, a slower rebuilt tail). No separate hit/miss ratio metric is needed; lower latency is the signal.

The metrics are **MCP-specific by design**. Generic transport signals — request counts, payload sizes, connection durations — already come from the `http`/`stream` groups on the binding that terminates the agent's transport, and the `mcp` group does not re-derive them. What it adds is the semantics the transport layer can't see: which method, which tool / resource / prompt, and the MCP-level outcome. Per-request detail that isn't on the wire (the field/entry counts a projection or authorization filter removed) is per-request log-event detail — a separate mechanism, not part of this metrics group.

---

## Proposed OTel metrics

These are deliberately **MCP-specific**, not generic. Stream and transport counters — request counts, payload byte sizes, connection durations — are already available from the `http`/`stream` groups on the binding that terminates the agent's transport; there is no value in the `mcp` group re-deriving them.

What those generic metrics *can't* give is anything about MCP semantics, because MCP multiplexes every logical operation over one endpoint: at the HTTP layer it's all `POST /mcp` — often many JSON-RPC messages over a single SSE stream — one indistinguishable request after another. The MCP layer has already parsed the method and the tool / resource / prompt name to route it (the `when` clause matches on `tool:` / `resource:` / `prompt:`), so those dimensions are available to a metric here and nowhere else. Exposing that breakdown is the whole job of this group:

```
telemetry:
  metrics:
    - mcp.initialize           # counter   — sessions initialized
    - mcp.tools.list           # counter   — tool-catalog requests
    - mcp.tools.call           # counter   — tool invocations, by tool + outcome
    - mcp.resources.list       # counter   — resource-catalog requests
    - mcp.resources.read       # counter   — resource reads, by resource + outcome
    - mcp.prompts.list         # counter   — prompt-catalog requests
    - mcp.prompts.get          # counter   — prompt resolutions, by prompt + outcome
    - mcp.<method>.duration    # histogram — per-method latency, same labels as the counter
```

The labels carry the MCP identity the transport layer never sees:

- **`tool`** on `mcp.tools.call` — the tool name exactly as carried in the begin extension, recorded verbatim; the metric doesn't interpret whether it's toolkit-prefixed. Per-tool call volume is the headline number, the first question every operator asks.
- **`resource`** on `mcp.resources.read` — the resource **catalog entry**, the direct analog of `tool`. A direct resource records its concrete URI (`config://app/settings`); a templated resource records its template (`repo://{owner}/{repo}`). Either way it's the declared identifier the binding matched to route the request (`when: resource:`), so cardinality is bounded by the catalog exactly as `tool` is bounded by the tool list. Template *parameter* values (`owner=acme`, `repo=widget-42`) are the resource's arguments — the analog of tool-call arguments — and are not metric labels for the same reason call arguments aren't; if needed they belong on a per-request log event or trace.
- **`prompt`** on `mcp.prompts.get` — the prompt name from the begin extension.
- **`outcome`** = `ok` | `error` on the invoking methods — MCP-level success vs failure, read from the **terminal frame**, never parsed from a body at the metric. There are two error sources and both are frame-native:
  - **Tool-domain error** (`tools/call` returns `isError: true`): the call protocol-succeeded, so the stream completes with a normal **END** carrying `McpEndEx.outcome` = `ok` | `error` (defaulting to `ok` when omitted). The result payload is delivered intact to the agent.
  - **JSON-RPC protocol error** (the response is a JSON-RPC `error` object — unknown method, bad params, dispatch/transport failure): the error block propagates as data, and the producing binding then completes the stream with **ABORT** instead of END. No dedicated extension is needed — ABORT *is* the error signal, and the error detail already rode through in the propagated JSON-RPC error block (which `mcp server` renders to the agent as a well-formed JSON-RPC error response).

  So the metric maps terminal frames directly: END (`outcome` omitted/`ok`) → `ok`; END (`outcome=error`) → `error`; ABORT → `error`. The producing binding — `mcp client` for a real upstream MCP server, or `mcp_openapi` / `mcp_http` / `mcp_kafka` mapping a non-MCP backend — is what lifts the upstream result onto `McpEndEx.outcome` or terminates with ABORT, so downstream bindings and the metric layer see one frame-native convention regardless of upstream. (If protocol-vs-tool errors need to be split in the metric, the terminal frame already distinguishes them — ABORT vs END+`outcome=error` — so an optional `error.kind` could derive from it.)

  This matters because it's invisible to HTTP: a failed `tools/call` returns HTTP `200` with `isError`, and a JSON-RPC error is also `200`, so an HTTP error-rate panel reads 0% while tools are failing. Capturing it is a primary reason the group exists.

Each method also carries a latency histogram. The list methods' latency is where a cache shows up: when a cache serves a `*/list`, the response is simply faster, so the latency distribution carries the cache benefit on its own — there's no need for a separate hit/miss ratio metric. Per-tool response *size*, where useful, is likewise MCP-specific (HTTP can't attribute bytes to one tool when several calls share a stream) and can be added as a histogram by `tool`.

These metrics are also **binding-agnostic**: the group attaches under any MCP binding's `telemetry.metrics`, so the same per-tool metric recorded at two points can be diffed — `mcp.tools.call{tool=…}` at ingress vs. at the fulfilling upstream localizes where a tool's calls or errors originate. Same MCP-specific metrics, distinguished by the binding they were recorded on.

What deliberately is **not** here: generic transport counts, sizes, and durations (use the `http`/`stream` groups), and the field/entry counts a projection or authorization filter removed (not on the wire — those are per-request log-event detail, a separate mechanism, not metrics).

Beyond these protocol labels, the group supports **attributes driven by token claims**: an operator can configure labels sourced from the validated authorization token — `tenant`, `sub`, or any claim — so per-tenant or per-subject breakdowns are available directly on the metrics. Claim-valued labels add cardinality, so they're opt-in per deployment.

---

## Where the metrics attach

The `mcp` group is bound under a binding's `telemetry.metrics`, the same way `http`/`stream` metrics are, and it works identically on whatever MCP binding it's listed under — no binding type is special, and one added later inherits it for free. There is no "correct" single placement: the operator instruments the points they care about, and each binding scopes its per-tool counts to its own identity. Typical placements and what diffing them reveals:

| Points compared | What the comparison shows |
|---|---|
| Ingress vs. fulfilling upstream — `mcp.tools.call{tool}` | where a tool's calls originate and how volume changes hop to hop |
| Any two points — `mcp.tools.call{outcome=error}` | which hop a tool failure is introduced at |

A cache's effect needs no comparison at all — it shows up as lower `*/list` latency at the binding that serves from cache. None of this requires per-binding metric variants — it's the same MCP-specific metrics, distinguished by the binding they were recorded on.

---

## Acceptance criteria

- [ ] `mcp` group registers the per-method metrics above, with `mcp.tools.call` labeled by `tool` + `outcome`, `mcp.resources.read` by `resource` + `outcome`, and `mcp.prompts.get` by `prompt` + `outcome`.
- [ ] `outcome` is read from the terminal frame, not a body parse: END with `McpEndEx.outcome` = `ok`/`error` (default `ok`) for tool-domain results, ABORT for JSON-RPC protocol errors (error block still propagated as data). The producing binding (`mcp client`, `mcp_openapi`, `mcp_http`, `mcp_kafka`) sets this so downstream sees one convention.
- [ ] Tool / resource / prompt names are recorded verbatim from the begin extension, without the metric layer interpreting or rewriting them.
- [ ] Each method has a latency histogram (including the `*/list` methods), so a cache's benefit is visible as lower list-method latency without any separate hit/miss metric.
- [ ] The group does not register generic transport metrics (counts, sizes, durations at the transport level) — those remain the province of the `http`/`stream` groups.
- [ ] The group attaches under `telemetry.metrics` on any MCP-compatible binding and records identically on each, scoped to that binding's identity, so the same metric at two bindings is distinguishable for diffing — one added later inherits it without changes.
- [ ] Attributes can optionally be driven by token claims (e.g., `tenant`, `sub`), resolved from the validated authorization token and opt-in per deployment.
- [ ] Metrics export cleanly via the existing OTLP / Prometheus exporters with no binding-specific coupling.
- [ ] Integration tests cover per-tool call counts, an `outcome=error` tool result that returned HTTP 200, and per-method latency including a cache-served `*/list` showing the expected lower latency.

jfallows:
## Description

This PR adds comprehensive metrics support for the MCP (Model Context Protocol) binding, enabling observability of MCP operations through counters and duration histograms.

### Changes

**New Modules:**
- `runtime/metrics-mcp`: Runtime implementation of MCP metrics collection
- `specs/metrics-mcp.spec`: Specification and test definitions for MCP metrics

**Core Metrics Implementation:**
- `McpMetricGroup`: Main metric group providing 14 MCP metrics (counters and histograms)
- `McpMetricContext`: Handles metric context and stream event processing
- `McpMethod`: Enum mapping MCP request types (initialize, tools.list, tools.call, resources.list, resources.read, prompts.list, prompts.get)
- `McpMeasure`: Enum for metric measures (COUNT, DURATION)
- `McpAttributeHelper`: Extracts metric attributes from MCP stream extensions
- `McpMetric`: Individual metric definition wrapper
- `McpMetricGroupFactorySpi`: Service provider for metric group factory

**Metrics Exposed:**
- `mcp.initialize` / `mcp.initialize.duration`
- `mcp.tools.list` / `mcp.tools.list.duration`
- `mcp.tools.call` / `mcp.tools.call.duration`
- `mcp.resources.list` / `mcp.resources.list.duration`
- `mcp.resources.read` / `mcp.resources.read.duration`
- `mcp.prompts.list` / `mcp.prompts.list.duration`
- `mcp.prompts.get` / `mcp.prompts.get.duration`

**Supporting Changes:**
- Extended MCP binding spec with `McpOutcome` enum (OK/ERROR) and `McpEndEx` structure for outcome reporting
- Updated `McpClientFactory` and `McpProxyItemFactory` to support outcome tracking in stream end extensions
- Added test specifications for error scenarios (tools.call.is.error)
- Updated parent pom.xml files to include new modules

### Test Coverage

- `McpMetricGroupTest`: 425 lines of comprehensive unit tests covering metric resolution, context creation, stream event processing, and attribute extraction
- `McpMethodTest`: Tests for MCP method type detection from stream extensions
- `McpMetricGroupFactoryTest`: Factory instantiation tests
- Integration tests in binding specs for error handling scenarios

### Integration

Metrics are automatically discovered via `MetricGroupFactorySpi` and integrated into the Zilla metrics framework. The implementation tracks both success and error outcomes for each MCP operation type.

Fixes #1825 

https://claude.ai/code/session_01L1p2VkKKCAZd7zxvGBFueS


jfallows (commented):
LGTM

PR #1841, opened 2026-06-05
feat(binding-mcp): per-route allow-set filtering for tools/prompts/resources (#1833)

### LINKED ISSUE #1833: binding-mcp: per-route primitive-name allow-set filtering for mcp proxy
## Summary

Add optional per-route **name allow-sets** — `tools`, `resources`, `prompts` — to the `mcp · proxy` route `when` condition, so an operator can expose a curated subset of an upstream toolkit's primitives (e.g. `github__create_*` but not `github__delete_repo`) rather than the whole toolkit. This is **identity-independent operator policy**, orthogonal to the per-identity authorization filtering tracked in #1831 and the toolkit/capability routing delivered in #1669.

## Motivation

Today the proxy route condition (`McpConditionConfig`) matches only on:
- `toolkit` — literal prefix match (`identifier.startsWith("github__")` for tools/prompts, `"github+"` for resources), and
- `capability` — the primitive *type* (`tools` / `prompts` / `resources`).

A single upstream MCP server exposes dozens of tools. Exposure is currently all-or-nothing per toolkit per type. Operators routinely need to publish a curated allow-list at the gateway, independent of who is connecting. There is no way to express this today.

Every sibling proxy binding already does pattern matching — `MqttKafkaConditionMatcher` (MQTT `+`/`#` → regex), `HttpKafkaConditionMatcher` (`*` glob + `{param}` → regex), `GrpcKafkaConditionConfig` (service/method/metadata) — so MCP is the outlier doing literal-prefix only.

## Relationship to #1831 and #1669

- **#1669** (toolkit + capability routing) — this layers *within* a matched toolkit; routing/exit-selection by toolkit prefix is unchanged. Name allow-sets are a projection, **not** a new exit-selection dimension.
- **#1831** (per-client `tools/list` filtering by per-tool **authorization** metadata) — a different axis: #1831 hides what a given identity *can't invoke*; this hides what the *operator* never wants exposed, for anyone. #1831 source #2 is already an "operator-declared mapping (tool name/glob → required scope)", so the two **must share a single glob engine**; this issue introduces the engine and the identity-independent allow-set, #1831 reuses it for the scope mapping.

### Field names and semantics

New optional fields on `McpConditionConfig`, plural to match the `capability` enum values and the `topics:` precedent in mqtt-kafka. Each is a list of glob patterns matched against the **prefix-stripped** name (the name after `McpRouteConfig.strip`, e.g. `create_pr`, not `github__create_pr`; resources matched against the stripped URI):

```yaml
routes:
  - exit: south_mcp_client_github
    when:
      - toolkit: github
        capability: [tools, resources]
        tools: ["create_*", "get_*"]   # allow-set within tools
        resources: ["repo://*"]         # allow-set within resources
        # prompts: [...]
```

Allow-set semantics (set membership; a name is selected iff it matches ≥1 pattern):

| Field state | Meaning |
|---|---|
| **absent** | no constraint — every primitive of that type passes |
| **present, non-empty** | the allow-set — select names matching at least one pattern |
| **present, empty (`[]`)** | empty allow-set — matches nothing (type negotiated but zero entries) |

This separates "I didn't say" (absent → all) from "I said: nothing" (`[]` → none).

### Two layers: capability negotiates, allow-set curates

`capability` drives `McpRouteConfig.serverCapabilities()` / `serves()` — the SERVER_TOOLS/PROMPTS/RESOURCES bits advertised at `initialize` and whether list ops federate through the route. The allow-set filters *contents within* a served type. These stay distinct:

- `capability: [resources, prompts]` (no `tools`) → tools capability **not advertised**; client never sends `tools/list`. The surface doesn't exist.
- `capability` omitted + `tools: []` → tools capability **advertised**; client sends `tools/list` and gets an **empty array**. The surface exists but is empty (and can later gain entries via `list_changed` without re-negotiation).

**Decision to lock in:** capability advertisement is driven **solely** by the `capability` field, and is never retroactively suppressed by an empty allow-set — otherwise the two configs above collapse and the useful "exists but empty" state is lost.

### Cross-field validation

A name field is legal **only if** its type is served — `capability` absent (all three) **or** `capability` contains that type. Specifying `tools: [...]` while `tools` is excluded from `capability` curates a surface declared not to exist and must be **rejected at parse time** (schema + `McpConditionConfigAdapter` guard), per validate-at-the-boundary.

### Enforcement points (dual)

Per the advisory-list / authoritative-call principle from #1810/#1831:
1. **list-merge (advisory):** the allow-set filters federated `tools/list` / `resources/list` / `prompts/list` results.
2. **invocation (authoritative):** `tools/call` / `prompts/get` / `resources/read` resolution in `McpBindingConfig.resolve` must reject a name outside the allow-set — list-only filtering is bypassable by a client that already knows the name.

### Key design change

`tools/list` filtering requires the federated list path to **honor `when`** (or apply the allow-set as a post-merge projection over the capability registry). This is a deliberate reversal of #1669's *"Federation for list operations fans out to all exits without inspecting individual route `when` clauses."* — and is the main cost of the feature.

### Matcher

Reuse the `HttpKafkaConditionMatcher` glob→regex approach (simpler than MQTT's `+`/`#` handling): `*` → `.*`, literal `.` escaped, compiled once per route. No named-parameter extraction in v1.

## Affected code

- `runtime/binding-mcp/.../config/McpConditionConfig.java` — add `tools`/`resources`/`prompts` fields
- `runtime/binding-mcp/.../internal/config/McpConditionConfigAdapter.java` — JSON (de)serialization + cross-field guard
- `runtime/binding-mcp/.../internal/config/McpRouteConfig.java` — extend ConditionMatcher with allow-set predicate; apply in `matches()` / list projection
- `runtime/binding-mcp/.../internal/config/McpBindingConfig.java` — enforce at `resolve` (invocation)
- federation/list path — honor `when` for list ops
- `specs/binding-mcp.spec/.../schema/mcp.schema.patch.json` — fields + `if/then` cross-field constraint
- `specs/binding-mcp.spec` — `.rpt` scripts (test-first)

## Acceptance criteria

- `tools`/`resources`/`prompts` allow-sets on a route filter both federated list responses and invocation routing
- Absent = unconstrained; non-empty = allow-set; `[]` = matches nothing
- Capability advertisement unaffected by an empty allow-set
- Name field with type absent from `capability` rejected at config parse
- Invocation of a name outside the allow-set is rejected even when the client supplies the name directly
- Glob patterns matched against the prefix-stripped name/URI
- IT coverage: curated allow-set, `[]`-empty-but-advertised, cross-field rejection, list∩call consistency

## References

- Parent/adjacent: #1669 (toolkit/capability routing), #1810 (per-toolkit elicitation), #1831 (per-identity auth filtering)
- Precedent: `HttpKafkaConditionMatcher`, `MqttKafkaConditionMatcher`

jfallows:
## Summary

Adds optional per-route **allow-sets** to the MCP proxy `routes[].when` condition, letting an operator restrict which tools/prompts/resources a federated upstream exposes. Two complementary behaviors:

1. **Authoritative invocation enforcement** — a `tools/call` / `prompts/get` / `resources/read` to a name outside the route's allow-set fails to resolve and is rejected, even when the client supplies the name directly.
2. **Advisory list filtering** — `tools/list` / `prompts/list` / `resources/list` federation results only advertise permitted primitives (live path and cache path).

Config shape (per `when` item, proxy kind only):

```yaml
routes:
  - exit: app1
    when:
      - toolkit: github
        capability: [tools, resources]
        tools: ["create_*", "get_*"]
        resources: ["repo://*"]
```

Semantics: field **absent** = no constraint; **present** = allow-set (match ≥1 glob); **empty** = matches nothing. Capability advertisement stays driven solely by `capability` (an empty allow-set does not suppress it).

## Implementation

- **Config/matcher**: new `tools`/`prompts`/`resources` glob fields on `McpConditionConfig`; the previously-nested condition matcher is extracted to a top-level `McpConditionMatcher` (mirrors `MqttKafkaConditionMatcher`) with the allow-set predicate. Names are matched via the JSON parser, so item field order doesn't matter.
- **Schema**: adds the three array fields plus cross-field validation rejecting a name field whose type is excluded from `capability`; fields are forbidden on non-proxy kinds.
- **Invocation enforcement**: falls out of route resolution — `match()` now gates the stripped name, so `McpProxyItemFactory` resolves no route and the stream is rejected.
- **List filtering** (`McpProxyListFactory`, the streaming hot path): in filtering mode an item is scanned (without emitting) until its name/uri is decoded, then emitted-from-start (begin + prefix-injected body, reusing the existing backpressure path) or skipped. **Non-filtering routes keep the exact original streaming path**, so unfiltered lists are byte-identical. Items too large to inspect before the slot boundary pass through unfiltered (advisory; invocation enforcement is the security boundary). The cache path inherits filtering because hydration runs through the same `McpListServer`.

## Tests (test-first)

- Unit: `McpConditionMatcherTest` (glob admit/deny, absent/empty/non-empty, `serves`/`admits`/`filters`), `McpConditionConfigAdapterTest` (round-trip incl. empty allow-set)
- Schema: valid filter config + cross-field rejection
- ITs: `McpProxyIT` invocation rejection, `McpProxyIT` live-path list filtering, `McpProxyCacheIT` cache-path list filtering at hydration
- Full `binding-mcp` + `binding-mcp.spec` suite passes (129 unit + 428 IT), 0 checkstyle violations, all JaCoCo coverage met

## Notes
- Docs for the new route fields live in the separate docs repo and are not included here.

Resolves #1833.

https://claude.ai/code/session_0115ZRrNA9iPG83DfsM1pduB

---
_Generated by [Claude Code](https://claude.ai/code/session_0115ZRrNA9iPG83DfsM1pduB)_

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import java.util.function.LongFunction;
 import java.util.function.LongSupplier;
 import java.util.function.LongUnaryOperator;
+import java.util.function.Predicate;
 
 import jakarta.json.stream.JsonParser;
 import jakarta.json.stream.JsonParserFactory;
@@ -243,8 +244,7 @@ abstract class McpProxyListFactory implements BindingHandler
         private final long originId;
         private final long routedId;
         private final String8FW prefix;
-        private final McpRouteConfig route;
-        private final boolean filtering;
+        private final Predicate<String> admits;
         private final McpLifecycleClient lifecycle;
         private long initialId;
         private long replyId;
@@ -280,14 +280,13 @@ abstract class McpProxyListFactory implements BindingHandler
             McpListServer server,
             long routedId,
             String8FW prefix,
-            McpRouteConfig route)
+            Predicate<String> admits)
         {
             this.server = server;
             this.originId = server.lifecycle.originId;
             this.routedId = routedId;
             this.prefix = prefix;
-            this.route = route;
-            this.filtering = route != null && route.filters(kind);
+            this.admits = admits;
             this.lifecycle = server.lifecycle.supplyClient(routedId);
         }
 
@@ -922,7 +921,7 @@ abstract class McpProxyListFactory implements BindingHandler
                 client.decodedItemProgress = decodedItemProgress - 1;
                 client.decodeItemDepth = 1;
                 client.itemBegun = false;
-                if (client.filtering)
+                if (client.admits != null)
                 {
                     client.itemDeferred = true;
                     client.decoder = decodeItemScan;
@@ -1019,7 +1018,7 @@ abstract class McpProxyListFactory implements BindingHandler
             if (event == JsonParser.Event.VALUE_STRING)
             {
                 final String name = parser.getString();
-                if (client.route == null || client.route.admits(kind, name))
+                if (client.admits == null || client.admits.test(name))
                 {
                     client.onDecodedItemBegin(traceId);
                     client.itemDeferred = false;
@@ -1565,7 +1564,11 @@ abstract class McpProxyListFactory implements BindingHandler
                 doEncodeEndItems(traceId);
                 return;
             }
-            client = new McpListClient(this, route.resolvedId(), route.prefix(), route.route());
+            final McpRouteConfig routeConfig = route.route();
+            final Predicate<String> admits = routeConfig != null && routeConfig.filters(kind)
+                ? name -> routeConfig.admits(kind, name)
+                : null;
+            client = new McpListClient(this, route.resolvedId(), route.prefix(), admits);
             client.doClientBegin(traceId);
         }
 
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
@@ -34,6 +34,7 @@ import org.agrona.concurrent.UnsafeBuffer;
 
 import io.aklivity.zilla.runtime.binding.mcp.internal.McpConfiguration;
 import io.aklivity.zilla.runtime.binding.mcp.internal.config.McpBindingConfig;
+import io.aklivity.zilla.runtime.binding.mcp.internal.config.McpRouteConfig;
 import io.aklivity.zilla.runtime.binding.mcp.internal.config.McpRoutePrefix;
 import io.aklivity.zilla.runtime.binding.mcp.internal.stream.McpProxyLifecycleFactory.McpLifecycleClient;
 import io.aklivity.zilla.runtime.binding.mcp.internal.stream.McpProxyLifecycleFactory.McpLifecycleServer;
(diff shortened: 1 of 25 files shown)


runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -267,16 +273,21 @@ private final class McpListClient implements McpRouteRequest
...
             long routedId,
-            String8FW prefix)
+            String8FW prefix,
+            McpRouteConfig route)
         {
             this.server = server;
             this.originId = server.lifecycle.originId;
             this.routedId = routedId;
             this.prefix = prefix;
+            this.route = route;
```
jfallows: Can we just pull admits as a functional interface?

jfallows: Done in a9ad5562 — `McpListClient` now takes a `Predicate<String> admits` instead of the `McpRouteConfig`. The predicate is built once per client in `onNextClient` (null when the route has no allow-set for the capability), and the deferred-scan path engages when it's non-null. Full MCP suite green.

---

jfallows (commented):
LGTM

PR #1844, opened 2026-06-06
fix(binding-mcp): defer reply content-type for elicitation-capable sessions

jfallows:
## Problem

When an `mcp` server fronts an arbitrary mcp-speaking upstream, the reply-side frames can arrive in the order: reply `BEGIN` → `elicitation/create` CHALLENGE → result. `McpRequestStream.onAppBegin` committed the HTTP response content-type **eagerly** (`application/json`, absent a `progressToken`), so a later `elicitation/create` CHALLENGE — which must stream over `text/event-stream` — could not be honored.

#1842 detects this illegal transition and fails loud (RESET/ABORT). This change prevents it from arising for elicitation-capable sessions, so url-mode elicitation can actually stream.

## Fix

Defer the reply content-type **only when the session can elicit** (`requestTimeout > 0` — the client advertised `elicitation.url` and the server `options.timeout` is set):

- `onAppBegin` no longer emits the net reply `BEGIN` for such sessions; a shared `ensureResponseBegun()` emits it exactly once, choosing the content-type from the first frame that determines it:
  - **JSON** on the first result `DATA` (`onAppData`)
  - **`text/event-stream`** when an `elicitation/create` CHALLENGE arrives (`server.onAppChallenge`) — `replyOpening` is still `false`, so the #1842 guard is correctly skipped and the upgrade succeeds.
- `ensureResponseBegun()` is also invoked from `onAppEnd`, `onAppAbort` and `doAppCancel` so every terminal path still emits a response.
- Sessions that **cannot** elicit (`requestTimeout == 0`) keep the original eager behavior, preserving the cancel/abort/shutdown choreography.

Invariants preserved:
- #1823 — a resumable flush on a non-SSE reply still yields a JSON result.
- #1842 — a result delivered **before** a challenge still commits JSON and the challenge still aborts.

## Test

Adds `tools.call.elicit.deferred.begin` (`McpServerIT.shouldCallToolElicitWhenReplyBeginsBeforeChallenge`) exercising the reply-`BEGIN`-before-CHALLENGE order — it **fails on develop** (JSON committed → #1842 abort) and **passes** with this change. Full binding-mcp suite: 223 ITs + 26 unit tests green.

> Note: this is a necessary server-side piece of end-to-end url-mode elicitation through the mcp proxy. A separate follow-up is still needed for the south client to parse/relay the upstream's `elicitation/create` as a CHALLENGE (it currently isn't recognized), and for the `requestTimeout`-set forwarding path.


runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpServerFactory.java
```diff
@@ -4639,28 +4639,29 @@ private void onAppBegin(
...
-            else
+
+            flushAppWindow(traceId, authorization, 0L, 0, 0, encodeMax);
+        }
+
+        private void ensureResponseBegun(
+            long traceId,
+            long authorization)
+        {
+            if (!McpState.replyOpening(server.state))
```
jfallows: Move this guard to server doEncodeResponseBegin and just call it directly.

Also, doEncodeResponseBegin can handle the httpBeginEx inside the method instead of requiring the caller to specify.

If explicit control is needed in some cases, then have an overload of doEncodeResponseBegin with no extension, so its defaulted same as below for success case.

specs/binding-mcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mcp/streams/application/tools.call.elicit.deferred.begin/server.rpt
```diff
@@ -0,0 +1,103 @@
...
+        '"name":"get_weather",'
+        '"arguments":'
+        '{'
+          '"location": "New York"'
+        '}'
+      '}'
+
+# reply opens before the elicitation challenge arrives — the server must NOT
+# commit the response content-type yet (no progressToken), otherwise the
+# subsequent elicitation/create cannot be streamed over text/event-stream
```
jfallows: Remove verbose comments from scripts.

specs/binding-mcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mcp/streams/application/tools.call.elicit.deferred/server.rpt
jfallows, line 1: `begin` is internal protocol and deferred begin is describing the implementation behavior, not the scenario itself.

I think we can just name this scenario `tools.call.elicit.deferred` since the challenge is actually deferred until after the write flush in the scenario, and yet the implementation handles it correctly.

jfallows:
Addressed in b1f7712e1:

- **`doEncodeResponseBegin`**: removed the `ensureResponseBegun` helper. Added a no-extension `McpServer.doEncodeResponseBegin(traceId, authorization)` overload that holds the `replyOpening` guard and builds the default success BEGIN (status 200 + content-type from `sseUpgrade`), delegating to the extension overload. `McpRequestStream` now calls `server.doEncodeResponseBegin(traceId, authorization)` directly from `onAppBegin`/`onAppData`/`onAppEnd`/`onAppAbort`/`doAppCancel`. The explicit-extension cases (initialize, ping, challenge) keep using the extension overload.
- **Scenario rename**: `tools.call.elicit.deferred.begin` → `tools.call.elicit.deferred` (IT `shouldCallToolElicitDeferred`).
- **Scripts**: removed the verbose comments.

223 ITs + 26 unit tests green.

jfallows (commented):
LGTM

PR #1850, opened 2026-06-08
fix(binding-http): reset streaming response on remote client disconnect

jfallows:
## Problem

For an HTTP/1.1 server exchange whose request is already complete but whose response is still **streaming** (e.g. Server-Sent Events), a graceful client disconnect arrives as a network `END`. `HttpExchange.onNetworkEnd` only aborted the request and left the open response untouched, so the application's response stream was never reset and was only reaped later by the inactivity timeout. This is inconsistent with HTTP/2, which resets the reply on `END` via cleanup.

## Change

- When the response is `OPEN` on network `END`, reset it and abort the network reply, mirroring the existing `onDecodeBodyInvalid` path. The `responseState == OPEN` guard preserves the valid HTTP half-close case (client closes its request side then reads the full response), which still completes via the existing `replyCloseOnFlush` path.
- Add server IT `rfc7230 .../client.close.during.response`: a streaming response is reset (write aborted) when the client closes the connection mid-response.
- Add the matching `rfc7540` (HTTP/2) scenario to lock in parity — HTTP/2 already handles this via `Http2Server` cleanup on network `END`; the test pins the behavior.

## Verification

`./mvnw clean verify -pl runtime/binding-http` is green. The new `shouldResetResponseWhenClientClosesDuringResponse` runs in both the rfc7230 and rfc7540 server `ConnectionManagementIT`s; all four `ConnectionManagementIT` classes pass.

### DIFF
--- a/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
+++ b/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpServerFactory.java
@@ -1900,7 +1900,7 @@ public final class HttpServerFactory implements HttpStreamFactory
 
                 if (exchange != null)
                 {
-                    exchange.onNetworkEnd(traceId);
+                    exchange.onNetworkEnd(traceId, authorization);
                 }
                 else
                 {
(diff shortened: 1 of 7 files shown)


jfallows (commented):
LGTM

PR #1851, opened 2026-06-08
feat(binding-mcp): SEP-1036 url-mode elicitation (passthrough + broker)

jfallows:
Implements SEP-1036 url-mode elicitation in the `mcp` binding, end to end, with a single north code path shared by passthrough and broker (guard/OAuth) topologies.

## What this adds

- **Passthrough url-mode elicitation** — the `mcp` server relays a server-initiated `elicitation/create` (mode `url`) to the client over the streaming response or the standalone session GET, relays the client's response back upstream, and relays `notifications/elicitation/complete`. Covers both **session-level** (lifecycle GET) and **request-coupled** (tool call / prompts get / resources read) elicitations.
- **Proxy aggregation** — the aggregating `mcp` proxy relays elicitation across toolkits, prefixing the JSON-RPC correlation id per toolkit so the response and OAuth callback route back to the originating downstream (replacing the prior `elicitClients` map / `context` breadcrumb).
- **Broker (guard/OAuth) convergence** — the north reacts to `elicitation/create` **identically** in broker and passthrough modes. Broker shrinks to its irreducible core: host the OAuth callback and layer routing onto the OAuth `state` (`sessionId.elicitationId.<guard-state>`). Discriminated by the create url's `redirect_uri` being anchored at the server's own callback. The auth-callback URL is plumbed north→south on the lifecycle BEGIN; the guard builds the DCR redirect_uri. Broker now honors decline (the gated tool call fails) instead of hanging.

## Notable internals

- `requestId` → `correlationId` rename (elicit-scoped only; generic request machinery untouched).
- North emits notification-style `elicitation/create` (no JSON-RPC id) when no correlation id is present (passthrough-remote), else a request with the id (quoted iff non-numeric).
- One uniform `notifications/elicitation/complete{elicitationId}`; dropped the broker-only complete/error data-events, the `status` field, and the `McpElicitStatus` enum.
- Supporting fixes folded in: `202 Accepted` for `notifications/initialized`, last-event-id propagation on session-level resume, standalone GET SSE relay, stop upstream keepalive on downstream suspend, bounded session-id minting, and guarding an unknown requestId in `notifications/cancelled`.

## Verification

`./mvnw clean verify` green for both modules: **runtime/binding-mcp 233 ITs** and **specs/binding-mcp.spec 212 ITs**, 0 failures, checkstyle and the coverage gate satisfied. Broker behavior is IT-covered via the engine test guard (no real OAuth guard required); the passthrough path is additionally exercised end-to-end by the `examples/mcp.proxy` demo (in a follow-up PR).


jfallows (commented):
LGTM

PR #1860, opened 2026-06-09
support guard injection for kafka client credentials

### LINKED ISSUE #1824: feat(binding-kafka): support guard injection for kafka client credentials
## Summary

Currently, Kafka client credentials (SASL username/password) are configured as static plaintext values in `zilla.yaml` under `options.sasl`. There is no way to resolve credentials dynamically — for example from a guard that manages token lifecycle, credential rotation, or AWS IAM-based authentication.

This issue tracks adding **guard injection** so that a guard can supply Kafka client credentials at connection time, following the same pattern already established in `binding-http`.

## Current behaviour

Credentials are configured statically in `KafkaSaslConfig`:

```yaml
options:
  sasl:
    mechanism: plain
    username: alice
    password: s3cr3t
```

`KafkaClientSaslHandshaker` reads `sasl.username` and `sasl.password` directly from the parsed config object. There is no hook for dynamic credential resolution.

## Desired behaviour

Allow a guard to be referenced in the Kafka binding options so credentials are resolved at handshake time:

```yaml
      options:
        authorization:
          guard0:
            credentials:
              mechanism: plain
              username: "{identity}"
              password: "{credentials}"
```

The guard would provide the username and password/token through the existing `GuardHandler` interface, enabling:

- **Token rotation** — short-lived tokens (e.g. AWS SigV4 / MSK IAM, OAuth) refreshed transparently without restarting Zilla
- **Credential vaulting** — credentials fetched from a secrets manager at runtime
- **Per-connection identity** — different credentials per downstream principal if needed

## Proposed implementation

Following the `binding-http` guard pattern:

1. **`KafkaOptionsConfig`** — add a `guard` reference field alongside `sasl` (analogous to `HttpOptionsConfig.authorization`)
2. **`KafkaSaslConfig` / builder** — allow `username`/`password` to be absent when a guard is wired
3. **`KafkaClientSaslHandshaker`** — inject `LongFunction<GuardHandler> supplyGuard`; at handshake time call the guard to resolve credentials instead of reading static config fields
4. **`KafkaClientMetaFactory` / `KafkaClientFetchFactory` / `KafkaClientProduceFactory`** — pass the guard supplier through (already receive `supplyBinding`)
5. **`KafkaOptionsConfigAdapter`** — read and write the new `guard` field

## Related

- `binding-http` guard wiring: `HttpAuthorizationConfig`, `HttpServerFactory.resolveGuard()`
- `KafkaClientSaslHandshaker.doEncodeSaslPlainAuthenticateRequest()` — the injection point for PLAIN
- SCRAM mechanisms will need the same injection at the corresponding encode methods

ankitk-me:
Fixes #1824

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/config/KafkaOptionsConfig.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/config/KafkaOptionsConfig.java
@@ -32,6 +32,7 @@ public final class KafkaOptionsConfig extends OptionsConfig
     public final List<KafkaTopicConfig> topics;
     public final List<KafkaServerConfig> servers;
     public final KafkaSaslConfig sasl;
+    public final KafkaAuthorizationConfig authorization;
 
     public static KafkaOptionsConfigBuilder<KafkaOptionsConfig> builder()
     {
@@ -48,13 +49,15 @@ public final class KafkaOptionsConfig extends OptionsConfig
         List<String> bootstrap,
         List<KafkaTopicConfig> topics,
         List<KafkaServerConfig> servers,
-        KafkaSaslConfig sasl)
+        KafkaSaslConfig sasl,
+        KafkaAuthorizationConfig authorization)
     {
         super(resolveModels(topics), List.of());
         this.bootstrap = bootstrap;
         this.topics = topics;
         this.servers = servers;
         this.sasl = sasl;
+        this.authorization = authorization;
     }
 
     private static List<ModelConfig> resolveModels(
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/config/KafkaSaslConfigBuilder.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/config/KafkaSaslConfigBuilder.java
@@ -26,7 +26,6 @@ public final class KafkaSaslConfigBuilder<T> extends ConfigBuilder<T, KafkaSaslC
     private String mechanism;
     private String username;
     private String password;
-    private long guardId;
 
     KafkaSaslConfigBuilder(
         Function<KafkaSaslConfig, T> mapper)
@@ -62,16 +61,9 @@ public final class KafkaSaslConfigBuilder<T> extends ConfigBuilder<T, KafkaSaslC
         return this;
     }
 
-    public KafkaSaslConfigBuilder<T> guardId(
-        long guardId)
-    {
-        this.guardId = guardId;
-        return this;
-    }
-
     @Override
     public T build()
     {
-        return mapper.apply(new KafkaSaslConfig(mechanism, username, password, guardId));
+        return mapper.apply(new KafkaSaslConfig(mechanism, username, password));
     }
 }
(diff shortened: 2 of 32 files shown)


runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/config/KafkaOptionsConfig.java
```diff
@@ -48,13 +49,15 @@ public static <T> KafkaOptionsConfigBuilder<T> builder(
         List<String> bootstrap,
         List<KafkaTopicConfig> topics,
         List<KafkaServerConfig> servers,
-        KafkaSaslConfig sasl)
+        KafkaSaslConfig sasl,
+        KafkaAuthorizationConfig authorization)
```
jfallows, line 53: I think we need to enforce that only one of these is specified (by schema), agree?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/config/KafkaSaslConfigBuilder.java
```diff
@@ -61,9 +62,16 @@ public KafkaSaslConfigBuilder<T> password(
         return this;
     }
 
+    public KafkaSaslConfigBuilder<T> guardId(
+        long guardId)
+    {
+        this.guardId = guardId;
+        return this;
+    }
+
```
jfallows: `guardId` is an internal concept, not part of external config, agree?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/config/KafkaBindingConfig.java
```diff
@@ -90,7 +102,25 @@ public KafkaTopicConfig topic(
...
+                sasl = options.sasl;
+            }
+            else if (options.authorization != null && guardId != 0L)
+            {
+                KafkaSaslCredentialsConfig credentials = options.authorization.credentials;
+                sasl = KafkaSaslConfig.builder()
+                    .mechanism(credentials.mechanism)
+                    .username(credentials.username)
+                    .password(credentials.password)
+                    .guardId(guardId)
```
jfallows: In `MqttBindingConfig` we store the `guard` handler directly. Perhaps we can do the same here?

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/schema/kafka.schema.patch.json
```diff
@@ -182,7 +182,12 @@
...
                                     "sasl":
                                     {
-                                        "$ref": "#/$defs/options/binding/kafka/sasl"
+                                        "$ref": "#/$defs/options/binding/kafka/sasl",
+                                        "deprecated": true
+                                    },
+                                    "authorization":
+                                    {
+                                        "$ref": "#/$defs/options/binding/kafka/authorization"
                                     }
```
jfallows, line 190: Only oneOf these.

jfallows (changes requested):

jfallows (approved):

PR #1872, opened 2026-06-11
Add common-avro module for streaming Avro codec

jfallows:
## Description

Introduces `common-avro`, a format-native, provider-free streaming Avro wire codec — a peer to
`common-json` and `common-protobuf` — for high-performance decode and canonical re-encode over Agrona
`DirectBuffer`s, with no full-document buffering and zero per-message allocation. It ships its own codec
and needs no Apache Avro library at runtime.

### Key Features

- **Schema-bound pull parser** (`AvroParser`/`AvroParserImpl`): walks a compiled schema in lockstep with
  the wire, emitting a typed `AvroEvent` stream with zero-copy value views over the caller's buffer.
- **Push pipeline** (`Avro.stream(parser).transform(…).into(sink)`): composes `AvroTransform` stages onto
  a terminal `AvroSink` (`AvroSchema#validator()` is a stage), backed by a buffer-bounded `AvroGenerator`
  for canonical re-encode.
- **Caller-owned zero-copy input**: the parser reads in place from the caller's window (no internal work
  buffer); `AvroPipeline#position()` is the consumed watermark the caller slides by.
- **Values larger than the buffer, streamed both directions**: a `STRING`/`BYTES`/`FIXED` value too large
  for the input window arrives as a run of the same event with `deferredBytes()` counting the remainder
  down to `0` (UTF-8-boundary-safe, so `getString()` never splits a character); the bounded
  `AvroGenerator` drains its output the same way via `writeSegment`.
- **Two back-pressure axes**: `STARVED` (input window consumed mid-datum) and `SUSPENDED` (bounded output
  full); a `last` flag distinguishes more-input-coming from a truncated datum (`REJECTED`).
- **Verbatim passthrough** via a single `SEGMENT` event (raw run bounded by `deferredBytes()` `UNBOUNDED`
  → `0`), opted into per `AvroSink.Delivery.SEGMENTABLE`.
- **Introspection SPI**: `AvroType`/`AvroField`/`AvroKind` expose named types, unions, logical types
  (decimal precision/scale), aliases, and defaults.

### Implementation Details

- `AvroParserImpl` is a pure cursor; `AvroPipelineImpl` owns the read-only `AvroSource` view and the
  `AvroController` stages see, driving the parser through the public interface with no downcast.
- `AvroSchemaCompiler` compiles JSON schemas into immutable `AvroNode` trees, cached per schema id.
- `AvroGeneratorImpl` frames arrays and maps as single-element blocks so values stream without buffering;
  `writeSegment`/`flush` stream a length-delimited value across a bounded output.

### Testing

- `AvroParserTest`, `AvroParserPullTest`, `AvroTypeTest`, `AvroSchemaTest`: event stream, direct pull API,
  type introspection, and schema compilation.
- `AvroGeneratorTest`: encoding and bounded-generator behavior.
- `AvroFragmentedParserTest`, `AvroChunkingTest`, `AvroValueStreamingTest`: windowed/fragmented input,
  bounded-output chunking, and value streaming (input + output, UTF-8 boundaries, `deferredBytes`,
  `STARVED`, truncation).
- `AvroMalformedTest`: malformed-binary rejection.
- `AvroConformanceTest`, `AvroInteroperabilityTest`, `AvroRoundTripTest`: conformance and round-trip
  against the Apache Avro reference implementation.
- `bench/AvroPipelineBM`: JMH; confirms zero per-message allocation under `-prof gc`.

https://claude.ai/code/session_01RcCcqXRpp2dRvSgTqPXzpY


runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/internal/AvroGeneratorImpl.java
```diff
@@ -0,0 +1,392 @@
...
+    private int bound;
+    private AvroNode[] nodeStack;
+    private int[] stateStack;
+    private int depth;
+    private int progress;
+
+    public AvroGeneratorImpl(
+        AvroSchema schema,
+        MutableDirectBuffer buffer,
+        int offset)
```
jfallows: Do we need these in the constructor, or can we just call.wrap(...) and change the wrap return type to AvroGenerator(Impl)? Then `retarget` can be inlined.

jfallows: Done. The constructor now calls `wrap(buffer, offset, buffer.capacity() - offset)`, `wrap` returns `AvroGenerator` (fluent, matching `ProtobufGenerator`), and `retarget` is inlined into it.

---

runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/internal/AvroGeneratorImpl.java
```diff
@@ -45,16 +45,32 @@ public AvroGeneratorImpl(
         this.nodeStack = new AvroNode[16];
         this.stateStack = new int[16];
         this.depth = 0;
-        retarget(buffer, offset, buffer.capacity());
+        wrap(buffer, offset, buffer.capacity() - offset);
```
jfallows: I meant at call site, not directly from inside constructor.
Also this is wrong, limit is absolute (not relative) so it should just pass buffer.capacity() as (maximum) limit.

jfallows: Fixed in `7cc476e1`. The constructor no longer wraps; `Avro.generator` wraps at the call site via `new AvroGeneratorImpl(schema).wrap(buffer, offset, buffer.capacity())`. And `wrap` now treats `limit` as the absolute bound (`this.bound = limit`), so the factory passes `buffer.capacity()` as the maximum.

---

jfallows (commented):
LGTM

jfallows (commented):
LGTM

PR #1877, opened 2026-06-13
feat(binding-mcp): validate tools/call arguments at the proxy

### LINKED ISSUE #1962: feat(binding-mcp): validate tools/call arguments at the proxy against cached tools/list inputSchema
## Problem

The `mcp proxy` forwards `tools/call` requests upstream without checking the call `arguments` against the tool's advertised JSON Schema. MCP tools publish an `inputSchema` in their `tools/list` entry, but today that schema is ignored, so malformed tool calls reach upstream unchecked. We want the gateway to reject structurally-invalid `tools/call` arguments at the edge with a JSON-RPC error, before fan-out.

## Proposed design

Validate in the `mcp proxy` binding, reusing its existing list cache as the schema source:

1. **Schema source = the proxy list cache.** The proxy already hydrates and refreshes `tools/list` from upstream (`cache.ttl` + `notifications/tools/list_changed`, store-backed). When the tools cache settles/refreshes, parse it once and build a `prefixedToolName → inputSchema` index, registering each schema into an in-memory, content-addressed catalog (CRC32C); rebuild on cache change.
2. **Apply on `tools/call`.** Resolve the called tool's schema from the index. Buffer the `arguments`, validate the complete value against the schema, then:
   - **valid** → forward upstream unchanged;
   - **invalid** → reply with JSON-RPC `-32602 "Invalid params"` correlated to the request id, and do **not** forward upstream.
3. **Best-effort, non-gating.** A `tools/call` for a tool not present in the cache passes through unvalidated (validation never blocks on the cache); upstream remains the authority.

## Why the proxy

The cache already settles the hard parts:

- **Freshness** — it actively refetches from upstream, so schemas are never stale (the consumer-facing edge cannot refetch on its own).
- **Scope** — a clear identity-independent, per-binding scope with per-toolkit name prefixing, so tool names are globally unique within the merged list.
- **Durability / shareability** — store-backed, survives gateway restart, and is shared across replicas.

An `inputSchema` is a property of the tool, not the caller, so cache-sourced validation is consistent for every session. Per-user tool visibility/authorization stays where it already is (route authorization + guard); this feature is a structural-correctness gate, not an authorization gate.

## Config surface

Add a `tools` validation block to the proxy `options`, referencing an in-memory catalog populated at runtime from the cache:

```yaml
catalogs:
  tools_schema:
    type: inline          # no subjects; populated at runtime from the cache
bindings:
  proxy0:
    type: mcp
    kind: proxy
    options:
      cache:
        store: cache
        ttl: PT5M
      tools:
        model: json
        catalog:
          tools_schema:
            - strategy: tool   # schema selected by (prefixed) tool name
    routes:
      - exit: mcp_client0
        when:
          - toolkit: everything
            capability: [tools]
```

Notes:
- The `inline` catalog must accept an empty/absent `subjects` list and support runtime `register`/`unregister` (content-addressed, refcounted) so the proxy can populate it from the cache.
- `strategy: tool` is a binding-interpreted marker meaning "schema selected by tool name"; schemas resolve through the normal content-addressed catalog path, so no engine/model SPI change is required.

## Out of scope

- No `maxAge`/expiry knob — freshness comes from the cache's own refresh (`ttl` + `tools/list_changed`).
- No engine/SPI change.

## Acceptance criteria

- `tools/call` with `arguments` that violate the cached `inputSchema` → JSON-RPC `-32602`, not forwarded upstream.
- Valid `arguments` → forwarded upstream unchanged.
- Unknown / not-yet-cached tool → pass-through (best-effort).
- Cache refresh / `tools/list_changed` re-indexes the schemas; catalog refcounts stay balanced across re-index.

jfallows:
## Summary

Validate `tools/call` arguments against each tool's advertised JSON Schema (`inputSchema`) in the mcp **proxy** binding, rejecting structurally-invalid calls at the edge before fan-out. Schemas are sourced from the proxy's existing tools/list cache, so freshness, scope, and durability come from the cache for free. This supersedes the earlier server-side approach on this branch — validation now lives on the proxy.

Closes #1962.

## How it works

- **Schema source = the proxy list cache.** When the tools cache settles/refreshes, the merged `tools/list` blob is parsed once and a `prefixedToolName → schemaId` index is built, registering each `inputSchema` into a content-addressed in-memory `inline` catalog. `McpProxyCache.ListChangedListener` is widened to hand the settled value to the rebuild.
- **Apply on `tools/call`.** For a tool present in the index, the proxy buffers the args, validates the complete value against the schema (reusing the existing `ModelPipeline`), then:
  - **valid** → forwards upstream unchanged;
  - **invalid** → resets the request stream with `McpResetEx KIND_ERROR(-32602, "Invalid params")` and aborts the upstream client. The fronting mcp server translates that reset into a client-facing JSON-RPC error (HTTP 200 + `-32602` body) via its existing `onAppReset`/`doNetRejectError` path — no new server code.
- **Best-effort, non-gating.** A `tools/call` for a tool not in the cache passes through unvalidated; upstream remains the authority.

No engine/SPI change — `strategy: tool` is interpreted by the binding and schemas resolve through the normal content-addressed catalog path. The `options.tools` validator surface moved from the `server` kind to the `proxy` kind in the schema patch.

### Config surface

```yaml
catalogs:
  tools_schema:
    type: inline          # populated at runtime from the cache
bindings:
  proxy0:
    type: mcp
    kind: proxy
    options:
      cache: { store: cache }
      tools:
        model: json
        catalog:
          tools_schema:
            - strategy: tool
    routes: [ ... ]
```

## Tests

- New `proxy.validation.yaml` + `cache.tools.call.{valid,invalid,no.schema}.input` scenarios, covered by `McpProxyCacheIT` (valid forwards, invalid → `-32602` with no args reaching upstream, uncached tool passes through).
- Removed the obsolete server-side validation scripts/tests (server no longer validates).
- Full `binding-mcp` module green: McpServerIT 80, McpProxyCacheIT 34, McpProxyIT 52, McpProxyLifecycleIT 7, McpClientIT 64, unit 60; spec ApplicationIT 115, NetworkIT 80; checkstyle clean; jacoco satisfied.

## Drive-by fix (separate commit)

Fixes three long-failing `McpServerIT` cases (`shouldRejectRequestParamsBeforeMethod`, `shouldRejectRequestParamsWithArray`, `shouldRejectToolsCallWithoutContentLength`). These predate this work and failed because `onDecodeParseError`/`onDecodeInvalidRequest` responded with an HTTP **400**. For MCP over Streamable HTTP the transport succeeded, so a malformed/invalid JSON-RPC request is now reported as a **valid JSON-RPC error response over HTTP 200** with the error in the body (`-32700 "Parse error"` / `-32600 "Invalid request"`) — consistent with the proxy `-32602` path and the upstream-error path. Updates the three `reject.*` script pairs (network client + server peers) to match; the unified contract is "every JSON-RPC error → HTTP 200 + error body".

https://claude.ai/code/session_015SCFtykHvSyqMSqe5VCBYo


runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
```diff
@@ -119,6 +149,206 @@ public McpBindingConfig(
...
+    public boolean validatesTools()
+    {
+        return toolsCatalog != null;
+    }
+
+    public int registerToolSchema(
+        String subject,
+        String schema)
+    {
+        return toolsCatalog != null ? toolsCatalog.register(subject, schema) : NO_SCHEMA_ID;
```
jfallows: Can we use a no-op CatalogHandler to skip the null checks?

runtime/binding-mcp/pom.xml
```diff
@@ -46,6 +46,18 @@
       <version>${project.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>model-json</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
```
jfallows: Is this the correct scope?
We don't compile directly against `model-json` APIs, agree?

runtime/binding-mcp/pom.xml
```diff
@@ -46,6 +46,18 @@
...
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>catalog-inline</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
```
jfallows: We should be using test catalog, not inline catalog for testing, agree?

runtime/catalog-inline/src/main/java/io/aklivity/zilla/runtime/catalog/inline/internal/InlineCatalogHandler.java
```diff
@@ -25,17 +25,46 @@
...
+        String subject)
+    {
+        Integer schemaId = schemaIds.remove(subject + VERSION_LATEST);
+        int[] removed = new int[0];
+        if (schemaId != null)
+        {
+            release(schemaId);
+            removed = new int[] { schemaId };
+        }
+        return removed;
```
jfallows, line 71: Should unregister not just return int instead of int[] ?

runtime/catalog-inline/src/main/java/io/aklivity/zilla/runtime/catalog/inline/internal/InlineCatalogHandler.java
```diff
@@ -25,17 +25,46 @@
 
 public class InlineCatalogHandler implements CatalogHandler
 {
+    private static final String VERSION_LATEST = "latest";
+
     private final Map<Integer, String> schemas;
     private final Map<String, Integer> schemaIds;
+    private final Map<Integer, Integer> references;
```
jfallows: Check agrona data structures to avoid auto-boxing.

specs/binding-mcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mcp/config/proxy.validation.yaml
```diff
@@ -0,0 +1,44 @@
...
+catalogs:
+  tools_schema:
+    type: inline
+stores:
+  memory0:
+    type: test
+    options:
+      entries:
+        tools: |-
+          {"tools":[{"name": "get_weather","title": "Weather Information Provider","description": "Get current weather information for a location","inputSchema": {"type": "object","properties": {"location": {"type": "string","description": "City name or zip code"}},"required": ["location"]},"icons": [{"src": "https://example.com/weather-icon.png","mimeType": "image/png","sizes": ["48x48"]}],"execution": {"taskSupport": "optional"}}]}
```
jfallows: Let's format this multi-line for improved readability.

specs/binding-mcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mcp/config/proxy.validation.yaml
```diff
@@ -0,0 +1,44 @@
...
+        resources: |-
+          {"resources":[{"uri": "file:///docs/welcome.md","name": "welcome","description": "Welcome document","mimeType": "text/markdown"}]}
```
jfallows: Let's format this multi-line for improved readability.

specs/binding-mcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/mcp/config/proxy.validation.yaml
```diff
@@ -0,0 +1,44 @@
...
+        prompts: |-
+          {"prompts":[{"name": "summarize","description": "Summarize a document"}]}
```
jfallows: Let's format this multi-line for improved readability.

runtime/catalog-inline/src/main/java/io/aklivity/zilla/runtime/catalog/inline/internal/InlineCatalogHandler.java
jfallows: Kept as `int[]` — it matches the existing `CatalogHandler.unregister` SPI signature, which is also implemented by `SchemaRegistryCatalogHandler` where a single subject can have multiple versions and the Confluent `DELETE /subjects/{subject}` returns the full array of removed version ids. Narrowing to `int` would lose that for schema-registry; inline only ever removing 0/1 is just a degenerate case of the same contract. Happy to revisit as a separate SPI-wide change if you'd prefer.

---

runtime/binding-mcp/pom.xml
```diff
@@ -50,12 +50,6 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>model-json</artifactId>
       <version>${project.version}</version>
```
jfallows, line 52: This should be removed also as we've removed the compile time dependency, agree?

jfallows: Keeping it at `test` scope (per follow-up). The compile-time coupling is gone — binding-mcp no longer references `JsonModelConfig` — but the proxy validation ITs (`McpProxyCacheIT`) exercise the real `json` model at runtime to validate tool arguments, so `model-json` is still needed on the test classpath. Removing it entirely makes those scenarios fail (no `json` model registered → no validation). At production runtime the model is provided by the engine image.

---

PR #1879, opened 2026-06-14
Add escape mode to JSON generator for JSON-in-JSON encoding

### LINKED ISSUE #1878: common-json: streaming JSON-in-JSON escaping (generator escape mode + consumption-driven segment writing)
## Motivation

When a binding needs to embed a projected JSON document as the **content of a JSON string** in another document — JSON-in-JSON — it currently has to escape the projected bytes itself, outside common-json. The `mcp-http` proxy does exactly this for the resource-read reply envelope:

```
{"contents":[{"uri":"…","mimeType":"…","text":"<projected JSON, escaped as a string>"}]}
```

Today the binding (`McpHttpProxyFactory`) hand-rolls the JSON string escaping (`stageEscaped` + `jsonString` + `hex`): the per-stream generator serializes the projected value into a scratch buffer (`projectBuffer`), then the binding copies it into the encode slot escaping byte-by-byte, gating encode-slot room with a conservative `ESCAPE_MAX = 6` worst-case factor.

Two problems with that:

1. **Escaping is JSON serialization logic living outside the library that owns JSON serialization.** The generator already escapes string *values* (`"`→`\"`, controls→`\uXXXX`); the binding re-implements the same table.
2. **It forces an intermediate buffer.** Because the generator can't escape, it must write unescaped bytes somewhere first, then the binding escape-copies into the slot. The request leg (raw body, no escaping) can wrap the encode slot directly; the response leg can't, purely because of this gap. (A naive attempt to drop the intermediate without solving escaping-in-the-generator regressed — see notes below.)

## Proposal

Give `common-json` a way to emit a generator's entire output **escaped as JSON string content**, so a consumer can do `projector → generator(escape) → encode slot` directly, with no intermediate buffer and no hand-rolled escaping. Capture two related pieces:

### 1. Generator "escape output" config option

An opt-in mode (e.g. set at `JsonGeneratorEx.wrap(...)`, defaulting off) where **every byte the generator emits is escaped** as it is written — structural bytes (`{ } : , [ ]`) pass through, `"` / `\` / control chars are escaped. This composes correctly with the generator's existing value-escaping: a value's `"` is first turned into `\"` by `writeString`, then the escape layer escapes those into `\\\"`, which is exactly the correct nested representation. So "escape the whole output stream" == "serialize this document as an escaped JSON string", which is what JSON-in-JSON needs.

The surrounding quotes and outer envelope are written by the caller; the escape-mode generator emits only the inner (escaped) content.

### 2. Consumption-driven segment writing

The blocker for escaping inside the generator is that **escaping is not length-preserving**, while the current bounded-output/`SUSPENDED` fragmentation contract assumes 1 input byte → 1 output byte. `JsonSinkImpl.writeChunk` slices the source by the generator's output `remaining()`:

```java
int length = Math.min(sliceLength - segmentWritten, generator.remaining());
generator.writeSegment(segment, segmentWritten, length, deferred);
```

With escaping, `length` source bytes can need up to ~6× that in output and overflow `limit`. The fix is to make the escape-aware segment path **consumption-driven**: the generator writes as many *source* bytes as fit the output limit (escaping as it goes) and reports how much it consumed / sets `deferred` itself, rather than the sink pre-slicing by an assumed 1:1 ratio.

## Implementation is intentionally open

Both pieces are described by **contract/intent**, not mechanism — leaving freedom on how to realize them. For example the escape mode could be a flag funneled through a single `putByte`/`putBytes` choke point, or a dedicated escaping byte-sink; the consumption-driven cut could change the `writeSegment` return contract, introduce a `writeStringSegment`-style entry point, or be handled entirely inside an escape-mode sink. No preference is being locked in here.

## Acceptance / test gate

- A chunking test mirroring `JsonPipelineChunkingTest` / `JsonProjectorScalarSegmentTest` but with **escape-heavy content** (values full of `"`, `\`, and control chars) driven through **tiny output windows**, so the non-1:1 ratio and the resume/`SUSPENDED` cascade are exercised. This is precisely the case that a botched ratio would corrupt.
- A test embedding a projected sub-document as an escaped string end-to-end (round-trips: escaped output re-parses to the original document).

## Downstream cleanup (separate, follow-up)

Once this lands, `binding-mcp-http` can:
- delete `stageEscaped` / `jsonString` / `hex` and the `ESCAPE_MAX` room heuristic,
- wrap the per-stream generator **directly over the encode slot** on the response leg (matching the request leg), dropping the `projectBuffer` intermediate entirely.

(Independently, `McpHttpProxyFactory.onUpstreamResponse` / `text(...)` should null-guard `body` so an aborted upstream yields a clean tool error / reply-abort instead of an NPE — noted here only as related context; not part of this issue.)

jfallows:
## Description

Adds an **escape mode** to `JsonGeneratorEx` that enables JSON-in-JSON encoding by escaping every emitted byte as JSON string content. When enabled, structural bytes and the generator's own value-escaping compose into the correct nested form, allowing documents to be serialized as escaped string values within outer JSON envelopes.

Fixes #1878 

### Key Changes

1. **New escape mode parameter** (`JsonGeneratorEx.wrap(buffer, offset, limit, escape)`):
   - When `escape=true`, every byte is escaped as JSON string content
   - Structural bytes (`{}:,[]`) and UTF-8 continuation bytes pass through
   - Quote, backslash, and control characters are properly escaped
   - Composes with existing value-escaping for correct JSON-in-JSON form

2. **Consumption-driven segment writing** (`writeSegment` redesign):
   - Changed from deferred-byte reporting to **consumption-driven** semantics
   - Returns the number of source bytes consumed (not output bytes)
   - Handles non-1:1 expansion ratio when escaping (one source byte → up to 6 output bytes)
   - Enables proper fragmentation of large values across bounded output windows

3. **Escape implementation**:
   - Added `escapeByte()` method handling all JSON escape sequences
   - Added `escapedWidth()` to calculate output expansion for each byte
   - Centralized byte output through `putByte()` to apply escaping uniformly
   - Supports short escapes (`\"`, `\\`, `\n`, etc.) and unicode escapes (`\uXXXX`)

4. **Updated `JsonSinkImpl`** to work with consumption-driven semantics:
   - Passes entire remaining slice to generator
   - Defers unconsumed bytes when output bound is reached
   - Properly resumes from deferred position on next write

### Testing

Added comprehensive test coverage:
- **`JsonGeneratorEscapeTest`**: Core escape functionality (quotes, backslashes, control chars, UTF-8, round-trip validation)
- **`JsonGeneratorEscapeChunkingTest`**: Escape mode under tiny output windows with non-1:1 expansion ratio, including projector transforms and end-to-end JSON-in-JSON embedding

All tests verify the defining property: escaping a document then decoding the result as JSON string content yields the document's plain serialization.

### Documentation

Updated `README.md` to reflect consumption-driven semantics and escape mode capabilities.

https://claude.ai/code/session_01X6xJ3m3xF5Q8Jn5fj5EziR


runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonGeneratorImpl.java
```diff
@@ -369,36 +384,18 @@ public JsonGeneratorImpl writeRaw(
...
-        int index,
-        int length,
-        int deferred)
-    {
-        assert progress + length <= limit;
-        buffer.putBytes(progress, source, index, length);
-        progress += length;
-        this.deferred = deferred;
-        return this;
+        return segmentWriter.write(source, index, length);
```
jfallows:
```suggestion
        return writeSegment.accept(source, index, length);
```

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonGeneratorEx.java
```diff
@@ -93,29 +93,20 @@ JsonGeneratorEx writeRaw(
...
+     * source byte may expand to several output bytes — and <em>returns the number of source bytes
+     * consumed</em>. A driver fragments a value larger than what fits by writing what it can, draining when
+     * the returned count is less than {@code length}, then continuing on resume from the unconsumed
+     * remainder. In the non-escape case the returned count is {@code min(length, remaining())}.
      */
-    JsonGeneratorEx writeRawContinue(
+    int writeSegment(
         DirectBuffer source,
         int index,
         int length);
```
jfallows: It's a shame to lose this fluent-style API for writeSegment.
Perhaps we can restore it but gain the information needed via length() instead?

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonGeneratorImpl.java
jfallows: Applied — the strategy field is now `writeSegment` with an `accept(...)` method, parallel to `putByte.accept(...)`.

---

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonGeneratorEx.java
jfallows: Restored the fluent `writeSegment` (returns the generator). Used `length()` won't work here, though: it's output-domain, and in escape mode one source byte expands to up to 6 output bytes, so it diverges from the *source* bytes the sink must advance by. Instead added `int consumed()` — a cumulative counter, the source-domain counterpart to `length()` (reset by `wrap`); the sink reads it as a delta around the call. Also relocated the config keys onto `JsonParserEx`/`JsonGeneratorEx` with a `JsonConfigEx` aggregator per the other thread.

---

jfallows (commented):
LGTM

PR #1880, opened 2026-06-14
Support fragmented JSON values across input windows

jfallows:
## Description

This change refactors the JSON tokenizer to support streaming large values (strings and numbers) that exceed a single input window without requiring reassembly across frames. Previously, the parser would rewind and retry when a value didn't fit in the current window. Now it can deliver values as fragments, enabling efficient streaming of arbitrarily large JSON values.

### Key Changes

**JsonTokenizer.java**
- Removed path filtering logic (`pathIncludes`, `pathExcludes`, `pathFiltering`) — this feature is no longer supported
- Removed `tokenMaxBytes` configuration parameter
- Added `windowLength` field to track the current input window size
- Added `numberLexeme` StringBuilder to accumulate fragmented number lexemes across windows
- Added `fragmenting` and `numberFragmented` flags to track fragmentation state
- Added `fragmentStart` and `unitStartOffset` to track fragment boundaries
- Replaced `valueReadable()` with `fragmenting()` to indicate ongoing fragmented delivery
- New `window(int length)` method to set the input window size per frame
- New `onScalarStarved()` method implementing the fragmentation logic:
  - For strings: rewinds to the last complete code-point/escape boundary and delivers decoded chars so far
  - For numbers: accumulates the lexeme and delivers digits scanned so far
  - For values fitting in a window: rewinds to reassemble whole in the next window
- Simplified constructor signatures (removed path and tokenMaxBytes parameters)

**JsonStrings.java** (new file)
- Extracted string unescaping logic into a reusable utility class
- Decodes JSON string bodies (between quotes) with escape sequence handling
- Supports UTF-8 multibyte characters and Unicode escapes

**JsonParserImpl.java**
- Removed path filtering configuration handling
- Added `window(int length)` call in `wrap()` to inform tokenizer of window size
- Added `position()` method to expose current stream offset
- Updated `deferredBytes()` to include `tokenizer.fragmenting()` state
- Removed `valueReadable()` check from `getString()` (path filtering removed)

**JsonGeneratorEx.java & JsonGeneratorImpl.java**
- Added `Completion` enum to distinguish incomplete vs. complete fragments
- New `write(CharSequence, Completion)` method for deferred string fragments
- New `writeNumber(CharSequence, Completion)` method for deferred number fragments
- Generator now tracks `stringOpen` and `numberOpen` state to handle fragmented values

**JsonSinkImpl.java**
- Updated to handle `Delivery.DECODED` mode for fragmented values
- Renders decoded values from fragments without concatenation
- Maintains verbatim segment delivery for `Delivery.STRUCTURED` mode

**Test Changes**
- Added `JsonStringsTest` covering string unescaping with escapes, Unicode, UTF-8 multibyte, and offset handling
- Added `JsonPipelineChunkingTest` tests for fragmented string/number reconstruction across windows
- Removed path filtering tests from `JsonTokenizerPathTest` (feature removed)
- Updated remaining path tests to use simplified tokenizer constructor

### Motivation

This enables efficient streaming of large JSON values without buffering them whole in memory. Values that exceed the input window are delivered as fragments, allowing the parser to process them incrementally while maintaining correct semantics for both decoded (unescaped) and verbatim (raw bytes) delivery modes.

### Breaking Changes

- Removed `JsonParserEx.PATH_INCLUDES` and `JsonParserEx.PATH_EXCLUDES` configuration options
- Removed `JsonTokenizer` constructors taking path and tokenMaxBytes parameters
- Removed `valueReadable()` method from tokenizer (replaced with `fragmenting()`)

https://claude.ai/code/session_01DCYLhQmFrTkg4UZaxZ6wvp

### DIFF
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonSink.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonSink.java
@@ -28,13 +28,18 @@ public interface JsonSink
 {
     /**
      * Delivery mode a terminal sink requests. {@link #STRUCTURED} consumes structured events and
-     * renders normalized output; {@link #SEGMENTABLE} opts in to verbatim segment delivery for kept
-     * values (best-effort, demand-gated) by calling {@link JsonController#segmentable()}.
+     * renders output by splicing each value's verbatim token bytes (chunked across bounded output);
+     * {@link #SEGMENTABLE} opts in to verbatim segment delivery for kept values (best-effort,
+     * demand-gated) by calling {@link JsonController#segmentable()}; {@link #DECODED} renders each
+     * scalar from its decoded {@code getString()} via {@link JsonGeneratorEx#write(CharSequence,
+     * boolean)}, letting the generator own quoting/escaping so a value delivered as fragments forms a
+     * single string without the sink concatenating.
      */
     enum Delivery
     {
         STRUCTURED,
-        SEGMENTABLE
+        SEGMENTABLE,
+        DECODED
     }
 
     JsonPipeline.Status feed(
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonSink.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonSink.java
@@ -14,8 +14,6 @@
  */
 package io.aklivity.zilla.runtime.common.json;
 
-import io.aklivity.zilla.runtime.common.json.internal.JsonSinkImpl;
-
 /**
  * The consume end of a {@link JsonStream} pipeline. Each {@link #feed(JsonController, JsonSource, JsonEvent)}
  * delivers one event (with {@code source} positioned to read its scalar, or its bytes when the event is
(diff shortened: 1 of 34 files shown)


runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonGeneratorImpl.java
```diff
@@ -56,6 +57,8 @@ public final class JsonGeneratorImpl implements JsonGeneratorEx
     private int depth;
     private int consumed;
     private boolean afterKey;
+    private boolean stringOpen;
+    private boolean numberOpen;
```
jfallows: Perhaps we can capture these 3 booleans as a single internal state, since only one of them can be true at a time.

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonSink.java
```diff
@@ -28,13 +28,18 @@ public interface JsonSink
...
+     * scalar from its decoded {@code getString()} via {@link JsonGeneratorEx#write(CharSequence,
+     * boolean)}, letting the generator own quoting/escaping so a value delivered as fragments forms a
+     * single string without the sink concatenating.
      */
     enum Delivery
     {
         STRUCTURED,
-        SEGMENTABLE
+        SEGMENTABLE,
+        DECODED
```
jfallows, line 46: `DECODED` feels like it doesn't need to be exposed in the public API and can be handled via source or generator state instead.

jfallows (commented):
LGTM

PR #1832, opened 2026-06-04
feat(binding-mcp-http): mcp_http · proxy binding

jfallows:
## Summary

New **`mcp_http · proxy`** binding — accepts `mcp` streams, produces `http` streams. It is the spec-free runtime primitive that `mcp_openapi` (#1673) compiles down to: it *terminates* `tools/list` / `resources/list` / `prompts/list` from config and *expands* `tools/call` / `resources/read` into upstream `http` requests, while *rendering* `prompts/get` from config templates. It also participates in the MCP **session lifecycle** — minting a worker-aligned session id on `initialize` and enforcing it on every subsequent request. This PR delivers the spec module **and** the runtime implementation.

## What's here

- **Spec** `specs/binding-mcp-http.spec/` (registered in `specs/pom.xml`), modeled on `binding-http-kafka.spec` — protocol-named `.rpt` script folders reusing `mcp:`/`http:` functions (no own IDL/Functions). **McpServerIT 19, HttpClientIT 10**, plus **SchemaTest 7** for config validation.
- **Runtime** `runtime/binding-mcp-http/` — `McpHttpProxyFactory` cross-protocol proxy: route resolution by `tool`/`resource`, `${args.x}`/`${params.x}` path interpolation, query/body projection, `${result.x}` result-summary templates, and upstream 2xx / error-body / ABORT propagation. **McpHttpProxyIT 26.**
- **Bounded streaming, both legs** — request and response stream through fixed `decodeSlot`/`encodeSlot` using `common-json`'s `SEGMENTABLE` delivery with consumed-pushback, so a single MCP message far larger than one buffer slot flows through `projector → sink → generator` with no whole-message accumulation and no per-fragment allocation. Verified end-to-end at an 8K slot capacity with the 10k **and 100k** flow-control scenarios passing. (This replaces the previously deferred data-path streaming; its `common-json` prerequisite — append-only consumed pushback — merged in #1882 and is now on `develop`.)
- **Lifecycle & session** — on the lifecycle stream the proxy mints a **worker-aligned session id** (reject-sampling against the local worker index), holds it in a per-worker session map, and replies with the lifecycle extension carrying the minted id. Every `tools`/`resources`/`prompts` request must carry a session id established via a prior lifecycle exchange; unknown or absent session ids are **rejected before route resolution**. `McpHttpConfiguration` exposes `session.id` (supplier) and `session.id.attempts`.
[...]

PR #1881, opened 2026-06-15
fix(common-json): project kept scalar leaves fragmented across input windows

jfallows:
## Summary

`JsonProjectorImpl.onScalar` finalized a kept (or dropped) scalar value on its **first** event — decrementing `depth` / setting `rootDone` — without checking `source.deferredBytes()`. Under `Delivery.DECODED` (added in #1880), an over-window scalar is fragmented into multiple `VALUE_STRING`/`VALUE_NUMBER` events, so finalizing on the first fragment corrupted the projector's descent bookkeeping (`depth` driven negative → `ArrayIndexOutOfBoundsException`). As a result a **retained scalar leaf larger than the input window could not be projected** — the projector only streamed kept *containers* (`onAwaiting`/`onForwarding`, which already check `deferredBytes()`), never a kept scalar *leaf*.

## Fix

Make `onScalar` fragment-aware, mirroring the container segment paths:

- track the in-flight scalar with `scalarPending` / `scalarEmit`;
- on the **first** fragment: enter the value, decide keep/drop, forward the kept key + fragment;
- on **continuation** fragments: forward (if kept) without re-entering;
- account for the consumed value **once**, on its closing fragment (`!deferredBytes()`), via `finishScalar()`.

This applies equally to kept and dropped scalar leaves (a dropped over-window scalar must also consume all its fragments without re-entering) and to scalar array elements.

## Tests

Added `JsonProjectorSegmentTest#shouldDecodeKeptScalarLeafFragmentedAcrossWindows`: a retained scalar leaf (and a dropped sibling) far larger than the input window, fed in 8-byte windows under `DECODED`, driven with the `pipeline.position()` watermark.

- Before the fix: `ArrayIndexOutOfBoundsException: Index -4` (depth corruption).
- After the fix: passes; output is the pruned document with the leaf rejoined.

Full `common-json` build green — 4768 tests, checkstyle, license, coverage.

## Motivation

This is the missing projector capability behind the deferred large-payload streaming for `binding-mcp-http` (#1832): both proxy legs prune a document to schema paths while one retained leaf (an order `status`, a PR `body`) far exceeds one buffer slot. With this fix the binding can drive a plain `projector → DECODED` pipeline for fragmented scalar leaves.

Refs #1880.

https://claude.ai/code/session_01AEpU8rooKSEqf6EuU8ri1i

---
_Generated by [Claude Code](https://claude.ai/code/session_01AEpU8rooKSEqf6EuU8ri1i)_

### DIFF
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonProjectorImpl.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonProjectorImpl.java
@@ -72,6 +72,8 @@ public final class JsonProjectorImpl implements JsonTransform
     private Status downstream;
     private SegMode segMode = SegMode.NONE;
     private JsonEvent deferredStart;
+    private boolean scalarPending;
+    private boolean scalarEmit;
 
     private enum SegMode
     {
@@ -99,6 +101,8 @@ public final class JsonProjectorImpl implements JsonTransform
         downstreamDemand = false;
         segMode = SegMode.NONE;
         deferredStart = null;
+        scalarPending = false;
+        scalarEmit = false;
     }
 
     private void onDownstreamSegmentable()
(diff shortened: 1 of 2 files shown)


jfallows (commented):
LGTM

PR #1882, opened 2026-06-15
feat(common-json): stream kept scalar leaves via append-only consumed pushback

jfallows:
## Summary

Makes the verbatim segment path of `common-json` append-only and stateless in the pipeline, drivable purely by `position()`, so a retained scalar leaf far larger than both the input window and the output bound streams through `projector → SEGMENTABLE sink → escape-mode generator` with **no per-fragment allocation** and **no sink-side write offset**.

Previously the sink tracked a per-value `segmentWritten` offset to span suspend/resume, which forced the value to be held whole (a `getString()` allocation) and left the pipeline stateful. Now the terminal stage reports the source bytes a segment write actually took via a new `consumed(int)` callback; the parser advances `position()` by exactly that count and re-exposes the value remainder on resume, so each pull is append-only.

## Changes

- **`JsonController`**: add default `consumed(int)` so a terminal stage reports the source bytes a segment write actually took; mediating transforms relay it (the same way they relay `segmentable()`).
- **`JsonGeneratorImpl`**: `writeRawSegment`/`writeEscapedSegment` advance only on whole UTF-8 characters and whole JSON escape sequences, never splitting a multibyte char or escape across a `consumed()` boundary, so a fragmented value can be re-presented cleanly.
- **`JsonSinkImpl`**: drop the per-value `segmentWritten` offset; write `getSegment()` whole and push `consumed()` back so the upstream re-exposes the remainder.
- **`JsonParserImpl` / `JsonTokenizer`**: track per-segment `consumed` to re-expose the value remainder; arm a kept scalar leaf value-string (`scalarSegment`) to stream verbatim as raw fragments with no decoded retention.
- **`JsonProjectorImpl` / `JsonSchemaImpl` Validator**: relay `consumed()` to upstream (the validator still declines `segmentable()`).

## Tests

- `JsonProjectorSegmentTest#shouldStreamKeptScalarLeafSegmentedAcrossInputAndOutputBounds` — a retained scalar leaf larger than both bounds, with escapes and a multibyte char; asserts chunked escaped output equals whole-shot and round-trips as JSON-in-JSON.
- `JsonProjectorSegmentTest#shouldStreamMultiLeafResourceLikeBinding` — a binding-shaped multi-leaf resource (small kept scalar, large kept scalar, kept number, dropped siblings) projected through `SEGMENTABLE` into an escape-mode generator; asserts chunked == whole-shot and round-trips to the projected plain document.

Full `common-json` build green (checkstyle, license, coverage). Validated end-to-end: the `mcp_http · proxy` binding streams both legs through bounded 8K slots with all 26 integration tests passing at `ENGINE_BUFFER_SLOT_CAPACITY=8192`, including the 10k/100k large-payload scenarios.

https://claude.ai/code/session_01AEpU8rooKSEqf6EuU8ri1i

---
_Generated by [Claude Code](https://claude.ai/code/session_01AEpU8rooKSEqf6EuU8ri1i)_

### DIFF
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonController.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonController.java
@@ -22,21 +22,14 @@ package io.aklivity.zilla.runtime.common.json;
  * caller determines which by observing the events that follow — there is no return value, because a
  * mediating upstream cannot know the outcome at call time. A mediating {@link JsonTransform} supplies
  * its own {@code JsonController} to its downstream; a non-mediating stage passes its own through.
- * <p>
- * A terminal stage that splices a kept value's verbatim bytes also calls {@link #consumed(int)} after
- * each segment chunk, reporting how many <em>source</em> bytes it actually took, so the upstream advances
- * its {@code position()} by exactly that count and re-exposes the value remainder on resume. A mediating
- * stage forwards {@code consumed} to its own upstream, the same way it relays {@code segmentable}.
  */
 public interface JsonController
 {
     void segmentable();
 
     /**
-     * Reports that {@code sourceBytes} source bytes of the current verbatim segment were consumed by the
-     * terminal write, always on a UTF-8 character and JSON escape-sequence boundary. The upstream advances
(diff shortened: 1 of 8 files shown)


jfallows (commented):
LGTM

PR #1885, opened 2026-06-15
feat(common-json): expose JsonParserEx.reset()

jfallows:
## Summary

Promotes `JsonParserImpl.reset()` to the public `JsonParserEx` interface so a caller holding only the interface type can rewind a reused parser to before a fresh document.

- A document is parsed across windows via `wrap()`; `reset()` returns the tokenizer to `DOC_START` (clearing the parse stack, scratch, and stream offset) for the *next* document.
- Window swaps still continue via `wrap()` and preserve in-flight state — `reset()` is per document, not per window.

## Why

Unblocks the `common-protobuf` `ProtobufJson` bridge (separate PR), whose `ProtobufParser` adapter reuses one `JsonParserEx` across messages and must rewind per message. Without a public `reset()` it would have to down-cast to the internal `JsonParserImpl`.

## Notes

This does **not** remove the `JsonStreamImpl` cast to `JsonParserImpl` — that path also needs the internal streaming event API (`hasNextEvent` / `nextEvent` / `position`), which is out of scope here.

## Test plan

- `./mvnw install -pl runtime/common-json` — full module build/tests pass (the change is a visibility/interface promotion of an already-covered method).

https://claude.ai/code/session_01KDCFMo1s8mqfAW1LRAoAPA

---
_Generated by [Claude Code](https://claude.ai/code/session_01KDCFMo1s8mqfAW1LRAoAPA)_

### DIFF
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonEx.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonEx.java
@@ -135,7 +135,7 @@ public final class JsonEx
     public static JsonStream stream(
         JsonParserEx parser)
     {
-        return new JsonStreamImpl((JsonParserImpl) parser);
+        return new JsonStreamImpl(parser);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


jfallows (commented):
LGTM

Issue #1857, opened 2026-06-08
model-protobuf: streaming converter on common-protobuf + common-json transcoder


jfallows:
### Motivation

Re-platform the `model-protobuf` converter (`runtime/model-protobuf`, today whole-message protobuf
↔ JSON) onto the format-native **`common-protobuf`** library (#1839) for the wire side and the
**`common-json`** transcoder (#1855) for the JSON side. Surfaced through the engine model SPI via
`supplyReadConverter` / `supplyWriteConverter`.

This is the **converter** layer, peer to `model-json` (#1835); the Protobuf library primitives and
the JSON primitives are tracked by their own issues.

### Scope

- `ProtobufRead`/`ProtobufWrite` converter handlers re-platformed: decode/encode the wire side with
  `common-protobuf`; drive the **`common-json` (#1855) transcoder** for the JSON side when
  converting protobuf ↔ JSON.
- Resolve the descriptor via `CatalogHandler`; compile and cache the `common-protobuf` model per
  `schemaId`.
- **Streaming + fragment-aware**: drive the fragment-aware handler (or the unified `convert` per
  #1836) with **streaming-validate-then-abort** (engine RESET on failure). Carry the
  **bounded-buffering contract** from #1839 (protobuf fields are length-delimited and may arrive
  out of order — bounded per-message field reassembly, documented and enforced) into the converter's
  behavior; zero per-message allocation beyond that bounded window.
- Preserve existing `model: protobuf` config + `view: json` semantics.

### Non-goals

- The `common-protobuf` library — streaming decode/encode, descriptor validation, the proto3 JSON
  **mapping** (64-bit ints→string, bytes→base64, enums→name, `oneof`, well-known types, json_name),
  and the out-of-order bounded-buffering design: **#1839**.
- The JSON-side reader/validator/generator: **#1855**.

### Relationship

- **Depends on #1839** (`common-protobuf`) + **#1855** (`common-json` transcoder).
- Relates to **#1836**: protobuf read converter = decode + JSON **generator**, the converter =
  validator + generator model.
- **Parallel goal** (not a hard requirement): native `common-protobuf` is the path to dropping the
  current protobuf/jackson dependencies from `model-protobuf`.

### Acceptance criteria

- protobuf ↔ JSON round-trip via `common-protobuf` + the #1855 transcoder, with the #1839 proto3
  JSON mapping; fragmented-input produces identical output to whole-buffer within the bounded-buffer
  contract.
[...]

PR #1886, opened 2026-06-15
feat(common-protobuf): protobuf↔JSON bridge via ProtobufJson

### LINKED ISSUE #1839: common-protobuf: streaming Protobuf decode/encode + descriptor validation, with JSON transcoding via common-json
### Motivation

Peer to `common-json`: a format-native Protobuf library for the hot path — streaming decode/encode and descriptor-based validation over `DirectBuffer` with no full-document buffering and no per-message allocation — reused by the `model-protobuf` converter. This is **not** a repurposing of the JSON validator; Protobuf validates against its **native descriptor**, with its own format-specific APIs in `common-protobuf`.

### Relationship to common-json

`common-protobuf` owns the Protobuf side only. When `model-protobuf` converts **protobuf ↔ JSON**, it decodes/encodes the wire side with `common-protobuf` and uses the **`common-json` transcoder** (the reader/validator/generator from #1855) for the JSON side. The two libraries compose in the model converter; neither depends on a shared cross-format validator or event model.

The `model-protobuf` **converter** that consumes this library — engine model SPI wiring (`supplyReadConverter`/`supplyWriteConverter`), catalog descriptor resolution, fragment-aware convert + streaming-validate-then-abort — is tracked separately in **#1857** (peer to `model-json` #1835).

### Scope
- Compile a Protobuf descriptor into an immutable model (cacheable per `schemaId`).
- **Streaming decode**: protobuf wire → a typed value/event stream, validating against the descriptor as it reads.
- **Streaming encode**: a typed value/event stream → protobuf wire.
- **protobuf ↔ JSON**: drive the `common-json` transcoder from decoded values to emit JSON; for JSON → protobuf, parse/validate JSON with `common-json` and encode the wire with `common-protobuf`.
- Define the Protobuf→JSON mapping (proto3 JSON mapping: 64-bit ints → string, bytes → base64, enums → name, `oneof`, well-known types, field names vs json_name) and document round-trip fidelity.

### Design risk to resolve
Protobuf fields are **length-delimited and may arrive out of order**, which complicates a strictly streaming decode → view under the no-full-document-buffer goal. The issue must settle how much bounded buffering is acceptable (e.g. per-message field reassembly within a bounded window) versus a documented bounded-buffer contract, distinct from JSON's clean forward streaming.

### Non-goals
- A universal/format-agnostic validator or shared event model.
- The JSON side itself — owned by `common-json` (#1855).
- The `model-protobuf` converter SPI wiring — tracked in #1857.

### Acceptance criteria
- Streaming Protobuf decode/encode against a descriptor, with the bounded-buffering contract documented and enforced.
- protobuf ↔ JSON round-trip via the `common-json` transcoder, with the proto3 JSON mapping.
- Unit tests covering nested messages, repeated/map fields, `oneof`, well-known types, out-of-order fields, and malformed input rejection.

### References
- `runtime/model-protobuf` (existing Protobuf model)
- `common-json` transcoder: #1855
- Consuming converter: #1857
- Model handler unification / generator concept: #1836

jfallows:
Fixes #1839.

## Summary

Adds the protobuf ↔ JSON bridge in a new `io.aklivity.zilla.runtime.common.protobuf.json` package (`ProtobufJson`), composing the `common-protobuf` wire layer with the `common-json` transcoder. It is built as **two adapters on the core wire interfaces**, so both directions drop into the existing `Protobuf` pipeline unchanged — as a pure `ProtobufParser` ↔ `ProtobufGenerator` pair, or via `Protobuf.stream(...).into(ProtobufSink.of(...))`:

- **`ProtobufJson.parser(JsonParserEx, schema, messageName)` → `ProtobufParser`** (JSON → protobuf): reads JSON one token at a time against the descriptor with a bounded frame stack, emitting the matching protobuf events.
- **`ProtobufJson.generator(JsonGeneratorEx, schema, messageName)` → `ProtobufGenerator`** (protobuf → JSON): renders the wire write calls (by field number) as JSON.

Both apply the proto3 JSON mapping: `repeated` → array, `map` → object, 64-bit and unsigned-64-bit integers → strings, `bytes` → base64, `enum` → value name (number when unknown), `float`/`double` → JSON numbers (`"NaN"`/`"Infinity"`/`"-Infinity"` as strings), proto3 json names, unknown fields ignored, `null` omitted.

## Streaming

- **JSON → protobuf** streams windowed input: `nextEvent` returns `null` on a partial window and `resume` continues, with no document buffer (only the per-message frame stack). Bound: one JSON *leaf value* must fit a single input window; message *structure* may split across windows at any token boundary.
- **protobuf → JSON** streams protobuf input across windows. The `ProtobufGenerator` interface has no root frame (the sink swallows the root `START`/`END_MESSAGE`), so the JSON root object opens on the first write and the consumer calls `generator.flush()` after `COMPLETED` to finalize it. The JSON output buffer must hold the whole document.

## Allocation — `ProtobufJsonPipelineBM` (`-prof gc`, both variants)

| variant | `gc.alloc.rate.norm` |
| --- | --- |
| `protobufToJson` | **≈ 0.01 B/op** |
| `jsonToProtobuf` | 104 B/op |

The adapters reuse all buffers/frames and add no per-op allocation. The generator formats numbers into a reused `StringBuilder` (via `writeNumber`/`write`) to avoid the JSON generator's internal `toString`. JSON values are read through `JsonParserEx.getStringView()` (a non-owning `CharSequence`), so the `jsonToProtobuf` residual is only the `float`/`double` parse — the JDK has no `CharSequence`-based float parser.

## Tests

`ProtobufJsonTest` (24 tests): scalars (all wire types), nested messages, `repeated` scalar/message, `map` (scalar- and message-valued), proto2 group, enum name/number, non-finite floats, base64 `bytes`, multibyte UTF-8, round-trips, windowed streaming, multi-message reuse, and rejection of malformed/non-object/unknown-message/unknown-enum/unsupported-write. Module build is green including the 90% JaCoCo gate.

## Dependencies

Rebased on `develop` after #1885 (`JsonParserEx` streaming accessors incl. `getStringView`) and #1895 (stateless wire sinks: pump-owned resume cursor + `consumed()` pushback) merged — both now drop out of this diff, leaving only the bridge.

## Out of scope / follow-ups

- Well-known-type canonical forms (Timestamp/Duration/wrappers/Any) — currently render as plain nested objects.
- The `model-protobuf` converter SPI wiring (#1857) that consumes this.

https://claude.ai/code/session_01KDCFMo1s8mqfAW1LRAoAPA

---
_Generated by [Claude Code](https://claude.ai/code/session_01KDCFMo1s8mqfAW1LRAoAPA)_

Fixes aklivity/zilla-plus#771
Fixes aklivity/zilla-plus#772
Fixes aklivity/zilla-plus#773


jfallows (commented):
LGTM

PR #1888, opened 2026-06-15
common-json: converge scalar value delivery — canonical structured rendering, unified consumed() flow control, remove Delivery.DECODED

### LINKED ISSUE #1887: common-json: unify scalar value delivery — tighten getSegment contract, fold DECODED into STRUCTURED, propagate flow control to the parser
## Summary

Tighten the `JsonStream` pipeline contract so that the raw-bytes accessor (`JsonSource.getSegment()`) is valid **only** in reaction to a segmented event, and the scalar value accessors (`getStringView()`/`getInt()`/`getLong()`/`getBigDecimal()`/`isIntegralNumber()`) are valid **only** in reaction to a structured value event. This makes each pipeline stage declare its capability by *which events it drives downstream*, rather than every stage that forwards a scalar implicitly having to support a later `getSegment()` call.

The follow-on simplifications:

1. The sink can no longer splice raw bytes (`getSegment()`) in reaction to a `VALUE_STRING`/`VALUE_NUMBER` event, so structured scalars render via the decoded strategy (`writeNumber`/`write(getStringView(), completion)`).
2. With that change **STRUCTURED ≡ DECODED**, so `Delivery.DECODED` is removed from the enum.
3. For the unified structured path to be the only scalar path, **end-to-end flow control must propagate back to the parser (the head source)** — including `consumed()` on output-window overflow mid-value — with no resumable partial-value state held in the sink or pipeline.

## Motivation

- **Simpler, composable contract.** Today a transform that forwards a `VALUE_NUMBER` implicitly signs up to support a downstream `getSegment()` on the source. That is fine for the parser-backed source but awkward for an intermediary transform that substitutes its own `JsonSource`. Splitting the accessors by event class lets a stage support only the getters relevant to the events it drives.
- **Stateless stages + uniform backpressure.** A single value-delivery contract with a `consumed()`-driven parser means no stage has to buffer partial values; backpressure is expressed once, at the head source. This unlocks arbitrary transform chains that compose without each stage re-implementing partial-value handling.
- **Orthogonal delivery semantics.** After convergence the model is clean: **structured = logical/canonical value**, **segmented = verbatim bytes**.

## Scope of the semantic change

Folding STRUCTURED into the decoded strategy changes **string-escape form only**: e.g. `&#34;A&#34;` → `&#34;A&#34;`, `&#34;a\/b&#34;` → `&#34;a/b&#34;`.

- **Numbers stay byte-exact** — the decoded path renders numbers from the lexeme (`getStringView()` = the scanned digits), so `1.50`, `-0`, `1E2` are preserved, fragmented or not. No precision/format loss.
- **Verbatim everything** (original escaping and insignificant whitespace) remains available via the segmented path (`SEGMENTABLE`).

## The hard part: unified `consumed()` flow control on the value path

Making `consumed()` the single backpressure mechanism for the decoded value path has to compose **two independent &#34;incomplete&#34; axes**:

- **Input-side:** `deferredBytes()` — more fragments of this value are still arriving across input windows (today: `Completion.INCOMPLETE`).
- **Output-side:** the bounded output filled *mid-fragment* — the value (or even one fragment) did not fit the current output window. Today only the raw/segment path handles this (`writeSegment` partial write + `consumed()`); the decoded path writes each fragment whole and would overrun the generator&#39;s `limit`.

Requirements to unify:

1. **Bounded scalar writer on the generator.** `writeNumber`/`write(CharSequence)` must emit only what fits, stop on a clean boundary (code point; never split a `\uXXXX`), and **return source units consumed** — the analog of what `writeSegment` already does for raw bytes (`escapedUnitWidth`/unit-boundary logic is the seed, but it currently works over bytes, not decoded UTF-16 chars).
2. **Parser holds the value cursor.** `consumed(n)` advances an offset into the current value; `getStringView()` returns the unconsumed remainder; the parser **does not advance past the value (or pull the next input fragment) until the cursor is fully consumed**.
3. **Sink/pipeline go stateless.** Drop `pendingSegment`; `resume()` re-reads the source view (which reflects the parser&#39;s advanced cursor) and re-drives the bounded write. Whether the next pump re-delivers the same partially-consumed value or advances is decided by the parser&#39;s cursor, not by sink state.

**Correctness trap:** scratch is reused and cleared between input fragments, so the parser must not advance/clear a fragment&#39;s decoded chars until they are output-consumed — the output-side cursor gates the input-side advance. Wrong ordering drops bytes or double-emits.

## Contract edge to settle

Under the tightened rule, scalar getters and `getSegment()` are mutually exclusive per event. We need to confirm no stage needs to **inspect the decoded value *and* forward the bytes verbatim** for the same scalar. Current stages look clear (the projector decides by path/key and keeps/drops whole subtrees via the segment path; validators read but do not forward), but a value-based-filter-then-passthrough stage would break the clean split.

## Decisions to confirm

1. **String-escape canonicalization** on the unified structured path is acceptable (verbatim escaping moves to the segmented path)?
2. **`SEGMENTABLE` stays** as the verbatim/byte producer?
3. Keep `getStringView`/`getInt`/`getLong` as the structured value getters (no new accessor needed)?

## Proposed sequencing (test-first)

1. Test asserting the tightened contract: `getSegment()` after a structured event is illegal/throws.
2. Test driving a **control-char-heavy string through a deliberately tiny output window**, asserting mid-value suspend → `consumed()` propagation → resume → byte-correct output (the case that currently overflows; the proof the new flow control works).
3. Test for a **fragmented number through a tiny output window** (both axes at once).
4. Add the bounded scalar writer to the generator (returns consumed units, clean boundaries).
5. Move the resumable value cursor into the parser; make the sink stateless (drop `pendingSegment`).
6. Repoint STRUCTURED scalars to the decoded strategy; remove `Delivery.DECODED`.
7. Audit transforms/consumers for reliance on verbatim string escaping via STRUCTURED.

## Context

Follow-up to a series of `common-json` allocation cleanups (allocation-free `JsonGeneratorImpl.write(int/long)`, `getInt/getLong` without materializing a `String`, decoded-sink rendering from a non-allocating `getStringView()`, and `getInt`/`getLong`/`getStringView` added to the `JsonSource` SPI). This issue captures the larger contract change those cleanups surfaced, for design discussion before implementation.

---

## Resolution (implemented in #1888)

All three decisions confirmed (canonicalization accepted; `SEGMENTABLE` stays the verbatim producer; `getStringView`/`getInt`/`getLong` reused) and the sequencing carried out:

- Tightened accessor contract **enforced by `assert`s in `JsonParserImpl`**, holding whether driven by the streaming pipeline or the raw `jakarta` parser (`getSegment()` segmented-only; scalar getters structured-only).
- `Delivery.DECODED` removed — **STRUCTURED ≡ decoded**; structured strings canonical, numbers byte-exact from the lexeme.
- **Unified `consumed()` flow control**: bounded scalar writers on the generator (clean code-point boundaries, report consumed source units) and a parser-held value cursor (`getStringView()` returns the unconsumed remainder); a verbatim scalar leaf is delivered as a `SEGMENT` so the byte path stays the single verbatim channel.
- **Hard-part item 3 satisfied** — the sink holds no resume state: `resume(JsonEvent)` derives continue-vs-advance from the parser&#39;s cursor (the source view remainder), and `JsonPipelineImpl` owns the single in-flight-event cursor.

Deviation from the original sketch: `resume()` was **kept** (with an added `JsonEvent` parameter) rather than removed. Full removal would require making every transform `feed()` idempotent under event re-delivery (the projector&#39;s path/index bookkeeping is not) — more complex, no behavioral payoff — so it is out of scope. The sink is otherwise **fully stateless** apart from the intrinsic `depth` (top-level completion): the generator owns the verbatim value&#39;s leading separator via a `writeSegment(…, Completion)` overload, so the sink&#39;s last per-value flag (`valueStarted`) is gone too.

jfallows:
Fixes #1887

## Summary

Converges the `JsonStream` scalar-value delivery model so a single, bounded `consumed()` flow-control mechanism drives the parser for every value, folds `Delivery.DECODED` into `STRUCTURED`, and tightens the accessor contract so each `JsonSource` accessor is valid only in reaction to its event class. The delivery semantics are now clean and orthogonal:

- **STRUCTURED** = logical/canonical value — each scalar is rendered from its decoded char view (`getStringView()`); strings are canonically escaped, numbers emit their lexeme.
- **SEGMENTABLE** = verbatim bytes — kept values (whole subtrees *and* scalar leaves) are delivered as `SEGMENT` events and spliced from `getSegment()`.

This began as allocation-free `common-json` cleanups (allocation-free `write(int/long)`, `getInt`/`getLong`/`getStringView` without materializing a `String`) and grew into the contract convergence those surfaced.

### Bounded, consumed-reporting scalar writers (generator)
`write(CharSequence, Completion)` and `writeNumber(CharSequence, Completion)` are now **bounded**: each emits only what fits the output bound, stops on a clean code-point boundary (never splitting a UTF-8 char or `\uXXXX`), reserves room for a string's closing quote, and reports the **source chars consumed** — the char-domain analog of the byte-based `writeSegment`. A `writeSegment(DirectBuffer, int, int, Completion)` overload owns a verbatim value's leading separator across fragments (emitted once on the first fragment, value ending on `COMPLETE`), mirroring the scalar writers.

### Char cursor + canonical rendering (parser/sink)
- `getStringView()` exposes the unconsumed decoded-char remainder via a reused, allocation-free view; `consumed()` advances a char cursor for a decoded scalar and the raw byte cursor for a segment.
- The sink renders structured strings **and numbers** canonically from the char view, suspending/resuming on the output bound; segments splice raw bytes.
- **The sink is fully stateless** apart from the intrinsic `depth` (top-level completion): `resume(JsonController, JsonSource, JsonEvent)` takes the in-flight event from the pump and continues only while the source still has an unwritten remainder (`getStringView().length()`/`getSegment().capacity()`); a boundary drain after a completed value is a no-op. `JsonPipelineImpl` owns the single resume cursor (the suspended event); the generator owns the segment leading-separator. `feed`/`resume` share one `writeValue` path.

### Tightened, enforced contract
- `getSegment()` is valid only on a **segmented** event; the scalar value getters (`getStringView`/`getInt`/`getLong`/`getBigDecimal`/`isIntegralNumber`/`getString`/`getKey`) only on their **structured** event. Enforced by `assert`s in `JsonParserImpl`, so the contract holds whether driven by the streaming pipeline or the raw `jakarta` parser.
- `lastEvent` is the canonical delivered event (set uniformly in `nextEvent`); `currentEvent` is derived from it as the `jakarta` projection (`null` for a segment or document framing). Shared `jakarta` getters assert `currentEvent`; streaming-only `getStringView`/`getKey`/`getSegment` assert `lastEvent`.
- A **verbatim scalar leaf** under SEGMENTABLE is now delivered as a `SEGMENT` event (not `VALUE_STRING`), so `getSegment()` is genuinely segmented-only and the sink routes values purely by event — no stage forwarding a `VALUE_NUMBER`/canonical `VALUE_STRING` implicitly signs up to support a downstream `getSegment()`.
- `JsonSink.resume`/`JsonTransform.resume` gain a `JsonEvent` parameter so the pump supplies the in-flight event and the stages keep no per-value resume state; the forwarding transforms (projector, validator) are otherwise unchanged.

### Removed `Delivery.DECODED`
With strings and numbers both rendered canonically from the char view, `DECODED` is identical to `STRUCTURED`; removed, all usages folded in. The four `McpHttpProxyFactory` STRUCTURED pipelines are unchanged (canonical, per the issue audit — none depend on byte-exact escaping; the verbatim response path already uses `SEGMENTABLE` + `GENERATE_ESCAPED`).

## Semantic change (accepted in #1887)

Structured string values are now **canonically escaped** (e.g. `\uA` → `A`, `a\/b` → `a/b`); **numbers stay byte-exact** (rendered from the lexeme). Byte-verbatim preservation remains available via the segmented/`SEGMENTABLE` path.

## Tests (test-first)

- Bounded-output chunking (added failing first, against the overrun): control-char-heavy, multibyte-UTF-8, and across-both-bounds decoded values — each suspends mid-value and reconstructs byte-correct.
- Contract enforcement: `getSegment()` after a structured event and a scalar getter after a `SEGMENT` both trip the assert.
- Deduped the now-equivalent decoded/verbatim STRUCTURED tests and the JMH fragment benchmarks; one projector test moved from verbatim to canonical; the `isIntegralNumber` misuse test moved to the assert.

## Verification

- `./mvnw clean install -pl runtime/common-json` — green (tests + checkstyle + jacoco 0.90).
- Full reactor `clean install -DskipITs -Djacoco.skip=true` (minus docker/helm, which need a daemon) — `BUILD SUCCESS`; the MCP pipelines compile and their unit tests pass.
- JMH `JsonPipelineBM` (`-prof gc`): every scenario stays effectively zero-allocation (≤ ~0.04 B/op, `gc.count ≈ 0`), including the segmented paths through the new `writeSegment(…, Completion)` overload.

## Not pursued

Full `resume()` removal (parser re-delivers a partial value through `feed()`, eliminating the method entirely) remains rejected: it requires every transform's `feed()` to be idempotent under event re-delivery (the projector's path/index bookkeeping is not), adding complexity for no behavioral payoff. The stages are already stateless via the `JsonEvent`-carrying `resume()`.

https://claude.ai/code/session_0111mS4B2sLNGVSq9NZEH1YT

### DIFF
--- a/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/transform/McpHttpRename.java
+++ b/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/transform/McpHttpRename.java
@@ -103,6 +103,12 @@ public final class McpHttpRename implements JsonTransform
         }
 
         @Override
+        public CharSequence getStringView()
+        {
+            return key;
+        }
+
+        @Override
         public CharSequence getKey()
         {
             return key;
@@ -116,6 +122,18 @@ public final class McpHttpRename implements JsonTransform
 
         @Override
         public boolean isIntegralNumber()
+        {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public int getInt()
+        {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public long getLong()
         {
             throw new UnsupportedOperationException();
         }
--- a/runtime/common-json/README.md
+++ b/runtime/common-json/README.md
@@ -130,13 +130,16 @@ one frame — and spliced across chunks; a string that fits is still re-encoded 
 never longer than the raw token, so the fit check is safe). So a giant `{"data":"…"}` value streams in
 both `STRUCTURED` and `SEGMENTABLE` delivery.
 
-Resumption is a **per-stage cascade**, not an event replay: `JsonSink.resume(control, source)` and
-`JsonTransform.resume(control, source, sink)` continue any in-flight fragment before the next event is
-pulled, with the same `control` and `source` context as `feed`. `JsonTransform.resume(control, source,
-sink)` defaults to `sink.resume(control, source)`, so a stage that only forwards events ignores it
-entirely; a stage that itself emits a value across chunks (substituting or expanding output) overrides
-it to continue its own emission, draining the downstream first. This keeps the transform contract
-simple — no stage has to be suspend/resume aware unless it originates `SUSPENDED`.
+Resumption is a **per-stage cascade**, not an event replay: `JsonSink.resume(control, source, event)`
+and `JsonTransform.resume(control, source, event, sink)` continue any in-flight fragment before the next
+event is pulled, with the same `control` and `source` context as `feed` plus the value `event` that
+suspended (supplied by the pump, which owns the single resume cursor, so a stage keeps no per-value
+resume state). The sink holds none: it continues only while the source view still has an unwritten
+remainder (`getStringView().length()`/`getSegment().capacity()`) and otherwise just advances.
+`JsonTransform.resume` defaults to `sink.resume(control, source, event)`, so a stage that only forwards
+events ignores it entirely; a stage that itself emits a value across chunks (substituting or expanding
+output) overrides it to continue its own emission, draining the downstream first. This keeps the
+transform contract simple — no stage has to be suspend/resume aware unless it originates `SUSPENDED`.
 
 ### Writing JSON directly
 
@@ -145,7 +148,10 @@ and quoting automatically from an internal context stack, emitting in source ord
 insignificant whitespace. It implements `jakarta.json.stream.JsonGenerator` with covariant returns for
 fluent chaining, plus the streaming-to-buffer extensions: `writeNumber(literal)` emits a numeric
 lexeme verbatim, `writeRaw` splices a pre-encoded value (emitting its leading separator once), and
-`writeSegment` appends a bounded, consumption-driven fragment of that value with no separator.
+`writeSegment(source, index, length)` appends a bounded, consumption-driven fragment of that value with
+no separator. Its `writeSegment(source, index, length, completion)` overload owns the leading separator
+across fragments — emitted before the first fragment, the value ending on `Completion.COMPLETE` — so a
+fragmented value splices without the caller tracking whether the separator was already written.
 
 `createGenerator(Map.of(JsonGeneratorEx.GENERATE_ESCAPED, true))` opts the generator into **escape mode**: every
 byte it emits is escaped as JSON string *content* (structural bytes and UTF-8 continuation bytes pass
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonGeneratorEx.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonGeneratorEx.java
@@ -140,6 +140,22 @@ public interface JsonGeneratorEx extends JsonGenerator
         int index,
         int length);
 
+    /**
+     * Appends a value's verbatim bytes that may arrive in fragments, owning the value's leading structural
+     * separator so the caller holds no per-value state. The separator is emitted before the first fragment;
+     * each fragment appends bounded, consumption-driven content exactly as {@link #writeSegment(DirectBuffer,
+     * int, int)} (track via {@link #consumed()}); the value ends — readying the next value's separator — on
+     * {@link Completion#COMPLETE} once the final fragment is fully written. Pass {@link Completion#INCOMPLETE}
+     * while the source reports more deferred bytes, so a value delivered across fragments forms one spliced
+     * value. {@code writeSegment(source, index, length, COMPLETE)} for a value that fits one fragment is
+     * equivalent to {@link #writeRaw(DirectBuffer, int, int)}.
+     */
+    JsonGeneratorEx writeSegment(
+        DirectBuffer source,
+        int index,
+        int length,
+        Completion completion);
+
     @Override
     JsonGeneratorEx writeStartObject();
 
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonSink.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonSink.java
@@ -49,16 +49,18 @@ public interface JsonSink
         JsonEvent event);
 
     /**
-     * Continues any output left in flight by a prior {@link JsonPipeline.Status#SUSPENDED} — a value
-     * being written across chunks — before the next event is fed, reading the in-flight value from
-     * {@code source} and steering the immediate upstream with {@code control}. Returns {@link
-     * JsonPipeline.Status#SUSPENDED} if the bounded output filled again, or {@link
-     * JsonPipeline.Status#ADVANCED} when nothing remains pending. A stage with no in-flight output
-     * returns {@code ADVANCED}; the default is sufficient for stages that only forward events.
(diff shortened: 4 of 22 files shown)


jfallows:
### Allocation benchmarks

`JsonPipelineBM`, decoded sink, `-prof gc` (allocation per op), before → after this PR:

| Benchmark | Before | After |
|---|---|---|
| `fragmentNumberDecoded` | 832.04 B/op | 0.04 B/op |
| `fragmentStringDecoded` | 840.04 B/op | 0.04 B/op |

Throughput flat (~180–190k ops/s). `JsonSinkImpl.writeDecoded` now sources the value from the non-allocating `getStringView()` instead of `getString()` (which materialized `scratch.toString()` per fragment), so the decoded render path drops the per-fragment `String` allocation to ≈0. Output bytes are unchanged for both numbers and strings.

Follow-up design (unify scalar delivery / propagate flow control to the parser) tracked in #1887.

---

jfallows (commented):
LGTM

PR #1893, opened 2026-06-16
refactor(common-avro): make AvroSink stateless (resume event + consumed() pushback)

### LINKED ISSUE #1932: Make common-avro / common-protobuf transforms consumption-driven for a uniform src-authoritative resume
## Summary

Follow-up to the per-stream `ModelPipeline` refactor (#1). Make the `common-avro` and `common-protobuf` `transform(src, dst)` implementations **consumption-driven** so their resume behaviour matches `common-json` and the `model-core` identity pipeline: on `OVERFLOW`, `consumed` is reported conservatively and the unconsumed input remainder stays in the caller's `src`, rather than the parser running ahead and draining buffered output from internal state.

## Background

All the re-encoding pipelines keep a small, bounded *structural* state across an `OVERFLOW`/`SUSPENDED` resume (generator nesting, "inside an open string", schema-walk cursor, pending separators). That state is the **generator's** concern and is not caller-visible — the source/caller contract is only ever `consumed`. The difference between the formats is *when input is consumed relative to output being emitted*:

- **Lazy / consumption-driven (`common-json`):** a field's wire bytes are consumed only as its value is emitted; string/bytes tails that don't fit use `write(.., INCOMPLETE)` + `deferredBytes()` so the un-emitted tail stays unconsumed and therefore stays in the caller's `src`. `consumed` tracks output progress, the remainder lives in `src`, and resume re-reads `src` for input plus the internal schema cursor for scaffolding. This is the same work that produced the consumption-driven keys + HEADROOM elimination in `common-json`.
- **Eager (`common-avro`, `common-protobuf` today):** the parser runs ahead and consumes the whole value; the generator drains leftover output across subsequent `OVERFLOW` calls from internal buffers. `consumed` jumps to the end, so on resume `src` is already empty — i.e. the transform "ignores the `src` window on resume".

Both obey "advance `src` by `consumed`", so the canonical driver loop is already correct for either. The practical asymmetry is only that the eager version *tolerates* a caller that forgets to advance `src` (because `src` is empty on resume anyway), whereas `model-core` and a lazy implementation *require* advancing. That makes the eager behaviour a latent footgun: a caller that relied on it would silently corrupt output against `model-core` (identity copy re-reads `src[srcIndex..]`).

## Goal

Convert the `common-avro` and `common-protobuf` generators to consume input lazily so that:

- `consumed` never runs ahead of the input whose output has actually been emitted (except the bounded case below);
- the unconsumed remainder of a value lives in the caller's `src`, not in an internal output buffer;
- the only retained state is the bounded, caller-invisible schema/generator cursor;
- the source contract becomes **byte-for-byte identical** to `model-core` / `common-json`, not merely compatible.

## Known bounded caveat

A single value whose **computed** output exceeds `dst` cannot keep `consumed` conservative, because the output is not a copy of the input:

- a numeric scalar rendered as decimal text (`long` → `"1234567890"`),
- a bytes field rendered as base64 (4 output chars per 3 input bytes).

These outputs are **bounded width**, so a small fixed reserve removes the case: do not begin emitting the scalar unless its maximum rendered width fits in the remaining `dst`; consume base64 in 3-byte-aligned groups. String/bytes-as-string values remain src-resumable via `deferredBytes()`.

## Acceptance criteria

- [ ] `common-avro` `transform` is consumption-driven: `OVERFLOW` reports conservative `consumed`, remainder recoverable from `src` on resume (no reliance on an internal output-drain across calls, except the bounded-scalar reserve path).
- [ ] `common-protobuf` `transform` likewise.
- [ ] Bounded-scalar reserve (numbers, base64) implemented so a single computed value never overflows `dst` mid-render.
- [ ] Unit tests assert that re-passing a non-advanced `src` on an `OVERFLOW` resume is **not** required, and that advancing by `consumed` reproduces the value exactly (mirrors the `model-core` OVERFLOW + interleaved-stream tests).
- [ ] `ModelPipeline.transform` Javadoc documents the contract: callers MUST advance `src` by `consumed` on every `OVERFLOW` re-call; pipelines may retain bounded internal structural state but never raw input across calls.

## Notes

- This is structural/perf-neutral cleanup, not a behaviour change for correct callers — the canonical driver loop already works against the eager implementation. The value is a uniform contract across all model formats and removal of the non-advancing-caller footgun before the Phase-4 binding callers are written.
- Reference implementation: `common-json` (`JsonGeneratorImpl` `write(CharSequence, Completion)` / `Pending` states / `deferredBytes`, and the HEADROOM-elimination / exact structural-prefix accounting).

jfallows:
## Summary

Makes the terminal `AvroSink` stateless, mirroring the design `common-json`'s `JsonSink` landed in #1888 — so the two streaming libraries share one model: the pump supplies the resume event, and value-write progress lives in the generator/parser via a `consumed()` pushback, not in the sink.

Two commits:

### 1. Drop resume state — pump supplies the event
- `AvroSink.resume(AvroController, AvroSource, **AvroEvent**)` — the pipeline re-supplies the event that suspended, so the sink keeps no `pending` field.
- `AvroTransform.resume` and `AvroStreamImpl.BoundSink` thread the event; `AvroPipelineImpl` remembers the suspending event (a re-suspend keeps the prior one).
- `AvroSinkImpl` drops `pending`.

### 2. Drop value-offset state — `consumed()` pushback
- New `AvroController.consumed(int)` and `AvroParser.consumed(int)` — the output-side back-pressure peer of `JsonController.consumed` / the JSON parser's segment cursor.
- `AvroParserImpl` tracks `valueConsumed`; `getSegment()` exposes only the unconsumed remainder after a bounded-output suspend (`getString()`/`decode()` still see the whole value). `AvroPipelineImpl`'s controller forwards `consumed()` to the parser.
- `AvroSinkImpl.writeValue` streams a length-delimited value by re-reading the source remainder and reporting `consumed()`, dropping `valueOffset`/`valueStarted`. It suspends on the existing **stateless** `atomic()` gate (`length()>0 && remaining()<HEADROOM`), which already guards the length-prefix write — so no generator change was needed.

Net: `AvroSinkImpl` keeps only structural `depth` (exactly as `JsonSinkImpl` does).

## Validation
`./mvnw verify -pl runtime/common-avro` — all 82 tests pass (structured + segmented transcode, fragmented values, malformed rejects), 0 checkstyle, JaCoCo coverage met. `AvroPipelineBM` structured/segmented transcode stay ~0 B/op (0.002–0.007), confirming no allocation regression. Added an `AvroGeneratorTest` case covering the `consumed()` pushback on a hand-fed value sink.

## Note for the AvroJson bridge
A parser that delivers length-delimited values in bounded pieces must override `AvroParser.consumed(int)` (the interface default is a no-op). `AvroParserImpl` does. The bridge's JSON-driven `AvroParser` (PR #1884) delivers whole values (`deferredBytes()==0`) so the no-op default is correct there; when #1884 merges this, no change is required, but it's worth a glance.

https://claude.ai/code/session_01HoyRbFKGExxejhKGacFfFH

---
_Generated by [Claude Code](https://claude.ai/code/session_01HoyRbFKGExxejhKGacFfFH)_

### DIFF
--- a/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/AvroController.java
+++ b/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/AvroController.java
@@ -26,4 +26,15 @@ package io.aklivity.zilla.runtime.common.avro;
 public interface AvroController
 {
     void segmentable();
+
+    /**
+     * Reports {@code sourceBytes} payload bytes consumed by a bounded value write, so the upstream advances
+     * its segment cursor and re-exposes the value's unconsumed remainder on resume — the output-side
+     * back-pressure pushback that lets a terminal sink stream a length-delimited value without keeping its
+     * own offset. The default does nothing for a non-mediating edge that forwards no value.
+     */
+    default void consumed(
+        int sourceBytes)
+    {
+    }
 }
--- a/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/AvroParser.java
+++ b/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/AvroParser.java
@@ -121,9 +121,21 @@ public interface AvroParser
 
     /**
      * Non-owning, on-stack view of the current contiguous raw Avro bytes — valid on {@link AvroEvent#BYTES}
-     * and {@link AvroEvent#FIXED} value events (the chunk available now) and on segment events.
+     * and {@link AvroEvent#FIXED} value events (the chunk available now) and on segment events. After
+     * {@link #consumed(int)} pushback (a bounded-output suspend) it exposes only the unconsumed remainder.
      */
     DirectBuffer getSegment();
+
+    /**
+     * Advances the current value chunk's read cursor by {@code sourceBytes} already written downstream, so a
+     * subsequent {@link #getSegment()} after a bounded-output suspend exposes only the remainder — letting a
+     * terminal sink resume a length-delimited value without tracking its own offset. The default does
+     * nothing, for a parser whose values are never delivered in bounded pieces.
+     */
+    default void consumed(
+        int sourceBytes)
+    {
+    }
 
     /**
      * The bytes still to come in later events of the run the current event belongs to, {@code 0} on its
--- a/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/AvroSink.java
+++ b/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/AvroSink.java
@@ -45,14 +45,16 @@ public interface AvroSink
 
     /**
      * Continues the value in flight when the previous {@link #feed} returned
-     * {@link AvroPipeline.Status#SUSPENDED}, after the caller has drained and re-wrapped the output. The
-     * default does nothing (a sink that never suspends never sees this); a sink that returns
-     * {@code SUSPENDED} overrides it to resume from where it paused, reading remaining bytes from
-     * {@code source}.
+     * {@link AvroPipeline.Status#SUSPENDED}, after the caller has drained and re-wrapped the output.
+     * {@code event} is the value event that suspended, supplied by the pump so the sink keeps no resume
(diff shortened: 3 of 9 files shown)


jfallows (commented):
LGTM

PR #1898, opened 2026-06-16
support `oauthbearer` `sasl` mechanism in kafka binding & support `identity` `guard` options with static identity and credentials

### LINKED ISSUE #1868: feat(binding-kafka): support SASL/OAUTHBEARER mechanism for Kafka client
## Summary

Add support for the SASL/OAUTHBEARER mechanism in the Kafka client binding, allowing Zilla to authenticate with Kafka brokers using OAuth 2.0 bearer tokens.

## Motivation

The Kafka binding currently supports SASL/PLAIN and SASL/SCRAM mechanisms. OAUTHBEARER is increasingly required by managed Kafka services (e.g., Confluent Cloud, Azure Event Hubs) and environments that use identity platforms (e.g., Keycloak, Auth0) as the authentication authority.

## Proposed behavior

- The bearer token should be injectable via a guard (similar to how `authorization` works today — see #1824), so that Zilla can obtain and refresh tokens from a configured JWT guard or identity provider

## Example configuration

```yaml
bindings:
  kafka_client:
    type: kafka
    kind: client
    options:
      servers:
        - broker:9092
      authorization:
          identity0:
            credentials:
              mechanism: oauthbearer
              token: "{credentials}"
    exit: tcp_client
```

### LINKED ISSUE #1897: support `identity` guard `options` with static identity and credentials
## Summary

The `identity` guard currently has no `options` support, so there is no way to configure static fallback values for `identity` and `credentials` when no session is active. This is needed to supply a fixed credential (e.g. a pre-signed JWT) to downstream bindings such as the Kafka client SASL/OAUTHBEARER mechanism.

## Proposed changes

- Add an `options` block to the `identity` guard with two optional string fields:
  - `identity` — a static identity value returned when no session is present
  - `credentials` — a static credential value returned when no session is present
- Wire the options through the JSON config adapter (`IdentityOptionsConfigAdapter`) and update the JSON schema patch
- Register the adapter via `OptionsConfigAdapterSpi` and declare the required `jakarta.json` / `jakarta.json.bind` module dependencies

## Example configuration

```yaml
guards:
  my_identity_guard:
    type: identity
    options:
      identity: zilla-service-account
      credentials: "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..."
```

## Motivation

Required to support the `http.kafka.oneway.oauthbearer` example where Zilla authenticates to Kafka using SASL/OAUTHBEARER with a static pre-signed JWT. Without static `credentials` in the guard options there is no mechanism to supply the token without an active upstream session.

ankitk-me:
Fixes #1868 #1897


jfallows (changes requested):
I think we also need a schema update for Kafka binding to allow token only for oauthbearer mechanism, and username/password only for the other mechanisms, agree?

jfallows (approved):

PR #1904, opened 2026-06-16
perf(common-json): cut hot-path allocations in JSON DOM and schema validation

jfallows:
## Summary

Removes remaining hot-path allocations in `common-json`, driven by the JMH benchmarks under `runtime/common-json/src/test/.../bench/`. The streaming pipeline (`JsonPipelineBM`) was already ~0 B/op; the residual allocations lived on two paths that lacked (or had misleading) coverage: the jakarta **DOM** path and the per-message **schema validation** path.

Benchmark-first: add/repair coverage so the allocations are measured, then optimize against `-prof gc` (`gc.alloc.rate.norm`, B/op).

## What changed

**DOM number path** — `NumberValue` is lexeme-backed: a parsed number retains its source lexeme and defers `BigDecimal` until a numeric/equality accessor needs it; parse-then-re-emit touches only `toString()`, preserving the source form. New `JsonReaderWriterBM` covers the DOM read/write path.

**Schema validation path** — `JsonSchemaImpl` is immutable and shareable; `validate(JsonParser)` is stateless (fresh evaluator per call), so a compiled schema may be validated concurrently. Mutable, reusable state lives on the validating parser from `newParser()` (`ValidatingParser`, now a resettable `JsonParserEx`), so a caller reuses one instance across documents.
- *Numbers*: skip `BigDecimal` for an unbounded typed integer; a long fast path checks an integer instance against integer bounds (precomputed `integerBounds`) with no `BigDecimal`; `isIntegralNumber()` is allocation-free.
- *Strings/keys*: value strings via `getStringView()`; a simple object (only properties/required/additionalProperties, no annotation/diagnostics) matches keys as a `CharSequence` against compiled key arrays and tracks `required` with a reused `boolean[]` — no key `String`, no `HashSet` node.
- *const/enum/uniqueItems*: canonicalized by the incremental `UniqueCanon` (no `Token` list); uniqueItems dedup via an open-addressed `CanonSet`. A scalar-string `const`/`enum` skips canonicalization entirely — the string instance is compared as a `CharSequence` against the decoded candidate(s).
- *Reuse*: item/contains/property evaluators reused across an array's elements; per-element collections cleared-and-reused; the whole evaluator reused across documents via the resettable validating parser.

### Schema validation — `JsonSchemaBM` (-prof gc, B/op, 2 forks × 5 iter)

`JsonSchemaBM` drives a **reused validating parser** (`reset()` per op) — the representative per-caller pattern (what `model-json` will do). Seven of eight cases are allocation-free; the rest is intrinsic:

| Benchmark | B/op | remaining |
|---|---:|---|
| validateFlatObject | ~0 | — |
| validateArrayObjects | ~0 | — |
| validateTypedIntegers | ~0 | — |
| validateBoundedNumbers | ~0 | — |
| validateContains | ~0 | — |
| validateOneOf | ~100 | two-branch oneOf structural cost |
| validateUniqueScalars | 768 | dedup canonical strings |
| validateUniqueObjects | 2048 | dedup canonical strings |

Validator-only improvement, measured with the stateless `validate()` drive (fresh evaluator per document, same reused parser) so it isolates the validator changes from the cross-document reuse above; the residual here is the per-document evaluator tree the reuse removes:

| Benchmark | baseline | per-call |
|---|---:|---:|
| validateBoundedNumbers | 5376 | 400 |
| validateTypedIntegers | 4616 | 400 |
| validateArrayObjects | 7552 | 1440 |
| validateUniqueScalars | 7896 | 1776 |
| validateContains | 6464 | 1616 |
| validateUniqueObjects | 9824 | 3304 |
| validateFlatObject | 1192 | 776 |
| validateOneOf | 2512 | 2240 |

### DOM read/write — `JsonReaderWriterBM` (-prof gc, B/op)

| Benchmark | baseline | this PR |
|---|---:|---:|
| writeNumbers | 1632 | 232 |
| writeNested | 216 | 216 |
| writeFlat | 200 | 152 |
| roundTripNumbers | 13600 | 10584 |
| readNumbers | 11968 | 10352 |
| readNested | 8704 | 8536 |
| readFlat | 6320 | 6168 |

The DOM reads stay in the KB range because building a `JsonValue` tree is inherently allocating (one node per value); the number pass-through removes the `BigDecimal` per number (most visible on `writeNumbers`, −86%).

## Behavior note for review

Parsed DOM numbers re-serialize **preserving their source lexeme** instead of canonicalizing (`100.0` stays `100.0`; `6.022e23` stays `6.022e23`). Lossless pass-through; conformance suite green — flagging it as observable.

## Tests

All 4786 `common-json` tests pass (4373-case conformance suite, uniqueItems, `$ref`/recursive, combinator, dependent, annotation/unevaluated, const/enum incl. scalar fast path and structural fallback, plus new `shouldReuseEvaluatorAcrossValidations`, `shouldReuseValidatingParserAcrossDocuments`, `shouldValidateScalarStringConstAndEnumAgainstNonStrings`). Checkstyle + license clean.

## Remaining allocations (intrinsic)

- uniqueItems: the canonical strings stored in the dedup set are the dedup keys.
- a structural (object/array) `const`/`enum` value: its `UniqueCanon` canonical, compared against the candidate.

## model-json follow-up

To realize the reuse in the gateway, `JsonValidatorHandler`/`JsonModelHandler` should hold one reused parser and one reused validating parser per schema id (`reset()` per message). Tracked separately.

https://claude.ai/code/session_013f5U8qqz2HjbGiHS91o1hR

### DIFF
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonEx.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonEx.java
@@ -51,7 +51,7 @@ public final class JsonEx
     /**
      * Returns an empty {@link JsonParserEx} to be fed via {@link JsonParserEx#wrap(
      * org.agrona.DirectBuffer, int, int)} (or {@link JsonPipeline#feed(org.agrona.DirectBuffer,
-     * int, int)} when driving a pipeline). Reuse a single instance per worker thread.
+     * int, int)} when driving a pipeline). Reuse a single instance per thread.
      */
     public static JsonParserEx createParser()
     {
@@ -69,9 +69,8 @@ public final class JsonEx
     }
 
     /**
-     * Mirrors {@link jakarta.json.Json#createParserFactory(Map)}. Construct once per
-     * factory class and reuse for the lifetime of the binding to avoid repeating config
-     * resolution on every stream.
+     * Mirrors {@link jakarta.json.Json#createParserFactory(Map)}. Construct once and reuse to
+     * avoid repeating config resolution on every stream.
      */
     public static JsonParserFactory createParserFactory(
         Map<String, ?> config)
@@ -82,7 +81,7 @@ public final class JsonEx
     /**
      * Returns a buffer-backed {@link JsonGeneratorEx} — a standard {@link
      * jakarta.json.stream.JsonGenerator} extended with the streaming-to-buffer methods. Reuse a
-     * single instance per worker thread, calling {@link JsonGeneratorEx#wrap(
+     * single instance per thread, calling {@link JsonGeneratorEx#wrap(
      * org.agrona.MutableDirectBuffer, int)} before each value. Requires no {@code jakarta.json}
      * provider on the classpath; {@code common-json} ships the implementation.
      */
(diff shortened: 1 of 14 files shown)


jfallows (commented):
LGTM

PR #1908, opened 2026-06-16
refactor(common-json): adapt JsonParserEx via Source/Control instead of casting

jfallows:
## Summary

Eliminates the `(JsonSource)` / `(JsonController)` casts of the parser in `JsonPipelineImpl`, replacing them with `Source` and `Control` inner adapters that delegate through the `JsonParserEx` interface — mirroring `AvroPipelineImpl`.

## Changes

- **`JsonParserEx`** — promotes `getSegment()`, `deferredBytes()`, and `consumed()` (default no-op) onto the interface, as `AvroParser` does. Adds a `Mode { STRUCTURED, SEGMENTED }` enum and `nextEvent(Mode)`, with `nextEvent()` as a default delegating to `STRUCTURED`. `segmentable()` is intentionally **not** on the interface.
- **`JsonPipelineImpl`** — `Source implements JsonSource` and `Control implements JsonController` delegate to the parser. `Control.segmentable()` records a one-shot flag; the pump threads `control.mode()` into `parser.nextEvent(...)`.
- **`JsonParserImpl`** — now `implements JsonParserEx` only (drops `JsonSource, JsonController`). The `segmentable()` method is removed; its arming logic is inlined into `nextEvent(Mode.SEGMENTED)`, matching `AvroParserImpl`.
- **`JsonSchemaImpl.ValidatingParser`** — updated to the new `JsonParserEx` surface (`nextEvent(Mode)`, `getSegment`, `deferredBytes`).

### Pump fix

JSON's pump previously called `parser.hasNextEvent()` in the loop condition, which advanced the tokenizer **past** the value boundary before the deferred `SEGMENTED` mode could arm — corrupting segment start offsets (e.g. `{ "a" : 1 }` → `" : 1 }`). The pump now mirrors Avro: pull with the requested mode first, `break` on a `null` event, so arming precedes the advance.

## Tests

Parser-level tests drive segmentation via `nextEvent(Mode.SEGMENTED)` instead of `parser.segmentable()`; the chunking probe reads `source.getBigDecimal()` / `getInt()` directly rather than casting `source` to `JsonParser`.

**4787 tests pass, 0 checkstyle violations.**

---

## Unrelated: CI build flag cleanup

Also drops `-U` from the `Build with Maven` step in `.github/workflows/build.yml`. The command carried both `-U` (force snapshot/metadata update) and `-nsu` (no snapshot updates), which are contradictory; `-U` forced a re-fetch of every group's `maven-metadata.xml` each build, adding remote chatter. Keeps `-nsu -ntp`.

Note: for `pull_request` events GitHub uses the workflow definition from the **base** branch, so this flag change takes effect once merged to `develop`, not on this PR's own build.

https://claude.ai/code/session_01XfJwE4PAk8wzfKVgLw4RXs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
           docker load -i ~/docker-base-cache/base-images.tar || true
         fi
     - name: Build with Maven
-      run: ./mvnw -B -U -nsu -ntp -Ddocker.logStdout -Dfailsafe.skipAfterFailureCount=1 -Ddocker.verbose install jacoco:report-aggregate
+      run: ./mvnw -B -nsu -ntp -Ddocker.logStdout -Dfailsafe.skipAfterFailureCount=1 -Ddocker.verbose install jacoco:report-aggregate
       env:
         GITHUB_ACTOR: ${{ github.actor }}
         GITHUB_TOKEN: ${{ github.token }}
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonParserEx.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonParserEx.java
@@ -83,10 +83,34 @@ public interface JsonParserEx extends JsonParser
     boolean hasNextEvent();
 
     /**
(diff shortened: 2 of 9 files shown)


jfallows (commented):
LGTM

PR #1910, opened 2026-06-16
feat(common-protobuf): message-index resolution + JsonFormat-compatible JSON options

jfallows:
## Motivation

Second foundation PR for the `model-protobuf` re-platform (#1857), after #1906. It adds the `common-protobuf` surface the re-platformed converter consumes so the converter PR can be a pure `model-protobuf` change. Everything here is additive; existing APIs and behavior are unchanged.

## Changes

- **`ProtobufSchema` — schema-registry message-index resolution.** `messageIndexes(record)` maps a package-relative dotted name (e.g. `DemoMessage.SimpleMessage`) or dotless full name to its message-index path (declaration order from top-level down through nested types, with synthetic `map_entry` messages ordered after explicit nested types, matching `protoc`); `messageByIndexes(path)` is the reverse. This is what the converter needs to encode/decode the protobuf message-index prefix (the magic-byte + schema-id + message-index framing originated by Confluent Schema Registry and re-implemented by Apicurio/Karapace).
- **`ProtobufField` declared defaults.** Carries the proto2 `[default = ...]` value, populated by **both** the binary `FileDescriptorSet` compiler (`default_value`, field 7) and the `.proto` source compiler (`[default = ...]` option).
- **`ProtobufJson.generator` JsonFormat-compatibility config**, configured via a `Map<String, ?>` (mirroring `JsonEx.createGenerator(Map)` and the `JsonGeneratorEx.GENERATE_ESCAPED` / `JsonSink.DELIVERY` precedents):
  - `ProtobufJson.FIELD_NAMES` → a `ProtobufJson.FieldNames` enum: `JSON` (proto3 lowerCamelCase json names, the default) or `PROTO` (snake_case proto field names, i.e. `preservingProtoFieldNames()`).
  - `ProtobufJson.INCLUDE_DEFAULTS` (a `Boolean`) → emit unset implicit-presence scalars/enums as their **declared (proto2) or type default**, empty `repeated` as `[]`, empty `map` as `{}`; fields with explicit presence (proto3 `optional`, message, `oneof`) remain omitted. Matched field-by-field against a **`JsonFormat.includingDefaultValueFields()` reference (the JSON library being replaced, v4.34.1)**.
  - Fields are emitted in **wire order**, which is semantically-equivalent JSON (RFC 8259 — object member order is not significant); the only non-semantic difference is key ordering, which no JSON consumer can observe. This keeps the read path fully streaming (no per-message materialization).
- **Docs:** removed the literal `protobuf-java` token from `common-protobuf` javadoc/READMEs — the library has no such dependency (it decodes descriptor sets and `.proto` source with its own reader) and the wording now states the property without naming the library.
- **Bug fix:** `ProtobufSourceCompiler` reversed the nesting prefix at depth ≥ 2 (an `ArrayDeque.push` + head-first iteration mismatch), producing wrong nested full names. Surfaced by the new message-index tests; fixed to append/iterate scope in declaration order.

## Performance

The default generator path (no config) is untouched and stays zero-allocation — JMH `gc` profiler reports `protobufToJson` at `0.008 B/op` after the change. The added per-`Scope` `BitSet` (written-field tracking for `INCLUDE_DEFAULTS`) is allocated once and reused via `clear()`; the default emission early-returns when `INCLUDE_DEFAULTS` is unset.

## Tests

- `ProtobufSchemaIndexTest` — top-level/nested index resolution, reverse lookup, explicit-before-map-entry ordering, unknown lookups.
- `ProtobufJsonDefaultsTest` — proto3 defaults + presence-field omission, every proto3 scalar default representation (incl. 64-bit-as-string), proto2 declared + type defaults, default recursion into a set message, and `FieldNames.PROTO` vs `FieldNames.JSON` keys — all pinned to the reference output.

`./mvnw clean install -pl runtime/common-protobuf -DskipITs` passes (unit tests, checkstyle, license, JaCoCo, moditect).

### DIFF
--- a/runtime/common-protobuf/README.md
+++ b/runtime/common-protobuf/README.md
@@ -13,7 +13,7 @@ the `.json` package pulls in `common-json`.
 ## Descriptor model
 
 `ProtobufSchema` is a compiled, immutable model — build one per `schemaId` and cache it. It is
-**provider-free**: there is no dependency on `protobuf-java`. Construct the model with the public
+**provider-free**: there is no dependency on a third-party protobuf library. Construct the model with the public
 builders (`ProtobufSchema.Builder`, `ProtobufMessage.Builder`, `ProtobufField.Builder`,
 `ProtobufEnum.Builder`); a consumer such as `model-protobuf` populates field metadata from its own
 `.proto` compilation.
@@ -32,7 +32,7 @@ synthetic map-entry message (`mapEntry(true)`) with `key` as field 1 and `value`
 
 A compiled `google.protobuf.FileDescriptorSet` (e.g. from `protoc --descriptor_set_out`) can be
 turned into a `ProtobufSchema` directly — `descriptor.proto` is itself Protobuf, so it is decoded
-with this library's own wire reader and needs no `protobuf-java`:
+with this library's own wire reader and needs no third-party protobuf library:
 
 ```java
 ProtobufSchema schema = Protobuf.schema(descriptorSet, offset, length);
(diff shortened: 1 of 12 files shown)


jfallows (commented):
LGTM

PR #1911, opened 2026-06-16
Parse YAML through a single-pass streaming scanner and retire the eager parser

### LINKED ISSUE #1905: common-yaml: incremental, buffer-backed YAML parser to remove eager-parse allocation
## Background

PR #1894 drove the `common-yaml` **generate** path to near-zero per-document allocation (96–99.6% reductions) and trimmed the **parse** path where it could without a redesign (per-scalar parser/regex churn, and `YamlJsonEvent` reuse via a 2-slot lookahead pool).

After that work, the remaining parse allocation (~17–27 KB/op on the benchmarks) is **architectural**, not avoidable churn. This issue captures the redesign that would address it, plus the smaller `readAll` options, so the analysis isn't lost.

Parsing is a config-load (cold) path today, so this is **not urgent** — but it's the correct end-state if a hot YAML path ever emerges, and it's the only thing that unlocks zero-copy scalar views.

## Where the remaining parse allocation comes from

Profiled with JFR allocation sampling on the `common-yaml` JMH benchmarks (current branch):

1. **`readAll(Reader)`** — slurps the entire input into a `char[4096]` read buffer + a growing `StringBuilder` + a final `String` copy, because the parser operates on a fully-materialized `String`.
2. **Eager document model** — the parser builds a complete `YamlNode` tree (`YamlScalarNode`/`YamlObjectNode`/`YamlEntry`) plus per-line structures (`YamlDocumentParser$Line`, `DocumentScanner.lines(String)`, backing `Object[]`/`HashMap`). Both the streaming `JsonParser` projection and the `readValue()` DOM are built on top of this.
3. **Parsed-content `String`s** — keys/values are substrings of the source. Inherent while the parser materializes them up front (see "zero-copy scalars" below).

## Proposed direction: incremental, buffer-backed parser

Align the parse path with `common-json`'s `JsonParserImpl` approach: a streaming, buffer-backed parser that

- tokenizes incrementally from the `Reader` (no `readAll`, no full-document `String`),
- emits parser events without first building a `YamlNode` tree,
- exposes zero-copy scalar views over the parse buffer.

### Zero-copy scalars via `getStringView()`

Once scalars are read directly from the parse buffer (rather than materialized into `YamlScalarNode.value` during an eager pass), the parser can implement `common-json`'s `JsonParserEx.getStringView()` to hand back a view instead of allocating a `String` per key/value. This requires `common-yaml` to depend on `common-json` (sibling modules) so the `JsonParserEx` type is visible — justified **only** in the context of this rewrite, since on today's eager parser the `String`s are already materialized and a view would wrap an existing `String` for no gain.

### Scope / caveats

- YAML is **indentation-sensitive with look-back** — multi-line and folded scalars, block vs. flow context, anchors/aliases, and merge keys (`<<`) that reference earlier nodes. A streaming parser that still supports these is a **substantial rewrite**, not a refactor (JSON is context-light by comparison, which is why `JsonParserImpl` can tokenize linearly).
- `readValue()` / DOM consumers would need a materialization layer on top of the streaming core.
- Conformance must be preserved — the existing `YamlConformanceTest` and `YamlJsonParserTest` suites are the contract.

## Smaller, non-redesign `readAll` options (if a streaming rewrite isn't pursued)

These are modest and only worth doing if a concrete call site benefits — captured for completeness:

1. **Skip the `toString()` copy** — let `YamlDocumentParser.parse` accept a `CharSequence` and read off the `StringBuilder` directly, removing one full-size copy.
2. **`String` fast-path** — add a `String`-accepting entry point so callers that already hold the config text in memory skip the `StringReader` → `readAll` → `String` round-trip. Worth it only if `EngineConfigReader` (or another caller) has a `String` in hand.
3. (Not recommended) tuning the `char[4096]` read buffer size — trades small-input cost against more `read()` calls on large files.

## Acceptance

- Parse allocation reduced materially below the current eager-parse floor on the `common-yaml` JMH parse benchmarks, captured before/after with `-prof gc`.
- `getStringView()` implemented and exercised, with the `common-json` dependency added per the runtime module-graph rules (fully modular).
- All existing parse/conformance tests pass unchanged.

Context: PR #1894 (generate-path allocation work).

jfallows:
## Description

Replaces Zilla's eager tree-building YAML parser with a single-pass
streaming scanner as the sole YAML 1.2 parser behind the jakarta.json API.

### Parsing
- `YamlScanner` classifies an entire YAML stream into a compact flat event
  buffer (parallel arrays into the source text) in one forward pass — no node
  tree, no intermediate line objects, plain scalars kept as zero-copy slices.
- `YamlParser` is a thin streaming cursor that frames the buffer into the YAML
  representation graph (stream / documents / balanced root nodes), exposing
  scalar value/view/style, node anchor/tag, flow vs block, and explicit
  document markers.
- `YamlJsonParser` projects the cursor directly into jakarta.json events,
  resolving aliases against already-projected step ranges — a single event
  buffer, no second materialized copy.
- The eager `YamlDocumentParser` and its tree model are removed; the scanner
  is complete for valid YAML 1.2 and is the only parser.

### Conformance
- Every parser event is verified verbatim against the canonical `test.event`
  fixtures for all 308 valid cases of the YAML test suite (data-2022-01-17),
  and every one of the 94 invalid cases is rejected with a located message
  pinned against a vendored snapshot.
- JSON projection is validated against `in.json` for the 279 JSON-mappable
  fixtures.

### Performance
- ~24% fewer bytes allocated per parse versus the prior eager path
  (measured with JMH `-prof gc`), from eliminating the tree and the
  redundant event buffering.

Fixes #1905 

https://claude.ai/code/session_01EAyLPRtkDUpaLB54eVAQ5Y


jfallows (commented):
LGTM

PR #1913, opened 2026-06-17
feat(common-protobuf): strict reject-unknown-fields option + pipeline rejection reason

jfallows:
## Motivation

Follow-up to the `model-protobuf` re-platform (#1912, part of #1857). Two `common-protobuf` additions the converter needs to restore behavior that the streaming library didn't yet expose:

1. **Spec-default strictness for unknown JSON fields.** The proto3 JSON→protobuf mapping says a parser should *reject* unknown fields by default (with an opt-in to ignore) — as `protobuf-java` `JsonFormat`, grpc-gateway, etc. do. `common-protobuf`'s `ProtobufJson` parser currently always *ignores* them, so the re-platformed `model-protobuf` write path lost the old strict rejection.
2. **Descriptive validation failures.** `ProtobufPipelineImpl.feed` caught `ProtobufException` and collapsed it to `REJECTED`, discarding the message — so consumers could only emit a generic failure event.

## Changes

- **`ProtobufJson.REJECT_UNKNOWN_FIELDS`** — a `Map<String, ?>` config key for `ProtobufJson.parser(...)`. When `true`, an unknown JSON field makes the message reach `REJECTED`; absent/`false` keeps the existing ignore behavior (so `ProtobufJsonTest.shouldIgnoreUnknownJsonField` is unchanged). The internal `ProtobufJsonParserImpl` gains a `rejectUnknownFields` flag; the existing 3-arg constructor delegates with `false`.
- **`ProtobufPipeline.reason()`** — a `default String reason()` returning a short human-readable reason for the most recent `REJECTED` (e.g. `unknown field nope`, `unknown enum value …`, `truncated json`), or `null` otherwise; cleared by `reset()`. `ProtobufPipelineImpl` records `ProtobufException.getMessage()` in the catch.

Both are additive; default behavior is unchanged.

## Tests

`ProtobufJsonStrictTest` — reject-unknown rejects with a reason containing the field name; default ignores (and `reason()` is null); `reason()` null on success; `reason()` cleared by `reset()`.

`./mvnw clean install -pl runtime/common-protobuf -DskipITs` passes (tests, checkstyle, license, JaCoCo, moditect).

## Next

#1912 will consume these: enable `REJECT_UNKNOWN_FIELDS` on the write path (reverting its `shouldWriteInvalidProtobufEventFormatJson` test to the unknown-field case), and emit `pipeline.reason()` as the descriptive validation-failure event message.

---

### DIFF
--- a/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufPipeline.java
+++ b/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufPipeline.java
@@ -76,6 +76,16 @@ public interface ProtobufPipeline
     void reset();
 
     /**
+     * A short, human-readable reason for the most recent {@link Status#REJECTED} — e.g. an unknown
+     * field, an unknown enum value, or truncated input — or {@code null} when the last feed did not
+     * reject. Cleared by {@link #reset()}.
+     */
+    default String reason()
+    {
+        return null;
+    }
+
+    /**
      * The number of input bytes committed since the message began — always at a whole-unit boundary. On
      * {@link Status#STARVED} everything at or after this position is the unconsumed tail: the caller
      * retains it (typically in its own reassembly slot) and re-presents it, contiguous with the next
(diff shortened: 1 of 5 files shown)


jfallows (commented):
LGTM

PR #1915, opened 2026-06-17
support `kind` in `guard`

ankitk-me:


specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/engine.schema.json
```diff
@@ -159,6 +159,11 @@
                     "type": "string",
                     "enum": []
                 },
+                "kind":
+                {
+                    "title": "Kind",
+                    "type": "string"
+                },
```
jfallows, line 166: Let's put `guard: false` on the guards that don't need `kind` so it doesn't show up in the docs schema check.

jfallows (changes requested):

jfallows (approved):

PR #1920, opened 2026-06-18
feat(common-json): surface verbatim schema diagnostics with accurate line/column

### LINKED ISSUE #1919: Surface verbatim schema-validation diagnostics through the streaming JSON pipeline (validator-throws + line/column tracking)
## Background

While migrating `model-json` to the streaming `common-json` pipeline (validator + converters), the validation/conversion **failure event message** no longer matches the verbatim message the non-streaming parser produced, e.g. `A message payload failed validation. [1,11][/id] expected string but was number`.

Two distinct gaps in the streaming pipeline are responsible:

1. **Schema-validation rejections don't reach the reporter.** `JsonSchemaImpl.Validator.feed(...)` signals a schema violation by returning a silent `Verdict.INVALID` → `Status.REJECTED`, with no message. `JsonPipelineImpl.feed(...)` only sets `diagnostic.message` from a caught `JsonParsingException` (malformed input), so for a schema violation the reporter fires with a `null` message. This is the "rejecting structurally with no message" anti-pattern that `ProtobufValidatorImpl` explicitly avoids — it throws a descriptive `ProtobufException` at the point of detection so `ProtobufPipelineImpl` maps it to `REJECTED` and pushes the message to the reporter.

2. **The streaming parser doesn't track line/column.** `JsonLocationImpl.getLineNumber()`/`getColumnNumber()` are hardcoded to `-1` (the parser is byte-offset based via `getStreamOffset()`), so the diagnostic renders `[-1,-1]` instead of `[1,11]`.

The current `model-json` branch keeps accurate messages via a cold-path re-parse (`validate(...)` over the non-streaming jakarta parser) on failure. This issue tracks replacing that with a proper streaming diagnostic.

## Proposed work (separate PR)

1. **Validator throws at the point of detection** — make `JsonSchemaImpl.Validator` collect diagnostics (active `Trace`) and `throw new JsonValidationException(diagnostics)` on `Verdict.INVALID`, mirroring `ProtobufValidatorImpl`. Broaden `JsonPipelineImpl`'s `catch (JsonParsingException)` to `catch (JsonException)` so both malformed and schema-invalid rejections set `diagnostic.message` and reach the `JsonReporter`.

2. **Streaming line/column tracking** — track line/column in the streaming tokenizer so `getLocation()` returns accurate values (and the partial-unit rewind logic adjusts them correctly). Because the parser always knows its position at the point of rejection, line/column are then **always** available — no "unknown" case, no conditional rendering. Apply the same to `common-avro` and `common-protobuf` as needed.

3. **Model handlers adopt the reporter** — `model-json` validator + read/write converters wire `JsonStream.reporting(...)`, log `diagnostic.message()` on REJECTED, and drop the re-parse fallback. (Equivalent for avro/protobuf model modules once their pipelines surface diagnostics.)

## Acceptance

- Validation/conversion failure events carry the verbatim `[line,col][pointer] message` produced entirely by the streaming pipeline (no re-parse).
- `model-json` `EventIT` passes against `[1,11][/id] expected string but was number` with the message sourced from the reporter.
- No conditional "unknown location" handling and no spec workaround.

## References

- Streaming model migration / pipeline uptake: #1907
- Reporter/diagnostic SPI: #1918
- Validator diagnostics: #1914

jfallows:
Fixes #1919

## Problem

The streaming `common-json` pipeline did not deliver the verbatim validation-failure diagnostic the non-streaming parser produced (e.g. `A message payload failed validation. [1,11][/id] expected string but was number`). Two distinct gaps were responsible:

1. **Schema-validation rejections never reached the reporter.** `JsonSchemaImpl.Validator.feed(...)` signalled a violation by returning a silent `Verdict.INVALID` → `Status.REJECTED`, with no message. `JsonPipelineImpl.feed(...)` only set `diagnostic.message` from a caught `JsonParsingException` (malformed input), so for a schema violation the reporter fired with a `null` message. This is the "rejecting structurally with no message" anti-pattern that `ProtobufValidatorImpl` explicitly avoids.

2. **The streaming parser didn't track line/column.** `JsonLocationImpl.getLineNumber()`/`getColumnNumber()` were hardcoded to `-1` (the parser is byte-offset based), so the diagnostic rendered `[-1,-1]`.

## Changes

1. **Validator throws at the point of detection** — `JsonSchemaImpl.Validator` collects diagnostics via an active `Trace` and throws a descriptive `JsonValidationException` on `Verdict.INVALID`, mirroring `ProtobufValidatorImpl`. `JsonPipelineImpl`'s `catch` is broadened from `JsonParsingException` to `JsonException`, so both malformed and schema-invalid rejections set `diagnostic.message` and reach the `JsonReporter`.

2. **Streaming line/column tracking** — `JsonTokenizer` now tracks 1-based line/column as bytes are consumed (counting characters, skipping UTF-8 continuation bytes), snapshots them at the value-start and complete-unit marks, and restores them on the window-edge rewinds so multi-window values stay accurate. `JsonLocationImpl` returns the tracked values. Because the parser always knows its position at the point of rejection, line/column are **always** available — there is no "unknown" case and no conditional rendering.

The reporter now delivers e.g. `[1,11][/id] expected string but was number` entirely from the streaming pipeline.

## Scope

- **JSON**: fixed (this PR).
- **Protobuf**: `ProtobufValidatorImpl` already throws a descriptive `ProtobufException`, so its reporter already receives the message; it's a binary format, so line/column don't apply. No change needed.
- **Avro**: binary format (no line/column); no analogous validator-transform gap. No change needed.

## Tests

- `JsonPipelineRejectTest` — schema-violation reporter diagnostics asserting the verbatim message **and** location, single line (`[1,11]`) and across newlines (`[2,12]`).
- `JsonLocationTest` — parser line/column tracking on a single line and across newlines (replacing the obsolete `-1` assertion).
- Full `common-json` suite green (4793 tests) with coverage met.

## Follow-up

Once this lands, the `model-json` pipeline-uptake PR (#1907) can adopt `JsonStream.reporting(...)` in its validator + converters and drop the cold-path re-parse, with the failure event sourced entirely from the reporter.

---
_Generated by [Claude Code](https://claude.ai/code/session_019YL2uqgwVoN3nqUxnfaj6g)_

### DIFF
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonLocationImpl.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonLocationImpl.java
@@ -29,13 +29,13 @@ final class JsonLocationImpl implements JsonLocation
     @Override
     public long getLineNumber()
     {
-        return -1;
+        return tokenizer.line();
     }
 
     @Override
     public long getColumnNumber()
     {
-        return -1;
+        return tokenizer.column();
     }
 
     @Override
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonPipelineImpl.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonPipelineImpl.java
@@ -16,8 +16,8 @@ package io.aklivity.zilla.runtime.common.json.internal;
 
 import java.math.BigDecimal;
 
+import jakarta.json.JsonException;
 import jakarta.json.stream.JsonLocation;
-import jakarta.json.stream.JsonParsingException;
 
 import org.agrona.DirectBuffer;
 
@@ -122,7 +122,7 @@ public final class JsonPipelineImpl implements JsonPipeline
                 status = last ? Status.REJECTED : Status.STARVED;
             }
         }
-        catch (JsonParsingException ex)
+        catch (JsonException ex)
         {
             status = Status.REJECTED;
             diagnostic.message = ex.getMessage();
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonSchemaImpl.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonSchemaImpl.java
@@ -2157,13 +2157,14 @@ public final class JsonSchemaImpl implements JsonSchema
         }
 
         private final JsonController decline = new Decline();
+        private final List<JsonSchemaDiagnostic> diagnostics = new ArrayList<>();
 
         private JsonController upstreamControl;
         private Eval eval;
 
         private Validator()
         {
-            this.eval = eval();
+            this.eval = eval(new Trace(diagnostics::add));
         }
 
         @Override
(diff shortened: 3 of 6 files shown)


jfallows (commented):
LGTM

PR #1921, opened 2026-06-18
feat(common-protobuf): stream bytes values across input windows without an adapter buffer

jfallows:
## Motivation

Follow-up to #1917. That PR dropped the JSON generator's internal `bytesCarry`, relying on output back-pressure for streaming. That works when the whole value is present (output windowing), but left one gap: a base64 `bytes` value spanning multiple **input** windows can have a 3-byte group straddle a window boundary. The 1–2 byte sub-group tail can't be emitted until the next input window arrives, but `ProtobufTypedSinkImpl.onValueLength` treated that under-consumption as output back-pressure (`SUSPENDED`, replay the *same* window) or threw `"value header exceeds output limit"` — so a `bytes` value larger than the input window couldn't stream.

The pure pipelines already stream scalars larger than both buffers (e.g. `ProtobufPipelineTest.shouldStreamLargeStringEndToEnd`, `JsonPipelineChunkingTest`) by leaving a partial unit **unconsumed** at the consumption boundary and re-presenting it via `remaining()` — the wire parser does this for partial UTF-8 code points. This wires base64 the same way: the alignment state lives in the input reassembly, not an adapter buffer.

## Changes

- **`ProtobufGenerator.deferring()`** (default `false`): the wire generator inherits `false` — it writes verbatim and is never input-bound, so `wire→wire` never takes the new path. `ProtobufJsonGeneratorImpl` returns `true` when `writeBytesSegment` emitted every whole group the input could form yet a 1–2 byte tail remains with `deferred > 0`.
- **`ProtobufTypedSinkImpl.onValueLength`**: on `written < available` → `deferring()` ⇒ `STARVED` (no `flush`; output isn't drained, the in-flight value keeps streaming into the same window); else `length > 0` ⇒ `SUSPENDED` (drain, replay — unchanged); else throw (unchanged). Base64 group arithmetic stays entirely in the adapter.
- **`ProtobufParserImpl` / `ProtobufReader`**: `leafChunk` eagerly skips the whole delivered chunk, but the sink may consume only the whole-group prefix. `ProtobufParser.giveback()` (default no-op) records the unconsumed streaming-leaf tail and `ProtobufReader.unread(int)` moves the cursor back, so `remaining()`/`resume` re-present it exactly once. `ProtobufPipelineImpl` calls `giveback()` on a sink-driven `STARVED` from **both** the `feed` and `resume` paths (the resume path was dropping a byte at each suspend-then-starve boundary).

## Tests

`ProtobufJsonChunkingTest` gains a two-axis (input + output windowed) case: a 15001-byte `bytes` value plus a large escaped string through a **64-byte input** window (not a multiple of 3, so groups straddle boundaries) and a **50-byte output** window, asserting both a `STARVED` and a `SUSPENDED` boundary occur and an exact round-trip (string equals; bytes base64-decode to the original). Confirmed failing-first (`REJECTED` with `deferring()` forced `false`), passing with the fix; the output-windowing cases are kept; no adapter carry reintroduced.

`./mvnw clean install -pl runtime/common-protobuf` → BUILD SUCCESS, no regressions (`ProtobufPipelineTest` 37, `ProtobufChunkingTest`, `ProtobufTypedSinkTest`, `ProtobufParserTest`, etc.; checkstyle, license, JaCoCo, moditect).

---

### DIFF
--- a/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufGenerator.java
+++ b/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufGenerator.java
@@ -69,6 +69,20 @@ public interface ProtobufGenerator
      */
     int consumed();
 
+    /**
+     * Whether the most recent {@link #writeSegment} left a sub-unit tail unconsumed because it needs
+     * <em>more input</em> to complete it, rather than because the bounded output filled. A generator that
+     * groups its encoding (e.g. base64, which encodes a whole 3-byte group at a time) cannot emit a 1-2 byte
+     * tail until the next input window supplies the bytes that complete the group; it leaves that tail
+     * unconsumed and reports {@code true} here so the driver fetches the next input window (input
+     * back-pressure) rather than draining and replaying the same one. A generator that writes its body
+     * verbatim is never input-bound and always reports {@code false}.
+     */
+    default boolean deferring()
+    {
+        return false;
+    }
+
     ProtobufGenerator writeInt32(
         int field,
         int value);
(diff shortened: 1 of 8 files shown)


jfallows (commented):
LGTM

PR #1922, opened 2026-06-19
perf(common-protobuf): allocation-free finite float/double parse on the JSON→protobuf write path

jfallows:
## Motivation

`ProtobufJsonPipelineBM.jsonToProtobuf` (JSON→protobuf write path) allocated **~104 B/op**, traced entirely to a single `double` field. The decode did:
```java
floatValue  = Float.parseFloat(parser.getStringView().toString());
doubleValue = Double.parseDouble(parser.getStringView().toString());
```
which allocates both the `.toString()` `String` **and** the JDK parser's internal buffer (the parse side has no thread-local cache, unlike the format side that the read path's `StringBuilder.append(double)` benefits from). Integer/string/bool values were already 0 B/op (`parseLong(CharSequence)`, manual `putUtf8`).

## Change

Add `parseDouble(CharSequence)` / `parseFloat(CharSequence)` to `ProtobufJsonParserImpl`, mirroring the existing `parseLong`/`parseUnsignedLong`, and point the `FLOAT`/`DOUBLE` cases at them (dropping `.toString()`). Each:

1. Scans the view once for sign, integer/fraction digits, and an optional `e`/`E` exponent into a `long` significand + decimal exponent — no allocation.
2. **Clinger exactly-rounded fast path** (single IEEE multiply/divide against a power-of-ten table): `double` when `digits ≤ 15` and `|exp| ≤ 22`; `float` when `digits ≤ 7` and `|exp| ≤ 10`, computed entirely in the **float domain** (`long * 10^n` / `long / 10^n`) so it is never narrowed from a `double` (no double-rounding).
3. **Fallback** to `Double/Float.parseXxx(view.toString())` for too-many-digits, out-of-range exponent, or non-numeric lexemes (`Infinity`/`-Infinity`/`NaN`) — still correct, allocates only on the rare path.

The string-form `enum` and base64 `bytes` `.toString()` paths are intentionally left untouched (out of scope; not on the number hot path).

## Tests

`ProtobufJsonNumberParseTest` round-trips both parsers **bit-exactly** against the JDK via `Double.doubleToLongBits` / `Float.floatToIntBits` (so `NaN` and `-0.0` are checked by bits) across literals, significand/exponent boundaries (`2^53±`, `2^24±`, `|exp|` at/beyond 22 and 10), subnormals, `MIN`/`MAX`, the non-finite strings, and ~300k randomized samples — driving **both** the fast path and the fallback.

## Result

`ProtobufJsonPipelineBM.jsonToProtobuf` `gc.alloc.rate.norm`: **104.020 → 0.019 B/op** (≈ zero). `./mvnw clean install -pl runtime/common-protobuf` → BUILD SUCCESS (tests, checkstyle, license, JaCoCo, moditect).

## Scope

Pre-existing allocation (predates #1917/#1921, in code those PRs didn't touch); split out as its own follow-up. Independent of the streaming work — this is in the JSON `ProtobufJsonParserImpl`, not the wire parser.

---

### DIFF
--- a/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/json/internal/ProtobufJsonParserImpl.java
+++ b/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/json/internal/ProtobufJsonParserImpl.java
@@ -665,10 +665,10 @@ public final class ProtobufJsonParserImpl implements ProtobufParser
             longValue = token == JsonEvent.VALUE_TRUE ? 1L : 0L;
             break;
         case FLOAT:
-            floatValue = Float.parseFloat(parser.getStringView().toString());
+            floatValue = parseFloat(parser.getStringView());
             break;
         case DOUBLE:
-            doubleValue = Double.parseDouble(parser.getStringView().toString());
+            doubleValue = parseDouble(parser.getStringView());
             break;
         case ENUM:
             longValue = token == JsonEvent.VALUE_STRING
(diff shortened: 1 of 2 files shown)


jfallows (commented):
LGTM

PR #1923, opened 2026-06-19
refactor(common-protobuf): retire the ProtobufJson staging buffers (key accumulator + value buffer)

jfallows:
## Motivation

Retire the two remaining staging buffers in the `common-protobuf` JSON adapters, so the protobuf↔JSON path holds no per-value growable buffer. Both are replaced without leaking JSON concerns (base64) into the wire codec.

## `ProtobufJsonGeneratorImpl.accumulator` — removed
`writeKeySegment` staged a **map key**'s bytes across fragments in a dedicated `ExpandableArrayBuffer` and materialized them via `new byte[]` + `new String`. The existing `scratch` `StringBuilder` already does cross-fragment staging for string values, so the key path reuses it: reset on the first fragment, `appendUtf8` each fragment, `captureKey(scratch.toString())` on completion (a key never touches the bounded output — it's emitted later via `writeKey` when the value's field is prepared — so each fragment is always fully consumed). Drops the buffer **and** the per-key `new byte[]`; the `segmentLength` counter falls out as dead.

## `ProtobufJsonParserImpl.valueBuffer` — replaced with bounded streaming
The parser staged the **whole** transcoded value (UTF-8-encoded string / base64-decoded bytes) in a growable `ExpandableArrayBuffer`. Replace it with a small **fixed** reused buffer (`valueChunk = new byte[4092]`, a multiple of 3 so a base64 4→3 group never straddles the end) and transcode incrementally, delivering the value as bounded chunks (mirroring the wire parser's `leafChunk`). **The transcode stays in the parser — `ProtobufGeneratorImpl` is untouched, so no base64/JSON concern leaks into the wire codec.**

State machine: the value's chars are already wholly in-window (`getStringView` requires the leaf present), so a source cursor drives transcode progress, not input back-pressure.
- `segment()` exposes `valueChunk[valueConsumed, decodedLength)`;
- `consumed(n)` advances `valueConsumed` only — within-chunk output back-pressure, never refills;
- `deferredBytes()` returns the bytes not yet transcoded (> 0 until the last chunk), so the generator's length prefix (`total = length + deferred` on the first `writeSegment`) is correct and the segment stays open across chunks;
- `streamValueStep()` (on the pump's ADVANCED re-entry) refills only once the chunk is fully drained: STRING encodes whole code points until the next won't fit; BYTES decodes whole 4→3 groups until the next won't fit, the final padded group on the last chunk; decoded length computed up front, allocation-free.

## Tests

- `accumulator`: a multi-character string map key (single feed) and a 43-char key driven through a 4-byte input window (the prior tests used only single-char keys, exercising neither the cross-fragment accumulation nor the reset).
- `valueBuffer` (`ProtobufJsonValueChunkingTest`): a 20k-char multibyte/surrogate string and ~15k padded bytes through a **40-byte** wire output window (within-chunk SUSPEND); a value straddling both the fixed buffer (refills) and the output window simultaneously; and a reflection check that the staging buffer is a fixed bounded `UnsafeBuffer` with no `ExpandableArrayBuffer` — all **bit-exact** against the single-feed wire.

`./mvnw clean install -pl runtime/common-protobuf` → BUILD SUCCESS (177 tests incl. `ProtobufJsonTest`/`ProtobufJsonChunkingTest`/`ProtobufPipelineTest`, checkstyle, license, JaCoCo, moditect). `ProtobufJsonPipelineBM.jsonToProtobuf` stays ~0 B/op.

## Result

No growable per-value buffer remains in the protobuf↔JSON adapters; both staging buffers are gone (generator) / bounded (parser), with the transcode in the correct layer.

---

### DIFF
--- a/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/json/internal/ProtobufJsonGeneratorImpl.java
+++ b/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/json/internal/ProtobufJsonGeneratorImpl.java
@@ -14,12 +14,9 @@
  */
 package io.aklivity.zilla.runtime.common.protobuf.json.internal;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
-
 import java.util.BitSet;
 
 import org.agrona.DirectBuffer;
-import org.agrona.ExpandableArrayBuffer;
 import org.agrona.MutableDirectBuffer;
 import org.agrona.concurrent.UnsafeBuffer;
 
@@ -62,7 +59,6 @@ public final class ProtobufJsonGeneratorImpl implements ProtobufGenerator
     private final boolean protoFieldNames;
     private final boolean includeDefaults;
     private final StringBuilder scratch;
-    private final ExpandableArrayBuffer accumulator;
     private final UnsafeBuffer byteView;
 
     private Scope[] scopes;
@@ -74,7 +70,6 @@ public final class ProtobufJsonGeneratorImpl implements ProtobufGenerator
 
     private boolean segmentActive;
     private boolean segmentOpened;
-    private int segmentLength;
     private int consumed;
     private boolean flushed;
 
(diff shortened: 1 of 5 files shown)


jfallows (commented):
LGTM

PR #1927, opened 2026-06-19
feat(common-json): consumer-driven value retention (accumulate on decline, incremental number validation, fail-closed cap, fragment-aware validator)

### LINKED ISSUE #1926: common-json: stream window-fragmented unconstrained values through the schema validator (forward-and-suppress)
## Background

The consumer-driven value-retention work in `common-json` (branch `claude/common-json-streaming-retention`) made the schema validator correct for window-fragmented values: `JsonSchemaImpl.Validator.feed` now **declines** a deferred scalar (`control.consumed(0)`, does not forward, returns `STARVED`) so the source accumulates it and re-presents it whole, then validates the complete value. See `feat(common-json): validate window-fragmented values by reassembling them whole`.

That is **Option A**: *every* scalar the validator sees is reassembled whole (bounded by `JsonParserEx.MAX_VALUE_SIZE`), because `Eval` is not fragment-aware.

## Problem

When the schema imposes **no value-content constraint** on a scalar position (e.g. `type`-only, or no applicable keyword), the validator still buffers the whole value just to forward it downstream. For large unconstrained payloads this defeats streaming through the validator stage — the value is held to the cap instead of flowing fragment-by-fragment to the sink.

## Why it wasn't done now

`Eval.feed` (`JsonSchemaImpl`) treats each `VALUE_*` event as one complete value (type/count/`combine`), and for nested values the parent eval routes to a child and counts the element. Feeding it fragments would mis-count / mis-evaluate. So streaming an unconstrained value through the validator can't simply pass fragments to `eval`.

## Proposed follow-up ("forward-and-suppress")

For a scalar position whose applicable schema needs **no value content**:
- **forward** each fragment to the downstream sink (so it streams), and
- **feed `eval` exactly once** at completion (type/count only — the content is already gone, which is fine because no content keyword applies),
- **suppress** `eval` on the intermediate fragments (so it isn't re-fed / doesn't double-count).

Scalars whose schema **does** need content (`pattern`/`enum`/`const`/`format`/`minLength`/`maxLength`/numeric bounds, or any `allOf`/`anyOf`/`oneOf`/`not`/`if`/`then`/`else` branch with one) keep the Option A decline-and-reassemble path.

### Requirements
1. A per-position **"needs scalar content"** signal. A static, per-`JsonSchemaImpl` flag is the natural form: this node's direct content keywords ∪ its combinator children, recursively; **conservatively `true` for `$ref`/`$dynamicRef`** (avoid cyclic precompute). It must reflect the schema applicable to the *current* value position (the eval's routing state), not just the root.
2. `Validator.feed`: for a deferred scalar at a no-content position, forward the fragment instead of declining; ensure `eval` is fed the scalar exactly once (at completion).
3. Confirm re-feed safety / element counting is unaffected (the "feed eval once" invariant).

### Out of scope
Per-keyword *incremental* validation of constrained values (e.g. running a regex across fragments) — that would require making `Eval` itself fragment-aware, a much larger redesign.

### Tests
- An unconstrained fragmented value larger than a small `MAX_VALUE_SIZE` must **not** be `REJECTED` (proves it streamed rather than buffered to the cap).
- A constrained fragmented value still validates correctly (the Option A behavior, preserved).

jfallows:
## Summary

Makes `common-json` retain whole values only when a consumer actually needs them. The source streams values as fragments and **forgets** each fragment once relayed; a consumer that needs a value whole **opts in** by declining its fragments via `JsonController.consumed(0)`, and the source accumulates them and re-presents the value whole. This removes unconditional whole-value buffering from the source while keeping correctness for consumers that need the whole value (regex/enum/numeric bounds, `getBigDecimal`).

Self-contained to `common-json`; no engine/binding changes.

## Motivation

Previously the tokenizer retained values regardless of whether any consumer needed them — most visibly `numberLexeme`, which accumulated every fragmented number's full lexeme unconditionally. The streaming pipeline already pushes back via `consumed()`; this work uses that same channel as the opt-in retention signal, so "free if not retained, buffered only if retained" falls out of the existing back-pressure mechanism.

## What changed (each increment test-first, suite-verified)

1. **`getBigDecimal()` completeness assert** — asserts the value is complete (`!deferredBytes()`) so a caller can't silently read a partial number.
2. **`consumed(0)` accumulation (strings)** — on resume the tokenizer drops only the *consumed* prefix of `scratch` and keeps the rest, so a decliner's fragments accumulate and the value is re-presented whole. Full consumption behaves exactly as before.
   - **fix**: ship a fragment only when new bytes were scanned (`streamOffset > fragmentStart`); the prior `scratch.length() > 0` guard re-shipped retained-but-unconsumed content and spun the pump.
3. **Incremental number validation** — RFC 8259 grammar validated by a state machine one char at a time (state persists across windows), so no whole lexeme is needed to validate a fragmented number.
4. **Delete `numberLexeme`/`numberFragmented`** — numbers accumulate in `scratch` via the same `consumed(0)` path; `getBigDecimal()` reads the whole value from `scratch`; `getInt()`/`getLong()` use a lexeme-length guard. **Contract change (approved):** reading a window-fragmented number whole now requires the consumer to decline its fragments — a forwarding/streaming consumer no longer gets the whole number retained for free.
5. **Fail-closed cap** — new `JsonParserEx.MAX_VALUE_SIZE` (Integer chars, default `1<<24`) bounds retained `scratch`; overflow throws a resource-exhaustion `JsonParsingException` surfaced as `Status.REJECTED`. Verbatim/segment values (raw-byte streamed, never accumulating) are exempt.
6. **Fragment-aware schema validator** — `JsonSchemaImpl.Validator` declines a deferred scalar until whole, then validates the complete value (`Eval` is not fragment-aware); validates a complete scalar **before** forwarding so eval reads the whole value, not the sink-consumed remainder; keeps structural events/keys forward-first to preserve emit-then-reject. Fixes a latent bug where a window-fragmented value was validated on its first fragment only.

## Follow-up

Per-keyword streaming of *unconstrained* values through the validator (forward fragments + feed eval once at completion) is deferred to #1926 — it needs a per-position "needs content" signal and is out of scope here.

## Testing

Full `common-json` suite (46 test classes) green, plus checkstyle and license, run under a Surefire fork-timeout guard. New tests pin: the `getBigDecimal` completeness assert, `consumed(0)` re-present for strings and numbers, the fail-closed cap, and window-fragmented validation against a `pattern` (match and violation).

---
_Generated by [Claude Code](https://claude.ai/code/session_019YL2uqgwVoN3nqUxnfaj6g)_

### DIFF
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonParserEx.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonParserEx.java
@@ -30,6 +30,15 @@ import org.agrona.DirectBuffer;
 public interface JsonParserEx extends JsonParser
 {
     /**
+     * Config key (for {@link JsonEx#createParser(java.util.Map)}) whose {@link Integer} value caps the
+     * number of decoded chars retained for a single value or key. A consumer that needs a value whole
+     * declines its fragments (consumed(0)) so the source accumulates them; this bounds that accumulation
+     * so an adversarially large value fails closed with a {@link JsonPipeline.Status#REJECTED} rather than
+     * exhausting memory. Absent ⇒ a generous default.
+     */
+    String MAX_VALUE_SIZE = "io.aklivity.zilla.runtime.common.json.parser.maxValueSize";
+
+    /**
      * Borrows {@code buffer} as the input window for the next pump, the half-open range
      * {@code [offset, limit)}. The buffer is read in place for the duration of the pump.
      * <p>
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonParserImpl.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonParserImpl.java
@@ -525,6 +525,10 @@ public final class JsonParserImpl implements JsonParserEx
     public BigDecimal getBigDecimal()
     {
         assert currentEvent == Event.VALUE_NUMBER;
+        // getBigDecimal() yields the whole value, so the lexeme must be complete; a caller that needs a
+        // value still spanning windows must first gather it (e.g. push back via consumed(0)) and only
+        // read it once the deferred bytes have arrived
+        assert !deferredBytes();
         return new BigDecimal(numberLexeme().toString());
     }
 
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonParserImpl.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonParserImpl.java
@@ -188,6 +188,13 @@ public final class JsonParserImpl implements JsonParserEx
         }
         else
         {
+            // before resuming a value that spans windows, tell the tokenizer how much of the prior
+            // fragment the consumer took (the char cursor) so it keeps the unconsumed remainder and
+            // accumulates rather than discarding a declined fragment
+            if (tokenizer.fragmenting())
+            {
+                tokenizer.markScratchConsumed(stringViewOffset);
+            }
             try
             {
                 result = tokenizer.advance(in);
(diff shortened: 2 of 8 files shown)


jfallows (commented):
LGTM

PR #1934, opened 2026-06-20
Introduce BudgetCredit and BudgetDebit handle APIs

jfallows:
## Description

Introduces new `BudgetCredit` and `BudgetDebit` handle-based APIs to replace the lower-level `BudgetCreditor` and `BudgetDebitor` interfaces in bindings. This is Phase 1 of a budget API modernization effort.

### Changes

**New Public APIs:**
- `BudgetCredit`: Consumer-side flow control handle for granting downstream capacity
- `BudgetDebit`: Producer-side flow control handle for claiming send capacity
- `BudgetFlusher`: Callback interface for resuming parked senders when capacity becomes available
- `EngineContext.supplyCredit(long, long)`: Factory method for obtaining `BudgetCredit` handles
- `EngineContext.supplyDebit(long, long, BudgetFlusher)`: Factory method for obtaining `BudgetDebit` handles

**New Facade Implementations:**
- `FacadeBudgetCredit`: Behavior-preserving adapter over `BudgetCreditor`
- `FacadeBudgetDebit`: Behavior-preserving adapter over `BudgetDebitor` that:
  - Owns the acquire/claim/release lifecycle
  - Folds declared per-frame padding into claimed frames
  - Composes declared minimum with each claim's minimum
  - Delegates to existing six-argument `BudgetDebitor.claim()` with deferred=0

**Deprecations:**
- `EngineContext.creditor()`: Marked `@Deprecated`
- `EngineContext.supplyDebitor(long)`: Marked `@Deprecated`

**Migration in MQTT Binding:**
- Updated `MqttServerFactory` and `MqttClientFactory` to use new handle APIs
- Removed direct `BudgetCreditor` and `BudgetDebitor` field dependencies
- Replaced manual acquire/claim/release dance with declarative handle calls
- Simplified budget index tracking by using handle objects instead of sentinel values

### Benefits

- **Unified API**: Single handle per stream unifies per-stream window and shared budget
- **Simplified Binding Code**: Eliminates hand-coded window arithmetic and lifecycle management
- **Backward Compatible**: Facade implementations preserve existing behavior during migration
- **Cleaner Semantics**: Explicit `declare()` and `claim()` methods replace implicit state management

### Test Coverage

Added comprehensive unit tests:
- `FacadeBudgetCreditTest`: Tests acquire/credit/release lifecycle and capacity delegation
- `FacadeBudgetDebitTest`: Tests acquire/claim/release lifecycle, padding folding, and minimum composition
- `FacadeBudgetTest`: Integration test verifying credit/debit interaction through real budget machinery

Existing MQTT binding tests continue to pass with the new handle-based implementation.

https://claude.ai/code/session_01MZnr4yxJ4NFdtvyhHWjwX3

### DIFF
--- a/runtime/binding-fan/src/main/java/io/aklivity/zilla/runtime/binding/fan/internal/stream/FanServerFactory.java
+++ b/runtime/binding-fan/src/main/java/io/aklivity/zilla/runtime/binding/fan/internal/stream/FanServerFactory.java
@@ -16,14 +16,11 @@
 package io.aklivity.zilla.runtime.binding.fan.internal.stream;
 
 import static io.aklivity.zilla.runtime.engine.budget.BudgetCreditor.NO_BUDGET_ID;
-import static io.aklivity.zilla.runtime.engine.budget.BudgetCreditor.NO_CREDITOR_INDEX;
-import static io.aklivity.zilla.runtime.engine.budget.BudgetDebitor.NO_DEBITOR_INDEX;
 import static java.util.function.UnaryOperator.identity;
 
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
-import java.util.function.LongFunction;
 import java.util.function.LongSupplier;
 import java.util.function.LongUnaryOperator;
 
@@ -43,8 +40,8 @@ import io.aklivity.zilla.runtime.binding.fan.internal.types.stream.WindowFW;
 import io.aklivity.zilla.runtime.engine.EngineContext;
 import io.aklivity.zilla.runtime.engine.binding.BindingHandler;
 import io.aklivity.zilla.runtime.engine.binding.function.MessageConsumer;
-import io.aklivity.zilla.runtime.engine.budget.BudgetCreditor;
-import io.aklivity.zilla.runtime.engine.budget.BudgetDebitor;
+import io.aklivity.zilla.runtime.engine.budget.BudgetCredit;
+import io.aklivity.zilla.runtime.engine.budget.BudgetDebit;
 import io.aklivity.zilla.runtime.engine.config.BindingConfig;
 import io.aklivity.zilla.runtime.engine.config.RouteConfig;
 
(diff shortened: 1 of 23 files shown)


runtime/binding-fan/src/main/java/io/aklivity/zilla/runtime/binding/fan/internal/stream/FanServerFactory.java
```diff
@@ -365,22 +360,22 @@ private void onGroupWindow(
...
                 long initialBud = budgetId;
                 if (initialBud == NO_BUDGET_ID)
                 {
                     initialBud = supplyBudgetId.getAsLong();
-                    initialBudIndex = creditor.acquire(initialBud);
+                    initialBudCredit = context.supplyCredit(initialId, initialBud);
                 }
                 this.initialBud = initialBud;
-                this.initialBudIndex = initialBudIndex;
+                this.initialBudCredit = initialBudCredit;
```
jfallows, line 371: Can this be simplified so we only need to track `BudgetCredit` and not also the budget id?

```suggestion
                if (initialBud == null)
                {
                    initialBud = context.supplyCredit(initialId);
                }
```
Perhaps this should even move to the constructor, making initialBud final and guaranteeing non-null, so some of the null guards can be removed where it tests if initialBudCredit != nulll.

jfallows: Agreed in spirit, and `BudgetCredit` can expose its id (`supplyCredit` already receives it), so the owned-budget path wouldn't need to store it separately.

The blocker is fan's credit-less mode: when the upstream WINDOW already advertises a `budgetId`, fan reuses that id and creates **no** `BudgetCredit` (`initialBudCredit` stays null, and the `!= null` guard skips crediting so it doesn't double-credit the parent budget). It still has to keep that inherited id to broadcast to fan-out members, including late joiners (lines 385 and 595) — and with no credit object in that mode, there's nothing to read the id from. So `initialBud` can't be fully removed today without making fan always create a credit and gate crediting on a separate "owns budget" flag.

That inherited-parent-budget case is exactly the budget-chaining scenario the engine will manage internally later, so I'd prefer to defer this tidy-up (field removal + dropping the null guards, and moving toward a final/non-null handle) to that work rather than change fan's flow-control semantics here. Tracking it for then.

---

PR #1936, opened 2026-06-20
fix(common-json): compile JSON Schema pattern with ECMA-262 brace semantics

### LINKED ISSUE #1935: common-json: JSON Schema `pattern` compiled with Java Pattern rejects valid ECMA-262 regexes
## Summary

`common-json`'s `JsonSchemaImpl.compilePattern(...)` compiles a JSON Schema `pattern` keyword using Java's `java.util.regex.Pattern.compile(...)`. JSON Schema regexes are **ECMA-262**, which is more lenient than Java's flavor — most notably, a `{` that does not form a well-formed quantifier is treated as a **literal** in ECMA, but throws `PatternSyntaxException` in Java. As a result, valid JSON Schema `pattern`s fail to compile and the whole engine configuration fails to load.

## Reproduction

A schema property with a pattern containing a literal brace group, e.g.:

```json
"pattern": "^[^:]+(?:-({[^}]+}|[^.]+))?(?::(\\d+)\\+)?$"
```

fails at config load with:

```
java.util.regex.PatternSyntaxException: Illegal repetition near index 12
^[^:]+(?:-({[^}]+}|[^.]+))?(?::(\d+)\+)?$
            ^
  at io.aklivity.zilla.runtime.common.json.internal.JsonSchemaImpl.compilePattern(JsonSchemaImpl.java:607)
  ...
  at io.aklivity.zilla.runtime.engine.internal.registry.EngineManager.onPathChanged(EngineManager.java:271)
io.aklivity.zilla.runtime.engine.config.ConfigException: Engine configuration failed
```

This surfaced when building `aklivity/zilla-plus` `binding-kafka-proxy` against a current `develop` engine: the `kafka-proxy` host schema pattern above is valid ECMA-262 (the `{...}` is a literal placeholder), but the engine now rejects it, so **every** kafka-proxy IT errors at config load. A prior released engine (`1.2.4`) did not exercise this path.

## Expected

JSON Schema `pattern` keywords should be evaluated with ECMA-262 semantics (or at least leniently enough that an ECMA-valid pattern compiles), so a literal `{`/`}` that is not a valid Java quantifier does not break config loading.

## Possible directions

- Pre-process schema patterns to escape lone/literal braces before `Pattern.compile`, or
- Use an ECMA-compatible regex evaluation path for the `pattern` keyword.

## Workaround (downstream)

In `aklivity/zilla-plus`, the kafka-proxy schema pattern was temporarily changed from `{[^}]+}` to `\{[^}]+\}` so it also compiles under Java `Pattern`. This should be reverted once the engine handles ECMA patterns.

jfallows:
## Summary

Fixes #1935.

`common-json`'s `JsonSchemaImpl.compilePattern(...)` compiled a JSON Schema `pattern` keyword with Java's `java.util.regex.Pattern.compile(...)`. JSON Schema regexes are **ECMA-262**, where a `{` that does not form a well-formed quantifier is treated as a **literal**, whereas Java throws `PatternSyntaxException`. As a result, valid schema patterns (e.g. a literal `{...}` placeholder group) failed to compile and the whole engine configuration failed to load.

## Changes

- `JsonSchemaImpl.compilePattern(...)` now pre-processes the pattern (after the existing `\p{...}` unicode-property translation) to escape braces that do not open a valid quantifier (`{n}`, `{n,}`, `{n,m}`).
- Unicode property escapes (`\p{...}`, `\P{...}`, `\x{...}`) and character classes are left untouched, and existing valid quantifiers are preserved.
- This path is shared by both the `pattern` keyword and `patternProperties`.

## Tests

Added two tests to `JsonSchemaTest` (test-first; confirmed they fail with `PatternSyntaxException` against the unfixed code, then pass with the fix):

- `shouldTreatNonQuantifierBraceAsLiteral` — a `{` not forming a quantifier is matched as a literal, while real quantifiers still apply.
- `shouldCompileEcmaLiteralBracePlaceholderPattern` — reproduces the kafka-proxy host schema pattern from the issue (`^[^:]+(?:-({[^}]+}|[^.]+))?(?::(\d+)\+)?$`).

`JsonSchemaTest` passes (27/27) and `checkstyle:check` is clean for the module.

## Follow-up

The downstream workaround in `aklivity/zilla-plus` (`{[^}]+}` → `\{[^}]+\}` in the kafka-proxy schema) can be reverted once an engine containing this fix is released.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QRF4YGLs99kMV5n6eryrU3)_

### DIFF
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonSchemaImpl.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonSchemaImpl.java
@@ -604,7 +604,103 @@ public final class JsonSchemaImpl implements JsonSchema
             matcher.appendReplacement(result, replacement);
         }
         matcher.appendTail(result);
-        return Pattern.compile(result.toString());
+        return Pattern.compile(escapeLiteralBraces(result.toString()));
+    }
+
+    // ECMA-262 treats a "{" that does not form a well-formed quantifier as a literal, whereas
+    // java.util.regex rejects it with PatternSyntaxException; escape such braces so an ECMA-valid
+    // pattern compiles. Valid quantifiers ({n}, {n,}, {n,m}), unicode property escapes (\p{...},
+    // \P{...}, \x{...}) and character classes are left untouched.
+    private static String escapeLiteralBraces(
+        String regex)
+    {
+        StringBuilder result = new StringBuilder(regex.length());
+        int length = regex.length();
+        int index = 0;
+        boolean inCharClass = false;
+        while (index < length)
+        {
+            char ch = regex.charAt(index);
+            if (ch == '\\' && index + 1 < length)
+            {
+                char escaped = regex.charAt(index + 1);
+                result.append(ch).append(escaped);
+                index += 2;
+                if ((escaped == 'p' || escaped == 'P' || escaped == 'x') &&
+                    index < length && regex.charAt(index) == '{')
+                {
+                    int close = regex.indexOf('}', index);
+                    int end = close < 0 ? length : close + 1;
+                    result.append(regex, index, end);
+                    index = end;
+                }
+            }
+            else if (ch == '[')
+            {
+                inCharClass = true;
+                result.append(ch);
+                index++;
+            }
+            else if (ch == ']')
+            {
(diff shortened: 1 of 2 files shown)


jfallows (commented):
LGTM

PR #1943, opened 2026-06-21
feat(common-json): generalize the verbatim transform pipeline — validate, prune, inject

### LINKED ISSUE #1926: common-json: stream window-fragmented unconstrained values through the schema validator (forward-and-suppress)

jfallows:
## Summary

Generalizes the verbatim primitive introduced for the validate-fidelity fix so a structure-inspecting
transform can preserve original bytes (whitespace included) across the full range of transform shapes:

- **Phase 1 — validate** (the original CI fix): a mediating transform forwards bytes verbatim while still
  inspecting structure.
- **Phase 2 — prune**: drop a value, forwarding the rest verbatim.
- **Phase 3 — inject**: splice a canonical value between verbatim runs, at any position including before a
  container's first member.

The converged design is documented in `runtime/common-json/VERBATIM-VALIDATE-DESIGN.md`.

## Phase 2 — prune (`JsonSource.skipValue()`)

`skipValue()` (with its `JsonParserEx` peer) is called on a matched object `KEY_NAME`: the source advances
the parser past the whole member (key, separator, value — scalar or container-to-close) and the verbatim
cursor past its bytes, so the transform never sees the dropped sub-events.

Occupancy is owned by the source — the tokenizer tracks `memberSeparated` (was this member reached after a
comma), and `skipValue` defers a one-shot leading-separator trim (whitespace + a single comma, bounded to
the parse frontier) onto the next verbatim pull. This settles the design's semantic fork: dropping a
container's only field renders `{}`, while dropping the first of many keeps the survivor's spacing
(`{ "b" : 2 }`).

## Phase 3 — inject (`getPosition()` + generator `seed()`)

`JsonSource.getPosition()` returns a container-anchored `JsonPosition` (a list of `JsonStep`:
`START_*` / `CONTINUE_*` / terminal `KEY_NAME`, occupancy folded into the step kind), built lazily from the
tokenizer's new per-level `pathOccupied` tracking. The generator gains `seed(JsonPosition)` (push one open
frame per step, no bytes emitted) with an A5 seed-baseline assertion on `writeEnd`. The terminal sink seeds
the generator from `getPosition()` on the first generated value after a verbatim copy, so an injected member
gets the correct leading separator without re-emitting the verbatim-written brackets; the following verbatim
run owns its own separator (no double comma). Seed-on-transition is a no-op in canonical and segment
delivery, so validate/prune/passthrough are unaffected.

**Inject before a container's first member** is the mirror of the prune-first trim, on the generator side:
the injected value takes the first slot, displacing the former-first member, whose verbatim bytes (having
been first in the source) carry no leading separator. The sink reads the seed occupancy — `START_*` means
the container was empty — and sets `separatorPending`, so the displaced member's verbatim run is preceded by
a generator-synthesized separator (`JsonGeneratorEx.writeVerbatimSeparator()`). The synthesized comma is
compact, matching the canonical injected member; the between-members case (`CONTINUE_*`) is unchanged.

## Tests & benchmark

- `JsonSkipTest` (renamed from `JsonPruneTest`): drop middle / last / first / object-value / only-field, each
  asserting whitespace-preserving output.
- `JsonInjectTest`: the design's worked example (number member between two verbatim runs), a string-member
  variant, and inject-before-first for the only-field and first-of-many cases.
- `JsonMutateBM`: benchmarks skip, inject, inject-before-first against a verbatim passthrough control;
  transforms reuse constant injected key/value and reused sources, so `-prof gc` measures only the pipeline.
  All paths ~0.005–0.006 B/op, matching the passthrough control — the verbatim primitives splice through
  reused buffer views and a reused position with no per-op allocation.

## Deferred

Injecting into a genuinely *empty* container (append — no member to displace) remains out of scope: this is a
different operation because `getPosition()` at the container's close reports the popped, outer context, so it
needs separate handling. See the resolution note in the design doc.

## Verification

- `common-json`: full build green (4820 tests), JaCoCo class + 90% instruction gate met, 0 checkstyle violations.
- `model-json`: 25 unit + EventIT green; `binding-mcp-http`: 11 unit green.
- Full reactor green except `cloud::docker-image` (no Docker daemon in CI sandbox).


jfallows (commented):
LGTM

PR #1946, opened 2026-06-22
test(engine): make store watch assertion deterministic via reply flush

jfallows:
## Problem

The `test` binding's `watch` store-assertion registered the listener and then
proceeded immediately, so nothing gated the k3po script on the asynchronous
watch notification. Whether the listener callback executed before test teardown
was a race between event delivery and test completion, leaving the watch-listener
code paths covered only intermittently — flaky store-watch coverage.

Every other store op (`get`/`put`/`lock`/`unlock`/`renew`/...) chains the next
step from inside its own async completion callback, so the assertion chain
naturally blocks on each completion. `watch` returns a `Closeable` rather than a
completion, so it alone had no synchronization point.

## Change

- Emit an observable advisory reply **FLUSH** from the watch callback when the
  notification arrives on the dispatch thread and matches the expectation. The
  new `engine.spec` `store.watch` scripts gate the script on
  `read advised zilla:flush`, so `k3po.finish()` blocks until the notification
  has actually fired — the determinism is now structural, not timing-based. This
  also upgrades the test from "passes unless a reset shows up" to a real positive
  assertion that the notification was observed.
- A reply FLUSH is only valid after the reply stream is established, so store
  assertions are now deferred until the reply stream is open and credited (run
  once from `onReplyWindow` instead of on initial begin). All other ops are
  unaffected beyond running one worker turn later.
- Repoint `MemoryStoreHandlerIT.shouldWatchKey` at the new gating scripts.

## Testing

Test-first: the repointed IT fails with a k3po flush-timeout against the current
code (no flush emitted), and passes after the change. The full
`MemoryStoreHandlerIT` (all store ops) passes and coverage checks are met;
checkstyle is clean. Only `store-memory` consumes store assertions in this repo.

---

### DIFF
--- a/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
+++ b/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
@@ -320,6 +320,7 @@ final class TestBindingFactory implements BindingHandler
         private TestTarget target;
         private boolean pendingBegin;
         private long pendingTraceId;
+        private boolean storeAssertionsStarted;
 
         private TestSource(
             MessageConsumer source,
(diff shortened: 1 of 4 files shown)


jfallows (commented):
LGTM

PR #1948, opened 2026-06-22
ci: pass docker image between build jobs via artifact instead of cache

jfallows:
## Problem

The `build` job saves the docker image to `actions/cache` keyed `zilla-develop-SNAPSHOT-${run_id}`, and the parallel `testing` matrix jobs each restore it before `docker load`. `actions/cache` is eventually consistent, so when the `testing` jobs start simultaneously some get a cache hit and some get "Cache not found" for the same key, failing at the `docker load` step before any example runs. Which jobs lose the race is effectively random.

(Same root cause and fix as the zilla-plus build workflow.)

## Change

- `build`: upload the docker image as an artifact (`zilla-image`) instead of caching it.
- `testing`: download the artifact and load it, instead of restoring from cache.
- Add a `cleanup` job (`needs: testing`, `if: always()`) that deletes the transient `zilla-image` artifact once the matrix completes — it is purely an intra-run transport, not a deliverable.

Artifacts are strongly consistent within a run and gated by the `needs` chain, so every example job gets the image deterministically. The Maven-packages, ZPM, and base-image caches are unchanged (those are legitimate cross-run caches).

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,12 +95,13 @@ jobs:
         mkdir -p ~/docker-cache
         docker save -o ~/docker-cache/docker.tar $(docker images --format "{{.Repository}}" | grep "zilla")
 
-    - name: Cache Docker image
-      uses: actions/cache@v5
+    - name: Upload Docker image
+      uses: actions/upload-artifact@v7
       if: success()
       with:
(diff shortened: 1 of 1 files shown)


jfallows (commented):
LGTM

PR #1951, opened 2026-06-22
fix(model): preserve content across bounded output windows on json/avro/protobuf view paths

jfallows:
## Problem

`#1933` migrated `binding-http` and the `binding-kafka` cache onto a per-stream `ModelPipeline` that drives model transforms into a **bounded output window** (`transform(src, …, dst, dstLimit)` with `SUSPENDED`/`OVERFLOW` resume). This exposed a class of byte-fidelity bugs on the `X → JSON` paths: content written without honouring the output bound overran the limit. It only manifested across bounded windows and with assertions disabled (production runs `-da`), so single-window unit tests stayed green.

## Changes (one commit per module)

- **`fix(common-json)`** — a byte-replay (validate-only) forward dropped the trailing insignificant bytes after the top-level value (e.g. the newline after `}`), so the forwarded body was one byte short of its `Content-Length` and the client saw a truncated transfer. The parser stays faithful/mode-neutral — `END_DOCUMENT` carries the trailing bytes — and the sink decides: a whole-document byte-replay delivery replays them (byte-identical), canonical / member-dropping deliveries normalize as before. **Fixes the `http.json.schema` example.**
- **`fix(common-avro)`** — a record field's structural key prefix (`,"name":`) was written unbounded and could overrun the window; `AvroJsonGeneratorImpl.remaining()` now reserves it, and the shared `JsonGeneratorImpl.putRaw` is hardened to throw `BufferOverflowException` instead of silently overrunning under `-da`.
- **`fix(common-protobuf)`** — the same key-prefix overrun on the length-delimited segment path, fixed proactively with coverage (no example exercised it).

## Testing

Test-first, each with a kept regression test that fails first and passes after:
- `JsonPipelineTransformTest.shouldForwardTrailingNewlineVerbatim`
- `AvroJsonChunkingTest.shouldPreserveStringValuesAcrossStructuralPrefixOverrun`
- `ProtobufJsonChunkingTest.shouldPreserveStringValuesAcrossStructuralPrefixOverrun`

Suites green under `-ea` (and `-da` for these regressions): `common-json` (4827), `common-avro` (120), `model-avro` (35), `common-protobuf`, `model-protobuf` (37); checkstyle clean.

## Related

The `http.kafka.avro.json` example failure had a **separate** root cause (the schema-registry framing prefix not being stripped on the `ModelPipeline` read path) — fixed in **#1953**. Until #1953 merges, the `http.kafka.avro.json` example job on this PR will remain red; the two PRs are complementary.

### DIFF
--- a/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/internal/json/AvroJsonGeneratorImpl.java
+++ b/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/internal/json/AvroJsonGeneratorImpl.java
@@ -140,7 +140,22 @@ public final class AvroJsonGeneratorImpl implements AvroGenerator
     @Override
     public int remaining()
     {
-        return Math.max(0, json.remaining() - RESERVE);
+        return Math.max(0, json.remaining() - RESERVE - pendingKeyWidth());
+    }
+
+    private int pendingKeyWidth()
+    {
+        int width = 0;
+        if (top > 0)
+        {
+            Frame frame = stack[top - 1];
+            if (frame.kind == RECORD && frame.fieldIndex < frame.fields.size())
+            {
+                String name = frame.fields.get(frame.fieldIndex).name();
+                width = name.length() + 3 + (frame.fieldIndex > 0 ? 1 : 0);
+            }
+        }
+        return width;
     }
 
     @Override
--- a/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/internal/json/AvroJsonGeneratorImpl.java
+++ b/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/internal/json/AvroJsonGeneratorImpl.java
@@ -87,6 +87,8 @@ public final class AvroJsonGeneratorImpl implements AvroGenerator
     private boolean valueOpen;
     private boolean valueQuoteOpen;
     private AvroKind valueKind;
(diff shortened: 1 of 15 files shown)


jfallows (commented):
LGTM

runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/internal/JsonGeneratorImpl.java
```diff
@@ -1033,7 +1034,10 @@ private static int escapedWidth(
     private void putRaw(
         int value)
     {
-        assert progress < limit;
+        if (progress >= limit)
+        {
+            throw new BufferOverflowException();
+        }
```
jfallows: If this is already covered by tests, then revert to assert.

jfallows: Reverted to `assert` (force-pushed). The real bound-defense is `AvroJsonGeneratorImpl.remaining()` reserving the next field's key-prefix width so the sink suspends before it — the throw was a redundant safety net, and the overrun discipline is covered by the `*JsonChunkingTest` cases under `-ea`. Follow-up in progress to make the assert a true by-construction invariant by fragmenting the avro/protobuf key prefix (like plain-JSON key names), so a prefix larger than the output window drains across windows rather than relying on the window being large enough.

---

PR #1955, opened 2026-06-23
fix(common-json): preserve document trailing bytes on the verbatim-forward path

jfallows:
## Problem

A validate-and-forward JSON pipeline (parser → validator → sink, no member-modifying transform) dropped a document's trailing insignificant bytes — e.g. a trailing newline. The HTTP response body then came up one byte short of its upstream `Content-Length` (40 of 41 bytes for the `http.json.schema` example), and the client hung waiting for the byte that never arrived.

Reproduced with the `http.json.schema` example: `GET /valid.json` returns a pretty-printed document whose `www/valid.json` source ends in `\n`. The validator forwarded the document body but not the trailing newline, truncating the response by one byte.

## Fix

Model document framing as verbatim steps rather than a bespoke sink path.

- `START_DOCUMENT` and `END_DOCUMENT` join `JsonStep` as **structurally inert** framing steps: they open or close no container and move no depth or member state.
- The parser aligns the verbatim cursor to the top-level close and tags the `END_DOCUMENT` step with the parse frontier, so its byte range is exactly the document's trailing `[documentEnd, frontier)` run — whether the body rode as `VERBATIM` or as a `SEGMENT` run.
- A byte-preserving sink splices that block through the same `writeVerbatim` path as any other run, suspending and resuming under a bounded output until the trailing bytes drain. A canonical (re-rendering) sink writes nothing here and drops the trailing bytes, as before.
- This replaces the `documentReplayed` flag and the bespoke `writeTrailing` path in `JsonSinkImpl`.

### Multi-document correctness

`END_DOCUMENT` is now held until the document boundary is known. On a non-terminal window the trailing bytes are still ambiguous (more could follow, or another document), so the parser starves before the boundary rather than emitting a premature `END_DOCUMENT` that could swallow a following document's leading bytes. On the terminal window (`last == true`) the boundary is decidable and `END_DOCUMENT` is emitted with the full trailing run.

## Tests

- `JsonValidatorVerbatimTest` — forwards a trailing newline verbatim, through a 16-byte bounded output, and across multiple documents preserving inter-document bytes.
- `JsonParserDocumentTest` — defers `END_DOCUMENT` until the boundary is known on a non-terminal window.
- Existing projector/inject/skip/validator-chain tests updated to expect trailing-byte preservation on pure-forward paths.
- Full `runtime/common-json` suite green (4831 tests).

---

### DIFF
--- a/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonStep.java
+++ b/runtime/common-json/src/main/java/io/aklivity/zilla/runtime/common/json/JsonStep.java
@@ -25,11 +25,15 @@ package io.aklivity.zilla.runtime.common.json;
  * copying bytes never needs a scalar's type, only that a value occupies a slot. {@link #SEPARATOR} marks a
  * value separator present in the block's bytes; only a block's leading/trailing separator is consulted (to
  * stitch an injected value at a block seam without doubling or dropping a comma), interior separators ride in
- * the bytes and move no state. There is no {@code START_DOCUMENT}/{@code END_DOCUMENT} — framing is not part of
- * a spliced run.
+ * the bytes and move no state. {@link #START_DOCUMENT} and {@link #END_DOCUMENT} are structurally inert framing
+ * steps — they open or close no container and move no depth or member state — present so a verbatim block can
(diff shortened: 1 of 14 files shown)


jfallows (commented):
LGTM

PR #1957, opened 2026-06-23
feat(engine): add identity() capability to ModelPipeline

### LINKED ISSUE #1956: engine: add identity() capability to ModelPipeline so callers can skip buffer-and-hold for validators
## Background

The model SPI migration collapsed the former `ValidatorHandler` / `ConverterHandler` distinction into a single `ModelPipeline`. That distinction implicitly told the caller something useful: a *validator* only validates (it is **identity on accept** — it never rewrites bytes, it only rejects), whereas a *converter* may change the value's bytes and/or length. Collapsing both into `ModelPipeline.transform(...)` lost the caller's ability to know, up front, that a given model will not alter the bytes.

## Problem

A caller that forwards a length-prefixed and/or CRC-protected wire format must, whenever a model is active, **buffer-and-hold the whole message, recompute the length framing, and recompute the CRC** — because it cannot know in advance whether the transform will change the bytes or their length. For a **validator** (identity on accept) this work is entirely unnecessary: the bytes are unchanged, so the original framing and CRC remain valid and the value could stream through untouched.

`HttpModel`, `KafkaCacheModel`, and `SseModel` each drive the pipeline and copy produced bytes even for validators, where a passthrough would suffice.

## Proposal

Add an `identity()` capability to `ModelPipeline` (and/or `ModelHandler`):

```java
default boolean identity()
{
    return false;
}
```

- **Default `false`** — this is a correctness requirement, not just a conservative choice. A transforming model that forgot to override must still be buffered/recomputed; defaulting to `true` would let it forward a stale CRC/length. Validators *opt in* to `identity() == true` to earn the fast path.
- Read validators override to `true` (json/avro/protobuf/test as appropriate). For the engine **test model**, `identity()` is `transformLength < 0` (i.e. no `transform: { length: N }` configured).
- Callers gate on it:
  - `identity()` → cheap passthrough: stream original bytes, no buffer, no CRC recompute (validators still reject as today).
  - otherwise → buffer-and-hold + reframe + recompute.

## Notes

There are really three tiers — (1) identity (bytes unchanged), (2) length-preserving but content-changing (CRC must be recomputed, no length resize), (3) length-changing (full reframe). A single `identity()` boolean cleanly separates tier 1 (validators → cheap path) from tiers 2–3 (must do work); distinguishing 2 from 3 is a finer optimization probably not worth extra API surface. Note `ModelPipeline.padding()` already exists but means "reserved overhead," not "won't change bytes," so it does not serve this purpose.

## Scope

- Engine: add `identity()` to the `ModelPipeline` SPI (default `false`); override `true` in the read validators; unit coverage.
- Sibling bindings (`HttpModel`, `KafkaCacheModel`, `SseModel`): gate their drivers on `identity()` to skip the produced-bytes copy for validators.

This is a deferred optimization layered on top of the existing (correct) buffer-and-hold behavior — callers that already buffer unconditionally for active models stay correct; the `identity()` fast-path simply lets them avoid the work for validators once the signal is available.

jfallows:
Fixes #1956

## Summary

Adds `ModelPipeline.identity()` so a caller can tell, up front, that a model is a *validator* — it leaves accepted values byte-for-byte unchanged — and skip the buffer-and-hold / reframe / checksum-recompute work that only a content-changing transform requires.

## Engine SPI

- `ModelPipeline.identity()` — new `default false` method. The default is a correctness requirement, not just conservatism: a transforming pipeline that forgot to override it must still be buffered and reframed, so a pipeline opts in to `true` only when it never alters an accepted value.

## Read-validator overrides

- `JsonReadModelPipeline` → `true` (a json read model only validates against its schema).
- `AvroReadModelPipeline` / `ProtobufReadModelPipeline` → `true` unless a converting `view: json` is configured (delegated to a handler-level `identity()`).
- Engine `TestModelPipeline` → `transformLength < 0`.

## Consumer: binding-mqtt

`MqttModel.identity()` plus a fragment-level `MqttModel.validate(...)`. `MqttServerFactory.decodePublishPayload` now gates on `identity()`: an identity model validates the publish payload fragment-by-fragment and streams the **original** bytes, so a payload larger than the decode slot no longer has to be buffered whole (and rejected with `PAYLOAD_FORMAT_INVALID`) just to recompute its length. The change is isolated to identity-active models — non-identity and unmodeled paths are byte-for-byte unchanged, and validation is gated on `canPublish` so rejection timing matches the existing modeled path.

`HttpModel`, `SseModel`, and `KafkaCacheModel` are intentionally **not** gated: HTTP/SSE do not reframe length, and the Kafka cache always stores the message, so `identity()` saves no work there.

## Tests

- Unit coverage for `identity()` added to `TestModelHandlerTest`, `JsonReadModelPipelineTest`, `AvroReadModelPipelineTest`, `ProtobufReadModelPipelineTest`, and `MqttModelTest` (including the streaming `validate`).
- MQTT `PublishIT` passes through k3po — v5 51/0, v4 25/0, client-v5 8/0 — covering both the identity validator config (`server.validator.yaml`, the new streaming path) and the content-transform configs (unchanged path).

## Follow-up

The new capability — an identity payload exceeding the decode slot streaming through instead of being rejected — is exercised by the existing in-slot identity IT but not yet by a dedicated oversized-payload spec script. A k3po script for that case would be a worthwhile addition.

https://claude.ai/code/session_01Py5AeLjwkYueJxYyP15GjA

---
_Generated by [Claude Code](https://claude.ai/code/session_01Py5AeLjwkYueJxYyP15GjA)_


jfallows (commented):
LGTM

jfallows (commented):
LGTM

PR #1972, opened 2026-06-28
Add scope-based filtering for MCP tools list with guard support

### LINKED ISSUE #1831: binding-mcp: per-client `tools/list` filtering by per-tool authorization metadata
## Summary

Filter the aggregated MCP proxy listing (`tools/list`, and likewise `prompts/list` /
`resources/list`) **per connecting client** so the gateway never advertises tools the
client can't invoke, using **per-tool authorization metadata**. Distinguish **acquirable**
tools (the client can elicit the required scope → list them and prompt on use) from
**non-acquirable** tools (the client could never obtain the scope → filter them out, the
dead-end case), and fall back to **per-identity live listing** where metadata is absent.

Split out of #1810 (was its Phase 8). #1810 delivers the per-toolkit OAuth signaling
(non-blocking `tools/list` + per-toolkit `elicitation/create` + `notifications/tools/list_changed`)
and per-`(identity, route)` execution; this issue is the **listing-accuracy / per-client
filter** layered on top of that, plus the cached-baseline ∪ per-identity hybrid serve.

## Motivation

Without per-tool authorization metadata the proxy may **advertise tools a client can never
invoke** (listing-accuracy / dead-ends). The shared cache is a **role-neutral catalog**; the
per-client view must be computed live. Today the `Tool` schema carries **no roles/scopes**
(only `annotations` hints + `_meta`), and SEP-1488 (`securitySchemes`, oauth2 + scopes) is
**draft**, so the filter must degrade gracefully across metadata sources.

## Design

The shared cache is a role-neutral catalog; the per-client view is computed live.

- **No cache credential** → the cache is a static shared list of auth-free tools. Routes
  requiring auth contribute nothing; when a client authorizes such a route, those tools are
  **per-identity** and the proxy fires `list_changed` to **that** client only. (Signaling
  delivered by #1810.)
- **List-only superset credential** → the cache holds the full catalog. To filter it per
  client without advertising un-invokable tools, we need **per-tool authorization metadata**.
  Sources, best-first (the tag is a tool property → cacheable):
  1. **SEP-1488 `securitySchemes`** when the upstream emits it;
  2. **operator-declared mapping** (tool name/glob → required scope/role) — authoritative;
  3. **annotation-derived default** (`readOnlyHint` / `destructiveHint`) — advisory only.
- Where metadata is absent, fall back to **per-identity live listing** (let the upstream
  filter by the client's token) or accept optimistic-list + reject-on-call.
- Distinguish **acquirable** (client can elicit the scope → list it, prompt on use) from
  **non-acquirable** (client could never obtain it → filter out, the dead-end case).

### Cached-baseline ∪ per-identity hybrid serve

`McpProxyListFactory.newStream` is currently cache-XOR-live: with a cache it serves the
shared baseline only (`McpCacheListServer`); the live path (`McpListServer`, no cache) fans
out per-identity using the inbound authorization. So a newly-authorized per-identity toolkit
appears on re-list only if (i) the route is cache-less (live per-identity fan-out works
today) **or** (ii) this issue adds the hybrid **"cached baseline ∪ live per-identity merge"**.
The per-identity execution token stays per `(identity, route)`, never in the shared store
(cross-user leakage).

### Enforcement note

Filtering the list is **advisory**. Restrictions the upstream can't model (gateway-level role
policy) must be enforced **authoritatively by Zilla at `tools/call`** — the upstream remains
the authoritative enforcer for its own scopes (a client lacking a scope gets `401/403`).

## Dependencies

Builds on #1810: the protocol bump + `elicitation.url` negotiation, the guard-driven
`preauthorize → elicit → callback → reauthorize` flow, per-`(identity, route)` execution
storage, and the non-blocking `tools/list` + `list_changed` signaling. Not required for the
Track A OSS-relay example (cache-less routes get per-identity listing free from the upstream).

## References

- SEP-1488 (per-tool `securitySchemes`, draft): https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1488
- MCP tool annotations (`readOnlyHint` / `destructiveHint`)
- Parent: #1810

jfallows:
## Description

This change adds scope-based filtering capabilities to the MCP binding's tools list functionality, enabling fine-grained access control based on security schemes and guard configurations.

### Key Changes

1. **New `McpScopeFilter` transform** (`McpScopeFilter.java`)
   - Implements JSON transformation to filter array items based on scope requirements
   - Parses `securitySchemes` from items and evaluates them against a configurable predicate
   - Drops items that don't meet scope requirements while preserving structure

2. **Guard integration in proxy list handling** (`McpProxyListFactory.java`)
   - Extended `McpListClient` to support guard-based authorization checks
   - Added scope extraction from `securitySchemes` in tools list items
   - Integrated `GuardHandler.verify()` to validate scopes against session authorization
   - Defers item emission until scope requirements are evaluated

3. **Security scheme injection in MCP client** (`McpClientFactory.java`)
   - Added role-based security scheme streaming injection for tools list responses
   - Injects OAuth2 schemes with appropriate scopes based on guarded route roles
   - Uses streaming injection at tool object boundaries without buffering

4. **Scope indexing in cache** (`McpProxyCache.java`)
   - Builds index of tool names to their required scopes during cache operations
   - Exposes `scopesByName()` for downstream filtering logic

5. **Guard API enhancement** (`GuardHandler.java`)
   - Added `verify(long sessionId, List scopes)` method to test scope membership
   - Allows bindings to perform dynamic, per-item authorization checks

6. **Route configuration updates** (`McpRouteConfig.java`, `McpBindingConfig.java`)
   - Added `getRoles()` method to expose role mappings from guarded route configs
   - Supports role-based security scheme generation

7. **Schema update for client routes**
   - Removed `"routes": false` restriction from client kind schema
   - Allows client binding to have routes with guarded entries for role configuration

### Test Coverage

- Added `ScopeFilterTest` with comprehensive unit tests for the scope filtering logic
- Added integration test `shouldServeToolsListFilteredBySecuritySchemes()` with guarded configuration
- Added integration test `shouldListToolsSecuritySchemes()` for client-side injection
- New test specification files for security scheme filtering scenarios

### Behavior

When a route is guarded:
- Tools list items are scanned for `securitySchemes` with scope requirements
- Each item's scopes are validated against the session's authorization via the guard
- Items with unmet scope requirements are dropped from the response
- Items without security schemes are always included
- Responses are transparently filtered without requiring client changes

Fixes #1831

https://claude.ai/code/session_017bkRD2Kvc815vhafFavV9F


runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
```diff
@@ -395,6 +395,21 @@ static String naturalAuthority(
             : authority;
     }
 
+    public Map<String, List<String>> getRoles(
+        long routeId)
+    {
+        Map<String, List<String>> result = Map.of();
```
jfallows: Hoist `Map.of()` to constant, avoid allocation.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpRouteConfig.java
```diff
@@ -65,6 +71,13 @@ public McpRouteConfig(
             .orElse(null);
     }
 
+    public Map<String, List<String>> getRoles()
+    {
+        return guarded.stream()
+            .filter(g -> !g.roles.isEmpty())
+            .collect(toMap(g -> g.name, g -> g.roles, (a, b) -> a, LinkedHashMap::new));
```
jfallows: Perhaps this can be done in the constructor instead of per call?

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -382,6 +399,7 @@ private void checkGet(
                 final long newChecksum = crc32.getValue();
                 changed = lastChecksum != -1L && lastChecksum != newChecksum;
                 lastChecksum = newChecksum;
+                rebuildScopeIndex(value);
```
jfallows:
```suggestion
                scopesByName = indexScopesByName(value);
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpProxyCache.java
```diff
@@ -403,5 +421,78 @@ private void checkPut(
             degraded = false;
             checkReady();
         }
+
+        private void rebuildScopeIndex(
```
jfallows: This approach to fully parse the string into a json document and walk is incorrect.

Use JsonParserEx to extract scopes using incremental approach.

The scopesByName should be empty or non-empty but not null.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
```diff
@@ -4804,10 +4806,32 @@ void doEncodeStreamedRequestEnd(
...
+                this.schemesLength = 0;
+                this.toolEndStreamOffsets = null;
+            }
+            else
+            {
+                final byte[] bytes = buildSchemesJson(roles);
+                this.schemesBuffer = new UnsafeBufferEx(bytes);
+                this.schemesLength = bytes.length;
+                this.toolEndStreamOffsets = new long[16];
+            }
```
jfallows: This should be done as a JsonPipeline on the payload body, with a JsonTransform to detect and inject the tool securitySchemes (based on route roles, if any) with the response, not allocating on the hot path.

When there are no roles to inject, the JsonPipeline would fall through efficiently, so no need to provide an alternate implementation for that case.

runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/stream/McpHttpProxyFactory.java
jfallows, line 2115: We should not be building the entire document like this, see the feedback on using JsonPipeline and JsonTransform instead.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -971,18 +993,63 @@ private int decodeItemScan(
...
+                    if (client.guard != null && client.decodedSchemesFound && !client.decodedNoAuth)
+                    {
+                        final List<String> scopes = client.decodedScopes != null
+                            ? client.decodedScopes
+                            : List.of();
+                        admitted = client.guard.verify(client.server.authorization, scopes);
+                    }
+                    if (admitted)
+                    {
+                        client.onDecodedItemBegin(traceId);
```
jfallows: Avoid so much if-else style coding.
Use strategy approach instead.
For example, instead of passing guard, pass a lambda for guard.verify as a functional interface, then when there is no guard, pass a lambda that behaves correctly for no guard, but the receiving code always calls the same way, with no special case if-else logic.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyListFactory.java
```diff
@@ -971,18 +993,63 @@ private int decodeItemScan(
...
                 }
                 break;
             case KEY_NAME:
-                if (client.decodeItemDepth == 1 &&
-                    client.idKey.equals(parser.getString()))
+                if (client.decodeItemDepth == 1)
                 {
-                    client.decoder = decodeItemName;
-                    break decode;
+                    final String key = parser.getString();
```
jfallows: We have `.getStringView()` on the parser to avoid unnecessary allocation, please use it.

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/transform/McpScopeFilter.java
```diff
@@ -0,0 +1,351 @@
...
+        }
+        mediatingControl.synthetic = false;
+        if (status == Status.ADVANCED)
+        {
+            status = sink.transform(mediatingControl, source, event);
+        }
+        return status;
+    }
+
+    private static final class MediatingController implements JsonController
```
jfallows:
```suggestion
    private static final class Control implements JsonController
```

runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/transform/McpScopeFilter.java
```diff
@@ -0,0 +1,351 @@
...
+    public Status transform(
+        JsonController control,
+        JsonSource source,
+        JsonEvent event,
+        JsonSink sink)
+    {
+        mediatingControl.delegate = control;
+        return inItems
+            ? onItems(source, event, sink)
+            : onOuter(source, event, sink);
```
jfallows: Use a state machine approach, with signature matching onItems / onOuter, similar to decoders.

That should eliminate most of the internal tracking booleans and make the code more readable.

runtime/binding-mcp/src/test/java/io/aklivity/zilla/runtime/binding/mcp/internal/transform/ScopeFilterTest.java
```diff
@@ -0,0 +1,141 @@
...
+import io.aklivity.zilla.runtime.common.agrona.buffer.DirectBufferEx;
+import io.aklivity.zilla.runtime.common.agrona.buffer.MutableDirectBufferEx;
+import io.aklivity.zilla.runtime.common.agrona.buffer.UnsafeBufferEx;
+import io.aklivity.zilla.runtime.common.json.JsonEx;
+import io.aklivity.zilla.runtime.common.json.JsonGeneratorEx;
+import io.aklivity.zilla.runtime.common.json.JsonParserEx;
+import io.aklivity.zilla.runtime.common.json.JsonPipeline;
+import io.aklivity.zilla.runtime.common.json.JsonPipelineResult;
+
+public class ScopeFilterTest
```
jfallows: This should already be exercised for code coverage by ITs, please confirm.

If so, then this unit test can be removed.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/guard/GuardHandler.java
```diff
@@ -169,6 +171,27 @@ String attribute(
...
+     * allow all items must override and return {@code true}.
+     * </p>
+     *
+     * @param sessionId  the session identifier
+     * @param scopes     the required scopes
+     * @return {@code true} if the session holds every scope in the list
+     */
+    default boolean verify(
+        long sessionId,
+        List<String> scopes)
```
jfallows:
```suggestion
        List<String> roles)
```

Guard terminology is roles, whereas scopes is an jwt implementation detail.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/guard/GuardHandler.java
```diff
@@ -169,6 +171,27 @@ String attribute(
...
+    {
+        return false;
+    }
```
jfallows: Let's remove the default and implement directly on each class instead.

PR #1974, opened 2026-06-28
Add offline mode support to ZPM cache resolution

jfallows:
## Description

This change adds support for offline artifact resolution in the ZPM (Zilla Package Manager) cache by introducing an `--exclude-remote-repositories` flag that enables strict local-only dependency resolution.

### Changes

- **ZpmCache**: Added `offline` parameter to constructor and `newRepositorySystemSession()` method. When offline mode is enabled, the repository system is configured to:
  - Set offline mode to prevent any network transfers
  - Ignore artifact descriptor repositories to prevent transitive POMs from declaring additional repositories that would require network access
  
- **ZpmInstall**: Pass the `excludeRemoteRepos` flag to the ZpmCache constructor to enable offline resolution when the flag is set

- **ZpmInstallTest**: Added new test `shouldInstallEngineExcludingRemoteRepositories()` that verifies the install command works correctly with the `--exclude-remote-repositories` flag, ensuring artifacts are resolved from the local cache only

### Motivation

This enables offline builds and air-gapped environments where network access is restricted or unavailable. The offline mode ensures that dependency resolution fails fast if required artifacts are not in the local cache, rather than attempting network operations that would stall or fail.

### Test Plan

Added unit test `shouldInstallEngineExcludingRemoteRepositories()` that verifies the offline installation flow works correctly and produces the expected artifacts in the output directory.

https://claude.ai/code/session_01JKheckDByW3Qh5GZAREPQr

### DIFF
--- a/manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
+++ b/manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
@@ -77,12 +77,13 @@ public final class ZpmCache
 
     public ZpmCache(
         List<RemoteRepository> repositories,
+        boolean offline,
         Path directory,
         ConsoleLogger logger)
     {
         this.logger = logger;
         this.repositorySystem = ZpmSupplierRepositorySystemFactory.newRepositorySystem();
-        this.session = newRepositorySystemSession(repositorySystem, directory);
+        this.session = newRepositorySystemSession(repositorySystem, directory, offline);
 
         this.repositories = repositories;
     }
@@ -172,8 +173,13 @@ public final class ZpmCache
 
     private RepositorySystemSession newRepositorySystemSession(
         RepositorySystem system,
-        Path dir)
+        Path dir,
+        boolean offline)
     {
+        // When remote repositories are excluded, resolve strictly from the local cache.
+        // Offline mode prevents any network transfer, and ignoring artifact-descriptor
+        // repositories stops repositories declared in transitive POMs from being contacted
+        // (e.g. for SNAPSHOT metadata), which would otherwise stall a network-less build.
         return new SessionBuilderSupplier(system)
             .get()
                 .withLocalRepositoryBaseDirectories(dir)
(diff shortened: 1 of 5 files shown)


manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
```diff
@@ -172,15 +173,23 @@ public List<ZpmArtifact> resolve(
 
     private RepositorySystemSession newRepositorySystemSession(
         RepositorySystem system,
-        Path dir)
+        Path dir,
+        boolean excludeRemoteRepositories)
```
jfallows:
```suggestion
        boolean excludeRemote)
```

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
```diff
@@ -77,12 +77,13 @@ public final class ZpmCache
 
     public ZpmCache(
         List<RemoteRepository> repositories,
+        boolean excludeRemoteRepositories,
```
jfallows:
```suggestion
        boolean excludeRemote,
```

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
```diff
@@ -172,15 +173,23 @@ public List<ZpmArtifact> resolve(
...
-        Path dir)
+        Path dir,
+        boolean excludeRemoteRepositories)
     {
+        // When remote repositories are excluded, ignore repositories declared in
+        // transitive dependency POMs so resolution stays within the explicitly
+        // configured (local file) repositories. Otherwise a POM-declared remote host
+        // would be contacted (e.g. for SNAPSHOT metadata) and stall a network-less build.
+        // Note: offline mode cannot be used here, as the local Maven repository is itself
+        // configured as a file-based remote repository, which offline mode would block.
```
jfallows: Remove verbose comment.

jfallows (commented):
LGTM

PR #1980, opened 2026-07-01
ci(release): cut 1.x maintenance releases from support/1.x with major/minor docker tags

jfallows:
## Description

Enables git-flow `support/1.x` releases so 1.x minor/patch versions can be cut from `support/1.x` while `develop` moves on to v2.

- The release base is the branch the workflow is dispatched from (`github.ref_name`) — no manual `base` input, so the ref and base can't diverge.
- A fail-fast guard rejects a version that does not match the branch's line: `support/N.x` only accepts `N.*`, and a `develop` run is blocked when a `support/<major>.x` branch already exists (so a 1.x release can't be cut from develop by mistake).
- Docker images publish moving `major` (`1`) and `major.minor` (`1.5`) tags alongside the exact `major.minor.patch`. The floating `latest` / `alpine` pointers move only for the highest stable release across all tags, so `support/1.x` keeps `latest` until a stable 2.x ships from develop; pre-releases publish only their exact version tag.
- A `support/*` release merges into the support branch and tags only — it never advances `main` or back-merges into `develop`. Mainline (`develop`) releases keep the existing git-flow finish.

Targets `support/1.x`. Companion PR against `develop` (`claude/release-workflow-support-branches-ss82ns`) brings the same workflow to the mainline so v2 releases also emit the new `2` / `2.0` tags.

Follow-up (deferred): switch `examples/*/compose.yaml` defaults on `support/1.x` from `${ZILLA_VERSION:-latest}` to `${ZILLA_VERSION:-1}`.

No associated issue.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,9 +2,9 @@ name: build
 
 on:
   push:
-    branches: [ develop, 'feature/**'  ]
+    branches: [ develop, 'support/**', 'feature/**'  ]
   pull_request:
-    branches: [ develop, 'feature/**' ]
+    branches: [ develop, 'support/**', 'feature/**' ]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,9 +24,6 @@ jobs:
       with:
         require: admin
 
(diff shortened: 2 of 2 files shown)


jfallows (commented):
LGTM

PR #1983, opened 2026-07-01
ci(release): force Maven to IPv4 and skip tests on release deploy

jfallows:
## Description

Mainline (`develop`) companion covering two release-workflow fixes:

1. **Force Maven to IPv4** — set `MAVEN_OPTS: -Djava.net.preferIPv4Stack=true` at the workflow level. `maven.packages.aklivity.io` publishes an AAAA (IPv6) record (`2a12:5240::1`) whose endpoint isn't routable from GitHub-hosted runners, so dependency downloads failed with "Network is unreachable"; forcing IPv4 uses the reachable `89.106.200.1`.

2. **Skip tests on release deploy** — add `-DskipTests` to `Deploy via Maven`. Those tests already run in `build.yml` CI, so this makes releases faster and avoids the k3po IPv6 spec tests (which bind `[::1]` inside the Maven JVM) conflicting with the `preferIPv4Stack` workaround above.

Note: the real fix for (1) is infra-side (drop/repair the AAAA record if the IPv6 endpoint isn't served); this is a reliable runner-side workaround in the meantime.

No associated issue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,12 @@ jobs:
     steps:
     - name: Checkout GitHub sources
       uses: actions/checkout@v6
+    - name: Pin maven.packages.aklivity.io to IPv4
+      run: |
+        ipv4=$(getent ahostsv4 maven.packages.aklivity.io | awk 'NR==1{print $1}')
+        if [ -n "$ipv4" ]; then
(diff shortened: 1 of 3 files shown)


jfallows (commented):
LGTM

PR #1984, opened 2026-07-02
ci(release): skip tests on release deploy

jfallows:
## Description

Add `-DskipTests` to the release `Deploy via Maven` step so the release doesn't re-run the full test suite (those tests already run in `build.yml` CI). This makes releases faster and, importantly, avoids the k3po IPv6 spec tests (e.g. `binding-tcp.spec` `client.connect.with.ipv6.extension`, which binds `[::1]` inside the Maven JVM) conflicting with the `-Djava.net.preferIPv4Stack=true` workaround that the package-repo download requires.

Context: a `1.2.5-rc1` release got past dependency resolution (IPv4 fix working) but failed in `binding-tcp.spec` because `preferIPv4Stack` disables IPv6 for the in-JVM k3po driver. Skipping tests on release removes that conflict while keeping the IPv4 download fix.

Follow-up to the merged IPv4 workaround on `support/1.x`.

No associated issue.

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -201,7 +201,7 @@ jobs:
 
     - name: Deploy via Maven
       run: |
-        ./mvnw -B -U -nsu -ntp clean deploy -Drelease -Ddocker.latest.tag=${{ steps.tag.outputs.latest }}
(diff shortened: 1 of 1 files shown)


jfallows (commented):
LGTM

PR #1985, opened 2026-07-02
ci: pin maven.packages.aklivity.io to IPv4 instead of forcing the JVM stack

jfallows:
## Description

Replaces the `-Djava.net.preferIPv4Stack=true` workaround (merged via #1982) with a transient `/etc/hosts` entry pinning only `maven.packages.aklivity.io` to its IPv4 address.

`preferIPv4Stack` fixed the release deploy's unreachable AAAA record, but it disables the JVM's IPv6 stack entirely — which `build.yml` cannot tolerate, since it runs the full k3po IT suite including tests that bind IPv6 loopback (`[::1]`). Pinning just that one hostname's DNS resolution to IPv4 fixes the actual problem (the host's unreachable IPv6 endpoint) without touching the JVM's networking stack at all, so it's safe to apply uniformly.

Applied to `release.yml` (`prepare` + `deploy`, replacing the removed `MAVEN_OPTS`), `build.yml`, and `codeql.yml` (the "scan" workflow for this repo) — everywhere `mvnw` resolves dependencies from that host.

Note: `-DskipTests` on the release deploy (#1984) is kept — it's independently faster and no longer strictly required to avoid the IPv6 conflict, but there's no reason to re-add tests to the release path.

No associated issue.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,12 @@ jobs:
     steps:
     - name: Checkout GitHub sources
       uses: actions/checkout@v6
+    - name: Pin maven.packages.aklivity.io to IPv4
+      run: |
+        ipv4=$(getent ahostsv4 maven.packages.aklivity.io | awk 'NR==1{print $1}')
+        if [ -n "$ipv4" ]; then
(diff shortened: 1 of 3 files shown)


jfallows (commented):
LGTM

PR #1988, opened 2026-07-02
fix(support/1.x): reset to 1.x-SNAPSHOT and prevent it recurring

jfallows:
## Description

`support/1.x`'s `pom.xml` was left at the exact released version (`1.2.5-rc3`, then `1.2.5`) after each support-branch release finish, instead of a SNAPSHOT placeholder like `develop`'s permanent `develop-SNAPSHOT`. This isn't cosmetic: the root `pom.xml` has an enforcer rule (`enforce-snapshot`, active on any build without `-Drelease`) that **fails the build** if the version isn't a SNAPSHOT. So every ordinary push/PR to `support/1.x` has been broken since the `1.2.5-rc3` release finish — confirmed live via #1987 failing `Build (25)` with `Snapshot version only!`, completely unrelated to that PR's actual (examples-only) content.

### Two commits

1. **One-time fix**: reset `support/1.x`'s `pom.xml` (all 114 modules, via `mvnw versions:set`) from `1.2.5` to `1.x-SNAPSHOT`.

2. **Prevent recurrence**: `develop` never hits this because `prepare` already does a `git merge -s ours` no-merge of the release branch into `develop` before the release deploys — this permanently protects `develop`'s tree from ever receiving a version bump, since git sees the release branch as already merged by the time `git flow release finish` tries its own (real) merge later. `support/*` never had this protection, so `git flow release finish`'s real `--no-ff` merge into the support branch brought the version bump in every time, with nothing to reset it.

   - `prepare`: generalize the no-merge step to run for **any** base branch (not just `develop`), so `support/*` gets the same protection.
   - `finalize`: since the no-merge protection makes the base-branch merge in `git flow release finish` a no-op, its own tag would land on the *protected* (snapshot) commit rather than the real release content. Fix: tag `release/<version>`'s own tip directly — the exact commit `deploy` built and published — and pass `-n`/`--notag` so `git flow release finish` skips creating its own (now-redundant, wrong-commit) tag.

Net effect: any base branch — `develop` or a `support/*` line — stays at its own SNAPSHOT version forever, while every release tag stays byte-accurate to what was actually built and deployed, regardless of what happens to the base branch afterward.

Scoped to `support/1.x` only per discussion; happy to mirror the `prepare`/`finalize` generalization onto `develop`'s own copy of `release.yml` as a follow-up so future support branches inherit it too.

No associated issue.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,7 @@ jobs:
       if: success()
       with:
         path: ~/docker-cache/docker.tar
-        key: zilla-develop-SNAPSHOT-${{ github.run_id }}
+        key: zilla-1.x-SNAPSHOT-${{ github.run_id }}
 
     - name: Conditional Artifact Upload
       uses: actions/upload-artifact@v7
@@ -104,7 +104,7 @@ jobs:
       - setup-examples
     runs-on: ubuntu-latest
     env:
-      ZILLA_VERSION: develop-SNAPSHOT
+      ZILLA_VERSION: 1.x-SNAPSHOT
     steps:
       - name: Checkout
         uses: actions/checkout@v6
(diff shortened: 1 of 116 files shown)


jfallows (commented):
LGTM

PR #1994, opened 2026-07-02
feat(common-openapi): generic access to x-* specification extensions

jfallows:
## Summary

`common-openapi` (hoisted in #1978) should stay OpenAPI-generic — like `common-json`/`common-avro`, it should know the OpenAPI *structure* only, not any consumer's `x-*` extension semantics. Today it violates that: `x-zilla-http-kafka` is hardcoded as a `@JsonbProperty("x-zilla-http-kafka")` field on `OpenapiOperation`, and the `model.extensions.http.kafka` package is exported from the module.

This adds a generic, strongly-typed way to read *any* specification extension, without `common-openapi` ever referencing a specific extension name or type:

```java
boolean hasExtension(String name);
<T> Optional<T> extension(String name, Class<T> type);
```

Both `name` and `type` are caller-supplied — the caller gets back a real typed object (via JSON-B), not a raw `JsonValue` it has to unwrap itself. Available on `OpenapiView` (root), `OpenapiOperationView`, and `OpenapiSecuritySchemeView`.

## How it works

- `OpenapiParser` now retains the parsed `JsonObject` document on the `Openapi` model (`transient`, not part of the OpenAPI JSON-B contract — JSON-B silently drops unmapped/unannotated fields during deserialization, and this field is never populated from input JSON).
- Each view resolves its own backing `JsonObject` node from that document (root document itself for `OpenapiView`; `paths.<path>.<method>` for `OpenapiOperationView`; `components.securitySchemes.<name>` for `OpenapiSecuritySchemeView`).
- `extension(name, type)` binds the requested `x-*` value to `T` via JSON-B **lazily**, only when called — no eager binding cost for extensions nobody reads.
- Directly-constructed `Openapi` models with no `document` (as in existing unit tests) simply report no extensions — no behavior change for existing callers.

## Dogfooding: migrate `x-zilla-http-kafka`

To validate the new API against a real, previously-hardcoded extension, and so `common-openapi` ends up carrying zero extension-specific code:

- Moved `OpenapiHttpKafkaOperationExtension` / `OpenapiHttpKafkaFilter` out of `common-openapi` into `binding-openapi-asyncapi` (its only consumer).
- `OpenapiAsyncapiProxyGenerator` now reads it via `httpOperation.extension("x-zilla-http-kafka", OpenapiHttpKafkaOperationExtension.class)` instead of a dedicated `httpOperation.httpKafka` field / `hasExtensionHttpKafka()`.
- Removed the `model.extensions.http.kafka` package and its `exports`/`opens` from `common-openapi`'s `module-info.java`.

## Testing

- `OpenapiParserTest` (new) — end-to-end: parses a spec with `x-zilla-sample` on both the root document and an operation, asserts `hasExtension`/`extension` detect and bind correctly through the real parse path, and that an unknown extension name returns `Optional.empty()`.
- `OpenapiViewTest` (extended) — unit-level coverage for `OpenapiView`, `OpenapiOperationView`, and `OpenapiSecuritySchemeView` extension access, using directly-constructed models (with and without a backing document) to confirm graceful no-op behavior when no document is present.
- `binding-openapi` and `binding-openapi-asyncapi` full test suites (unit + k3po ITs) verified green against the updated views, confirming the `x-zilla-http-kafka` migration is a faithful behavioral port.
- No NOTICE changes required — no external dependency changed.

## Verification

- `./mvnw clean verify -pl runtime/common-openapi` — 10 tests, green
- `./mvnw clean verify -pl runtime/binding-openapi` — 11 tests, green
- `./mvnw clean verify -pl runtime/binding-openapi-asyncapi` — 8 tests, green
- `./mvnw checkstyle:check` clean on all touched modules

---
_Generated by [Claude Code](https://claude.ai/code/session_013Li1mTsigtRqijhbtm4Xp5)_


runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/view/OpenapiExtensions.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Optional;
+
+import jakarta.json.JsonObject;
+import jakarta.json.JsonValue;
+import jakarta.json.bind.Jsonb;
+import jakarta.json.bind.JsonbBuilder;
+
+import io.aklivity.zilla.runtime.common.yaml.json.YamlJson;
+
+final class OpenapiExtensions
```
jfallows: I feel like this class belongs in the model, providing access to the named extension in raw form as a JsonValue (not just JsonObject is possible).

Then `OpenapiExtensionsView` would need to resolve against the `OpenapiExtensions` model.

This should also support $ref resolution, similar to other model vs view APIs.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiOperation.java
```diff
@@ -29,7 +25,4 @@ public class OpenapiOperation
     public Map<String, OpenapiResponse> responses;
     public List<Map<String, List<String>>> security;
     public List<OpenapiServer> servers;
-
-    @JsonbProperty("x-zilla-http-kafka")
-    public OpenapiHttpKafkaOperationExtension httpKafka;
```
jfallows, line 34: Perhaps like this for the model layer?
```
public OpenapiExtensions extensions;
```

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/view/OpenapiExtensions.java
jfallows: Reworked in 4ff6bdac: extension capture now lives in the model (`OpenapiExtension` holds a raw `JsonValue value`), populated by `OpenapiOperationDeserializer`/`OpenapiSecuritySchemeDeserializer` as JSON-B visits each node during its own parse pass — no separate post-parse walk. `$ref` resolution happens in `OpenapiExtensionAdapter` via `JsonPointer` against the raw document, with cycle detection, so a reusable extension like `x-http-kafka` referenced via `$ref` resolves the same way the rest of the spec does. Deleted this `OpenapiExtensions` view helper; views call `OpenapiExtension.has`/`get` directly against the model's `Map<String, OpenapiExtension>`.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiOperation.java
jfallows: Landed as `public transient Map<String, OpenapiExtension> extensions;` (one `OpenapiExtension` per name, each holding a `JsonValue value`) rather than a single `OpenapiExtensions` wrapper — settled on that shape after discussing tradeoffs. Populated in a single parse pass by `OpenapiOperationDeserializer` in 4ff6bdac.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiExtension.java
```diff
@@ -0,0 +1,73 @@
...
+        T value = null;
+
+        if (extensions != null && extensions.containsKey(name))
+        {
+            Jsonb jsonb = JsonbBuilder.create();
+            value = jsonb.fromJson(extensions.get(name).value.toString(), type);
+        }
+
+        return Optional.ofNullable(value);
+    }
```
jfallows: These static methods feel like they belong on the view layer instead of the model layer APIs.
Perhaps we need to have an OpenapiExtensibleView super type for each of the extensible views, such as OpenapiOperationView, etc?

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiExtensionAdapter.java
```diff
@@ -0,0 +1,93 @@
...
+        return extension.value;
+    }
+
+    @Override
+    public OpenapiExtension adaptFromJson(
+        JsonValue json)
+    {
+        OpenapiExtension extension = new OpenapiExtension();
+        extension.value = resolve(json, new LinkedHashSet<>());
+        return extension;
```
jfallows: $ref resolution is typically done when the view layer is being created over the model layer.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiJsonValues.java
```diff
@@ -0,0 +1,105 @@
...
+import jakarta.json.JsonArrayBuilder;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonObjectBuilder;
+import jakarta.json.JsonValue;
+import jakarta.json.stream.JsonParser;
+import jakarta.json.stream.JsonParser.Event;
+
+// materializes a JsonObject using only the primitive JsonParser event API (next/getString/
+// getBigDecimal); the convenience JsonParser.getObject()/getValue() lose position tracking
+// when invoked from a JsonbDeserializer dispatched for a Map value, so they cannot be used here
```
jfallows: Why are extensions a special case? Shouldn't this follow the same approach as the rest of the model?

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiOperation.java
```diff
@@ -30,6 +26,6 @@ public class OpenapiOperation
     public List<Map<String, List<String>>> security;
     public List<OpenapiServer> servers;
 
-    @JsonbProperty("x-zilla-http-kafka")
-    public OpenapiHttpKafkaOperationExtension httpKafka;
+    // populated by OpenapiParser after the main parse; see Openapi.extensions
+    public transient Map<String, OpenapiExtension> extensions;
```
jfallows: Can this not be done as part of the main parse?

runtime/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/model/extensions/http/kafka/OpenapiHttpKafkaOperationExtension.java
jfallows:
```suggestion
public class OpenapiHttpKafkaOperationEx
```

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/config/OpenapiParser.java
```diff
@@ -74,9 +93,14 @@ public Openapi parse(
 
             Jsonb jsonb = JsonbBuilder.newBuilder()
                     .withProvider(provider)
+                    .withConfig(new JsonbConfig()
+                        .withDeserializers(
+                            new OpenapiOperationDeserializer(extensionTypes),
+                            new OpenapiSecuritySchemeDeserializer(extensionTypes)))
```
jfallows: We need to support extensions on each of the supported extension point parts of the OpenAPI specification.

Check the spec to see where x-* names are supported.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/config/OpenapiParser.java
```diff
@@ -74,9 +93,14 @@ public Openapi parse(
...
                     .build();
 
             openapi = jsonb.fromJson(openapiText, Openapi.class);
+            openapi.extensions = OpenapiExtension.capture(document, extensionTypes);
```
jfallows: Not sure we still have this call to .capture(...), seems like it is not the same approach as the deserializers.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/Openapi.java
```diff
@@ -24,4 +24,12 @@ public class Openapi
...
     public List<Map<String, List<String>>> security;
     public OpenapiComponents components;
+
+    // populated by OpenapiParser right after the main parse, not via JSON-B field binding,
+    // since x-* extensions are siblings of the known properties, not nested under
+    // a JSON property named "extensions"; captured as-is with no $ref resolution, since
+    // OpenAPI defines no $ref semantics for specification extension values (only for
+    // specific schema/component positions), and cross-document reuse is already handled
+    // by YAML anchors/aliases, resolved transparently before parsing reaches here
+    public transient Map<String, OpenapiExtension> extensions;
```
jfallows: Can we not use a deserializer for the top level Openapi model object?
Perhaps we can use an adapter with a shadow DTO plus dynamic fields?

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiExtension.java
```diff
@@ -0,0 +1,66 @@
...
+import jakarta.json.JsonObject;
+import jakarta.json.JsonValue;
+import jakarta.json.bind.Jsonb;
+import jakarta.json.bind.JsonbBuilder;
+
+public class OpenapiExtension
+{
+    private static final Jsonb JSONB = JsonbBuilder.create();
+
+    public JsonValue value;
```
jfallows: Do we still need this OpenapiExtension class now that we're using dynamic deserializers?

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiExtension.java
```diff
@@ -0,0 +1,66 @@
...
+                {
+                    extension.bound = JSONB.fromJson(extension.value.toString(), type);
+                }
+
+                extensions.put(name, extension);
+            }
+        }
+
+        return extensions;
+    }
```
jfallows: This feels incorrect, we should be able to use Jsonb deserialization directly without this capture method.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiOperation.java
```diff
@@ -30,6 +26,7 @@ public class OpenapiOperation
     public List<Map<String, List<String>>> security;
     public List<OpenapiServer> servers;
 
-    @JsonbProperty("x-zilla-http-kafka")
-    public OpenapiHttpKafkaOperationExtension httpKafka;
+    // populated during the main parse by OpenapiOperationAdapter, since x-* extensions
+    // are siblings of the known properties, not nested under a property named "extensions"
+    public transient Map<String, OpenapiExtension> extensions;
```
jfallows: If we use the deserializers approach, then this should be populated in the same way as the other fields, not special, so no comment needed and also not transient.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiOperationDeserializer.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public OpenapiOperation deserialize(
+        JsonParser parser,
+        DeserializationContext ctx,
+        Type type)
+    {
+        JsonObject object = parser.getObject();
+        OpenapiOperation operation = PLAIN.fromJson(object.toString(), OpenapiOperation.class);
+        operation.extensions = OpenapiExtension.capture(object, extensionTypes);
```
jfallows: Would expect no to need the capture method.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiSecurityScheme.java
```diff
@@ -23,4 +25,8 @@ public class OpenapiSecurityScheme extends AbstractOpenapiResolvable
     public String bearerFormat;
     public String openidConnectUrl;
     public Object flows;
+
+    // populated during the main parse by OpenapiSecuritySchemeAdapter, since x-* extensions
+    // are siblings of the known properties, not nested under a property named "extensions"
+    public transient Map<String, OpenapiExtension> extensions;
```
jfallows: Same feedback as operation.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiSecuritySchemeDeserializer.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public OpenapiSecurityScheme deserialize(
+        JsonParser parser,
+        DeserializationContext ctx,
+        Type type)
+    {
+        JsonObject object = parser.getObject();
+        OpenapiSecurityScheme scheme = PLAIN.fromJson(object.toString(), OpenapiSecurityScheme.class);
+        scheme.extensions = OpenapiExtension.capture(object, extensionTypes);
```
jfallows: Same feedback, didn't expect capture method needed.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/view/OpenapiExtensibleView.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Optional;
+
+import jakarta.json.bind.Jsonb;
+import jakarta.json.bind.JsonbBuilder;
+
+import io.aklivity.zilla.runtime.common.openapi.model.OpenapiExtension;
+
+abstract class OpenapiExtensibleView
+{
+    private static final Jsonb JSONB = JsonbBuilder.create();
```
jfallows: This static jsonb doesn't seem right, it should already be parsed correctly at model layer.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiExtensionAdapter.java
jfallows: Dropped `$ref` resolution for extensions entirely — verified against the OpenAPI spec that `$ref`/Reference Object isn't defined for specification extension values, only for `components/*`, Path Item, Operation `parameters`/`requestBody`/`callbacks`, and Schema Objects. Cross-document reuse for extensions is already handled by YAML anchors/aliases, resolved before parsing reaches here. `OpenapiExtensionAdapter` and the whole `$ref`-for-extensions concept are gone.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiJsonValues.java
jfallows: Agreed — deleted `OpenapiJsonValues` after the common-yaml cursor-advance fix (#1998) landed; every deserializer now calls `parser.getObject()` directly like the rest of the model, no hand-rolled event materialization.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiOperation.java
jfallows: Yes — `OpenapiOperationDeserializer` populates `extensions` inline during its own parse pass, no separate post-parse walk.

---

runtime/binding-openapi-asyncapi/src/main/java/io/aklivity/zilla/runtime/binding/openapi/asyncapi/internal/model/extensions/http/kafka/OpenapiHttpKafkaOperationExtension.java
jfallows: Done in 97b0e936.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/config/OpenapiParser.java
jfallows: Broadened to every OpenAPI object type the spec allows `x-*` on: Server, ServerVariable, Components, Path, Parameter, RequestBody, MediaType, Encoding, Schema, Response, Header, Link, plus the root Openapi — 15 total, each with its own `JsonbDeserializer`.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/config/OpenapiParser.java
jfallows: Removed — the root `Openapi` type now goes through its own `OpenapiDeserializer` like every other extensible type, no post-parse capture step.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/Openapi.java
jfallows: Added `OpenapiDeserializer` for the root type — confirmed empirically that JSON-B honors a custom deserializer registered for the exact type requested by `Jsonb.fromJson(text, T.class)`. Kept the same per-type-deserializer shape as the rest of the model rather than a shadow DTO, since it lets every extensible type materialize consistently: each deserializer builds its own fields via a Jsonb registered with every sibling deserializer except itself, so nested extensible types (e.g. an Operation inside a Path, or a Schema inside a Schema) still get their own extensions captured without infinite recursion. `OpenapiSchema`, the one genuinely self-referential type, additionally walks its own nested schema fields by hand to apply that same scan recursively.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiExtension.java
jfallows: No — deleted.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiExtension.java
jfallows: Agreed, removed. Each deserializer now binds its own type via `Jsonb.fromJson` directly and inlines the `x-*` scan, no shared capture helper.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiOperation.java
jfallows: Done — `extensions` is now a plain `Map<String, Object>` field with no `transient` and no comment, populated the same way as every other field.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiOperationDeserializer.java
jfallows: Removed.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiSecurityScheme.java
jfallows: Same fix applied to `OpenapiSecurityScheme`.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/model/OpenapiSecuritySchemeDeserializer.java
jfallows: Removed.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/view/OpenapiExtensibleView.java
jfallows: Removed — `OpenapiExtensibleView` is now a pure lookup + cast against the already-bound `Map<String, Object>`, no Jsonb at the view layer. One consequence: an extension name never registered via `OpenapiParserFactory.withExtension(...)` is now invisible rather than lazily bound on first access — that's an intentional trade-off (covered by a new regression test), flag if it needs revisiting.

---

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/view/OpenapiHeaderView.java
```diff
@@ -29,6 +29,8 @@ public final class OpenapiHeaderView
         String name,
         OpenapiHeader model)
     {
+        super(resolver.headers.resolve(model).extensions);
+
         this.name = name;
 
         OpenapiHeader resolved = resolver.headers.resolve(model);
```
jfallows, line 37: This approach requires duplicate call to resolver.headers.resolve(model).
Perhaps eliminate the inheritance and just inline the extensions related methods.

runtime/common-openapi/src/main/moditect/module-info.java
```diff
@@ -22,11 +22,9 @@
 
     exports io.aklivity.zilla.runtime.common.openapi.config;
     exports io.aklivity.zilla.runtime.common.openapi.model;
-    exports io.aklivity.zilla.runtime.common.openapi.model.extensions.http.kafka;
     exports io.aklivity.zilla.runtime.common.openapi.model.resolver;
     exports io.aklivity.zilla.runtime.common.openapi.view;
 
     opens io.aklivity.zilla.runtime.common.openapi.model;
-    opens io.aklivity.zilla.runtime.common.openapi.model.extensions.http.kafka;
     opens io.aklivity.zilla.runtime.common.openapi.view;
```
jfallows: Do we need this one?
I understand we still need opens for the model to support jsonb reflection based parsing.

runtime/common-openapi/src/test/java/io/aklivity/zilla/runtime/common/openapi/config/OpenapiParserTest.java
```diff
@@ -0,0 +1,198 @@
...
+        "    }," +
+        "    \"securitySchemes\": {" +
+        "      \"bearerAuth\": {" +
+        "        \"type\": \"http\"," +
+        "        \"scheme\": \"bearer\"," +
+        "        \"x-zilla-sample\": { \"key\": \"scheme-value\" }" +
+        "      }" +
+        "    }" +
+        "  }" +
+        "}";
```
jfallows: Use multi-line strings `"""` for this instead.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/view/OpenapiHeaderView.java
jfallows: Eliminated the inheritance in 3446ffda — `OpenapiExtensibleView` is gone entirely. Every extensible view now has its own private `Map<String, Object> extensions` field with inlined `hasExtension`/`extension` methods. This also fixed the duplicate-resolve bug: `OpenapiHeaderView` (and the same pattern in `OpenapiSecuritySchemeView`, `OpenapiLinkView`, `OpenapiRequestBodyView`, `OpenapiParameterView`) now calls `resolver.headers.resolve(model)` exactly once and reuses the result.

---

runtime/common-openapi/src/main/moditect/module-info.java
jfallows: Removed in 3446ffda. Confirmed it isn't needed: the `view` package classes are plain POJOs, always constructed via explicit `new` calls, never via JSON-B reflection — only `model` needs `opens` for reflective field binding. Verified empirically (full build + tests + checkstyle + license all green with it removed).

---

runtime/common-openapi/src/test/java/io/aklivity/zilla/runtime/common/openapi/config/OpenapiParserTest.java
jfallows: Done in 3446ffda — `SPEC` is now a text block.

---

runtime/common-openapi/src/main/moditect/module-info.java
```diff
@@ -22,11 +22,8 @@
 
     exports io.aklivity.zilla.runtime.common.openapi.config;
     exports io.aklivity.zilla.runtime.common.openapi.model;
-    exports io.aklivity.zilla.runtime.common.openapi.model.extensions.http.kafka;
     exports io.aklivity.zilla.runtime.common.openapi.model.resolver;
     exports io.aklivity.zilla.runtime.common.openapi.view;
```
jfallows, line 25: Double check how much we actually need to export now, possibly just the view package?
If so then let's move the other packages to internal, such as model to internal.model, etc.

jfallows: Checked actual cross-module usage in `binding-openapi`/`binding-openapi-asyncapi`: they import `config.OpenapiParser`/`OpenapiParserFactory`/`OpenapiCatalogConfig`/`OpenapiSchemaConfig`/`OpenapiServerConfig(Builder)`/`OpenapiSpecificationConfig`, `model.Openapi` (parse() return type, and the input to `OpenapiView.of(...)`), and a dozen `view.*` types. So `config` and `model` are genuine public API, not internal — `model.Openapi` in particular has to stay exported since it's the type consumers hold between `parser.parse(...)` and `OpenapiView.of(...)`.

The only package with zero external references was `model.resolver` (only used internally by the view layer's own resolution) — removed that export in 4e1b6e3d. `config`/`model`/`view` stay exported, matching the already-merged sibling `common-asyncapi` module's shape (which exports `config`, `model` (+ sub-packages), `model.resolver`, and `view` too). Happy to revisit if you want `common-openapi` to diverge from that precedent, but wanted to flag the inconsistency rather than rename silently.

---

jfallows (commented):
LGTM

PR #1995, opened 2026-07-02
feat(manager): support enabling incubator modules via --incubator

jfallows:
## Description

`zpm install` (`ZpmInstall`) already links a fixed extra set of modules into the generated runtime image via `--debug` (`jdk.jdwp.agent`) and `--instrument` (`java.instrument`).

Some plugins need to link an optional JDK incubator module into the generated runtime image. Rather than hardcoding specific incubator module names into the shared install code, this adds a repeatable `--incubator <name>` option, following the same pattern as `--debug`/`--instrument`:

- `--incubator <name>` is repeatable and maps each name to `jdk.incubator.<name>`.
- The resulting module names are added to `linkModules()`'s existing `extraModuleNames` list, so `jlink` includes them in the generated runtime image.
- The option defaults to an empty list, so when it isn't passed, the generated image and launcher are unchanged from today.
- `generateLauncher()` is untouched: a jlink'd image's own module resolution (`ServiceLoader` `uses`/`provides` binding, plus transitive `requires`) already picks up a linked incubator module automatically at launch, so no extra `--add-modules` needs to be baked into the launcher script.

A new test (`shouldLinkIncubatorModuleIntoImage`) exercises `--incubator vector` end-to-end through the real `zilla install` CLI (`jdk.incubator.vector` is currently the only incubator module shipped in the JDK, so it's used to validate the mechanism actually links) and asserts the linked image's `--list-modules` output contains `jdk.incubator.vector`.

## Test plan

- [x] New unit test added covering the CLI option and the linked image's module list, run locally end-to-end (confirmed `jlink` links `jdk.incubator.vector` into the image and the image resolves it at launch with no extra runtime flags)
- [x] Existing `ZpmInstallTest` tests unaffected (option defaults to empty, no behavior change when unset)
- [ ] Full CI build (`./mvnw install`)

Fixes # N/A — no linked issue.

### DIFF
--- a/manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
+++ b/manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
@@ -123,6 +123,11 @@ public final class ZpmInstall extends ZpmCommand
         hidden = true)
     public Boolean instrument = false;
 
+    @Option(name = {"--incubator"},
+        description = "Enable an incubator module (repeatable), e.g. --incubator vector",
+        hidden = true)
+    public List<String> incubator = new ArrayList<>();
+
     @Option(name = {"--exclude-local-repository"},
         description = "Exclude the local Maven repository")
     public boolean excludeLocalRepo;
@@ -732,6 +737,13 @@ public final class ZpmInstall extends ZpmCommand
         }
     }
 
+    private List<String> incubatorModuleNames()
+    {
+        return incubator.stream()
+            .map(name -> String.format("jdk.incubator.%s", name))
+            .collect(toList());
+    }
+
     private void linkModules(
         Collection<ZpmModule> modules) throws IOException
     {
@@ -748,6 +760,7 @@ public final class ZpmInstall extends ZpmCommand
         {
             extraModuleNames.add("java.instrument");
         }
+        extraModuleNames.addAll(incubatorModuleNames());
 
         extraModuleNames.add("java.management");
         extraModuleNames.add("jdk.management");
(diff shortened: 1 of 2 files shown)


manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
```diff
@@ -732,6 +737,13 @@ private void generateDelegating(
         }
     }
 
+    private List<String> incubatorModuleNames()
```
jfallows:
```suggestion
    private List<String> incubatorModules()
```

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
```diff
@@ -793,12 +805,23 @@ private void generateLauncher() throws IOException
...
-                    "-m io.aklivity.zilla.runtime.command/io.aklivity.zilla.runtime.command.internal.ZillaMain \"$@\"")),
-                imageDir)));
+                "--sun-misc-unsafe-memory-access=%s\"".formatted(unsafeMemoryAccess)));
+
+        List<String> incubatorModuleNames = incubatorModuleNames();
+        if (!incubatorModuleNames.isEmpty())
+        {
+            lines.add(String.format("JAVA_OPTIONS=\"$JAVA_OPTIONS --add-modules %s\"",
+                String.join(",", incubatorModuleNames)));
+        }
```
jfallows: Is this actually needed after we have already specified the modules via jlink?

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
jfallows: Done in e590af6c.

---

manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/ZpmInstall.java
jfallows: Good catch — no, it isn't needed. Verified with a standalone multi-module jlink spike: a jlink'd image launched via `-m root.module/Main` with no `--add-modules` at the `java` command line still resolves a separate module via `ServiceLoader` `uses`/`provides` binding, and then transitively resolves that provider module's own `requires` (including an incubator module) automatically. Since `linkModules()` already puts `--incubator` modules into the image via `extraModuleNames`, that's sufficient — the module graph resolution at launch handles the rest.

Reverted `generateLauncher()` to its original form in e590af6c and retargeted the test to assert against the linked image's module list instead of launcher script content.

---

jfallows (commented):
LGTM

PR #2015, opened 2026-07-04
Fix UnsafeBufferEx.asNative() to preserve offset for sub-range wraps

jfallows:
## Description

Fixes an issue where `UnsafeBufferEx.asNative()` did not correctly preserve the offset when called on a buffer created with a sub-range wrap (i.e., `new UnsafeBufferEx(byteBuffer, offset, length)`).

### Changes

**UnsafeBufferEx.java**: Modified `asNative()` to slice the segment to the sub-range before creating the Native wrapper. This ensures that the Native buffer's address offset correctly reflects the original sub-range offset, rather than always starting from the beginning of the underlying ByteBuffer.

The fix accounts for the fact that `Native.wrapAdjustment()` always reports 0, so the segment passed to Native must already be adjusted to the correct sub-range boundaries.

**UnsafeBufferExNativeTest.java**: Added `asNativeOnSubRangeWrapPreservesOffset()` test to verify that:
- A Native wrapper created from a sub-range buffer has the correct capacity
- The address offset is preserved from the original sub-range
- Data written through the Native wrapper is correctly positioned in both the sub-range and whole buffers

### Test Plan

The new unit test `asNativeOnSubRangeWrapPreservesOffset()` validates the fix by creating a sub-range wrap, converting it to Native, writing data, and verifying correct positioning in both the sub-range and whole buffer views.

https://claude.ai/code/session_012aTZ6zyEtnxCRPUfWdbhsa


jfallows (commented):
LGTM

PR #2026, opened 2026-07-05
fix(common-openapi): resolve full servers[] precedence in OpenapiOperationView

### LINKED ISSUE #2020: common-openapi: resolve full servers[] precedence (operation > path-item > root) in OpenapiOperationView
## Summary

`OpenapiOperationView.servers` currently reflects only an operation's own `servers[]` array, if declared. Extend it to resolve the full OpenAPI server precedence so callers always get a correct, ready-to-use list without reimplementing the precedence rule themselves.

## Precedence to implement

Per the OpenAPI specification, `servers[]` can be declared at three levels, with the most specific taking precedence:

1. Operation-level `servers[]` — used if declared, otherwise fall through
2. Path-item-level `servers[]` — used if declared, otherwise fall through
3. Root/document-level `servers[]` — the final fallback

`OpenapiOperationView.servers` should always resolve to the effective server list per this precedence, so any caller reading `operation.servers` gets the correct answer regardless of which level actually declared it.

## Server variable resolution

Server URL templates may declare `{variable}` placeholders with defaults and/or enums. When resolving a server URL, if a caller-supplied override value doesn't match the variable's allowed pattern, the current behavior silently substitutes the variable's declared default. This silent-fallback-on-mismatch should be made explicit — either reject the mismatched value at resolution time and surface that as an error, or otherwise make the fallback behavior an observable, intentional decision rather than a silent substitution.

## Note

This is expected to be a building block for other in-progress work that currently duplicates parts of this precedence logic in a downstream binding — that downstream simplification is out of scope for this issue.

jfallows:
## Description

`OpenapiOperationView.servers` previously only reflected an operation's own `servers[]` array, falling straight back to the root/document-level `servers[]` when absent — it never looked at path-item-level `servers[]`. This change wires the path-item's `servers[]` through `OpenapiPathView` into `OpenapiOperationView`, so `operation.servers` now resolves the full OpenAPI precedence: operation-level > path-item-level > root-level.

While implementing this, `OpenapiOperationView` (and, by extension, the new path-item level) also now mirrors `OpenapiView`'s handling of an empty `configs` list — previously, an operation or path declaring its own `servers[]` combined with an empty server-config list would silently resolve to zero servers instead of falling back to using the model's own server URL directly.

The issue also raised the silent-fallback-on-mismatch behavior in `OpenapiServerView.VariableMatcher` (a caller-supplied server URL override that doesn't match the variable's allowed pattern silently substitutes the declared default). After investigation, this PR intentionally leaves that fallback as-is: `binding-mcp-openapi` already treats a configured server URL as an unconditional override elsewhere (its `resolveServerOverride` bypasses `common-openapi`'s matcher entirely), and `OpenapiOperationView.servers` cross-products every model server against every configured override, so a non-matching pairing is the expected outcome whenever more than one server or config is declared — not a misconfiguration to reject.

This is a building block for downstream work (e.g. `binding-openapi-asyncapi`) that currently duplicates parts of the servers[] precedence logic; simplifying that downstream duplication is left for a follow-up change.

## Testing

Added unit tests in `OpenapiViewTest` covering:
- Operation-level `servers[]` taking precedence over path-item and root
- Path-item-level `servers[]` taking precedence over root when the operation declares none
- Root-level `servers[]` fallback when neither operation nor path-item declare any
- Server URL variable resolution using the declared default when no override is supplied
- Server URL variable resolution using a matching caller-supplied override
- Server URL variable resolution falling back to the default when a caller-supplied override doesn't match the variable's allowed pattern

`./mvnw test -pl runtime/common-openapi` passes, along with the existing test suites for `binding-openapi`, `binding-openapi-asyncapi`, and `binding-mcp-openapi`.

Fixes #2020

### DIFF
--- a/runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/view/OpenapiOperationView.java
+++ b/runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/view/OpenapiOperationView.java
@@ -20,9 +20,11 @@ import static java.util.stream.Collectors.toMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import io.aklivity.zilla.runtime.common.openapi.config.OpenapiServerConfig;
 import io.aklivity.zilla.runtime.common.openapi.model.OpenapiOperation;
+import io.aklivity.zilla.runtime.common.openapi.model.OpenapiServer;
 import io.aklivity.zilla.runtime.common.openapi.model.resolver.OpenapiResolver;
 
 public final class OpenapiOperationView
@@ -50,6 +52,7 @@ public final class OpenapiOperationView
         OpenapiResolver resolver,
         String method,
         String path,
+        List<OpenapiServer> pathServers,
         OpenapiOperation model)
     {
         this.specification = specification;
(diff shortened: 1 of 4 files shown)


jfallows (commented):
LGTM

PR #2027, opened 2026-07-05
Optimize MemorySegment.copy calls by avoiding wrapper allocation

jfallows:
## Description

Optimize `getBytes()` methods in `UnsafeBufferEx` by using the direct array overload of `MemorySegment.copy()` instead of wrapping the destination array in a `MemorySegment.ofArray()` wrapper.

### Changes

- Updated 4 `getBytes()` call sites (2 in the outer class, 2 in an inner class) to pass the byte array directly to `MemorySegment.copy()` instead of wrapping it first
- Removed unnecessary `MemorySegment.ofArray(dst)` wrapper construction and corresponding `BYTE_LAYOUT` parameter
- Added explanatory comments documenting the performance rationale

### Motivation

Profiling identified that per-call `MemorySegment.ofArray()` wrapper construction was a dominant cost for this method when processing small documents with many verbatim key/value deliveries. The wrapper creates a fresh heap-session-scoped `MemorySegment` with no cacheable form for arbitrary caller-supplied arrays, making it wasteful for frequent small operations.

The direct array overload of `MemorySegment.copy()` copies directly into the destination array without constructing an intermediate wrapper, eliminating this allocation overhead.

### Test Plan

Existing tests cover these code paths. No new tests required for this optimization.

https://claude.ai/code/session_01FdmT7MqTozanMnH2rLiE8F

### DIFF
--- a/runtime/common-agrona/src/main/java/io/aklivity/zilla/runtime/common/agrona/buffer/UnsafeBufferEx.java
+++ b/runtime/common-agrona/src/main/java/io/aklivity/zilla/runtime/common/agrona/buffer/UnsafeBufferEx.java
@@ -1032,11 +1032,7 @@ public final class UnsafeBufferEx implements AtomicBufferEx, DirectBufferViewEx
         int offset,
         int length)
     {
-        // the MemorySegment-to-array overload copies directly into dst without ever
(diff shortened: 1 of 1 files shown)


jfallows (commented):
LGTM

PR #2029, opened 2026-07-06
Support bulk operation selection by tag and glob pattern

### LINKED ISSUE #2021: mcp_openapi: bulk/tag/pattern route selection, when.capability, deterministic tool naming
## Summary

`mcp_openapi` routes currently require every route to name one explicit `spec` + `operation` pair — there is no way to select multiple operations from a single route. Add bulk/output-side selection forms, an orthogonal inbound-primitive-kind filter, and deterministic default naming for tools generated without an explicit name.

## Bulk route selection forms

Add three output-side selectors to a route&#39;s `with` block, each usable without a `when` clause (a route with `with` and no `when` is a defined bulk-selection form, not an error):

- **By OpenAPI tag** — `with: { spec: , tag:  }` selects every operation in the spec whose `tags` array contains ``.
- **By operationId pattern** — `with: { spec: , operation:  }` selects every operation whose `operationId` matches the glob pattern.
- **Select-all** — `with: { spec:  }` with neither `tag` nor `operation` selects every operation in the spec.

Each remaining unmatched operation becomes a generated tool (bulk expansion produces tools only, never resources — a resource requires an operator-authored URI, which is an explicit, not bulk, decision).

## Route precedence

Adopt first-match-wins, ordered top-to-bottom evaluation, consistent with how route matching already works elsewhere: an operation claimed by an earlier route is removed from the candidate set of every later route. This means an explicit single-operation route placed before a bulk route effectively &#34;claims&#34; that operation and overrides what the bulk route would otherwise have produced (rename, output projection, or promotion to a resource in a separate explicit route). No separate specificity-ranking logic is needed — it falls out of ordered evaluation.

A route using a bulk selector (`tag`, an `operation` pattern, or select-all) must not also declare `when` — validate this at config load.

## `when.capability` filter

Add an inbound-side `capability` condition (e.g. `when: { capability: [tool] }`) to a route, orthogonal to the output-side selection axis above. This constrains which kind of MCP primitive a route (particularly a bulk route) is allowed to emit, independent of which operations it selects.

## Deterministic tool naming

- Default tool name for an operation selected without an explicit name = its `operationId`, converted to MCP-safe snake_case (e.g. `listWidgets` → `list_widgets`; path/slash characters become underscores).
- Two operations that normalize to the same generated name, or an operation with no `operationId` at all, must fall back deterministically: first to a method+path-derived slug, then to a numeric suffix if still colliding. The result must be stable across repeated runs against the same spec.

## Routes are required

`routes` is a required, non-empty field in the `mcp_openapi` config schema (enforced via `required` + `minItems: 1`). A binding can therefore never load with zero routes — config validation rejects a missing or empty `routes` list at load time with a normal schema error, the same way any other missing required property would be rejected. Since this makes the empty-routes state unreachable, no separate load-time warning event is needed for it.

jfallows:
Fixes #2021

## Description

This PR extends the MCP OpenAPI binding to support bulk selection of operations from OpenAPI specifications using tags and glob patterns, in addition to explicit operation IDs.

### Key Changes

**Configuration &amp; Schema Updates:**
- Added `tag` field to `McpOpenapiWithConfig` to select operations by OpenAPI tag
- Extended `operation` field to support glob patterns (e.g., `pulls/*`) for wildcard matching
- Added `capability` field to `McpOpenapiConditionConfig` to filter routes by capability type (`tool` or `resource`)
- Updated JSON schema to enforce constraints: `tag` and `operation` are mutually exclusive, and a route using a bulk selector (`tag`, a glob `operation` pattern, or spec-only) cannot also declare `when.tool`/`when.resource`
- `routes` is now required and non-empty in the schema, so a binding with no routes fails config validation at load time instead of loading silently with nothing exposed

**Core Generator Logic:**
- Refactored `McpOpenapiCompositeGenerator` to support three bulk selection modes:
  - **Spec-only routes**: Select all operations from a spec
  - **Tag-based routes**: Select operations matching a specific OpenAPI tag
  - **Glob pattern routes**: Select operations matching a wildcard pattern (e.g., `pulls/*`)
- Implemented operation claiming mechanism to prevent duplicate tool generation when multiple routes match the same operation
- Earlier explicit routes now take precedence over later bulk routes for the same operation

**Tool Naming:**
- Created `McpOpenapiToolNamer` utility to generate tool names from operation IDs
- Converts operation IDs to snake_case (e.g., `repos/get` → `repos_get`)
- Falls back to a method+path slug on name collision (e.g., `get_widgets_b`)
- Handles remaining name conflicts with a numeric suffix

**Config Object Conventions:**
- `McpOpenapiWithConfig`, `McpOpenapiConditionConfig`, `McpOpenapiSpecificationConfig`, `McpOpenapiCatalogConfig`, and `McpOpenapiToolConfig` now follow the same package-private all-args constructor + builder convention already used by `McpOpenapiOptionsConfig`
- `McpOpenapiOptionsConfigBuilder.spec()`/`.tool()` and `McpOpenapiSpecificationConfigBuilder.catalog()` are fluent nested builders (mapper-based), matching the engine's `NamespaceConfigBuilder.catalog()` pattern, so callers chain `.spec().catalog()...build().build()` instead of pre-building objects
- `McpOpenapiOptionsConfigAdapter` updated to build config objects through these fluent builders

**Test Coverage:**
- Added comprehensive test cases covering:
  - Default snake_case tool naming from operation IDs
  - Spec-only route expansion (all operations)
  - Tag-based operation filtering
  - Glob pattern matching with wildcards
  - Operation claiming (earlier explicit routes take precedence)
  - Capability-based filtering
  - Name collision fallback to method-path slug
  - Schema validation of invalid configurations, including missing/empty `routes`
- Test style: each test method builds its own `BindingConfig`/`RouteConfig`/options fixtures inline via the fluent builders (including the engine's `BindingConfigBuilder.route()`) rather than through shared setup helpers, so each test reads linearly without jumping elsewhere; JSON fixtures use text blocks

### Validation &amp; Constraints

- Bulk routes (with `tag`, glob `operation`, or spec-only) cannot declare `when.tool` or `when.resource` — enforced by the JSON schema
- `tag` and `operation` fields are mutually exclusive in route configuration
- `capability` values must be either `tool` or `resource`
- Operations are claimed in route order; once claimed, they cannot be used by later routes
- `routes` must be present and non-empty

### Test Plan

All changes are covered by new unit tests in `McpOpenapiCompositeGeneratorTest` and schema validation tests in `SchemaTest`. The test suite validates:
- Correct tool name generation and collision handling
- Proper operation selection by tag and glob pattern
- Operation claiming and precedence rules
- Capability-based filtering
- Configuration validation and error cases, including missing/empty `routes`

https://claude.ai/code/session_01VMbC4ary5AZ6npJXfb3UnW

### DIFF
--- a/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiConditionConfig.java
+++ b/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiConditionConfig.java
@@ -14,18 +14,30 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.openapi.config;
 
+import java.util.List;
+
 import io.aklivity.zilla.runtime.engine.config.ConditionConfig;
 
 public final class McpOpenapiConditionConfig extends ConditionConfig
 {
     public final String tool;
     public final String resource;
+    public final List<String> capability;
 
     public McpOpenapiConditionConfig(
         String tool,
         String resource)
     {
+        this(tool, resource, null);
+    }
+
+    public McpOpenapiConditionConfig(
+        String tool,
+        String resource,
+        List<String> capability)
+    {
         this.tool = tool;
         this.resource = resource;
+        this.capability = capability;
     }
 }
--- a/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiConditionConfig.java
+++ b/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiConditionConfig.java
@@ -15,6 +15,7 @@
 package io.aklivity.zilla.runtime.binding.mcp.openapi.config;
 
 import java.util.List;
+import java.util.function.Function;
 
 import io.aklivity.zilla.runtime.engine.config.ConditionConfig;
 
@@ -24,14 +25,18 @@ public final class McpOpenapiConditionConfig extends ConditionConfig
     public final String resource;
     public final List<String> capability;
 
-    public McpOpenapiConditionConfig(
-        String tool,
-        String resource)
+    public static McpOpenapiConditionConfigBuilder<McpOpenapiConditionConfig> builder()
     {
-        this(tool, resource, null);
+        return new McpOpenapiConditionConfigBuilder<>(McpOpenapiConditionConfig.class::cast);
     }
 
-    public McpOpenapiConditionConfig(
+    public static <T> McpOpenapiConditionConfigBuilder<T> builder(
+        Function<ConditionConfig, T> mapper)
+    {
+        return new McpOpenapiConditionConfigBuilder<>(mapper);
+    }
+
+    McpOpenapiConditionConfig(
         String tool,
         String resource,
         List<String> capability)
--- a/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiWithConfig.java
+++ b/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiWithConfig.java
@@ -20,12 +20,22 @@ public final class McpOpenapiWithConfig extends WithConfig
 {
     public final String spec;
     public final String operation;
+    public final String tag;
 
     public McpOpenapiWithConfig(
         String spec,
         String operation)
     {
+        this(spec, operation, null);
+    }
+
+    public McpOpenapiWithConfig(
+        String spec,
+        String operation,
+        String tag)
+    {
         this.spec = spec;
         this.operation = operation;
+        this.tag = tag;
     }
 }
--- a/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiWithConfig.java
+++ b/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiWithConfig.java
@@ -14,6 +14,8 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.openapi.config;
 
+import java.util.function.Function;
+
 import io.aklivity.zilla.runtime.engine.config.WithConfig;
 
 public final class McpOpenapiWithConfig extends WithConfig
@@ -22,14 +24,18 @@ public final class McpOpenapiWithConfig extends WithConfig
     public final String operation;
     public final String tag;
 
-    public McpOpenapiWithConfig(
-        String spec,
-        String operation)
+    public static McpOpenapiWithConfigBuilder<McpOpenapiWithConfig> builder()
     {
-        this(spec, operation, null);
+        return new McpOpenapiWithConfigBuilder<>(McpOpenapiWithConfig.class::cast);
     }
 
-    public McpOpenapiWithConfig(
+    public static <T> McpOpenapiWithConfigBuilder<T> builder(
+        Function<WithConfig, T> mapper)
+    {
+        return new McpOpenapiWithConfigBuilder<>(mapper);
+    }
+
+    McpOpenapiWithConfig(
         String spec,
         String operation,
         String tag)
--- a/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/McpOpenapiRouteConfig.java
+++ b/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/McpOpenapiRouteConfig.java
@@ -20,6 +20,7 @@ import java.util.List;
 
 import io.aklivity.zilla.runtime.binding.mcp.openapi.config.McpOpenapiConditionConfig;
 import io.aklivity.zilla.runtime.binding.mcp.openapi.config.McpOpenapiWithConfig;
+import io.aklivity.zilla.runtime.engine.config.ConfigException;
 import io.aklivity.zilla.runtime.engine.config.RouteConfig;
 
 public final class McpOpenapiRouteConfig
@@ -36,5 +37,26 @@ public final class McpOpenapiRouteConfig
             .map(McpOpenapiConditionConfig.class::cast)
             .collect(toList());
         this.with = (McpOpenapiWithConfig) route.with;
+
+        if (this.with != null && isBulk(this.with) && namesToolOrResource(this.when))
+        {
+            throw new ConfigException(
+                "mcp_openapi route with a bulk selector (with.tag, a glob with.operation pattern, or spec-only) " +
+                "must not declare when.tool or when.resource");
+        }
+    }
+
+    public static boolean isBulk(
+        McpOpenapiWithConfig with)
+    {
+        return with.tag != null ||
+            with.operation == null ||
+            with.operation.indexOf('*') != -1;
+    }
+
+    private static boolean namesToolOrResource(
+        List<McpOpenapiConditionConfig> when)
+    {
+        return when.stream().anyMatch(w -> w.tool != null || w.resource != null);
     }
 }
--- a/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/McpOpenapiRouteConfig.java
+++ b/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/McpOpenapiRouteConfig.java
@@ -20,7 +20,6 @@ import java.util.List;
 
 import io.aklivity.zilla.runtime.binding.mcp.openapi.config.McpOpenapiConditionConfig;
 import io.aklivity.zilla.runtime.binding.mcp.openapi.config.McpOpenapiWithConfig;
-import io.aklivity.zilla.runtime.engine.config.ConfigException;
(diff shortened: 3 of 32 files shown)


runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiWithConfig.java
```diff
@@ -20,12 +20,22 @@ public final class McpOpenapiWithConfig extends WithConfig
...
         String spec,
         String operation)
+    {
+        this(spec, operation, null);
+    }
+
+    public McpOpenapiWithConfig(
+        String spec,
+        String operation,
+        String tag)
```
jfallows: Single, all-args, package-private constructor, called from builder.

runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/config/McpOpenapiConditionConfig.java
```diff
@@ -14,18 +14,30 @@
...
         String tool,
         String resource)
+    {
+        this(tool, resource, null);
+    }
+
+    public McpOpenapiConditionConfig(
+        String tool,
+        String resource,
+        List<String> capability)
```
jfallows: Single, all-args, package-private constructor, called from builder.

runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/McpOpenapiRouteConfig.java
```diff
@@ -36,5 +37,26 @@ public McpOpenapiRouteConfig(
...
+        }
+    }
+
+    public static boolean isBulk(
+        McpOpenapiWithConfig with)
+    {
+        return with.tag != null ||
+            with.operation == null ||
+            with.operation.indexOf('*') != -1;
+    }
```
jfallows: Why static method, can we not just have a method isBulk directly on McpOpenapiRouteConfig since with is a field already?

runtime/binding-mcp-openapi/src/test/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGeneratorTest.java
```diff
@@ -199,6 +202,29 @@ public class McpOpenapiCompositeGeneratorTest
         "  }" +
         "}";
```
jfallows: Use multi-line strings `"""` in all tests.

runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/McpOpenapiRouteConfig.java
```diff
@@ -36,5 +37,25 @@ public McpOpenapiRouteConfig(
             .map(McpOpenapiConditionConfig.class::cast)
             .collect(toList());
         this.with = (McpOpenapiWithConfig) route.with;
+
+        if (this.with != null && isBulk() && namesToolOrResource(this.when))
+        {
+            throw new ConfigException(
+                "mcp_openapi route with a bulk selector (with.tag, a glob with.operation pattern, or spec-only) " +
+                "must not declare when.tool or when.resource");
+        }
```
jfallows: This validation should be in the schema, not here.

runtime/binding-mcp-openapi/src/test/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGeneratorTest.java
jfallows: Prefer duplication in each test method to make them linear and more readable.

runtime/binding-mcp-openapi/src/test/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGeneratorTest.java
```diff
@@ -255,9 +255,35 @@ public void initMocks()
...
-        BindingConfig binding = bindingWithRoutes(
-            route(0, "create_pr", null, "pulls/create"),
-            route(1, null, "repo://{owner}/{repo}", "repos/get"));
+        BindingConfig binding = BindingConfig.builder()
+            .namespace("test")
+            .name("mcp_openapi0")
+            .type("mcp_openapi")
+            .kind(CLIENT)
+            .options(McpOpenapiOptionsConfig.builder()
+                .spec(new McpOpenapiSpecificationConfig("openapi_github0",
```
jfallows: McpOpenapiOptionsConfigBuilder.spec() should return a McpOpenapiSpecificationConfigBuilder with mapper to add spec to internal specs field, then use fluent style for spec here too.

runtime/binding-mcp-openapi/src/test/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGeneratorTest.java
```diff
@@ -255,9 +255,35 @@ public void initMocks()
...
+                    "https://api.github.com",
+                    of(new McpOpenapiCatalogConfig("catalog0", "rest-api", "latest")),
```
jfallows: McpOpenapiSpecificationConfigBuilder.catalog() should return a McpOpenapiCatalogConfigBuilder, then use fluent style for catalog here too.

McpOpenapiSpecificationConfig (and all configuration objects) should have package-private all-args constructor and builder.

runtime/binding-mcp-openapi/src/test/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGeneratorTest.java
```diff
@@ -255,9 +255,35 @@ public void initMocks()
...
+                    Map.of("bearerAuth", "guard0", "oauthScheme", "guard0")))
+                .build())
+            .routes(of(
+                RouteConfig.builder()
+                    .order(0)
+                    .when(McpOpenapiConditionConfig.builder().tool("create_pr").build())
```
jfallows: Always use multi-line for builders to capture structure more clearly in tests.

runtime/binding-mcp-openapi/src/test/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGeneratorTest.java
```diff
@@ -255,9 +255,35 @@ public void initMocks()
...
+            .name("mcp_openapi0")
+            .type("mcp_openapi")
+            .kind(CLIENT)
+            .options(McpOpenapiOptionsConfig.builder()
+                .spec(new McpOpenapiSpecificationConfig("openapi_github0",
+                    "https://api.github.com",
+                    of(new McpOpenapiCatalogConfig("catalog0", "rest-api", "latest")),
+                    Map.of("bearerAuth", "guard0", "oauthScheme", "guard0")))
+                .build())
+            .routes(of(
```
jfallows: BindingConfigBuilder.should have route() method returning RouteConfigBuilder with mapper to add route to internal routes field on binding, then use fluent style for route here too.

runtime/binding-mcp-openapi/src/test/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGeneratorTest.java
```diff
@@ -255,9 +255,35 @@ public void initMocks()
...
+                    .when(McpOpenapiConditionConfig.builder().resource("repo://{owner}/{repo}").build())
+                    .with(McpOpenapiWithConfig.builder().spec("openapi_github0").operation("repos/get").build())
+                    .build()))
+            .build();
+        binding.resolveId = name -> switch (name)
+        {
+        case "catalog0" -> 1L;
+        case "guard1" -> 3L;
+        default -> 2L;
+        };
```
jfallows: This can be changed to a lambda fixture private final `resolveId` field and reused per test without impacting readability, actually improving it to reduce noise.

runtime/binding-mcp-openapi/src/test/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGeneratorTest.java
```diff
@@ -770,78 +1264,20 @@ private BindingConfig bindingOf(
...
-        return RouteConfig.builder()
-            .order(order)
-            .when(McpOpenapiConditionConfig.builder().capability(of(capability)).build())
-            .with(McpOpenapiWithConfig.builder().spec(apiId).tag(tag).build())
-            .build();
+        return composite.namespaces.get(0).bindings.stream()
+            .filter(b -> "mcp_http0".equals(b.name))
+            .findFirst()
+            .orElse(null);
     }
```
jfallows: All of these static helper methods are inconsistent with the linear test style we want, please inline them instead.

jfallows (commented):
LGTM

Issue #2022, opened 2026-07-05
mcp_openapi: params: rebinding and authored tool input schemas


jfallows:
## Summary

`mcp_openapi` currently derives a tool's input schema directly from its operation's parameters and request body, with every parameter bound by passthrough (the generated argument name always equals the OpenAPI parameter name). Add support for an operator-authored input schema with a different shape, plus an explicit parameter-binding map to reconcile the two.

## `params:` binding map

Add a `params:` map to a route's `with` block. It is OpenAPI-major: each key is one of the operation's parameter names, and each value is an expression drawing from the inbound MCP side — `${args.x}` / `${args.x.y}` for a tool route, `${params.x}` for a resource route's URI captures.

Rules:

- Omitting `params:` entirely preserves today's behavior: every parameter binds by passthrough from the like-named `${args.x}` (tool) or `${params.x}` (resource).
- Parameter placement (path / query / header / cookie) always comes from the OpenAPI spec's `in:` field for that parameter, never from `params:`. The map only supplies the value source, not where it lands on the outbound request.
- The request body is not decomposed field-by-field through `params:`. It continues to be bound as a single unit via the existing body-schema mechanism.

## Authored `schemas.input`

Add a `schemas.input` override (tools only) alongside the existing `schemas.output`, in both the options schema and the config adapter, using the **same shape as `schemas.output`** — a `ModelConfig` (`$ref: #/$defs/models`, e.g. `model: json` + `catalog:`). Fixed a pre-existing dangling `$ref: #/$defs/converter` (nonexistent; the engine's base schema defines `#/$defs/models`) for both `output` and the new `input`.

## Validation strategy — none of this is checkable at config-load time, and that's fine

`Binding.validate(BindingConfig binding)` (#2031/#2034) only ever receives one binding's own already-parsed config — never the surrounding `NamespaceConfig`, never a `CatalogHandler`. All `${args.x}`/`${params.x}` path-validity checking relies entirely on the existing, unmodified `mcp_http` `unsatisfiedAccessors()` / `schemaAccessorUnresolved` mechanism. No new validation code was added anywhere. `schemas.input` on a resource route is rejected for free by JSON-schema shape alone.

## Implementation status — done, verified end-to-end against a live engine

- `params:` binding map and `schemas.input` parse, round-trip, and flow through `McpOpenapiCompositeGenerator` into the generated `mcp_http` composite, for **both tool routes and resource routes**.
[...]

PR #2034, opened 2026-07-06
feat(engine): add Binding.validate(BindingConfig) SPI hook for config load-time validation

### LINKED ISSUE #2031: engine: add Binding.validate(BindingConfig) SPI hook for cross-field config validation at load time
## Summary

Bindings have no principled place to run cross-field / multi-object validation over their own parsed config tree before the engine starts routing traffic.

`*ConfigAdapter` classes (e.g. `McpOpenapiOptionsConfigAdapter`, `McpOpenapiWithConfigAdapter`) are pure JSON-shape parsers — they build DTO-like `*Config` objects from `zilla.yaml` and should not carry semantic, cross-referencing logic (e.g. "does this expression navigate a valid path through a schema declared elsewhere in the same binding"). That kind of check spans multiple parts of a single binding's already-parsed config tree, but today there is nowhere for it to live:

- `EngineManager.parse()` only runs JSON-schema shape validation (`EngineConfigReader`) plus Jsonb deserialization via the `*ConfigAdapter` classes.
- `EngineManager.process(...)` resolves namespace-qualified ids and computes routing/affinity bookkeeping — no semantic validation.
- `register()` → the binding's `*BindingContext`/`*ClientFactory.attach()` is the first point a binding sees its fully-resolved `BindingConfig`, but by then the namespace is already being registered. Any validation there is ad hoc and inconsistent — most bindings do none, and where one does (`McpOpenapiClientFactory.attach()` → `McpOpenapiCompositeGenerator.generate()`), it currently fails silently (skips the route) rather than raising a config error.
- No binding's `*Config.java` (as opposed to `*ConfigAdapter.java`) throws `ConfigException` for a cross-field issue today; `ConfigException` is only thrown by `EngineConfigReader` (schema errors) and `EngineManager`'s generic wrap of any load failure.

## Proposal

Add a default no-op extension point to the `Binding` SPI (`runtime/engine/.../binding/Binding.java`):

```java
default void validate(BindingConfig binding) throws ConfigException
{
}
```

Call it once per binding from `EngineManager.process(...)`, right after that binding's routes/options are fully resolved into `BindingConfig`/`RouteConfig` objects and before `register()`/`attach()`. At that point all the data such a check needs — the binding's own parsed config tree — is already in memory, so a thrown `ConfigException` can propagate through the same path JSON-schema violations already use, failing the whole config load deterministically and synchronously, instead of being silently dropped or only discovered later inside `attach()`.

**Explicitly out of scope:** validation that depends on data resolved asynchronously from outside the config document itself — e.g. an OpenAPI spec fetched through a `CatalogHandler`, whose `resolve(...)` contract is non-blocking and may not have a value yet at load time (see #2022). That class of check cannot be a hard load-time guarantee under the current catalog model and needs a different strategy (event-log + protocol-level error at the call boundary), not this hook.

## Motivating case

#2022 (`mcp_openapi` `params:` binding map) wants two checks that are pure config-tree-internal (no OpenAPI spec / catalog data needed) to fail at config load:
- every `${args.x}` / `${args.x.y}` expression in `params:` resolves against the tool's own `inputSchema` (authored or derived)
- every `${params.x}` expression in `params:` corresponds to a capture actually present in the route's own resource URI template

Both are good first callers of this hook once it exists. Validation of `params:` expressions in #2022 is being deferred until this hook (or an equivalent) lands, rather than being bolted onto the JSON config adapters.

jfallows:
## Description

Adds a `default void validate(BindingConfig binding) throws ConfigException` no-op hook to the `Binding` SPI (`runtime/engine/.../binding/Binding.java`), giving bindings a principled place to run cross-field / multi-object validation over their own already-parsed config tree before the engine starts routing traffic.

- `EngineManager.process(...)` now calls `typed.validate(binding)` once per binding, immediately after that binding's routes/options are fully resolved into `BindingConfig`/`RouteConfig` objects, and before the binding is registered/attached.
- A thrown `ConfigException` propagates through the same path JSON-schema violations already use, failing the whole config load deterministically and synchronously.
- The hook defaults to a no-op, so it's backward compatible with every existing `Binding` implementation.
- `TestBinding` (engine test scaffolding) now overrides `validate()` to throw for a sentinel binding name, and `EngineTest` has a new `shouldNotConfigureBindingValidationFailure` regression test exercising the hook end-to-end. Verified the test fails without the `EngineManager` wiring and passes with it.

This lands only the generic hook, per the issue's stated scope. Wiring it up for the `mcp_openapi` `params:` checks from #2022 is left as follow-up work.

### Testing

- `./mvnw test -pl runtime/engine` — full unit test suite passes (245 tests), aside from one pre-existing, environment-related failure (`TrustedTest.shouldConfigureTrustViaCacerts`) confirmed to fail identically on the unmodified base branch (no CA certs in this sandbox).
- `./mvnw checkstyle:check -pl runtime/engine` — clean.

Fixes #2031

---
_Generated by [Claude Code](https://claude.ai/code/session_01DzmC3AxYNy2HVtagPZjun5)_

### DIFF
--- a/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/binding/Binding.java
+++ b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/binding/Binding.java
@@ -19,6 +19,8 @@ import java.net.URL;
 
 import io.aklivity.zilla.runtime.engine.EngineContext;
 import io.aklivity.zilla.runtime.engine.EngineController;
+import io.aklivity.zilla.runtime.engine.config.BindingConfig;
+import io.aklivity.zilla.runtime.engine.config.ConfigException;
 import io.aklivity.zilla.runtime.engine.config.KindConfig;
 
 /**
@@ -143,5 +145,30 @@ public interface Binding
         KindConfig kind)
     {
         return Integer.MAX_VALUE;
+    }
+
+    /**
+     * Validates a binding's fully-resolved configuration at load time.
+     * <p>
+     * Called once per binding, after its options and routes have been resolved into
+     * {@link BindingConfig} and {@code RouteConfig} objects, but before the binding is
+     * registered and begins routing traffic. This is the place for cross-field or
+     * multi-object checks that span the binding's own already-parsed config tree — the
+     * kind of validation a {@code ConfigAdapter} should not carry because it spans more
+     * than the single JSON shape being adapted.
+     * </p>
+     * <p>
+     * Does nothing by default. Implementations that need this hook should throw
+     * {@link ConfigException} to fail the configuration load; validation that depends on
+     * data resolved asynchronously from outside the config document itself does not belong
+     * here.
+     * </p>
+     *
+     * @param binding  the fully-resolved binding configuration
+     * @throws ConfigException  if the binding configuration is invalid
+     */
+    default void validate(
+        BindingConfig binding) throws ConfigException
+    {
     }
 }
--- a/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
+++ b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
@@ -502,6 +502,8 @@ public class EngineManager
                 }
             }
 
+            typed.validate(binding);
+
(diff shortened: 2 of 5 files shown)


jfallows (commented):
LGTM

Issue #2033, opened 2026-07-06
binding-mcp: support resources/templates/list as a distinct JSON-RPC method (server, proxy, client kinds)


jfallows:
## Summary

Follow-up to #2023. That issue's `mcp_openapi`/`mcp_http`-side work (resource vs. resource-template classification, `options.resources` overrides, query-capture URI suffixes) is done, but the underlying `binding-mcp` protocol layer still has no distinct `resources/templates/list` JSON-RPC method — resource templates are folded into `resources/list` today, discriminated only by a `uriTemplate` vs. `uri` key on each item. Per the MCP spec, `resources/templates/list` is a separate method with its own response shape (`{"resourceTemplates":[...]}`), gated by the same `resources` capability rather than a new one.

This issue tracks adding real protocol-level support for that method across all three `binding-mcp` kinds (`server`, `proxy`, `client`), so that once it lands, `mcp_openapi`/`mcp_http` can be wired to actually split their already-classified resources into the two distinct listings.

## What exists today

- `McpHttpResourceConfig` (binding-mcp-http) already carries an explicit `template` boolean (added in #2023's follow-up work), computed by `mcp_openapi` from whether the OpenAPI path has a `{param}` capture, and by `mcp_http`'s own YAML adapter via a brace-presence check for hand-authored resources.
- `McpHttpProxyFactory.buildResourcesList()` uses that flag today only to pick the JSON key (`uri` vs `uriTemplate`) per item — it still emits every resource (concrete or template) into the single `resources/list` response.
- `runtime/binding-mcp`'s `McpServerFactory` JSON-RPC dispatch switch recognizes exactly `tools/list`, `tools/call`, `prompts/list`, `prompts/get`, `resources/list`, `resources/read` — no `resources/templates/list` case exists; a request for it today falls through to `onDecodeParseError` as an unknown method.
- The shared `mcp.idl` (`specs/binding-mcp.spec/src/main/resources/META-INF/zilla/mcp.idl`) is the single source of truth for `McpBeginEx`/`McpFlushEx` types — both `binding-mcp` and `binding-mcp-http` independently regenerate their own local copies of these types from it via the `flyweight-maven-plugin`'s `scopeNames` config, so a type change there propagates to both modules on rebuild.
[...]

PR #2035, opened 2026-07-06
feat(binding-mcp-openapi): options.resources overrides, resource/template classification, optional query param omission

### LINKED ISSUE #2023: mcp_openapi: options.resources overrides, resource/template discrimination, query param omit-when-absent
## Summary

Three related gaps in how `mcp_openapi` exposes OpenAPI operations as MCP resources: there is no way to override a resource's description or output schema, no distinction between a concrete resource and a resource template, and optional query parameters are always emitted on the generated outbound request even when the caller didn't supply a value.

## `options.resources` override map

Add an `options.resources` map, keyed by resource URI template, analogous to the existing `options.tools` override map for tools. Each entry may override:

- `description` — replaces the spec-derived description.
- `schemas.output` — a response projection, narrowing the emitted response surface (extra fields dropped), the same mechanism already used for tool output.

An operation selected by a route but not listed in `options.resources` gets spec-derived defaults, same as tools today.

## Resource vs. resource-template discrimination

Introduce a distinction between a concrete resource (enumerable, listed in a resource listing) and a resource template (parameterized by a path capture, listed separately). The discriminator is the presence of a `{}` path-parameter expression in the resource's URI — one that binds, via `params:` or passthrough, to an `in: path` operation parameter:

- No such path-parameter expression in the URI → concrete resource. A query-only parameterization (e.g. a `{?limit}`-style optional query capture) does not disqualify it — a query-parameterized collection is still a fixed, addressable resource.
- Any path-parameter expression in the URI → resource template.

This requires corresponding support in the underlying MCP binding(s) this composite generates into, since a resource-template listing distinct from a plain resource listing does not currently exist anywhere in that layer — this issue's scope includes whatever downstream capability is needed to actually distinguish the two in the protocol responses, not just the `mcp_openapi`-side classification.

Each resource route continues to bind exactly one operation; a resource's entire input surface is its URI template (path and query captures), never `${args.x}`.

## Query parameter omit-when-absent

Currently, every declared query parameter on an operation is unconditionally appended to the generated outbound `:path`, regardless of whether it's required or optional, and regardless of whether the caller actually supplied a value. Change this so an optional query parameter with no supplied value is omitted entirely from the generated query string, rather than being emitted as a literal empty value or an unresolved expression.

Decide and implement the exact composition rule before implementation is considered complete:

- How a single optional parameter is omitted when absent.
- How multiple optional parameters compose into one query string when some are present and others are absent (no dangling `&` or empty segments).
- How an explicitly-empty-string value is distinguished from an absent value, if that distinction matters for a given parameter.

Required query parameters continue to be always emitted, as today.

jfallows:
## Description

Implements two of the three gaps described in #2023, plus the classification groundwork for the third:

- **`options.resources` override map** — a new map keyed by resource URI, analogous to the existing `options.tools` map. Each entry may override `description` and `schemas.output`, applied the same way tool overrides already are.
- **Resource vs. resource-template classification** — a resource is classified as a template based on whether its OpenAPI path contains a `{param}` path capture; a query-only parameterization no longer disqualifies it from being concrete. Concrete resources with query parameters now advertise an RFC 6570 `{?name1,name2}` suffix on their URI. `McpHttpResourceConfig` carries this as an explicit `template` boolean rather than inferring it from brace-presence in the URI (which the new query-capture suffix would otherwise make ambiguous).
- **Query-param omit-when-absent** — optional query parameters are now embedded as a `${?expression=name}` marker at generation time; `McpHttpProxyFactory` resolves each query fragment independently at request time and drops absent optional parameters (and their separator) entirely, instead of emitting `name=` with an empty value or an unresolved expression. An explicitly empty-string value is still emitted.

All three changes are covered by new unit tests and k3po spec/IT scenarios across `binding-mcp-openapi`, `binding-mcp-http`, and their spec modules. `./mvnw verify` (including JaCoCo coverage gates) passes on all four touched modules.

**Not included here:** a distinct `resources/templates/list` JSON-RPC method at the `binding-mcp` protocol layer — resource templates still surface via `resources/list` with a `uriTemplate` key, same as before. That requires new IDL types, server/proxy/client dispatch, and a new fan-in list-cache factory in `binding-mcp`, which is tracked separately in #2033 so `mcp_openapi`/`mcp_http` can be wired to it once it lands.

Fixes #2023 (partially — see #2033 for the remaining protocol-layer follow-up)

---
_Generated by [Claude Code](https://claude.ai/code/session_01WXGys1DUG73JHVG9Nhjcw1)_

### DIFF
--- a/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/config/McpHttpResourceConfig.java
+++ b/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/config/McpHttpResourceConfig.java
@@ -14,6 +14,8 @@
  */
 package io.aklivity.zilla.runtime.binding.mcp.http.config;
 
+import java.util.function.Function;
+
 import io.aklivity.zilla.runtime.engine.config.ModelConfig;
 
 public final class McpHttpResourceConfig
@@ -25,17 +27,18 @@ public final class McpHttpResourceConfig
     public final String mimeType;
     public final ModelConfig output;
 
-    public McpHttpResourceConfig(
-        String name,
-        String uri,
-        String description,
-        String mimeType,
-        ModelConfig output)
+    public static McpHttpResourceConfigBuilder<McpHttpResourceConfig> builder()
     {
-        this(name, uri, uri != null && uri.indexOf('{') >= 0, description, mimeType, output);
+        return new McpHttpResourceConfigBuilder<>(McpHttpResourceConfig.class::cast);
     }
 
-    public McpHttpResourceConfig(
+    public static <T> McpHttpResourceConfigBuilder<T> builder(
+        Function<McpHttpResourceConfig, T> mapper)
+    {
+        return new McpHttpResourceConfigBuilder<>(mapper);
+    }
+
+    McpHttpResourceConfig(
         String name,
         String uri,
         boolean template,
--- a/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/stream/McpHttpProxyFactory.java
+++ b/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/stream/McpHttpProxyFactory.java
@@ -794,7 +794,7 @@ public final class McpHttpProxyFactory implements BindingHandler
             doMcpWindow(traceId);
 
             final McpHttpWithConfig with = route.with;
-            final String path = interpolate(with.headers.get(HEADER_PATH), this::resolveParam);
+            final String path = buildPath(with.headers.get(HEADER_PATH), this::resolveParam, this::resolveRawParam);
             credentials.clear();
             binding.resolveCredentials(authorization, credentials);
             delegate.doHttpBegin(traceId, with.headers, credentials, path, null);
@@ -810,6 +810,21 @@ public final class McpHttpProxyFactory implements BindingHandler
             {
                 final String captured = params.get(expression.substring(7));
                 value = encode(captured != null ? captured : "");
+            }
+            return value;
+        }
+
+        // Unlike resolveParam (used for path segments, which are always present), this reports true
+        // absence: null when the referenced capture was never supplied, distinct from an explicitly
+        // captured empty string. Used only to decide whether an optional query fragment (see buildPath)
+        // is included in the outbound request at all.
+        private String resolveRawParam(
+            String expression)
+        {
+            String value = null;
+            if (expression.startsWith("params."))
+            {
+                value = params.get(expression.substring(7));
             }
             return value;
         }
@@ -1251,7 +1266,8 @@ public final class McpHttpProxyFactory implements BindingHandler
             long traceId)
         {
             final McpHttpWithConfig with = route.with;
-            String path = interpolate(with.headers.get(HEADER_PATH), this::resolveStreamingRequest);
+            String path = buildPath(with.headers.get(HEADER_PATH),
+                this::resolveStreamingRequest, this::resolveRawStreamingRequest);
             if (queryCaptured != null)
             {
                 final String query = buildQueryString(queryCaptured);
@@ -1293,6 +1309,24 @@ public final class McpHttpProxyFactory implements BindingHandler
             {
                 final String captured = params.get(expression.substring(7));
                 value = encode(captured != null ? captured : "");
+            }
+            return value;
+        }
+
+        // See McpProxy.resolveRawParam: reports true absence (null) rather than falling back to an
+        // empty string, so buildPath can tell an omitted optional argument apart from one the caller
+        // explicitly set to "".
+        private String resolveRawStreamingRequest(
+            String expression)
+        {
+            String value = null;
+            if (expression.startsWith("args."))
+            {
+                value = requestArgs.get(expression.substring(5));
+            }
+            else if (expression.startsWith("params."))
+            {
+                value = params.get(expression.substring(7));
             }
             return value;
         }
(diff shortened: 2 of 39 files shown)


runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/config/McpHttpResourceConfig.java
```diff
@@ -30,9 +31,21 @@ public McpHttpResourceConfig(
...
+    }
+
+    public McpHttpResourceConfig(
+        String name,
+        String uri,
+        boolean template,
+        String description,
+        String mimeType,
+        ModelConfig output)
     {
```
jfallows: Configs should have only all-args package-private constructors, called by builder.

runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/stream/McpHttpProxyFactory.java
```diff
@@ -814,6 +814,21 @@ private String resolveParam(
             return value;
         }
 
+        // Unlike resolveParam (used for path segments, which are always present), this reports true
+        // absence: null when the referenced capture was never supplied, distinct from an explicitly
+        // captured empty string. Used only to decide whether an optional query fragment (see buildPath)
+        // is included in the outbound request at all.
```
jfallows: Remove these comments.

runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/stream/McpHttpProxyFactory.java
```diff
@@ -2474,6 +2508,69 @@ private static String compact(
...
+            final String raw = rawResolver.apply(expression);
+            result = raw != null ? name + "=" + encode(raw) : null;
+        }
+        else
+        {
+            result = interpolate(fragment, resolver);
+        }
+        return result;
+    }
+
```
jfallows: None of this belongs here, should be on McpBindingConfig instead.

runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGenerator.java
```diff
@@ -338,14 +354,22 @@ private McpHttpOptionsConfig mcpHttpOptions(
...
             else
             {
-                final String uri = entry.operation.path;
+                final ModelConfig output = entry.resourceConfig != null && entry.resourceConfig.output != null
+                    ? entry.resourceConfig.output
+                    : jsonModel("%s-output".formatted(name));
+                final String description = entry.resourceConfig != null ? entry.resourceConfig.description : null;
+                // a {param} path-parameter expression in the OpenAPI path is what makes this a resource
+                // template rather than a concrete, enumerable resource; a query-only parameterization
+                // (appended below as a separate {?...} suffix) does not disqualify it from being concrete
```
jfallows: Remove these comments.

runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/stream/McpHttpProxyFactory.java
```diff
@@ -814,6 +815,17 @@ private String resolveParam(
...
+            String expression)
+        {
+            String value = null;
+            if (expression.startsWith("params."))
+            {
+                value = params.get(expression.substring(7));
+            }
+            return value;
+        }
+
```
jfallows: These and other helper methods belong on McpBindingConfig.

runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGenerator.java
jfallows, line 1: Review this class for conditionals preventing full unbroken use of fluent api, and instead consider using .inject(...) with method reference to perform conditional logic without breaking the fluent api usage.

runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGenerator.java
```diff
@@ -226,6 +228,21 @@ private McpOpenapiToolConfig toolConfig(
...
+        McpOpenapiResourceConfig result = null;
+        if (resource != null && binding.options != null && binding.options.resources != null)
+        {
+            result = binding.options.resources.stream()
+                .filter(r -> resource.equals(r.uri))
+                .findFirst()
+                .orElse(null);
+        }
+        return result;
+    }
```
jfallows, line 272: Shouldn't this also be an inject method?

runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGenerator.java
```diff
@@ -842,13 +913,15 @@ private RoutedOperation(
             String resource,
             OpenapiOperationView operation,
             McpOpenapiToolConfig toolConfig,
+            McpOpenapiResourceConfig resourceConfig,
```
jfallows: Wouldn't it make sense for the McpOpenapiToolConfig to know its own name, and for the McpOpenapiResourceConfig to know its own uri, then we can just pass McpOpenapiToolConfig tool, and McpOpenapiResourceConfig resource, instead of trying to disambiguate them because they are currently passed separately.

Issue #2040, opened 2026-07-07
common-json/common-avro/common-protobuf: JsonGeneratorImpl's atomic write methods have no room check


jfallows:
## Background

While fixing a real crash in `binding-mcp-http` (window credit propagation letting a JSON generator be wrapped with very little or zero destination room left), we found that `JsonGeneratorImpl.putRaw` has no real bounds check — only `assert progress < limit`, a no-op without `-ea`. `write(CharSequence, Completion)` and `writeKey(CharSequence, Completion)` each write a comma/quote unconditionally outside `writeStringBody`'s own per-codepoint bound check, so a caller that legitimately wraps the generator with little/no room left gets an uncaught write past `limit` — silent memory corruption in production (no exception, no crash signal), occasionally manifesting later as an unrelated `SIGSEGV` once the corrupted memory is touched elsewhere.

`-ea` has been enabled for surefire/failsafe test runs (root `pom.xml`) so this class of bug now fails loudly (`AssertionError`) in tests instead of corrupting memory silently — but that's a test-time safety net, not a production fix.

## Fixed so far

`write(CharSequence, Completion)` and `writeKey(CharSequence, Completion)` in `JsonGeneratorImpl` now check `remaining()` before their opening quote/comma, and before their closing quote (handling the empty-value edge case, where `writeStringBody`'s loop never runs to enforce its own reserve). Verified via a minimal direct repro (zero-width wrap silently corrupted before the fix, cleanly deferred after) and new regression tests in `JsonGeneratorExTest`.

## Remaining work (this issue)

The same "call `preValue()` (or equivalent) then write unconditionally" shape exists in:

- `JsonGeneratorImpl.write(int)`, `write(long)`, `write(boolean)`, `writeNull()`, `writeStartObject()`/`writeStartObject(String)`, `writeStartArray()`/`writeStartArray(String)`, `writeEnd()` — all atomic/non-resumable (no `Completion`-based partial-write contract like the string/key path has).
  - `write(int)`/`write(long)` currently have **no callers anywhere in the repo** (every numeric path routes through `writeNumber(CharSequence, Completion)` instead, which is already safe except for `preValue()`'s comma — see below). Low urgency, but part of the public `JsonGeneratorEx`/`jakarta.json.stream.JsonGenerator` contract.
  - `writeNumber(CharSequence, Completion)`'s body-writing is already safe (`writeAsciiBounded` checks `progress < limit` per character); only its `preValue()` call on entry is unguarded.
[...]

PR #2041, opened 2026-07-07
fix(common-json): guard JsonGeneratorImpl's unconditional quote/comma writes

jfallows:
## Description

While fixing a window-credit propagation bug in `binding-mcp-http` (`fix/mcp-http-coverage`, #2039), a crash surfaced that traced back to a real, separate defect in `common-json`: `JsonGeneratorImpl.putRaw` has no real bounds check — only `assert progress < limit`, a no-op without `-ea`. `write(CharSequence, Completion)` and `writeKey(CharSequence, Completion)` each write a comma/opening-quote unconditionally, outside `writeStringBody`'s own per-codepoint bound check. A caller that legitimately wraps the generator with little or no destination room left (e.g. a proxy draining its own outbound buffer only as fast as the wire allows) hits an uncaught write past `limit` — silent memory corruption in production, no exception, no crash signal at the point of corruption. It later manifests unpredictably (a `SIGSEGV` somewhere unrelated, once the corrupted memory is touched) rather than failing at the actual fault.

Confirmed via a minimal direct repro: wrapping the generator at a zero-width region (`offset == limit`) and calling `write("x")` silently wrote one byte past the boundary with exit code 0; the identical call under `-ea` threw a clean `AssertionError` at the exact write site instead.

### Changes

1. **Enable `-ea` for surefire/failsafe test runs** (root `pom.xml`) — converts this class of violation from silent corruption into a loud, catchable `AssertionError` during tests. This alone doesn't fix production behavior (assertions are typically off there too) but makes the existing `assert` statements actually useful for catching regressions going forward.

2. **`write(CharSequence, Completion)`** now checks real room before its two unconditional writes:
   - The opening quote (plus a possible preceding comma) checks `remaining()` up front. An already-complete *empty* value also reserves its closing quote atomically here, since `writeStringBody`'s loop never runs for a zero-length value and so can't independently enforce that reserve the way it does for a non-empty one — deferring the close afterwards would be ambiguous (`consumed()` reads the same, zero, whether the value is genuinely done or the close was merely deferred).
   - The closing quote gets its own direct check, covering the resumed-continuation case where the opening already happened in a prior call.

3. **`writeKey(CharSequence, Completion)`** gets the identical treatment (comma + opening quote, then closing quote + key separator), since it has the same shape.

On insufficient room, these methods now write nothing and leave `pending` state unchanged, so a caller (already required to check `consumed()` for the existing chunking contract) correctly resumes on retry — no new caller-visible contract, no behavior change for the already-common case of ordinary write patterns with adequate room.

### Not included (follow-up filed as #2040)

The same "call `preValue()` (or equivalent) then write unconditionally" shape also exists in `write(int)`, `write(long)`, `write(boolean)`, `writeNull()`, `writeStartObject`/`writeStartArray`/`writeEnd`, and in `AvroSinkImpl`/`AvroJsonGeneratorImpl` (`common-avro`) and `ProtobufJsonGeneratorImpl` (`common-protobuf`), which call these same methods unconditionally with no room check of their own. Fixing those properly requires giving them a deferred-write/retry contract and correctly wiring it through each pipeline's resume logic (notably `JsonSinkImpl.resume()`, which currently has no way to re-attempt a structural/boolean/null write that never happened — naively extending today's contract would silently drop data rather than corrupt memory, which is worse). Scoped out here given the cross-module surface and correctness risk of getting resume semantics wrong under time pressure; tracked in #2040.

### Testing

- New regression tests in `JsonGeneratorExTest`: `shouldNotWriteStringOpeningQuoteWithoutRoom`, `shouldDeferEmptyCompleteStringWithoutRoomForBothQuotes`, `shouldWriteEmptyCompleteStringWhenExactRoomForBothQuotes`, `shouldResumeStringOpeningQuoteOnceRoomIsAvailable`.
- Verified test-first: these tests fail with `AssertionError` (3 of 4) against the pre-fix code, and pass against the fix.
- Full `common-json` module suite (4836 tests) passes with `-ea` enabled, no regressions — including the pre-existing `shouldNotWriteBeyondLimit` test, which already expected `AssertionError` and is now genuinely exercised with `-ea` on.
- Checkstyle and license checks pass.

Fixes the root cause behind the crash seen in #2039; that branch will rebase onto this once merged.

---
_Generated by [Claude Code](https://claude.ai/code/session_016EtF5FFTbYY4jTAYLd1DoH)_

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -437,7 +437,7 @@
           <artifactId>maven-surefire-plugin</artifactId>
           <version>3.0.0-M5</version>
           <configuration>
-            <argLine>@{jacoco.java.option} -Xshare:off -Djdk.attach.allowAttachSelf=true -XX:+StartAttachListener --add-opens java.base/jdk.internal.misc=ALL-UNNAMED --enable-native-access=ALL-UNNAMED --sun-misc-unsafe-memory-access=deny -javaagent:${settings.localRepository}/org/mockito/mockito-core/${mockito.version}/mockito-core-${mockito.version}.jar</argLine>
+            <argLine>-ea @{jacoco.java.option} -Xshare:off -Djdk.attach.allowAttachSelf=true -XX:+StartAttachListener --add-opens java.base/jdk.internal.misc=ALL-UNNAMED --enable-native-access=ALL-UNNAMED --sun-misc-unsafe-memory-access=deny -javaagent:${settings.localRepository}/org/mockito/mockito-core/${mockito.version}/mockito-core-${mockito.version}.jar</argLine>
           </configuration>
           <dependencies>
             <dependency>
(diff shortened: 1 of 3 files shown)


jfallows (commented):
LGTM

PR #2045, opened 2026-07-07
feat(binding-mcp-openapi): params: binding map, schemas.input, and path/query/header/cookie rebinding

### LINKED ISSUE #2022: mcp_openapi: params: rebinding and authored tool input schemas

jfallows:
## Description

Adds `mcp_openapi` support for an operator-authored tool input schema (`schemas.input`) whose shape may diverge from an OpenAPI operation's own parameter names, plus a `params:` map on a route's `with:` block to rebind those diverged argument names back to the real OpenAPI parameters — for path, query, header, and cookie parameters alike.

Investigation (recorded on the issue) established that none of the originally-proposed load-time validation checks are implementable as new validation code: `Binding.validate(BindingConfig)` only ever sees one binding's own parsed config, never the surrounding `NamespaceConfig` or a `CatalogHandler`, and `schemas.input`'s catalog reference lives in a sibling `catalog:` declaration no `validate()` can see. This ended up being a pure generator feature — `binding-mcp-http`'s existing `unsatisfiedAccessors()`/`schemaAccessorUnresolved` mechanism already catches any mismatch at request time, uniformly, regardless of whether the input schema is authored or derived.

For headers and cookies specifically: a header is included when its accessor resolves and omitted otherwise (`McpHttpRouteConfig.resolveHeaders`, all-or-nothing per header). Cookies needed a different mechanism — multiple `in: cookie` parameters must aggregate into a single `Cookie:` header, so cookies get their own first-class `cookies:` map (parallel to `headers:`) that resolves each cookie independently and joins the survivors (`resolveCookies`), rather than reusing header semantics that can't represent "drop just this one cookie."

Also fixes two pre-existing bugs found along the way:
- `ModelConfigAdapter.adaptToJson` returning `null` (NPE) when `adaptType` wasn't called first, for both `schemas.output` and the new `schemas.input`.
- Authored `schemas.input`/`schemas.output` catalog references crossing into the generated `mcp_http` composite's own namespace unqualified, resolving against the wrong (same-named) catalog — fixed by qualifying with the caller's namespace-qualified `catalog.qname`.

### Key changes

- `mcp_openapi` schema patch: `params:` on `with:`, `schemas.input` on `options.tools.*`, fixed a dangling `$ref` typo.
- `mcp_openapi` config POJOs/adapters: `McpOpenapiWithConfig.params`, `McpOpenapiToolConfig.input`.
- `McpOpenapiCompositeGenerator`: threads `params:` through path/query/header/cookie parameter emission; prefers an authored `schemas.input` over the derived one; qualifies authored model catalog references across the namespace boundary.
- `binding-mcp-http`: `McpHttpRouteConfig.resolveHeaders`/`resolveCookies` for per-accessor resolve-or-omit (headers) and aggregate-or-omit (cookies); `McpHttpArguments` now captures nested tool-call arguments under dotted keys (e.g. `repository.owner`).

### Test plan

- [x] `./mvnw verify -pl runtime/binding-mcp-http` — 36 unit + 52 IT, checkstyle clean, coverage clean
- [x] `./mvnw verify -pl runtime/binding-mcp-openapi` — 39 unit + 21 IT, checkstyle clean, coverage clean
- [x] `./mvnw verify -pl specs/binding-mcp-http.spec` — clean
- [x] `./mvnw verify -pl specs/binding-mcp-openapi.spec` — clean (includes config-schema validation coverage for `schemas.input` rejection on resource routes)

Fixes #2022

---
_Generated by [Claude Code](https://claude.ai/code/session_01HjxpDJbm1YCtE68eXBpnvw)_


jfallows:
Added `body:` request-body field rebinding to this PR, extending the same nested-argument-divergence support from `params:` (path/query/header/cookie) to request body fields.

`binding-mcp-http` already supports a complete `body: {template: {...}}}` override on a route, so this folds that in from `mcp_openapi` rather than inventing new mechanics: a route's `with:` block can now carry a `body:` map (target OpenAPI requestBody field -> `${args...}` expression), parallel to but distinct from `params:` — kept separate to avoid overloading `params:` with body semantics. When present, `body:` is the **complete** requestBody template (every field the operator wants sent) and is passed straight through as `bodyTemplate` on the generated `mcp_http` binding; when absent, behavior is unchanged (the derived `body` ModelConfig projector).

While implementing this, fixed a pre-existing limitation in `binding-mcp-http`'s underlying rename mechanism (`McpHttpRename`), which only supported renaming depth-1 keys. Replaced it with `JsonTransforms.flatten` (new in `common-json`), a resumable `JsonTransform` that hoists arbitrarily-nested dotted accessor paths (e.g. `pr.branch`) up to their own top-level target key, discarding ancestor wrappers along the way — so `body:` (and `bodyTemplate:` generally) can now target nested tool-call arguments like `${args.pr.branch}`, not just flat ones.

### Key changes
- `mcp_openapi` schema patch: `body:` on `with:`, parallel to `params:`.
- `mcp_openapi` config POJOs/adapters: `McpOpenapiWithConfig.body`.
- `McpOpenapiCompositeGenerator`: emits `bodyTemplate` verbatim from `body:` instead of the derived `body` ModelConfig when present.
- `common-json`: new `JsonTransforms` utility class (moved `projector()` here from `JsonEx`, which is now pipeline plumbing only) plus `JsonFlattenerImpl`/`JsonTransforms.flatten()` for arbitrary-depth key rename.
- `binding-mcp-http`: `McpHttpRouteConfig`/`McpHttpProxyFactory` wired to `JsonTransforms.flatten`; deleted the now-superseded `McpHttpRename`.

### Test plan
- [x] `./mvnw verify -pl runtime/common-json` — 4842 tests, checkstyle clean
- [x] `./mvnw verify -pl runtime/binding-mcp-http` — 36 unit + 53 IT (incl. new nested `bodyTemplate` rename scenario), checkstyle clean, coverage clean
- [x] `./mvnw verify -pl runtime/binding-mcp-openapi` — 43 unit + 22 IT (incl. new `body:` end-to-end scenario), checkstyle clean, coverage clean
- [x] `./mvnw verify -pl specs/binding-mcp-http.spec` — clean (54 tests, incl. new `McpServerIT` peer coverage)
- [x] `./mvnw verify -pl specs/binding-mcp-openapi.spec` — clean (17 tests)

---

jfallows:
Flattened `options.tools.*`/`options.resources.*` and added `summary:`, per discussion — this is a **breaking config-schema change** to this PR's own `schemas.input`/`schemas.output` shape (acceptable since neither has shipped in a release yet; `mcp_openapi` is unreleased/v2.0).

- `options.tools.<name>.schemas: { input, output }` → flat `options.tools.<name>.input` / `.output`, siblings of `description`.
- `options.resources.<uri>.schemas: { output }` → flat `options.resources.<uri>.output`.
- Added `options.tools.<name>.summary: <string>` — an authored override for the MCP tool summary, same idea as `schemas.input` letting an authored shape diverge from the OpenAPI operation; previously `summary` was always derived from `operation.summary` or a `"Call <operationId>"` fallback with no way to override it.

Rationale: `mcp_http`'s own tool/resource config already has `input`/`output` as flat fields, because it mirrors the MCP wire protocol's own tool descriptor shape (`description`/`inputSchema`/`outputSchema` as siblings in `tools/list`). The `schemas:` grouping in `mcp_openapi` was just an authoring-convenience choice with no protocol constraint behind it, so flattening removes an asymmetry between the two config surfaces for no functional cost.

Updated all 6 fixtures in `specs/binding-mcp-openapi.spec` that used the old `schemas:` shape, added `summary:` coverage to both the adapter test and a new generator test (`shouldOverrideSummary`), and confirmed the resource-route rejection test (`schemas.input`/now `input` on a resource is invalid) still rejects correctly one level shallower.

### Test plan
- [x] `./mvnw verify -pl runtime/binding-mcp-openapi` — 44 unit + 22 IT, checkstyle clean, coverage clean
- [x] `./mvnw verify -pl specs/binding-mcp-openapi.spec` — clean (17 tests, incl. `shouldRejectResourceInputSchema`)

---

jfallows (commented):
LGTM

PR #2047, opened 2026-07-07
fix(binding-mcp): support per-tool guarded routes for client kind

### LINKED ISSUE #2046: binding-mcp: client kind lacks per-tool scope guarding (when/guarded + securitySchemes injection)
## Summary

The `mcp` binding's `client` kind (forwards MCP requests to a single upstream MCP server) doesn't support per-tool `guarded:` routes, unlike `binding-mcp-http`'s proxy (`when: [{tool: name}]` + `guarded:`, enforced via `McpHttpBindingConfig.toolGuarded(name)`) and unlike `binding-mcp`'s own `proxy` kind (`when: [{toolkit, tools: [glob...]}]` + `guarded:`, enforced via `McpBindingConfig.resolve(beginEx, authorization)` in `McpProxyItemFactory`).

Today, `client`-kind route resolution is tool-blind: `McpClientFactory.newStream()` calls `binding.resolve(long authorization)` — picking any route the caller's authorization satisfies, without looking at which tool/capability is being requested. Two consequences:

1. **`tools/call` can't be gated per tool.** A `client` binding with multiple routes intended for different tools would let a caller authorized for *any* route call *any* tool, since the resolved route is chosen before the request kind is even parsed.
2. **`securitySchemes` injection is blanket, not per-tool.** `McpSchemeInjector` (added in #1972 for advertising OAuth scopes on `tools/list` responses) stamps the *same* `securitySchemes` block onto every tool in the list, taken from whichever single route got resolved — it never looks at each tool's own name the way the sibling `McpScopeFilter` (used by the `proxy` kind) does.

For context, the rest of the picture already works as intended and does *not* need changing:
- `binding-mcp` `proxy` kind already correctly "only honors what it receives" — `McpScopeFilter` filters `tools/list` using whatever `securitySchemes` is already present in the payload, via `GuardHandler.verify()`, without maintaining its own tool→scope config.
- `binding-mcp` `server` kind already strips `securitySchemes` from the final reply to the real MCP client via `McpSchemeExcluder`, pending SEP-1488 finalization.

## Additional bug found while investigating

`McpBindingConfig`'s constructor unconditionally builds `routeByPrefix`/`aggregateRoutes` via `McpAggregateEventId.computePrefixes(toolkits)` whenever a binding has more than one route, regardless of `kind`. `computePrefixes` → `encodeCrc32c(toolkit)` calls `toolkit.getBytes(UTF_8)`, which throws an NPE the instant any route's `toolkit` is null. Since `client`-kind routes have no `toolkit` concept (there's only one upstream `options.server` per `client` binding, no toolkit aggregation), a `client` binding configured with two or more per-tool routes hits this NPE today at binding-attach time.

## Proposed fix

- `McpConditionConfig`/`McpConditionMatcher` already tolerate `toolkit == null` combined with a `tools: [glob...]` list (verified: `McpConditionMatcher` builds an empty-string prefix rather than crashing), so `client`-kind routes can reuse the existing `tools: [glob...]` condition shape without requiring `toolkit`.
- Fix the NPE by skipping toolkit-less routes when building the prefix-aggregation map (that aggregation is a `proxy`-only concept anyway).
- Add `McpRouteConfig.matchesTool(name)` and `McpBindingConfig.getRoles(String toolName)`, mirroring `McpHttpBindingConfig.toolGuarded(name)`, for per-tool route lookup independent of the caller's current authorization (advertising a required scope must not depend on whether the caller already has it).
- Restructure `McpClientFactory.newStream()` to parse the `McpBeginExFW` before resolving a route, and use the existing beginEx-aware `binding.resolve(beginEx, authorization)` overload (already used by the `proxy` kind) for every non-lifecycle request kind — so a per-tool `guarded:` route is enforced through the existing `route.authorized()` gate at `tools/call`, and `McpSchemeInjector` can look up each tool's own matching route by name instead of using one blanket scheme set for the whole list.
- Existing single-route (or no explicit `routes:`) `client` configs are unaffected — this only changes behavior once multiple per-tool routes are configured.
- Add a `routes` property to the `client`-kind block in `specs/binding-mcp.spec`'s `mcp.schema.patch.json` (currently the schema doesn't mention `routes` at all for `client`), reusing the `proxy` block's `when.items` shape without the `required: ["toolkit"]` constraint.

Test-first per this repo's discipline: a unit test on `McpBindingConfig` covering the NPE fix + new lookup method, and new `McpClientIT` scenarios covering (a) two tools receiving different `securitySchemes` from their own matching route, and (b) `tools/call` being accepted/rejected per tool based on its own route's `guarded:` roles.

Note: because a route's id becomes the actual network-side `routedId` for the client's outbound stream, per-tool routes on one `client` binding should share the same `exit:` (natural, since there's exactly one upstream `options.server` per `client` binding) — will call this out in docs/schema description.

jfallows:
## Description

The `mcp` binding's `client` kind resolved routes tool-blind, so it couldn't express different guarded scopes per tool the way `binding-mcp-http` and `binding-mcp`'s `proxy` kind already do. This made `McpSchemeInjector` stamp one blanket `securitySchemes` block onto every tool in a `tools/list` response, and left `tools/call` unable to enforce a per-tool guarded route at all.

### Key changes

- `McpClientFactory.newStream` now resolves via the existing beginEx-aware overload (`McpBindingConfig.resolve(beginEx, authorization)`, already used by the `proxy` kind) for every non-lifecycle request kind, so a per-tool `guarded:` route is enforced through the existing `route.authorized()` gate at `tools/call`.
- Added `McpRouteConfig.matchesTool(name)` and `McpBindingConfig.getRoles(String toolName)` for per-tool route lookup, mirroring `McpHttpBindingConfig.toolGuarded(name)`.
- Rewrote `McpSchemeInjector` to resolve `securitySchemes` per tool item (by name) instead of stamping one fixed scheme set on the whole list.
- Updated the `mcp` binding's JSON schema patch so `client`-kind routes can declare `when: [{tools: [...]}]` + `guarded:` without requiring `toolkit` (which has no meaning for a client with a single upstream `options.server`).
- Fixed a latent NPE: `McpBindingConfig` unconditionally built a toolkit-prefix aggregation map whenever a binding had more than one route, which crashed the moment any route's `toolkit` was null (always true for `client`-kind routes).

### Follow-up cleanup: removed the redundant `capability` condition field

While adding the `client`-kind schema, we noticed `capability: [tools, prompts, resources]` and the per-capability `tools`/`prompts`/`resources` glob allow-sets were doing overlapping jobs — `capability` predates the globs (it originally only drove the proxy's aggregated-capability advertisement), and layering the glob fields onto the same condition meant a route scoped by `tools:` alone would, absent an explicit `capability` restriction, silently also admit *any* name for prompts/resources (an unspecified allow-set defaults to "admits everything" once its capability is considered active).

Since none of this has shipped yet, we removed `capability` entirely: a capability is now active purely based on whether its own glob field is present (omit all three to match everything, same as a bare `{toolkit: X}` today). To activate a capability with no name restriction, use a wildcard allow-set (`tools: ["*"]`) instead of the old `capability: [tools]` with no matching glob. This updates every existing proxy fixture and the new per-tool client fixtures, and drops the now-moot `allOf` cross-field validation (and its negative test) that only existed to catch a capability/glob mismatch.

### Testing

Test-first per this repo's discipline:
- 6 new unit tests on `McpBindingConfig` covering the NPE fix and the new per-tool lookup.
- 2 new `McpClientIT` integration scenarios: per-tool `securitySchemes` injection (two tools, two different required scopes) and per-tool call-time rejection (an unguarded tool succeeds, a guarded tool without the required role is rejected).
- Matching `ApplicationIT` peer-consistency scripts for both new scenarios.
- `McpConditionMatcherTest`/`McpConditionConfigAdapterTest` rewritten for the `capability` removal, including new coverage for the "no allow-sets given" and wildcard-allow-set cases.

Full `verify` passes clean on both modules:
- `runtime/binding-mcp`: 67 unit + 248 integration tests, checkstyle clean, coverage gates met.
- `specs/binding-mcp.spec`: 103 unit + 223 integration tests, checkstyle clean, coverage gates met.

Fixes #2046


jfallows (commented):
LGTM

PR #2061, opened 2026-07-08
fix(binding-mcp): thread inherited authorization into guard fallback calls

jfallows:
## Description

`binding-mcp`'s client-side credential fallback tries `guard.credentials(authorization)` before falling back to `guard.reauthorize(...)`, so a guard that already recognizes the inherited `authorization` value doesn't need to redo the full reauthorize/preauthorize dance on every request. Both fallback call sites now pass the inherited `authorization` (rather than `initialId`) into `reauthorize`/`preauthorize`, since that's the value guards use to look up an already-established session.

`GuardHandler.preauthorize` gains an explicit `sessionId` parameter alongside the existing `contextId`. Previously a guard that delegates identity resolution to another guard (e.g. one configured with a `via` reference to a second guard) had no parameter to receive the caller's already-established session id at pre-authorization time, other than overloading `contextId` — which is documented as an opaque, caller-optional token used only to correlate an async completion, with no guaranteed meaning. This gave `binding-mcp` (and any other caller) a well-defined parameter to pass an existing session id through, instead of repurposing `contextId` for something it isn't meant to carry.

Since the full set of `GuardHandler` implementations lives in this repo, `preauthorize`'s signature is changed directly (not via an additive overload); `TestGuardHandler` and its test call sites are updated to match.

### Test plan

- [x] `./mvnw -pl runtime/engine clean verify` — unit + integration tests pass (aside from a pre-existing, unrelated `TrustedTest.shouldConfigureTrustViaCacerts` environment failure)
- [x] `./mvnw -pl runtime/binding-mcp clean verify` — 68 unit + 251 integration tests pass, coverage checks met

Fixes # (issue)

---
_Generated by [Claude Code](https://claude.ai/code/session_016FBuPeUhS7zhs8ezqB6BnQ)_

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
@@ -2783,7 +2783,7 @@ public final class McpClientFactory implements McpStreamFactory
             if (sessionId == GuardHandler.NEEDS_PREAUTHORIZE && session.binding.needsCredentials)
             {
                 final String preauthorizeUrl =
-                    guard.preauthorize(traceId, session.binding.id, authorization, session.authCallback);
+                    guard.preauthorize(traceId, session.binding.id, initialId, authorization, session.authCallback);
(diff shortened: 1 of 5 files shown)


jfallows (commented):
LGTM

PR #2075, opened 2026-07-09
feat(examples): demonstrate mcp.proxy tool search and eager tool loading

jfallows:
## Description

Updates the `examples/mcp.proxy` example to demonstrate two `mcp(proxy)` cache features: keyword tool search and eager tool loading.

- `etc/zilla.yaml`: adds `options.cache.tools.eager` (policy `explicit`, matching a handful of frequently used tools) and `options.cache.tools.search` (`zilla__search_tools`) on `north_mcp_proxy`. `tools/list` now stays short — a fixed eager set plus the synthesized search tool — while every other tool, most visibly the `everything` reference server's dozen-plus demo tools, becomes "cold": omitted from listings but still discoverable by keyword and directly callable by name.
- `tools-list-client/client.mjs`: renders `tool_reference` search results as `tool_reference:<name>` instead of a blank line; updates the header comment to explain eager/cold behavior.
- `README.md`: new "Search tools by keyword instead of listing them all" section walking through listing tools with the shortened set, searching for a cold tool by keyword, and calling it directly to show cold does not mean inaccessible.
- `.github/test.sh`: three new assertions — a cold tool (`everything__get-sum`) is absent from a fully-authorized `tools/list`, discoverable via `zilla__search_tools`, and still callable directly by name.

The eager-tool match list was checked against the existing scope-based assertions (no-token / partial-scope / full-scope `tools/list` checks) so none of them regress. Actual `@modelcontextprotocol/server-everything` tool names and input schemas (`get-sum`, `echo`, etc.) were confirmed by unpacking the npm package rather than guessed.

Fixes # (n/a — no tracking issue)

## Test plan

- [x] `sh -n .github/test.sh` — script parses
- [x] `node --check tools-list-client/client.mjs` — script parses
- [x] `python3 -c "yaml.safe_load(...)"` — `etc/zilla.yaml` parses
- [ ] CI: `docker compose up -d` + `./.github/test.sh` against the updated config (existing + 3 new assertions)

---
_Generated by [Claude Code](https://claude.ai/code/session_0188UdZgQEPmdHspAmDTxzTi)_

### DIFF
--- a/examples/mcp.proxy/.github/test.sh
+++ b/examples/mcp.proxy/.github/test.sh
@@ -232,49 +232,20 @@ else
 fi
 
 # WHEN: that same caller calls zilla__search_tools for "sum"
-# THEN: the cold everything__get-sum tool comes back as a tool_reference --
-#       proving a tool omitted from tools/list is discoverable by keyword,
-#       not gone. Asserted over raw HTTP (initialize -> notifications/initialized
-#       -> tools/call, reusing the Mcp-Session-Id header) rather than through
-#       tools-list-client's MCP SDK client: "tool_reference" is a Zilla-specific
-#       content type with no entry in the SDK's own CallToolResult content-block
-#       schema (text/image/audio/resource_link/resource only), so the strict
-#       client-side validator rejects it before this script ever sees the
-#       result -- a known limitation of the search feature's response shape,
-#       independent of anything this example configures. This check verifies
-#       what Zilla itself puts on the wire, consistent with every other
-#       assertion in this script
+# THEN: the cold everything__get-sum tool comes back in structuredContent.tools --
+#       proving a tool omitted from tools/list is discoverable by keyword, not gone
 search_cold_tool() {
-  _session_headers=$(mktemp)
-  curl -sS -N --max-time 10 -D "$_session_headers" \
-      -H "Authorization: Bearer $JWT_FULL" \
-      -H "Content-Type: application/json" \
-      -H "Accept: application/json, text/event-stream" \
-      "http://localhost:$PORT/mcp" \
-      -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{},"clientInfo":{"name":"curl","version":"0"}}}' >/dev/null
-  _session_id=$(grep -i "mcp-session-id" "$_session_headers" | tr -d '\r' | cut -d' ' -f2)
-  rm -f "$_session_headers"
-  curl -sS -N --max-time 10 \
-      -H "Authorization: Bearer $JWT_FULL" \
-      -H "Content-Type: application/json" \
-      -H "Accept: application/json, text/event-stream" \
-      -H "Mcp-Session-Id: $_session_id" \
-      "http://localhost:$PORT/mcp" \
-      -d '{"jsonrpc":"2.0","method":"notifications/initialized"}' >/dev/null
-  SEARCH_OUT=$(curl -sS -N --max-time 10 \
-      -H "Authorization: Bearer $JWT_FULL" \
-      -H "Content-Type: application/json" \
-      -H "Accept: application/json, text/event-stream" \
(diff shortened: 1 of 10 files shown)


jfallows:
CI was hanging on the new `zilla__search_tools` assertion. Root-caused and fixed with a follow-up commit:

**1. Hang in `McpToolSearchServer` (fixed)**

`McpToolSearchServer` gated response-readiness on a `contentLength` declared on the stream's `BEGIN` extension. That value comes from `McpServerFactory`'s `paramsLength = contentLength - decodedParamsProgress - 1`, which assumes `params` is the last key in the inbound JSON-RPC object — true for k3po's own scripts, but not for the real MCP TypeScript SDK client, which serializes `jsonrpc`/`id` *after* `params`. That overstates the declared length, so the search server waited forever for bytes that were never coming.

Fixed by dropping reliance on `contentLength` entirely for this stream and using the `DATA` frame's own FIN flag as the completion signal instead (matching how other single-shot streams in this codebase already signal end-of-message). Confirmed against a real `docker compose` stack built from source (`ZILLA_VERSION=develop-SNAPSHOT`) that the real client sends exactly one `DATA` frame with `flags=3` (INIT|FIN) and never an `END` — the k3po specs were updated to `write close` before reading the response so they exercise the same shape.

I didn't touch `McpServerFactory`'s `paramsLength` formula itself — it's shared by three RPC methods and fixing the general case is a bigger, separate change than this example PR warrants.

**2. Empty search results for unguarded tools (fixed)**

Once the hang was gone, searches were returning `content: []` for every query. `McpToolSearchServer.buildResponse()` was calling `filterGuard.verify()` even when a tool had no `roles` entry in `scopesByName` (i.e. no security scheme at all), unlike `McpScopeFilter`'s established convention for the same map ("null roles = always admit, don't call the guard"). Fixed to match that convention.

**3. Known, unfixed limitation — flagging for visibility, not fixing here**

`zilla__search_tools` results use `{"type":"tool_reference","tool_name":"..."}`. That content-block `type` has no entry in the real MCP `CallToolResult` schema (`text`/`image`/`audio`/`resource_link`/`resource` only), so the official `@modelcontextprotocol/sdk` TypeScript client's Zod validation rejects it client-side before application code ever sees the result. This is pre-existing behavior of the search feature itself, not something introduced by the example — I didn't think it was my call to redesign the wire format inside an "update the example" PR, so I left it as-is and instead asserted the example's new search test over raw HTTP (bypassing the strict SDK client) and updated the README's search example to do the same, with a note explaining why. Worth a maintainer decision on what a "tool reference" result should actually look like (`resource_link`? a plain `text` block with the name?) — happy to open a tracking issue if useful.

All 12 `test.sh` assertions pass locally against a from-source build; k3po `McpProxyCacheIT` (44 tests) also green.

---

jfallows:
Follow-up to the third finding above (`tool_reference` isn't a valid MCP content-block type) — fixed rather than just flagged, per your go-ahead.

`zilla__search_tools` now returns matches via `structuredContent` (`{"tools":[{"name":"...","description":"..."}]}`), with a serialized-JSON `text` block alongside for clients that predate `structuredContent` — the pattern the MCP spec itself recommends for exactly this case. `McpProxyCache` retains each tool's `description` from the same `McpToolSearchDocumentScanner` pass already feeding the search index, so no re-parsing of the cached `tools/list` is needed to populate it.

Updated the two `cache.serve.tools.search*` k3po specs to the new response shape (both `runtime/binding-mcp`'s engine-backed IT and `specs/binding-mcp.spec`'s peer-to-peer self-consistency IT pass — 262 and 26 tests respectively, plus the full `binding-mcp` JaCoCo coverage check). Also reverted the mcp.proxy example's `test.sh`/README/`client.mjs` back to calling `zilla__search_tools` through the real `@modelcontextprotocol/sdk` client instead of raw HTTP — no longer necessary now that the response validates cleanly against the SDK's own schema. Verified end-to-end against a from-source build: the SDK client no longer throws, and prints the expected tool name from `structuredContent.tools`.

---

PR #2105, opened 2026-07-10
feat(binding-mcp-openapi): support operator-declared guarded routes

### LINKED ISSUE #2103: binding-mcp-openapi: support operator-declared guarded routes independent of OpenAPI security schemes
## Summary

`mcp_openapi`'s own `routes[]` schema currently disallows `guarded:` (`additionalProperties: false` on the route item alongside `when`/`with`). The only way a tool/resource route ends up guarded today is by resolving the bundled OpenAPI spec's own `security`/`securitySchemes` declarations against `options.specs[label].security` (see `McpOpenapiCompositeGenerator.resolveAlternative`/`injectGuarded`).

That's a real limitation: many OpenAPI specs declare no security schemes at all, so there's currently no way to guard or scope individual `mcp_openapi` tools/resources when the underlying spec doesn't already carry that information — even though the operator may still want to require a guard (JWT, API key, etc.) on specific tools.

## Proposal

Allow `guarded:` directly on `mcp_openapi`'s own `routes[]`, using the same shape every other binding's routes already support (`guarded: [{name, roles}]`, referencing a guard declared in the enclosing `zilla.yaml`). When present, thread it into the corresponding generated `mcp_http0` route's `guarded:` — in addition to (or independent of) whatever the bundled spec's own security requirements resolve to.

```yaml
bindings:
  app0:
    type: mcp_openapi
    kind: client
    options:
      specs:
        github:
          catalog: ...
    routes:
      - when:
          - tool: create_pr
        with:
          spec: github
          operation: create_pr
        guarded:
          - name: jwt0
            roles: [write]
```

## Motivating use case

`mcp_schema_registry` (composite `kind: client` wrapping `mcp_openapi`, see #2068) exposes a fixed set of tools with no OpenAPI security schemes in its bundled spec at all. To let operators guard/scope individual tools (e.g. require a guard on `register_schema`/`delete_subject` but leave read-only tools open), its own composite generator needs to pass an operator-declared route's `guarded:` straight through to the `mcp_openapi0` binding it generates — which requires `mcp_openapi` to accept `guarded:` on its own routes independent of spec-derived security. Once this lands, `mcp_schema_registry` can add its own `routes[].when.tool` (name or glob) + `guarded:` on top of it, resolving guard references to absolute `namespace:name` qnames since the generated composite lives in a different namespace than where the operator declares the guard.

## Notes

- No schema change needed beyond removing the `additionalProperties: false` exclusion and documenting `guarded` on the route item (the property shape itself is the engine's existing generic `$defs/route` `guarded` — nothing new to define).
- Should compose with the existing spec-derived guarding rather than replace it — an operator-declared route `guarded:` and a spec-derived one aren't mutually exclusive concepts.

jfallows:
## Description

Allows `guarded:` directly on `mcp_openapi`'s own `routes[]`, independent of the bundled OpenAPI spec's own `security`/`securitySchemes` declarations. Previously `guarded:` was rejected by `mcp_openapi`'s route schema (`additionalProperties: false` on the route item omitted it), so the only way a tool/resource route could end up guarded was by resolving the spec's `security` requirements against `options.specs[label].security` — which doesn't help when the underlying OpenAPI spec declares no security schemes at all.

When both an operator-declared `guarded:` and a spec-derived guard are present, they compose rather than one replacing the other:
- Same guard referenced by both → roles are unioned into a single entry
- Distinct guards → both entries are emitted on the generated `mcp_http0` route, which the engine already ANDs together

Changes:
- `specs/binding-mcp-openapi.spec` — schema patch now allows a `guarded` property on the route item, matching the engine's generic `$defs/binding.routes.items.properties.guarded` shape
- `runtime/binding-mcp-openapi` — `McpOpenapiRouteConfig` now carries the route's `guarded` list; `McpOpenapiCompositeGenerator` merges it with the existing spec-derived guard resolution before emitting the generated `mcp_http0` route's `guarded:`

## Testing

- Added 3 unit tests to `McpOpenapiCompositeGeneratorTest` covering: operator guard with no spec security, operator guard composing with a distinct spec-derived guard, and operator guard unioning roles into the same spec-derived guard
- `./mvnw clean verify -pl specs/binding-mcp-openapi.spec,runtime/binding-mcp-openapi` — all green (schema tests, unit tests, k3po ITs including the pre-existing spec-derived `proxy.guarded.yaml` scenarios)
- `./mvnw clean install -DskipTests` from the repo root — all modules build successfully (the only unrelated failure is `cloud::docker-image`, which requires a Docker daemon not present in the build environment)

Fixes #2103

---
_Generated by [Claude Code](https://claude.ai/code/session_01D9pVSm36B2zdo5z9KxcBvm)_

### DIFF
--- a/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/McpOpenapiRouteConfig.java
+++ b/runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/McpOpenapiRouteConfig.java
@@ -20,6 +20,7 @@ import java.util.List;
 
 import io.aklivity.zilla.runtime.binding.mcp.openapi.config.McpOpenapiConditionConfig;
 import io.aklivity.zilla.runtime.binding.mcp.openapi.config.McpOpenapiWithConfig;
+import io.aklivity.zilla.runtime.engine.config.GuardedConfig;
 import io.aklivity.zilla.runtime.engine.config.RouteConfig;
 
 public final class McpOpenapiRouteConfig
@@ -27,6 +28,7 @@ public final class McpOpenapiRouteConfig
     public final long id;
     public final List<McpOpenapiConditionConfig> when;
     public final McpOpenapiWithConfig with;
+    public final List<GuardedConfig> guarded;
 
     public McpOpenapiRouteConfig(
         RouteConfig route)
@@ -36,6 +38,7 @@ public final class McpOpenapiRouteConfig
             .map(McpOpenapiConditionConfig.class::cast)
             .collect(toList());
         this.with = (McpOpenapiWithConfig) route.with;
+        this.guarded = route.guarded;
     }
 
     public boolean isBulk()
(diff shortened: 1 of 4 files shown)


jfallows (commented):
LGTM

PR #2115, opened 2026-07-10
feat(engine): support connect aborted zilla:reset.ext matcher

jfallows:
## Description

Wires zilla's k3po test transport into the `connect aborted zilla:reset.ext ...` matcher syntax added by [k3po#20](https://github.com/aklivity/k3po/pull/20) (fixes k3po#17), so scripts can assert the RESET extension bytes sent when a handshake is rejected, not just that the connect aborted.

- `ZillaTarget.Throttle.onResetBeforeHandshake` now captures the RESET extension bytes and writes them into the read-ext buffer, mirroring the existing mid-stream `onReset` handling.
- `AbstractReadExtHandler` now implements k3po driver's `DecodingHandler` marker interface, exposing its internal `ChannelDecoder` so `ConnectAbortedHandler` can reuse it to verify the extension.
- Updated the three `handshake.reset.ext.rejected` client scripts (duplex, half-duplex, simplex) to assert the RESET extension content instead of just `connect aborted`.
- Bumps `k3po.version` to `3.4.5`, which contains k3po#20.

## Test plan

- [x] `./mvnw -pl runtime/engine clean verify -Dit.test=DuplexIT,SimplexIT,HalfDuplexIT` — 0 failures against released k3po 3.4.5, including genuine byte-level verification of the RESET extension (confirmed with a negative control: a deliberately mismatched extension fails with a clear diff)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@
     <byteman.version>4.0.26</byteman.version>
     <jmock.version>2.6.0</jmock.version>
     <mockito.version>5.23.0</mockito.version>
-    <k3po.version>3.4.4</k3po.version>
+    <k3po.version>3.4.5</k3po.version>
     <jmh.version>1.37</jmh.version>
     <helm.version>3.16.4</helm.version>
   </properties>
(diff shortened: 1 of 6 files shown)


jfallows (commented):
LGTM

PR #2136, opened 2026-07-11
feat(command-logs): add zilla logs command for engine event readiness checks

### LINKED ISSUE #2131: Add a `zilla logs` command to tail engine events, for use as a reliable readiness/health check
## Problem

Every example's `compose.yaml` healthcheck (`examples/*/compose.yaml`) uses the same bare TCP-connect probe:

```yaml
healthcheck:
  interval: 5s
  timeout: 3s
  retries: 5
  start_period: 15s
  test: ["CMD", "bash", "-c", "echo -n '' > /dev/tcp/127.0.0.1/7114"]
```

This only proves the kernel accepted a TCP connection on the configured port — it says nothing about whether the engine has actually finished attaching all bindings on all workers. That gap caused an intermittent CI flake in `testing (asyncapi.http.kafka.proxy)` (`curl: (52) Empty reply from server` immediately after Docker Compose reported the `zilla` container `Healthy`) — the healthcheck passed a moment before the HTTP binding pipeline was fully wired up. Seen while investigating aklivity/zilla#2116 / PR #2129 (that PR's fix is unrelated to this flake — confirmed by code path and by `testing (http.kafka.cache)` passing in the same run).

## What already exists

Turns out the engine already has everything needed for a reliable readiness signal — it's just not exposed as a CLI command:

- `EngineEventContext.started()` fires an `engine.started` event, but only from `EngineManager.start()` **after** `watchTask.submit()` synchronously drives the initial config load and `register(namespace)`, which does `boss.attach(namespace).join()` and `workers.stream().map(w -> w.attach(namespace)).reduce(...).join()`. That `.join()` blocks until every worker thread has finished attaching every binding before `register()` (and therefore `events.started()`) fires. So `engine.started` is already guaranteed to mean "all bindings on all workers are live," not just "process launched."
- The read/format plumbing already exists: `EventReader` (`runtime/engine/.../internal/event/io/EventReader.java`), `EngineContext.supplyEventReader()`, and `EngineEventFormatter`/`EventFormatterFactory` for turning raw events into text.
- There's a direct architectural precedent to copy: `runtime/command-metrics/.../ZillaMetricsCommand.java` builds a second, **readonly** `Engine` instance pointed at the same `ENGINE_DIRECTORY` the running engine writes to, and reads its live state out-of-process via `MetricsReader` over the mmap'd metrics buffer.

## Proposal

Add a `runtime/command-logs` module (mirroring `command-metrics`'s structure: module-info, SPI registration, docker-image + `zpm.json.template` wiring) providing a `zilla logs` command that:
- Builds a readonly `Engine` against the same `ENGINE_DIRECTORY` as the running instance (same pattern as `ZillaMetricsCommand`)
- Reads and formats events via `EventReader` + `EngineEventFormatter`
- Supports filtering/waiting for a specific event type (e.g. `zilla logs --grep engine.started` or a dedicated `--wait-for` flag) so it can be used directly as a Docker healthcheck command: `healthcheck.test: ["CMD", "zilla", "logs", "--wait-for", "engine.started"]`

This is preferable to the alternative of using the existing `type: stdout` exporter + `-l`/`--logs` (`ENGINE_VERBOSE_EVENTS`) flag, since that only prints to the engine process's own stdout — a Docker healthcheck command runs as a separate process inside the container and would need that output redirected to a file it can grep, an extra moving part. Reading the mmap'd event log directly avoids that.

## Follow-up

Once available, update the healthcheck in all `examples/*/compose.yaml` files to use `zilla logs` instead of the bare TCP-connect probe.

jfallows:
## Description

Adds `runtime/command-logs`, providing a `zilla logs` CLI command that attaches a readonly `Engine` to a running instance's `ENGINE_DIRECTORY` (mirroring the existing `zilla metrics` command's pattern), reads its event ring buffers, and formats them via the existing `EventFormatter` pipeline.

- `--format text|json` — text (default, human-readable) or newline-delimited JSON, suitable for piping to `jq`
- `-f, --follow` — models `docker logs -f`: without it, prints the current log and exits; with it, keeps printing new events as they arrive
- No `--grep`/`--filter`/`--wait` flags — composes with external `grep`/`jq` instead of reinventing filtering, and a one-shot "did event X already happen" check doesn't need in-process waiting (a Docker `HEALTHCHECK` re-invokes the command on its own interval/retries cadence)
- `ZillaLogsCommandSpi` is a regular (non-`@Incubating`) command, available without `ZILLA_INCUBATOR_ENABLED` — promoted from `incubator/` to `runtime/` once the design settled
- Resolves its default `zilla.properties`/engine directory the same way `zilla start` does (via the `zilla.directory` system property the launcher script sets), rather than a CWD-relative default that never matches inside the Docker image

JSON output is produced with `jakarta.json`'s streaming `JsonGenerator` (not `JsonObjectBuilder`), avoiding an intermediate object tree per event on what can be a `--follow` polling loop. Trace IDs are encoded as zero-padded hex strings (not JSON numbers) to avoid precision loss for unsigned 64-bit values in `jq`/JS consumers.

Also adds `Engine.supplyQName()`/`Engine.supplyEventFormatter()` delegates (mirroring the existing `supplyLocalName()`/`supplyLabelId()`/`supplyEventReader()` pattern) since the public `Engine` facade used by external readonly CLI tools didn't previously expose per-worker qualified-name resolution or event formatting.

Updates the healthcheck in **every** `examples/*/compose.yaml` (39 examples) from the bare TCP-connect probe (`echo -n '' > /dev/tcp/127.0.0.1/<port>`, which only proves the kernel accepted a connection) to `zilla logs | grep -q engine.started`, fully addressing the follow-up called out in #2131. The `ZILLA_INCUBATOR_ENABLED` env var is removed from `asyncapi.http.kafka.proxy`'s compose.yaml since it's no longer needed there (other examples that still require it do so for unrelated incubator bindings, e.g. `binding-amqp`, and are left untouched).

Filed/commented on #1012 to track adding a matching `--format json` (or equivalent) option to the stdout exporter, since this PR establishes the JSON-lines convention for one output path but not the other.

### Example output

```
# --format text (default)
engine:events [13/Jul/2026:23:32:46 +0000] [0000000000000000] engine.started Engine Started.

# --format json
{"namespace":"engine:events","timestamp":1783985566587,"traceId":"0000000000000000","event":"engine.started","message":"Engine Started."}
```

### Testing

- Test-first: `LogsPrinterTest`, `LogsReaderTest` (7 tests) — `LogsReader` is a small extracted class so the event-decode logic is unit-testable with a fake `MessageReader`/`MessageFormatter`, without needing a live `Engine`.
- `./mvnw clean install` passes locally for `command-logs` (compile, checkstyle, license headers, tests, JaCoCo, moditect module-info weaving), and a full-repo `./mvnw validate`/`install` confirms the `incubator` → `runtime` module move didn't break the reactor.
- Verified end-to-end against a real Docker container: rebuilt the image locally, confirmed `zilla help logs`/`zilla logs` work with no `ZILLA_INCUBATOR_ENABLED` set, and exercised both `--format text` and `--format json` (including `--follow`) against a running engine's real `engine.started` event.
- Verified the updated healthcheck end-to-end with `docker compose up` against the local image for a representative example (`tcp.echo`): container reaches Docker's `healthy` state within `start_period`.

Fixes #2131

---
_Generated by [Claude Code](https://claude.ai/code/session_015gEjFBcrpmGxgfvigBdCSa)_


jfallows (commented):
LGTM

PR #2152, opened 2026-07-13
Canonicalize/un-canonicalize request path at server/client stream-factory boundary

### LINKED ISSUE #2145: Canonicalize/un-canonicalize request path and location values at the server/client stream-factory boundary
## Context

Follow-up to #2113 / PR #2128. Stages 1-2 of that work (merged) established:

- `options.specs.<name>.server` is a singular, connection-only override that exists solely on `server`/`client`-kind bindings (`binding-openapi`, `binding-asyncapi`) — never on `proxy`-kind bindings, which hold no real connection.
- `when.servers`/`when.tag` are pure, portable route-selection conditions that match against each spec's own literal declared values (`OpenapiServerView.literalUrl` / `AsyncapiServerView.literalHost`/`literalUrl`), decoupled from whatever deployment override is configured — available uniformly across server, client, and proxy kinds.

This closed the abstraction leak in `options.*` and route-condition config. What remains out of scope for that PR is the deeper, runtime-stream-handler-level leak identified during design discussion:

In the pipeline `openapi server -> openapi-asyncapi proxy -> asyncapi client` (and symmetrically for `openapi client`, and for `asyncapi server`/`client`/`proxy`), when `openapi server` has `options.specs.*.server` configured for a spec with a path prefix, the decoded HTTP request carries the connection-oriented `:scheme`/`:authority`/`:path` (including that path prefix) as an embedded `OpenapiBeginEx` → `HttpBeginEx` extension. `openapi-asyncapi` then needs to route via the generated `http_kafka_proxy0` composite binding, whose `when.path` conditions are baked in path-prefix-aware at generation time. Today this works only because `openapi-asyncapi` (incorrectly, per Stages 1-2) held its own copy of the connecting binding's server config to reconstruct the same path shape — an abstraction leak, since `openapi-asyncapi` already has the canonical `apiId`/`operationId` and shouldn't need to know anything about how the connecting binding chose to serve requests.

## Proposal

Canonicalize the request (strip the deployment-specific server/path shape down to the spec's own canonical form) on the way in at `OpenapiServerFactory`/`AsyncapiServerFactory`, and un-canonicalize (reapply the configured server's shape) on the way out at `OpenapiClientFactory`/`AsyncapiClientFactory`. Everything in between — `openapi-asyncapi` proxy, `asyncapi` proxy (mqtt-kafka/sse-kafka/http-kafka mapping) — would then operate purely in canonical terms (`apiId`/`operationId`), with no need to reference `options.specs.*.server` at all, closing the leak completely.

This is real-time stream-handler work (not just config generation), materially riskier than Stages 1-2, and depends on Stages 1-2 already being in place (a clean, singular `server` override to canonicalize against). It was deferred from #2128 for that reason.

## Scope

- `OpenapiServerFactory` / `OpenapiClientFactory` (`runtime/binding-openapi`)
- `AsyncapiServerFactory` / `AsyncapiClientFactory` (`runtime/binding-asyncapi`)
- Verify `openapi-asyncapi`'s generated `http_kafka_proxy0` routing and `asyncapi` proxy's mqtt-kafka/sse-kafka/http-kafka mapping no longer need any server-shaped config once canonicalization is in place.

jfallows:
## Description

Fixes #2145.

Closes the abstraction leak identified in #2145: the deployment-specific
connection shape (`:scheme`/`:authority`/path prefix) no longer needs to
leak past the server/client stream-factory boundary for either `openapi` or
`asyncapi`.

**`binding-openapi`** (`OpenapiServerFactory`/`OpenapiClientFactory`):
canonicalizes the decoded `HttpBeginEx` down to the spec's own literal,
unresolved operation path template on the way in, and un-canonicalizes
(reapplies the resolved deployment `:scheme`/`:authority`/path prefix) on
the way out. `options.specs.*.server` became a plural, relocatable
`servers[]` list with round-robin client selection; `when.servers`/`when.tag`
route conditions decoupled from the deployment override, matching the
spec's own literal declared values instead.

**`binding-asyncapi`** (`AsyncapiServerFactory`/`AsyncapiClientFactory`):
mirrors the same canonicalize/un-canonicalize pattern for HTTP and SSE
(MQTT carries no non-relocatable server info in its `BeginEx` and needs no
change — verified against the IDL and left untouched, protected by the
existing byte-for-byte MQTT scripts). Also fixes a related, previously
latent bug: `AsyncapiServerGenerator` parsed a spec's declared
`servers.*.pathname` but never combined it with `channel.address` when
generating HTTP/SSE request paths and routes, so a spec-compliant client
prefixing requests with that pathname would 404 — fixed test-first
alongside the canonicalization work.

Verified per the issue's scope item: `openapi-asyncapi`'s generated
`http_kafka_proxy0` routing and `asyncapi`'s own proxy (mqtt-kafka/sse-kafka/
http-kafka) route generation and runtime dispatch already key exclusively on
canonical `channel.address`/operation data, never on the deployment
`server`/`servers` config — no changes were needed there, confirming the
abstraction leak is fully closed.

## Test plan

- [x] `./mvnw -pl runtime/binding-openapi clean verify` — all unit tests + k3po ITs green
- [x] `./mvnw -pl runtime/binding-asyncapi clean verify` — all unit tests + k3po ITs green (18 ITs)
- [x] `./mvnw -pl runtime/binding-openapi-asyncapi clean verify` — green except one pre-existing,
      unrelated failure (`OpenapiAsyncapiIT.shouldCreatePetWithServerPathPrefix`), confirmed via
      `git stash` to reproduce identically on the pre-branch baseline
- [x] `./mvnw checkstyle:check` / `license:check` clean on all touched modules
- [x] New test-first k3po fixtures proving: HTTP/SSE canonicalize+un-canonicalize round-trip,
      pathname-combination bug fix, MQTT no-op (existing byte-for-byte scripts unchanged)

---
_Generated by [Claude Code](https://claude.ai/code/session_01HzJgppmUNpSfzpJbzgxf8k)_


jfallows (commented):
LGTM

PR #2164, opened 2026-07-15
fix(manager): preserve uses clauses from real modules merged into the zpm delegate (#2162)

jfallows:
## Description

Backport of #2162 to `support/1.x`.

`zpm install` folds every automatic (non-modular) dependency into a synthetic `io.aklivity.zilla.manager.delegate` module, and recursively drags in any of that dependency's own dependencies too — including real, correctly-modularized ones. `ZpmInstall.generateDelegate()` derives the delegate's `uses` clauses only from `provides ... with ...;` pairs it can resolve within the delegate itself. `jdeps --generate-module-info` never proactively detects arbitrary `ServiceLoader.load()` call sites, so it emits zero bare `uses` clauses on its own.

The practical effect: when a real module with its own authored `uses` declaration gets merged into the delegate via this recursive process, that declaration is silently dropped. Any code from that module that calls `ServiceLoader.load()` for the first time then throws an uncaught `ServiceConfigurationError`, which crashes the whole engine.

### Fix

Preserve the `uses` declarations of every real (non-automatic) module folded into the delegate — read via `ModuleFinder.of(path).findAll()` over each constituent artifact path already tracked on the delegate `ZpmModule`, filtered to non-automatic descriptors, unioned with whatever `uses` clauses were already derived from `provides` matching.

`support/1.x`'s `generateDelegate()` predates the optional-dependency-tree/strict-jdeps-fallback refactor present on `develop`, so a verbatim cherry-pick of the `develop` commit conflicts on unrelated surrounding context. This PR applies the same fix logic directly onto this branch's own simpler flow instead.

### Verification

New regression test `ZpmInstallTest.shouldPreserveUsesClauseWhenMergingModuleIntoDelegate` uses a small, fully self-contained two-artifact fixture (an automatic module depending on a real module that declares `uses`) resolved from a local `file:` repository under `src/conf/install-delegate-uses/`. This avoids depending on any real-world package's actual transitive dependency graph — empirically, this branch's `ZpmCache` does not perform Maven-style version conflict resolution across separately-requested coordinates for the same artifact (confirmed directly: a `jose4j` + explicit `slf4j-api` fixture resolved both `slf4j-api:1.7.36` and `slf4j-api:2.0.17` as distinct, unmerged graph nodes, with only the transitively-requested 1.7.36 — non-modular — actually reaching the delegate).

Confirmed the new test fails (`uses` empty) against the pre-fix code and passes after. Full `manager` module `verify` (unit tests, checkstyle, license headers, JaCoCo) passes clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_011LfmguC5M5mwdLJ64Dxruk)_


jfallows (commented):
LGTM

Issue #2182, opened 2026-07-21
guard-identity: rename to guard-inline, add Guard.aliases() for backwards-compatible identity alias


jfallows:
## Problem

`guard-identity` (`type: identity`) is a minimal guard that authorizes any session unconditionally and vends inline-configured `identity`/`credentials` option values -- there's no external validation, no token, no wire protocol. The type name `identity` overloads with the guard SPI's own `identity(sessionId)` concept and the `${guarded['name'].identity}` template placeholder used across `binding-http`, `binding-mqtt`, `kafka-proxy`, etc. `type: identity` reads as "the guard that validates identity claims," when what it actually is is "a guard whose config is inlined into `options` rather than externally sourced." `inline` names what the guard does; `identity` doesn't.

## Scope -- rename on `develop`

- Rename the module/package/classes: `runtime/guard-identity` -> `runtime/guard-inline`, `specs/guard-identity.spec` -> `specs/guard-inline.spec`, package `io.aklivity.zilla.runtime.guard.identity` -> `io.aklivity.zilla.runtime.guard.inline`, `IdentityGuard`/`IdentityGuardHandler`/`IdentityGuardContext`/`IdentityGuardFactorySpi`/`IdentityOptionsConfig*` -> `Inline*`. `IdentityGuard.NAME = "identity"` (`IdentityGuard.java:36`) becomes `InlineGuard.NAME = "inline"`.

## Backwards compatibility -- `Guard.aliases()` via `Aliasable`

`Vault` and `Catalog` already support multiple type names per plugin instance via `io.aklivity.zilla.runtime.engine.factory.Aliasable` (`interface Vault extends Aliasable`, `Vault.java:39`) -- a mixin with a single default method `Set<String> aliases()`, defaulting to empty. `Guard` (`io.aklivity.zilla.runtime.engine.guard.Guard`) does not currently extend it.

- Add `Guard extends Aliasable`.
- `EngineWorker.doInit()` currently builds `guardsByType` via a plain `guards.stream().collect(toMap(Guard::name, ...))` (`EngineWorker.java:995-996`) -- not alias-aware. Rewrite it to loop like the existing `vaultsByType`/`catalogsByType` construction (`EngineWorker.java:998-1024`), registering both `guard.name()` and every `guard.aliases()` entry against the same `GuardContext` instance.
- `InlineGuardFactorySpi` overrides `aliases()` to return `Set.of("identity")`, so `type: identity` keeps resolving to the same guard.
[...]

PR #2198, opened 2026-07-21
docs(examples): update http.kafka.oneway.oauthbearer to use inline guard

jfallows:
## Description

`guard-identity` was renamed to `guard-inline` in #2182 / #2195. `type: identity` still resolves via the backwards-compatible alias added in that PR, but the `http.kafka.oneway.oauthbearer` example was the only example in the repo still configured with the old type name — update it to `type: inline` so examples demonstrate the current primary type.

No functional change; the guard's `credentials`-passthrough behavior is identical either way.

## Test plan

- [x] Confirmed no other `examples/` config in the repo references `type: identity` (only this one)
- [x] `type: inline` is the renamed guard's primary type (validated against the schema in #2195); no other changes needed to this example

---
_Generated by [Claude Code](https://claude.ai/code/session_01AeEyCoZoGvqWfCVovcHdzT)_

### DIFF
--- a/examples/http.kafka.oneway.oauthbearer/etc/zilla.yaml
+++ b/examples/http.kafka.oneway.oauthbearer/etc/zilla.yaml
@@ -2,7 +2,7 @@
 name: example
 guards:
   authn_kafka:
-    type: identity
+    type: inline
     options:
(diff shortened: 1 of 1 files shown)


jfallows (commented):
LGTM

PR #2211, opened 2026-07-23
binding-kafka: generic per-API request/response envelope with ApiVersions negotiation

### LINKED ISSUE #2121: binding-mcp-kafka: topic administration (create_topics, delete_topics)
## Summary

Sub-issue of #1671. Adds admin tools for topic lifecycle management:

- `create_topics` → `KafkaRequestBeginEx/createTopics` + `KafkaResponseBeginEx/createTopics`
- `delete_topics` → `KafkaRequestBeginEx/deleteTopics` + `KafkaResponseBeginEx/deleteTopics`

See input/output schemas in the &#34;Topic management&#34; section of #1671.

## Dependencies

None — `createTopics`/`deleteTopics` already exist in `binding-kafka`. Ready to build.

## Notes

Mutating/admin-risk tools — keep this PR scoped narrowly to these two tools so the diff is reviewable on its own, separate from the read-only topic-metadata tools (tracked in a separate sub-issue since those need a `binding-kafka` prerequisite).

jfallows:
Fixes #2121

## Description

Adds `KafkaClientApiFactory`, a generic Kafka client binding that forwards an arbitrary `api`/`version`/`clientId` request/response pair over one shared connection via `KafkaApiRequestBeginEx`/`KafkaApiResponseBeginEx`, instead of requiring a dedicated per-API factory (`KafkaClientCreateTopicsFactory`, `KafkaClientDeleteTopicsFactory`, etc.) for every Kafka API a caller wants to reach.

Built out over several commits:

- **Generic envelope** — `KafkaApiStream`/`KafkaApiClient` forward raw request/response bytes for any `api`/`version`, proven against `createTopics`/`deleteTopics` v3.
- **Implicit ApiVersions negotiation** — before sending a caller's first real request on a connection, transparently probe the broker (ApiVersions v0), cache the supported version range per API key, and proactively reject out-of-range requests via `RESET` instead of forwarding a request the broker can't handle.
- **Explicit ApiVersions short-circuit** — a caller can drive the ApiVersions probe itself through the same generic envelope; when it does, the hidden implicit probe is skipped and the real response is forwarded back instead of discarded.
- **Reactive `UNSUPPORTED_VERSION` recovery** — real Kafka brokers reply to a request whose version they don't support with just `correlationId + errorCode=35` (the request header is deliberately version-stable across all APIs so this degraded reply is always parseable). Detected generically on any response and translated into a `RESET`, invalidating the cached version range so the next request re-negotiates. This is now the *sole* invalidation trigger for the cached range — a disconnect/reconnect no longer clears it (see below), since a reconnect carries no information about whether the broker's supported versions actually changed.
- **Cache-preserving reconnect, consistently** — `apiVersionRangeByApiKey` survives every reconnect regardless of whether the range was learned via an implicit probe or a caller-driven explicit one; only a real `UNSUPPORTED_VERSION` response invalidates it.
- **Narrow reconnect on network disconnect** — only the request actually in flight when a connection drops gets `RESET`; everything else still queued survives and resumes once the connection reconnects, instead of failing the whole pending queue.
- **Progressive reconnect backoff** — reconnect attempts use `Signaler`-scheduled exponential backoff with a configurable ceiling (`client.reconnect`, default 5s), mirroring the existing pattern in `KafkaCacheServerDescribeFactory`, instead of looping tightly against a genuinely-dead broker.
- **`client.api.versions` config toggle** (default `true`) — disables ApiVersions negotiation entirely for a plain passthrough, useful for other APIs' fixtures that don't need to model the ApiVersions exchange.
- **`binding-mcp-kafka` `create_topics`/`delete_topics` tools** — built directly on the generic envelope above, closing #2121: handles the CHALLENGE/FLUSH re-negotiation round trip, allocates request/response buffers from pooled slots (no per-call allocation), and exposes a configurable request timeout (default 30s).

This started as infrastructure work in `binding-kafka`; the generic envelope is now consumed by `binding-mcp-kafka`'s `create_topics`/`delete_topics` tools, and is intended to let further bindings reach arbitrary Kafka APIs generically rather than needing a bespoke factory per API.

Test-first throughout: every scenario has a runtime IT (`KafkaApiIT`) against the live engine, plus peer-to-peer network/application IT pairs (`KafkaCreateTopicsIT`, `KafkaDeleteTopicsIT`, application `KafkaApiIT`) verifying the `.rpt` scripts are mutually consistent. All new ITs use the `Kafka*` prefix to distinguish them from the legacy per-API-factory ITs during the migration.

## Test plan

- [x] `./mvnw clean verify -pl specs/binding-kafka.spec` — 434 tests, 0 failures
- [x] `./mvnw clean verify -pl runtime/binding-kafka` — 392 tests, 0 failures
- [x] `./mvnw checkstyle:check -pl specs/binding-kafka.spec,runtime/binding-kafka` — 0 violations
- [x] Legacy `CreateTopicsIT`/`DeleteTopicsIT` (unrelated per-API factories, shared scripts) unaffected
- [x] CI fully green — all 42 checks passing on the latest commit, including the full `Build (25)` reactor build and CodeQL
- [x] All review feedback addressed — 17/17 review threads resolved (fixes applied, one intentionally deferred as out-of-scope follow-up, noted inline)

## Status

Ready for review/merge — CI green, no merge conflicts, all outstanding review threads resolved.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ShEikNDTdSoYi5i5JqVDxN)_


examples/mcp.proxy/etc/zilla.yaml
```diff
@@ -608,8 +609,18 @@ bindings:
...
+    # -- so its route matches on tool name alone, gated by its own
+    # tool-specific kafka:write scope layered under the toolkit-level
+    # kafka:tools guard above (the same layering register_schema demonstrates
+    # for the kafka_sr toolkit)
     routes:
+      - when:
+          - tool: create_topics
+        guarded:
+          authn_jwt:
+            - kafka:write
```
jfallows:
```suggestion
        guarded:
          authn_jwt:
            - kafka:admin
```

examples/mcp.proxy/etc/zilla.yaml
```diff
@@ -608,8 +609,18 @@ bindings:
...
       - when:
           - tool: produce
             topics: [orders]
```
jfallows, line 632:
```suggestion
      - when:
          - tool: produce
            topics: [orders]
        guarded:
          authn_jwt:
            - kafka:write
```

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/api/KafkaCreateTopicsResponsePipeline.java
```diff
@@ -0,0 +1,130 @@
...
+                topic.replicationFactor(),
+                configs));
+        }
+
+        final CreateTopicsResponsePart2FW responsePart2 = createTopicsResponsePart2RO.wrap(buffer, progress, limit);
+        progress = responsePart2.limit();
+
+        assert progress <= limit;
+
+        return new Response(throttleTimeMillis, topics);
```
jfallows: The pipeline design should be closer to ModelPipeline or Json/Avro/Protobuf Pipeline, where events are delivered by the generator to indicate decode progress, pulled into the generator by the terminating Sink and supporting a Transform in between to change the timing an content of the events.

This is allocating an intermediary representation on the hot path.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientApiFactory.java
```diff
@@ -0,0 +1,2582 @@
...
+
+        if (resolved != null)
+        {
+            final long resolvedId = resolved.id;
+            final short api = kafkaApiRequestBeginEx.api();
+            final short version = kafkaApiRequestBeginEx.version();
+            final int requestLength = kafkaApiRequestBeginEx.length();
+            final String16FW clientId = kafkaApiRequestBeginEx.clientId().length() != -1
+                ? new String16FW(kafkaApiRequestBeginEx.clientId().asString())
+                : null;
```
jfallows:
```suggestion
            final String16FW clientId  = new String16FW(kafkaApiBeginEx.clientId().asString());
```
Does this work unconditionally?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientApiFactory.java
```diff
@@ -0,0 +1,2582 @@
...
+        final long initialId = begin.streamId();
+        final long affinity = begin.affinity();
+        final long authorization = begin.authorization();
+        final OctetsFW extension = begin.extension();
+        final ExtensionFW beginEx = extensionRO.tryWrap(extension.buffer(), extension.offset(), extension.limit());
+        final KafkaBeginExFW kafkaBeginEx = beginEx != null && beginEx.typeId() == kafkaTypeId ?
+                kafkaBeginExRO.tryWrap(extension.buffer(), extension.offset(), extension.limit()) : null;
+
+        assert kafkaBeginEx.kind() == KafkaBeginExFW.KIND_API_REQUEST;
+        final KafkaApiRequestBeginExFW kafkaApiRequestBeginEx = kafkaBeginEx.apiRequest();
```
jfallows:
```suggestion
        final KafkaApiRequestBeginExFW kafkaApiBeginEx = kafkaBeginEx.apiRequest();
```

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientApiFactory.java
```diff
@@ -0,0 +1,2582 @@
...
+                replyMax = 0;
+
+                nextCorrelationId = 0;
+                responseBytesRemaining = 0;
+                apiVersionKeysRemaining = 0;
+                saslMechanismsRemaining = 0;
+
+                attached = false;
+                requestInFlight = false;
+                apiVersionsRequestExplicit = false;
```
jfallows: Review these to see if they can be inferred by a combination of `state` and `requests` count.

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientApiFactory.java
```diff
@@ -0,0 +1,2582 @@
...
+            requestInFlight = false;
+
+            final KafkaApiStream head = pending.poll();
+            if (head != null)
+            {
+                if (errorCode == ERROR_UNSUPPORTED_VERSION)
+                {
+                    event.apiVersionRejected(traceId, routedId, head.api, head.version);
+                    apiVersionsResolved = false;
+                    apiVersionRangeByApiKey.clear();
```
jfallows: Is `apiVersionsResolved` inferred by `!apiVersionRangeByApiKey.isEmpty()` ?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientCreateTopicsFactory.java
```diff
@@ -535,10 +545,42 @@ private int decodeCreateTopicsResponse(
...
+                if (topicPart2 == null)
+                {
+                    client.decoder = decodeIgnoreAll;
+                    break decode;
+                }
+
+                progress = topicPart2.limit();
+
                 responseTopics.add(new CreateTopicsResponseInfo(
                     topic.name().asString(), topic.error(), topic.message().asString()));
```
jfallows: This is allocating on the hot path.

runtime/binding-mcp-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mcp/kafka/internal/McpKafkaConfiguration.java
```diff
@@ -30,13 +31,18 @@ public class McpKafkaConfiguration extends Configuration
 
     public static final PropertyDef<SessionIdSupplier> MCP_KAFKA_SESSION_ID;
     public static final PropertyDef<String> MCP_KAFKA_CACHE_CLIENT_EXIT;
+    public static final PropertyDef<String> MCP_KAFKA_DIRECT_CLIENT_EXIT;
```
jfallows:
```suggestion
public static final PropertyDef<String> MCP_KAFKA_CACHE_CLIENT_EXIT;
    public static final PropertyDef<String> MCP_KAFKA_CLIENT_EXIT;
```

runtime/binding-mcp-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mcp/kafka/internal/McpKafkaConfiguration.java
```diff
@@ -30,13 +31,18 @@ public class McpKafkaConfiguration extends Configuration
...
+    public static final PropertyDef<Duration> MCP_KAFKA_REQUEST_TIMEOUT;
 
     static
     {
         final ConfigurationDef config = new ConfigurationDef("zilla.binding.mcp.kafka");
         MCP_KAFKA_SESSION_ID = config.property(SessionIdSupplier.class, "session.id",
             McpKafkaConfiguration::decodeSessionIdSupplier, McpKafkaConfiguration::defaultSessionIdSupplier);
         MCP_KAFKA_CACHE_CLIENT_EXIT = config.property("cache.client.exit", "");
+        MCP_KAFKA_DIRECT_CLIENT_EXIT = config.property("direct.client.exit", "");
```
jfallows:
```suggestion
MCP_KAFKA_CACHE_CLIENT_EXIT = config.property("cache.client.exit");
        MCP_KAFKA_DIRECT_CLIENT_EXIT = config.property("direct.client.exit");
```

Leave default as `null`, avoid explicitly setting blank to indicate unset.

runtime/binding-mcp-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mcp/kafka/internal/stream/McpKafkaClientFactory.java
```diff
@@ -36,7 +36,7 @@ public McpKafkaClientFactory(
     {
         super(config, context);
         this.context = context;
-        this.generator = new McpKafkaClientGenerator(config.cacheClientExit());
+        this.generator = new McpKafkaClientGenerator(config.cacheClientExit(), config.directClientExit());
```
jfallows: Why do we need to override these via config?

runtime/binding-mcp-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mcp/kafka/internal/McpKafkaConfigurationTest.java
```diff
@@ -53,6 +57,24 @@ public void shouldDecodeSessionIdSupplierFromMethodReference() throws Exception
...
+    @Test
+    public void shouldOverrideRequestTimeout() throws Exception
+    {
+        final Properties properties = new Properties();
+        properties.setProperty(MCP_KAFKA_REQUEST_TIMEOUT_NAME, "PT10S");
+        final McpKafkaConfiguration config = new McpKafkaConfiguration(new Configuration(properties));
+
+        assertEquals(Duration.ofSeconds(10), config.requestTimeout());
+    }
+
```
jfallows: We don't need to retest generic configuration property functionality, this will get picked up by the ITs implicitly.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/network/create.topics.v7.sasl.handshake.v1/create.topics.sasl.oauthbearer/client.rpt
```diff
@@ -58,41 +58,62 @@ read 20                                 # size
...
-write 144                               # size
+write 125                               # size
       19s                               # create topics
-      3s                                # v3
+      7s                                # v7
       ${newRequestId}
       5s "zilla"                        # client id
-      2                                 # topic count
-        6s "events"                     # name
+      [0x00]                            # tagged fields
```
jfallows: use ${kafka:var...} functions instead for readability

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/network/create.topics.v7.sasl.handshake.v1/create.topics.sasl.oauthbearer/client.rpt
```diff
@@ -58,41 +58,62 @@ read 20                                 # size
...
+      [0x03]                            # topic count
```
jfallows: use ${kafka:var...} functions instead for readability

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/network/create.topics.v7.sasl.handshake.v1/create.topics.sasl.oauthbearer/client.rpt
```diff
@@ -58,41 +58,62 @@ read 20                                 # size
...
+        [0x07] "events"                 # name
```
jfallows: use ${kafka:varstring(...)} function instead for readability

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/api/create.topics.v7.reconnect.sticky/client.rpt
```diff
@@ -0,0 +1,146 @@
...
+    option zilla:window 8192
+    option zilla:transmission "half-duplex"
+    option zilla:byteorder "network"
+    option zilla:affinity 0
+
+write zilla:begin.ext ${kafka:beginEx()
+                               .typeId(zilla:id("kafka"))
+                               .apiRequest()
+                                 .length(110)
+                                 .api(19)
```
jfallows: Let's have .api("create_topics") mapping internally to 19 (similar for other Kafka wire protocol APIs) so that the scripts are more readable.

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/api/create.topics.v7.reconnect.no.probe/client.rpt
jfallows, line 1: Explain the "sticky" scenario and how it it not already covered by other scenarios.

jfallows (commented):
We seem to be missing input and output tool schemas for create_topics tool.
Note: produce and consume cannot be fully specific on schema because they are sending and receiving messages, but create_topics etc are structurally defined by the mco-kafka binding.

jfallows:
`testing (mcp.proxy)` failed on the latest push, but not because of this PR's `create_topics`/`delete_topics` work — both passed cleanly against the real Kafka broker in that same run:

```
✅ kafka__create_topics created a new topic on the real Kafka broker
✅ kafka__delete_topics deleted the topic from the real Kafka broker
```

The job's overall failure is from an unrelated, pre-existing scenario in the same script — `kafka_sr__describe_subject` (Karapace schema registry toolkit, untouched by this PR) got no result content back across all 5 retries:

```
❌ kafka_sr__describe_subject did not list the registered version
```

`register_schema` and `list_subjects` immediately before it succeeded, so this looks like an eventual-consistency race against Karapace rather than a real regression. Leaving it as-is since it's out of scope for this PR — flagging here so it doesn't get mistaken for a `create_topics`/`delete_topics` regression.

Run: https://github.com/aklivity/zilla/actions/runs/30568894377/job/90966764948

---

specs/binding-kafka.spec/src/main/scripts/io/aklivity/zilla/specs/binding/kafka/streams/application/api/create.topics.v7.reconnect.no.probe/client.rpt
jfallows: Renamed `reconnect.sticky` → `reconnect.no.probe` — asserts what the client actually does (no ApiVersions re-probe on reconnect once the range is cached), rather than naming an internal flag.

This also surfaced a real inconsistency the rename alone wouldn't have fixed: `apiVersionRangeByApiKey` survival across a reconnect was gated on `apiVersionsExplicit` (whether the caller's *first* request was itself an explicit ApiVersions call vs. Zilla's implicit probe). A disconnect carries no information about whether the broker's supported versions changed, so that gate had no correctness basis — it was an artifact of how the sticky cache was originally scoped in, not a real distinction. Removed `apiVersionsExplicit` entirely; the cache now survives every reconnect unconditionally regardless of origin, and the existing reactive `UNSUPPORTED_VERSION` handling remains the sole (and correct) invalidation trigger, since it's the only point where a real response actually contradicts the cached assumption.

Also fixed the scenario itself to exercise two requests of the *same* API key across the reconnect boundary instead of mixing `create_topics`/`delete_topics`, so it isolates the behavior under test.

---

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaClientCreateTopicsFactory.java
jfallows: Deferring this — `KafkaClientCreateTopicsFactory` is slated to be replaced by the generic `KafkaClientApiFactory` pipeline in a follow-up, so reworking its allocation profile now would be throwaway effort. Marking out of scope for this PR; tracking separately.

---

PR #2215, opened 2026-07-29
Add AsyncAPI channel-binding SPI, writer, and typed document builders

akrambek:
## Summary

Extends `common-asyncapi` in three connected pieces, driven by zilla-platform's need to both read AsyncAPI 3.0 specs (with Kafka channel bindings) and programmatically generate them:

- Kafka channel-binding SPI + `asyncapi.3.0.1.schema.json` support for kafka bindings 0.5.0 (`cd3b22f67a`)
- `AsyncapiWriter` to serialize the `Asyncapi` model back to YAML/JSON text, with insertion-order preservation and vendor-extension flattening to sibling `x-*` keys (`52b2da6a35`, `2393cdcacf`), plus an explicit `JsonProvider` pin so `Jsonb` instances stop resolving nondeterministically off classpath order (`aadcbdfcc4`)
- Typed builders for the AsyncAPI document model (`Asyncapi`, `AsyncapiInfo`, `AsyncapiServer`, `AsyncapiChannel`, `AsyncapiOperation`, `AsyncapiComponents`, `AsyncapiMessage`, `AsyncapiMultiFormatSchema`) so consumers building a doc don't have to hand-assemble every node via `new X(); x.field = y;` (`5910e17444`)

The builders follow the existing `ConfigBuilder<T,B>` idiom from `config/*.conf` (e.g. `OtlpEndpointConfigBuilder`) — CRTP self-type, nested builders composed via a parent setter method-reference, keyed variants for the `Map<String,X>` fields adapting the list-append pattern (`NamespaceConfigBuilder.binding()`) to map-insert-by-name. The CRTP base is kept local to `common-asyncapi` rather than depending on `config/engine.conf`, avoiding a new `runtime`→`config` cross-group dependency that nothing else in the reactor has today.

Scoped to the 8 model classes actually exercised by a real write path (zilla-platform's Kafka-metadata-to-AsyncAPI generator) — no builders added for classes never manually constructed (parameters, security schemes, correlation ids, traits, replies, tags); same pattern extends cleanly if that changes.

## Test plan

- [x] `./mvnw -pl runtime/common-asyncapi -am test` — all green, including new `AsyncapiBuilderTest`, `AsyncapiWriterTest`, `AsyncapiFragmentTest`
- [x] `./mvnw -pl runtime/common-asyncapi -am install -DskipTests` — installs cleanly
- [x] Downstream consumer (zilla-platform's `KafkaAsyncapiGenerator`) refactored onto the new builders; its golden-fixture tests pass with zero output diffs, confirming the builders are behavior-preserving vs hand construction

### DIFF
--- a/runtime/common-asyncapi/src/main/java/io/aklivity/zilla/runtime/common/asyncapi/config/AsyncapiFragment.java
+++ b/runtime/common-asyncapi/src/main/java/io/aklivity/zilla/runtime/common/asyncapi/config/AsyncapiFragment.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2021-2026 Aklivity Inc
+ *
+ * Licensed under the Aklivity Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ *   https://www.aklivity.io/aklivity-community-license/
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package io.aklivity.zilla.runtime.common.asyncapi.config;
+
+import java.io.StringReader;
+
+import jakarta.json.JsonReader;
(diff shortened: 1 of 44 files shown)


jfallows (approved):
Builders look great!

PR #2241, opened 2026-07-31
fix(binding-tcp): enforce per-route exit requirement for kind: server

### LINKED ISSUE #2239: binding-tcp: kind: server's per-route exit requirement doesn't actually enforce anything
## Summary

`binding-tcp`'s schema patch (`specs/binding-tcp.spec/.../schema/tcp.schema.patch.json`) requires `kind: server` to declare `exit` either top-level or per-route:

```json
{
    "properties": { "kind": { "const": "server" } },
    "anyOf":
    [
        { "required": [ "exit" ] },
        {
            "properties": { "routes": { "required": [ "exit" ] } },
            "required": [ "routes" ]
        }
    ]
}
```

The second `anyOf` branch — the per-route alternative — looks like it doesn't actually enforce anything. `"required"` only constrains **object** instances (per JSON Schema semantics, object-only keywords like `required`/`properties`/`additionalProperties`/`patternProperties` are vacuously satisfied against instances of any other type). `routes` is declared `"type": "array"` elsewhere in the same schema, so `"routes": { "required": [ "exit" ] } }` is checking `required: ["exit"]` against an *array* instance, which trivially passes regardless of content. It needed an `"items"` wrapper to actually reach each route object: `"routes": { "items": { "required": [ "exit" ] } } }`.

Net effect: only the top-level `"required": ["exit"]` branch does real work. A `kind: server` binding that declares `routes:` with per-route `exit:` on each item, but no top-level `exit:`, should be valid (that's clearly the intent — the whole point of the `anyOf` is to allow either style) — but a `kind: server` binding with `routes:` and **no** `exit` anywhere (top-level or per-route) also currently passes schema validation, when it should be rejected. Any route lacking a resolvable `exit` string will resolve to `route.id = 0` at engine-attach time (`EngineManager.NameResolver.resolve` returns `0L` for a `null` name, no error) — silently broken routing instead of a clear startup validation error.

## Repro / how to confirm

Add a `SchemaTest` fixture: `kind: server` with `routes: [{ when: [...] }]` (no `exit` anywhere, top-level or per-route) and assert it's currently accepted (it shouldn't be). Also confirm the *intended* valid case — `routes` with per-route `exit:` and no top-level `exit:` — is separately still accepted after the fix (it already is today, coincidentally, since the broken branch's vacuous truth means the `anyOf` always succeeds via that branch; after fixing the nesting, it needs to keep succeeding via a real check that each route actually has `exit`).

## Suggested fix

```json
{
    "properties": { "routes": { "items": { "required": [ "exit" ] } } },
    "required": [ "routes" ]
}
```

## Context

Found incidentally while designing `binding-mcp-http`'s own analogous `kind: proxy` exit requirement (#2238), which was going to mirror this exact pattern — flagging here instead of silently copying the same mistake. Not fixing as part of #2238; this is `binding-tcp`'s own pre-existing schema, independent of the mcp-http work.

jfallows:
## Description

`binding-tcp`'s schema patch requires `kind: server` to declare `exit` either top-level or per-route, via an `anyOf` with two branches. The per-route branch checked `"routes": { "required": ["exit"] }`, but per JSON Schema semantics `required` only constrains object instances — `routes` is a `type: array`, so the check was vacuously true against any array and never actually inspected route contents.

Net effect: a `kind: server` binding with `routes:` and no `exit` anywhere (top-level or per-route) passed schema validation, when it should be rejected. Such a route resolves to `route.id = 0` at engine-attach time (`EngineManager.NameResolver.resolve` returns `0L` for a `null` name), silently breaking routing instead of failing at startup with a clear error.

Fix: wrap the per-route check in `items` so it actually reaches each route object: `"routes": { "items": { "required": ["exit"] } }`.

Added `SchemaTest` fixtures:
- `server.routes.exit.yaml` — routes with per-route `exit`, no top-level `exit` — confirms the intended valid case still passes.
- `server.routes.exit.missing.invalid.yaml` — routes with no `exit` anywhere — confirms it is now rejected. Verified this test fails for the right reason against the pre-fix schema.

Fixes #2239

## Test plan

- [x] `./mvnw -pl specs/binding-tcp.spec test -Dtest=SchemaTest` — all 15 cases pass
- [x] Confirmed the new negative test fails against the pre-fix schema (reverted the schema change locally, re-ran, saw the expected `JsonException` was not thrown)
- [x] `./mvnw -pl specs/binding-tcp.spec license:check checkstyle:check` passes

---
_Generated by [Claude Code](https://claude.ai/code/session_01EZSeQpusk33eebeZGi9Mnz)_

### DIFF
--- a/specs/binding-tcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tcp/config/server.routes.empty.invalid.yaml
+++ b/specs/binding-tcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tcp/config/server.routes.empty.invalid.yaml
@@ -1,0 +1,26 @@
+#
+# Copyright 2021-2026 Aklivity Inc.
+#
+# Aklivity licenses this file to you under the Apache License,
+# version 2.0 (the "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at:
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+---
+name: test
+bindings:
+    net0:
+        type: tcp
+        kind: server
+        options:
+            host: 0.0.0.0
+            port: 12345
+        routes: []
--- a/specs/binding-tcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tcp/config/server.routes.exit.missing.invalid.yaml
+++ b/specs/binding-tcp.spec/src/main/scripts/io/aklivity/zilla/specs/binding/tcp/config/server.routes.exit.missing.invalid.yaml
@@ -1,0 +1,28 @@
+#
+# Copyright 2021-2026 Aklivity Inc.
+#
+# Aklivity licenses this file to you under the Apache License,
+# version 2.0 (the "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at:
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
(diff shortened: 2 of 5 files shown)


jfallows:
Pushed a follow-up commit: added `minItems: 1` to the per-route `exit` branch. Without it, `routes: []` (empty array) with no top-level `exit` still passed vacuously, since `items` has nothing to iterate over an empty array. Now the requirement is exactly: top-level `exit`, OR `routes` with at least one item and `exit` on every item. New `SchemaTest` case (`server.routes.empty.invalid.yaml`) confirms this is now rejected, and I verified it fails without `minItems` first.

---

PR #2243, opened 2026-07-31
binding-mcp-http: add kind: client and require exit for kind: proxy

### LINKED ISSUE #2238: binding-mcp-http: add kind: client (implicit sys:http_client exit) and require exit for kind: proxy
## Summary

`mcp_http` currently supports only `kind: proxy`, and its own schema never validates `exit` at all (neither top-level nor per-route). Two related changes:

1. Add a `kind: client` variant that disallows `exit` in schema and implicitly routes every request to the real, engine-provided `sys:http_client` binding at runtime — no operator-declared `exit` needed.
2. Fix a silent gap in the existing `kind: proxy`: `exit` should be required (top-level or per-route), not left optional.

## Background

This came out of discussion on #2104/#2231/#2237: `mcp_openapi`'s own composite generator always builds an inner `mcp_http0` with `kind: proxy`, defaulting its exit to `sys:http_client` via the `zilla.binding.mcp.openapi.http.client.exit` Configuration property when no override is supplied. That's a workable pattern for `mcp_openapi` because it *has* a composite generator to vary the exit in. `binding-mcp-http` itself has no such generator — it's the leaf/terminal binding — so for `mcp_http` to offer the same "self-contained, real HTTP" vs. "operator-declared exit" split that `mcp_openapi`/`mcp_schema_registry` have, the default has to be filled in at the binding's own config-construction time, not by a generator.

### `kind: client`

- Schema: `kind` enum widened to `["client", "proxy"]`. `client` branch sets `"exit": false` both top-level and per-route (`routes[].items.properties.exit: false`) — disallowed entirely, since the exit is implicit.
- Runtime: `McpHttpBindingContext.attach()`/`detach()` widen their `kind == PROXY` gates to also accept `CLIENT`, delegating to the same `McpHttpProxyFactory` (already kind-agnostic internally, same relay behavior either way — mirrors `McpSchemaRegistryClientFactory`/`McpSchemaRegistryBindingContext`'s existing client+proxy dual-kind pattern from #2231).
- Mechanism (traced and confirmed against `runtime/engine`):
  - `EngineManager.java:452-456` resolves **every** route's exit generically for every binding: `route.id = resolver.resolve(route.exit)`, before any binding-specific code runs.
  - `RouteConfig.id` is `public transient long id` — mutable, not `final`.
  - `binding.resolveId = resolver::resolve` (`EngineManager.java:413`) — the exact same resolver, exposed to every binding's config wrapper, and it already handles qualified `"namespace:name"` strings.
  - So in `McpHttpBindingConfig`'s constructor, before it maps `binding.routes` into `McpHttpRouteConfig` (which captures `route.id` at construction time), add: when `binding.kind == KindConfig.CLIENT`, overwrite every route's `route.id = binding.resolveId.applyAsLong("sys:http_client")` — a hardcoded literal, no new Configuration property, directly mirroring `binding-openapi`'s `OpenapiClientGenerator`/`binding-asyncapi`'s `AsyncapiClientGenerator`, which already hardcode `exit("sys:tcp_client")` with no configuration knob.
- `"sys:http_client"` is confirmed to be a real, always-present engine construct, not an operator-authored convention: `EngineManager.buildSystemNamespaceIfNecessary()` assembles the built-in `sys` namespace at startup from `system()` patches contributed by installed `Binding`/`Exporter` SPIs — `binding-http` contributes `sys:http_client` (routing `https`→`tls_client`/`http`→`tcp_client`), `binding-tls` contributes `sys:tls_client`, `binding-tcp` contributes `sys:tcp_client` (the real socket connect). `examples/mcp.proxy/etc/zilla.yaml` already references `exit: sys:http_client` directly with no such namespace authored anywhere in that file.
- `McpHttpBinding` itself needs no change — it doesn't override `originType`/`routedType` (defaults to accepting any kind, like `TcpBinding`).

### `kind: proxy` — require `exit`

Traced `EngineManager.NameResolver.resolve(String name)` (`EngineManager.java:612-628`): when `name` (i.e. `route.exit`) is `null`, it returns `0L` with no error — no exception, no config-time validation failure. That `0L` becomes `route.id`, which `McpHttpProxyFactory` uses directly as the routed exit target. So today, a `kind: proxy` `mcp_http` binding with routes but no `exit` anywhere silently gets a dead/invalid routing target instead of a clear startup error.

Fix: require `exit`, top-level or per-route, mirroring `binding-tcp`'s `kind: server` pattern — but written correctly. `binding-tcp`'s own per-route branch (`"routes": {"required": ["exit"]}`) looks like a latent bug: `"required"` only constrains object instances, and `routes` is `type: array`, so that check is vacuously true and doesn't actually enforce anything (only tcp's top-level `"required": ["exit"]` branch does real work). `mcp_http`'s version should nest correctly: `"routes": {"items": {"required": ["exit"]}}`, so it actually validates each route object.

(Not proposing to touch `binding-tcp`'s existing schema in this issue — separate, pre-existing, out of scope here — just flagging it so mcp-http's version doesn't copy the same mistake.)

## Notes

- Independent of `kind: proxy`'s existing runtime behavior otherwise — no change for existing valid configs that already declare `exit` somewhere.
- Needs test coverage: `SchemaTest` cases for the valid/invalid `kind: client`/`kind: proxy` combinations (including the newly-required `exit` for `proxy`), a unit test on `McpHttpBindingConfig` covering the `route.id` resolution for `kind: client` (config-POJO logic, not a stream-handler state machine, so a mocked-`EngineContext` unit test is appropriate here per `runtime/AGENTS.md`), and ideally a real end-to-end IT exercising `kind: client` against the genuine `sys:http_client` pipeline (would need `binding-http`/`binding-tls`/`binding-tcp` as real — not test-scope — dependencies of `runtime/binding-mcp-http`, mirroring how `binding-openapi` depends on them and its `OpenapiClientIT` does a real network round-trip with no exit override needed).

jfallows:
## Description

Adds a `kind: client` variant to `binding-mcp-http` that implicitly routes every request to the engine-provided `sys:http_client` binding — no operator-declared `exit` needed — mirroring the self-contained "real HTTP" pattern already used by `mcp_openapi`/`mcp_schema_registry`. Also fixes a schema gap where `kind: proxy` allowed `exit` to be omitted entirely, silently producing an unresolved (`0`) routing target at runtime.

### Runtime (`runtime/binding-mcp-http`)
- `McpHttpBindingConfig`: for `kind: client`, every route's `id` is overwritten to resolve to the configured client-exit target before routes are wrapped into `McpHttpRouteConfig` (which captures `id` at construction time).
- `McpHttpBindingContext`: `attach`/`detach` widened to accept both `PROXY` and `CLIENT` kinds, delegating to the same kind-agnostic `McpHttpProxyFactory` — mirroring `McpSchemaRegistryClientFactory`/`McpSchemaRegistryBindingContext`'s existing client+proxy dual-kind pattern.
- Added a `zilla.binding.mcp.http.client.exit` `Configuration` property (default `sys:http_client`), so the implicit client exit target is overridable for testing — mirroring `zilla.binding.mcp.openapi.http.client.exit` in `binding-mcp-openapi`.

### Schema (`specs/binding-mcp-http.spec`)
- `kind` enum widened to `["client", "proxy"]`.
- `kind: proxy` now requires `exit` on any route that also declares `with` (the routes that actually resolve to an upstream target) — correctly nested under `routes.items.if/then`, unlike `binding-tcp`'s vacuous `routes: {required: [exit]}` version, which doesn't constrain anything since `routes` is an array. Routes that only carry `guarded` (layering-only, no `with`) are unaffected, since they never resolve to a routed exit.
- `kind: client` disallows `exit` entirely, top-level and per-route (`exit: false`), since the exit is implicit.

### Tests (test-first)
- Unit tests in `McpHttpBindingConfigTest` covering the `kind: client` route-id resolution (default and overridden `client.exit`), and that `kind: proxy` routes are left untouched.
- `McpHttpConfigurationTest` covering the new `client.exit` property default and override.
- `SchemaTest` + new YAML fixtures for valid `kind: client` config and invalid combinations (client with exit, client route with exit, proxy route with `with` but no exit).
- Full `binding-mcp-http` k3po IT suite (62 tests) and spec `SchemaTest` (18 tests) verified green via `mvn verify`.

Fixes #2238

---
_Generated by [Claude Code](https://claude.ai/code/session_01CKueUWZbnpZ6ABHUVn7xkV)_

### DIFF
--- a/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/McpHttpBindingContext.java
+++ b/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/McpHttpBindingContext.java
@@ -38,7 +38,7 @@ final class McpHttpBindingContext implements BindingContext
     {
         BindingHandler handler = null;
 
-        if (binding.kind == KindConfig.PROXY)
+        if (binding.kind == KindConfig.PROXY || binding.kind == KindConfig.CLIENT)
         {
             factory.attach(binding);
             handler = factory;
@@ -51,7 +51,7 @@ final class McpHttpBindingContext implements BindingContext
     public void detach(
         BindingConfig binding)
     {
-        if (binding.kind == KindConfig.PROXY)
+        if (binding.kind == KindConfig.PROXY || binding.kind == KindConfig.CLIENT)
         {
             factory.detach(binding.id);
         }
--- a/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/McpHttpConfiguration.java
+++ b/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/McpHttpConfiguration.java
@@ -32,6 +32,7 @@ public class McpHttpConfiguration extends Configuration
 
     public static final PropertyDef<SessionIdSupplier> MCP_HTTP_SESSION_ID;
     public static final IntPropertyDef MCP_HTTP_SESSION_ID_ATTEMPTS;
+    public static final PropertyDef<String> MCP_HTTP_CLIENT_EXIT;
 
     static
     {
@@ -40,6 +41,7 @@ public class McpHttpConfiguration extends Configuration
             McpHttpConfiguration::decodeSessionIdSupplier, McpHttpConfiguration::defaultSessionIdSupplier);
         MCP_HTTP_SESSION_ID_ATTEMPTS = config.property("session.id.attempts",
             McpHttpConfiguration::defaultSessionIdAttempts);
+        MCP_HTTP_CLIENT_EXIT = config.property("client.exit", "sys:http_client");
         MCP_HTTP_CONFIG = config;
     }
 
@@ -62,6 +64,11 @@ public class McpHttpConfiguration extends Configuration
     public int sessionIdAttempts()
     {
         return MCP_HTTP_SESSION_ID_ATTEMPTS.getAsInt(this);
+    }
+
+    public String clientExit()
+    {
+        return MCP_HTTP_CLIENT_EXIT.get(this);
     }
 
     @FunctionalInterface
--- a/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/config/McpHttpBindingConfig.java
+++ b/runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/config/McpHttpBindingConfig.java
@@ -55,7 +55,6 @@ public final class McpHttpBindingConfig
 {
     private static final String CREDENTIALS_TOKEN = "{credentials}";
     private static final String IDENTITY_TOKEN = "{identity}";
-    private static final String SYS_HTTP_CLIENT = "sys:http_client";
 
     public final long id;
     public final McpHttpOptionsConfig options;
@@ -79,7 +78,8 @@ public final class McpHttpBindingConfig
 
     public McpHttpBindingConfig(
         BindingConfig binding,
-        EngineContext context)
+        EngineContext context,
+        String clientExit)
     {
         this.id = binding.id;
         this.context = context;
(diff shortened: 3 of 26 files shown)


jfallows (commented):
LGTM

PR #2269, opened 2026-08-01
refactor(binding-mcp,binding-mcp-kafka,binding-mcp-openapi,binding-mcp-schema-registry): unify route condition naming

### LINKED ISSUE #2253: mcp bindings: unify route condition naming to singular tool/prompt/resource, with glob + array support where safe
## Summary

The mcp-family binding route conditions (`mcp`, `mcp_kafka`, `mcp_http`, `mcp_openapi`, `mcp_schema_registry`) currently disagree on naming and shape for the same concept:

- Base `mcp` binding (`kind: proxy`/`client`): `tools`/`prompts`/`resources` — all plural, all `List<String>` glob-pattern allow-sets (`McpConditionConfig`)
- `mcp_kafka`, `mcp_http`, `mcp_openapi`, `mcp_schema_registry`: `tool`/`resource` — singular `String`, exact match only, no glob

An operator writing one `zilla.yaml` with both a base `mcp` proxy in front and typed bindings behind it has to remember which syntax group each binding belongs to. We're pre-GA on all of these, so this is a clean-break rename + shape unification, not a deprecate-and-migrate — same approach already used in #2047 when the redundant `capability` condition field was removed from `McpConditionConfig` ("None of this has shipped yet, so clean it up").

## Proposed shape

Singular field name everywhere: `tool:`, `prompt:`, `resource:`. Type is `string | string[]` (shorthand-permissive, same pattern as `exit`) where safe; glob support only for `tool`, never for `resource`/`prompt`.

### `tool:`

| Binding | Shape | Glob | Reason |
|---|---|---|---|
| `mcp` (base) | `string \| string[]` | yes | Rename only (`tools:` → `tool:`) — already supports this today |
| `mcp_kafka` | `string \| string[]` | yes | Tool routes are commonly `with:`-free; no ambiguity |
| `mcp_schema_registry` | `string \| string[]` | yes | Its `with.operation` is auto-derived internally from a fixed, hardcoded tool list, not from the operator's route value — safe despite generating `mcp_openapi` routes underneath |
| `mcp_openapi` | `string` only | no | Every functional route pairs 1:1 with an operator-authored `with.operation` |
| `mcp_http` | `string` only | no | Every functional route pairs 1:1 with an operator-authored `with.headers`/`:path` template |

**Future-proofing rule** (relevant mainly to `mcp_kafka`, if/when it adds a tool-specific `with:` later): a route may pair a tool-varying `with:` field **only when `tool` is the bare-string form** — not the single-element array form. Even though `tool: [x]` and `tool: x` match the same single tool today, only the string form is eligible once a tool-varying `with:` exists; the array form is reserved for the "no `with:`, grouping" case regardless of how many elements it has. This is a deliberate simplification over "exactly one non-glob array element counts too" — one rule, no length-counting, and it lets an author's choice of `string` vs `array` signal intent rather than relying on incidental array length.

### `prompt:`

Base `mcp` binding only (`prompts:` → `prompt:`, `string | string[]`, glob-capable, mirroring `tool:`). Trivial — none of the four typed proxies expose MCP prompts today, so there's nothing to reconcile.

### `resource:`

No glob support anywhere (mcp-kafka's URI-template capture, e.g. `kafka://{topic}/messages` → `${params.topic}`, already does more than a glob would; the rest just need exact matching).

| Binding | Shape | Reason |
|---|---|---|
| `mcp` (base) | `string \| string[]` | Resource routing is `with:`-free at the aggregator/client layer |
| `mcp_kafka` | `string` only | Resource match pairs with a captured-value `with.topic: ${params.topic}` — same coupling problem `mcp_openapi`/`mcp_http` have for tools |
| `mcp_openapi` | `string` only | Same coupling — operator-authored `with:` per matched resource |
| `mcp_http` | `string` only | Same coupling |

Note the split here has different binding membership than `tool:` — `mcp_kafka` is unrestricted for tools but restricted for resources, because its resource routes (unlike most of its tool routes) carry a captured-value `with:` binding. `mcp_schema_registry` has no resource concept — unaffected either way.

`mcp_kafka`'s resource condition has no shipped implementation yet (tracked separately in #2123, part of the #1671 tracking issue) — this design is for when that lands, not a retrofit.

## Explicitly deferred, not part of this change

Whether base-`mcp`'s own resource matching should gain URI-template capture (to reach parity with `mcp_kafka`'s `{topic}` capture) is a separate, harder design question, noted here so it isn't conflated with this rename/shape-alignment work.

## Affected modules

- `config/binding-mcp.conf` — `McpConditionConfig` (`tools`/`prompts`/`resources` → `tool`/`prompt`/`resource`)
- `config/binding-mcp-kafka.conf` — `McpKafkaConditionConfig` (`tool: String` → `tool: string | string[]`, glob-capable)
- `config/binding-mcp-schema-registry.conf` — `McpSchemaRegistryConditionConfig` (`tool: String` → `tool: string | string[]`, glob-capable)
- `config/binding-mcp-http.conf`, `config/binding-mcp-openapi.conf` — no schema change; confirming these intentionally stay as-is
- Matching runtime matcher classes (`McpConditionMatcher`, `McpKafkaConditionMatcher`) and JSON schema patches for the three changing bindings
- Existing YAML fixtures across `specs/binding-mcp.spec`, `specs/binding-mcp-kafka.spec`, `specs/binding-mcp-schema-registry.spec` need renaming; new fixtures for array/glob coverage

## Related

- #2047 / #2046 — established the per-tool guarded-route pattern for `kind: client` and the precedent for a clean-break schema cleanup pre-GA
- #2123 — `mcp_kafka`'s `kafka://{topic}/messages` resource, not yet implemented; this design applies once it lands

jfallows:
## Description

Unifies the mcp-family binding route condition naming/shape, per the design in #2253:

- **`mcp` (base)**: `tools`/`prompts`/`resources` → `tool`/`prompt`/`resource`, now accepting a bare string as shorthand for a single-element array (in addition to the existing glob-capable array form)
- **`mcp_kafka`**: `tool` widened from exact-match `String` to glob-capable `string | string[]`
- **`mcp_schema_registry`**: `tool` widened the same way; its composite generator's existing glob-matching logic now iterates the new list shape instead of a single string
- **`mcp_openapi`**: removes the leftover `capability` condition field — a pre-GA cleanup that should have landed alongside #2047/#2046 when the equivalent field was removed from the base `mcp` binding, but was missed for this binding. Verified via the composite generator's actual logic that `capability` was a redundant route-level gate that could never do more than admit-or-reject an already-uniformly tool-shaped bulk match (no bulk resource generation is actually implemented), so removing it changes no observable behavior — routing is now driven purely by presence of `tool`/`resource`, mirroring the base `mcp` binding's existing pattern
- **`mcp_http`**: unchanged, as scoped by the issue

Also extracts two small shared utilities used by the above, to stop the glob-compile and `string | string[]` adapter logic from being copy-pasted per binding:
- `Matchers` (`runtime/common-lang`) — glob-pattern compilation and null-means-unrestricted allow-list matching
- `JsonStrings` (`runtime/common-json`) — reads/writes a `string | string[]` shorthand-permissive JSON property as a `List<String>`

All existing spec fixtures were renamed in place; new fixtures were added covering the bare-string shorthand and array/glob forms for each binding.

## Test plan

- [x] `./mvnw test` — unit tests pass for all touched modules (`common-lang`, `common-json`, all four `*.conf` modules, all four runtime binding modules)
- [x] `./mvnw verify` — k3po integration tests pass for `specs/binding-mcp.spec`, `specs/binding-mcp-kafka.spec`, `specs/binding-mcp-openapi.spec`, `specs/binding-mcp-schema-registry.spec`, and their corresponding `runtime/binding-mcp*` modules
- [x] Full reactor `./mvnw clean install -DskipTests` succeeds end-to-end (only the final `docker-image` module fails, due to no Docker daemon being available in the sandbox this was built in — unrelated to this change)

Fixes #2253

---

A companion docs update is at aklivity/zilla-docs#(pending) to keep the `mcp` binding reference pages in sync.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dn8Qvoir4yDrwyuXTucLS4)_

### DIFF
--- a/config/binding-mcp-kafka.conf/NOTICE
+++ b/config/binding-mcp-kafka.conf/NOTICE
@@ -15,7 +15,9 @@ This project includes:
   JSON-B API under Eclipse Public License 2.0 or GNU General Public License, version 2 with the GNU Classpath Exception
   zilla::config::binding-kafka.conf under Aklivity Community License Agreement
   zilla::config::engine.conf under Aklivity Community License Agreement
+  zilla::runtime::common-agrona under The Apache Software License, Version 2.0
   zilla::runtime::common-feature under The Apache Software License, Version 2.0
+  zilla::runtime::common-json under Aklivity Community License Agreement
   zilla::runtime::common-lang under Aklivity Community License Agreement
 
 
--- a/config/binding-mcp-kafka.conf/pom.xml
+++ b/config/binding-mcp-kafka.conf/pom.xml
@@ -62,7 +62,6 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>common-json</artifactId>
       <version>${project.version}</version>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/config/binding-mcp-kafka.conf/src/main/java/io/aklivity/zilla/config/binding/mcp/kafka/McpKafkaConditionConfig.java
+++ b/config/binding-mcp-kafka.conf/src/main/java/io/aklivity/zilla/config/binding/mcp/kafka/McpKafkaConditionConfig.java
@@ -21,7 +21,7 @@ import io.aklivity.zilla.config.engine.ConditionConfig;
 
 public final class McpKafkaConditionConfig extends ConditionConfig
 {
-    public final String tool;
+    public final List<String> tool;
     public final String resource;
     public final List<String> topics;
 
@@ -37,7 +37,7 @@ public final class McpKafkaConditionConfig extends ConditionConfig
     }
 
     McpKafkaConditionConfig(
-        String tool,
+        List<String> tool,
         String resource,
         List<String> topics)
     {
--- a/config/binding-mcp-kafka.conf/src/main/java/io/aklivity/zilla/config/binding/mcp/kafka/McpKafkaConditionConfigBuilder.java
+++ b/config/binding-mcp-kafka.conf/src/main/java/io/aklivity/zilla/config/binding/mcp/kafka/McpKafkaConditionConfigBuilder.java
@@ -24,7 +24,7 @@ public final class McpKafkaConditionConfigBuilder<T> extends ConfigBuilder<T, Mc
 {
     private final Function<ConditionConfig, T> mapper;
 
-    private String tool;
+    private List<String> tool;
     private String resource;
     private List<String> topics;
 
@@ -42,7 +42,7 @@ public final class McpKafkaConditionConfigBuilder<T> extends ConfigBuilder<T, Mc
     }
 
     public McpKafkaConditionConfigBuilder<T> tool(
-        String tool)
+        List<String> tool)
     {
         this.tool = tool;
         return this;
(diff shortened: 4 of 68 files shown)


jfallows (commented):
LGTM

Issue #2260, opened 2026-08-01
mcp proxy: disambiguate tool `title` across toolkits


jfallows:
## Problem

The `mcp` proxy binding namespaces proxied tool **names** with a `toolkit__` prefix (`McpToolNames.effectiveName`, delimiter `__`) so that names never collide across aggregated toolkits. However, when a proxied tool also declares a `title`, MCP clients typically display the `title` in place of the raw (prefixed) name. Titles are free-text, everyday-language phrases (e.g. "Send Message", "Create Issue") and are not namespaced at all, so two toolkits that each expose a similarly-titled tool reintroduce the exact duplicate-display problem the `toolkit__` name prefix was meant to solve.

## Proposed direction

Discussed a couple of approaches:

1. **Always suffix every title** with a toolkit disambiguator, e.g. `Send Message (toolkit)`. Simple and predictable, but:
   - `(toolkit)` as a literal string doesn't say *which* toolkit, so it doesn't actually disambiguate two colliding titles from different toolkits.
   - Applies noisy suffixing even in the common case where no collision exists (e.g. a single-toolkit deployment).

2. **Collision-only disambiguation** (leaning toward this): compute the title-collision set across all tools aggregated for a given proxy/session, and only for titles that actually collide, append a suffix naming the *specific* toolkit, e.g. `Send Message (GitHub)` / `Send Message (Slack)` — using the same toolkit identifier already used for the `toolkit__` name prefix, so name-space and title-space stay consistent.
   - Keeps titles clean in the common (no-collision) case.
   - Tradeoff: a tool's displayed title becomes dependent on what else is aggregated alongside it in a given proxy config — it isn't a fixed, standalone property of the tool anymore. Clients that cache titles across sessions could see a title change if the aggregate tool set changes.

## Next steps

Discuss which approach (or a variant) to adopt before implementation. Once agreed, this should land in `runtime/binding-mcp` (likely near `McpToolNames`/`McpProxyItemFactory`), following test-first discipline (spec/unit tests first) per `AGENTS.md`.

PR #2271, opened 2026-08-01
fix(binding-mcp): fall back to annotations.title for tool title collisions

### LINKED ISSUE #2270: binding-mcp-http/mcp-openapi: tool title nested in annotations, but McpToolTitleDisambiguator reads it top-level
## Problem

`McpHttpProxyFactory.buildToolsList()` (`runtime/binding-mcp-http/.../stream/McpHttpProxyFactory.java`, `annotationsObject()`) emits a configured tool `title` nested inside the `annotations` object:

```json
{"name":"some_tool","annotations":{"title":"Some Tool","readOnlyHint":true}}
```

This matches the operator-facing config schema (`options.tools[].annotations.title`, see `McpHttpToolAnnotationsConfig`) and predates #2260/#2262.

`McpToolTitleDisambiguator` (added in #2262, `runtime/binding-mcp/.../stream/cache/McpToolTitleDisambiguator.java`) does collision-only title disambiguation across aggregated toolkits, but reads `title` as a **top-level field** on the tool object (`item.getString(TITLE, null)`) — confirmed by its own test fixture using `{"name":"send_message","title":"Send Message"}` with no `annotations` wrapper at all.

Net effect: any tool exposed through `binding-mcp-http` or `binding-mcp-openapi` (including `binding-mcp-schema-registry`, which is generated on top of `mcp_openapi`) that sets a `title` will never have that title participate in cross-toolkit collision disambiguation, silently. Two aggregated toolkits with colliding `annotations.title` values will both display the same title with no `(toolkit)` suffix.

## Where

- `runtime/binding-mcp-http/src/main/java/io/aklivity/zilla/runtime/binding/mcp/http/internal/stream/McpHttpProxyFactory.java` (`buildToolsList`, `annotationsObject`)
- `runtime/binding-mcp-openapi/src/main/java/io/aklivity/zilla/runtime/binding/mcp/openapi/internal/config/composite/McpOpenapiCompositeGenerator.java` (would need to thread a title through from tool config / OpenAPI operation, similar to how `description` is derived — see the `description`/`summary` fallback chain around line 439)
- `config/binding-mcp-http.conf/src/main/java/io/aklivity/zilla/config/binding/mcp/http/McpHttpToolAnnotationsConfig.java` (existing `annotations.title` config field)
- `runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpToolTitleDisambiguator.java` (the top-level reader)

## Open questions to settle before implementing

- Should `buildToolsList()` start emitting a **top-level** `title` (in addition to or instead of `annotations.title`)? Removing `annotations.title` is a wire-format change for existing `binding-mcp-http`/`binding-mcp-openapi` operators; keeping both means redundant bytes on every `tools/list` response.
- If `binding-mcp-schema-registry`'s 9 fixed tools get titles, where do they come from — an author-controlled override in `McpSchemaRegistryCompositeGenerator`/`McpOpenapiCompositeGenerator`, since the bundled OpenAPI operations have no natural "title" field (unlike `description`/`summary`)?

## Context

Split out from #2256 (`binding-mcp-kafka`/`binding-mcp-schema-registry` description + title work) — `binding-mcp-kafka` builds its own tools/list JSON directly and got a self-contained top-level `title` fix there. This issue covers the broader, shared `binding-mcp-http`/`binding-mcp-openapi` path, which affects every `mcp_http`/`mcp_openapi` consumer, not just schema-registry, so it's tracked separately for its own discussion.

Refs #2256, #2260, #2262

jfallows:
## Description

Follow-up to #2262 / #2260. `McpToolTitleDisambiguator` previously only checked the top-level `title` field. Per the MCP spec's own documented display precedence — `title`, then `annotations.title`, then `name` (schema/2025-11-25/schema.ts, `Tool.annotations` doc comment: "Display name precedence order is: title, annotations.title, then name.") — a tool with no top-level `title` can still have a client-visible display label via `annotations.title`, which was previously invisible to collision detection entirely.

This adds `annotations.title` as a fallback:
- The effective title used for collision detection is `title` when present, else `annotations.title`.
- When disambiguating, the field actually rewritten is whichever one held the effective title — top-level `title` when present, `annotations.title` only when the tool has no top-level title.
- Behavior is otherwise unchanged: collision-only, cache path only (`options.cache.tools`), toolkit-less routes never contribute to or receive disambiguation.

## Testing

- New unit tests in `McpToolTitleDisambiguatorTest`: fallback to `annotations.title` when top-level `title` is absent, preference for top-level `title` over `annotations.title` when both are present, cross-field collision (one tool's top-level `title` colliding with another's `annotations.title` fallback), and items with neither field left unchanged.
- `./mvnw clean verify -pl runtime/binding-mcp` passes in full (165 unit tests, 285 integration tests, checkstyle, license headers, JaCoCo coverage).

Fixes #2270
Refs #2260

---
_Generated by [Claude Code](https://claude.ai/code/session_01GigP5EY5KusjbJASWMhm7k)_

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpToolTitleDisambiguator.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/cache/McpToolTitleDisambiguator.java
@@ -40,12 +40,16 @@ import jakarta.json.JsonWriter;
 // distinct toolkits. A route with no toolkit configured has no identifier to disambiguate with, so
 // its items never contribute to collision detection and are never themselves rewritten -- a title
 // shared with only one other toolkit-identified route (or with a toolkit-less route) is left alone.
+// Per the MCP spec's own display precedence order (title, then annotations.title, then name), the
+// effective title checked for a collision -- and the one rewritten -- is the top-level "title" when
+// present, falling back to "annotations.title" only when the tool has no top-level title at all.
 // This runs once per cache hydration cycle (not on the request hot path), so the jakarta.json object
 // model is used freely, mirroring McpProxyCacheHydrater's own per-item JSON rewrites (nameFirst,
 // injectToolScopes).
 final class McpToolTitleDisambiguator
 {
     private static final String TITLE = "title";
+    private static final String ANNOTATIONS = "annotations";
 
     private McpToolTitleDisambiguator()
     {
@@ -75,7 +79,7 @@ final class McpToolTitleDisambiguator
             {
                 for (JsonObject item : parsed)
                 {
-                    final String title = item.getString(TITLE, null);
+                    final String title = effectiveTitle(item);
                     if (title != null)
                     {
                         toolkitsByTitle.computeIfAbsent(title, t -> new LinkedHashSet<>()).add(toolkit);
@@ -120,7 +124,7 @@ final class McpToolTitleDisambiguator
 
         for (JsonObject item : items)
         {
-            final String title = item.getString(TITLE, null);
+            final String title = effectiveTitle(item);
             if (title != null && collidingTitles.contains(title))
             {
                 rewritten.add(withDisambiguatedTitle(item, title, toolkit));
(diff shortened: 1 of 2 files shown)


jfallows:
The `testing (asyncapi.http.kafka.proxy)` check failure is unrelated to this PR — it's a transient timing flake in the example's round-trip test (POST succeeded, the immediate follow-up GET saw an empty body, likely Kafka consumer lag), not a binding-mcp issue. The same job passed on `develop`'s latest run (same base commit `b2097b7e`) moments earlier. I don't have permission to re-run the failed job via the API (403) — a maintainer re-run should clear it. Will keep watching.

---

PR #2286, opened 2026-08-02
fix(binding-mcp-kafka): fix create_topics schema gaps, drop timeout arg

### LINKED ISSUE #2285: mcp-kafka: tool schema completeness gaps (create_topics sub-schemas, timeout availability)
Same audit as the field-naming normalization issue surfaced two gaps that need an actual design decision rather than a pure rename:

**1. `create_topics.assignments` has no `items` sub-schema**
The `assignments` array property in `create_topics`'s input schema declares no `items` schema at all, so it's effectively untyped/unvalidated. Needs a decision on what shape a manual partition-assignment entry should take (likely `{partition, replicas: [broker_id...]}` or similar, matching Kafka's `CreateTopics` assignment semantics) before this can be filled in.

**2. `create_topics.configs` has no `additionalProperties` sub-schema**
Unlike `alter_configs.configs`, which declares an `additionalProperties` sub-schema for its config-value map, `create_topics.configs` doesn't. Should be brought in line with `alter_configs` once the value type/shape is agreed (probably `string`, matching Kafka topic-config semantics, but worth confirming rather than assuming).

**3. `timeout` present only on `create_topics`/`delete_topics`**
No other tool (`alter_configs`, `produce`/`produce_message`, `consume`/`consume_messages`, etc.) exposes a caller-supplied timeout. Needs a decision: should `timeout` be a general-purpose argument across all tools that make a blocking Kafka call, or is it deliberately scoped to just the two topic-admin operations? Not filing a fix until that's decided.

**Considered and not included:** the `create_topics` (rich object items) vs `delete_topics` (plain string items) request-item shape asymmetry was also flagged during the audit, but this looks intentional — `create_topics` needs per-topic config/partition detail that a delete doesn't. Noting it here so it doesn't get rediscovered as a "bug" later, but no action proposed.

jfallows:
## Description

Addresses all three gaps flagged in #2285:

**1 & 2 — `create_topics`'s untyped `assignments`/`configs`:**
- `assignments` items now match the shape the runtime parser (`McpKafkaToolCreateTopicsSource`) already expects and the `create.topics` k3po fixture already exercises: `{partition: integer, brokers: integer[]}`, both required.
- `configs` now declares `additionalProperties: {"type": "string"}`, matching `alter_configs.configs`.

**3 — `timeout` only on `create_topics`/`delete_topics`:**
Rather than generalizing `timeout` to every blocking tool (a much larger change touching ~12 more schemas/parsers), it's removed from `create_topics`/`delete_topics` for consistency with every other tool, none of which exposes a per-call override. The underlying Kafka request still uses `zilla.binding.mcp.kafka.request.timeout` (default `PT30S`) via `Source.timeoutMs()`, just no longer caller-configurable. Whether a general-purpose timeout argument is worth adding later is tracked separately in #2290.

The `tools.list` / `tools.list.security.schemes` k3po fixtures (which assert the exact `inputSchema` JSON literally) are updated to match each change — confirmed they failed against the old schema before touching `McpKafkaProxyFactory` or the tool sources.

Fixes #2285.

---
_Generated by [Claude Code](https://claude.ai/code/session_0153dSFGyFL2tpxZJGBRvvhC)_


jfallows:
The `testing (mcp.proxy)` failure is unrelated to this diff. The actual failing assertion is `kafka_sr__get_compatibility` (Karapace schema-registry compatibility lookup) never returning `Compatibility level is FULL` across 10 retries (~30s) — a timing flake in the Kafka Connect/Karapace part of that example's test suite, nowhere near `binding-mcp-kafka`'s `create_topics` schema this PR touches. The same job (`testing (mcp.proxy)`) passed on `develop` at this exact base commit (`285fd12`) in the run right before this PR opened, which rules out a base-branch regression too.

I don't have permission to re-run the workflow via the API (403 from the GitHub App integration), so I'm pushing an empty commit to retrigger CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_0153dSFGyFL2tpxZJGBRvvhC)_

---

PR #2293, opened 2026-08-03
docs: add MIGRATING.md covering 1.x to 2.x backwards-incompatible changes

jfallows:
## Description

Adds `MIGRATING.md` documenting the backwards-incompatible changes between Zilla 1.x and the upcoming 2.0.0 release, based on a survey of the `develop` and `support/1.x` branches. Scoped to the core `zilla` repository only (no `zilla-plus` content).

Covers:
- `zilla.json` config fallback removal
- `guard-identity` → `guard-inline` rename (the `identity` type alias still resolves today; flagged as a removal candidate rather than removed here)
- `binding-kafka`: `routes[].when` now forbidden for `kind: client` (single unconditional exit only); removal of the unused `routes[].when[].groupId` condition
- `binding-grpc`: `options.services` (local `.proto` file list) removed in favor of catalog-sourced service definitions
- `binding-mqtt-kafka`: `options.server` removed in favor of the engine's `service.hostname` configuration property
- `binding-openapi` / `binding-asyncapi` / `binding-openapi-asyncapi`: new route-dispatch vocabulary (`spec`/`operation`), including a note for anyone on an early `2.0.0-alpha` build that used `api-id`/`operation-id`
- A statement of the Apache Kafka 2.8.0+ compatibility floor
- Engine SPI changes relevant to binding/guard/vault plugin authors (`ConfigAdapterContext` removal, `Configuration.directory()` removal, in-progress `ValidatorHandler` folding)
- A "Still deprecated" table of items that are marked deprecated but intentionally **not** removed by this change (e.g. the `identity` guard alias, `binding-kafka`'s `options.sasl`, a few Java SPI accessor overloads) — called out for maintainers to decide on deliberately before 2.0.0 ships, not acted on here

No code changes — documentation only.

## Test plan

- [ ] Maintainer review of accuracy against the intended 2.0.0 scope
- [ ] Confirm whether any "Still deprecated" items should be removed before 2.0.0 GA

---
_Generated by [Claude Code](https://claude.ai/code/session_01FYcBnir6XPXyHxWn2AHDcC)_

PR #2296, opened 2026-08-03
fix(examples/mcp.proxy): retry resource-subscribe-client on missing everything+ resource

jfallows:
## Description

The `mcp.proxy` example's CI job was observed stalling for ~6 minutes on the `subscribe_resource` step before eventually recovering. Root cause: Zilla's mcp(proxy) only advertises the `resources.subscribe` capability — and lists the "everything" toolkit's resources at all — once that south binding's own handshake with its upstream has completed. A freshly started gateway's very first session can race that hydration.

`resource-subscribe-client/client.mjs` already retries on one symptom of this race (`does not support resource subscriptions`), but a second symptom of the same race — `resources/list` simply not yet including an `everything+...` resource, surfacing as `no everything+ resource found in resources/list` — was not covered by that retry check. When that happens on a single attempt, the whole `docker compose run` for that step exits 1 immediately (no internal retry), and the example's `test.sh` only recovers because of its own outer `retry_until 3 5` shell-level retry loop. If it happened on the last of those 3 attempts, the whole job would fail outright even though it's the same known, benign startup race the script already accounts for.

This PR broadens the retry check in `client.mjs` to treat both error messages as the same transient hydration race, so a single slow/racy attempt no longer risks failing the job.

## Test plan

- [ ] CI: `testing (mcp.proxy)` job in `.github/workflows/build.yml` passes
- [x] `node --check client.mjs` (syntax)

---
_Generated by [Claude Code](https://claude.ai/code/session_01TosP2YNe5P499HRjtthgsn)_


jfallows:
CI status after catching up to `develop`:

- ✅ `Build (25)`, `cache-images`, `CodeQL`, `Analyze (java)` — green
- ✅ `testing (mcp.proxy)` — **green**, ~9m17s end-to-end (was previously stalling ~6min on a single retry attempt before this fix)
- ✅ every other `testing (*)` example job in the matrix — green
- ⚠️ `testing (http.kafka.oneway)` — **cancelled**, and `cleanup` cancelled as a result

Diagnosed the `http.kafka.oneway` job: it queued for ~17 minutes before starting, then hung indefinitely on its very first `curl -w "%{http_code}" http://localhost:7114/events ...` call (`.github/test.sh:20`) — zero bytes received back from Zilla for 4+ minutes until the run was cancelled. That `curl` has never had a `--max-time` guard (pre-existing, unrelated to this PR), but the hang itself lines up with a different change: the `develop` commit this branch just merged in (5b05b6b3d, "docs: add MIGRATING.md ... (#2293)") migrated this exact example's `zilla.yaml` off deprecated `options.sasl` onto `options.authorization` + an inline guard for Kafka SASL credentials. That's a real behavioral change to this example, landed via an unrelated, already-merged PR — not part of this PR's diff (which only touches `examples/mcp.proxy/resource-subscribe-client/client.mjs`).

I don't have a definitive root cause (the job was cancelled before the docker logs upload step ran, so no container logs are available, and I don't have permission to trigger `rerun-failed-jobs` via the API — got a 403). Given it's unrelated to this PR's content, I'm not attempting a fix here. Worth a maintainer either re-running this job to check for a one-off queue/flake, or checking the new `options.authorization` SASL wiring in `examples/http.kafka.oneway/etc/zilla.yaml` if it reproduces.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TosP2YNe5P499HRjtthgsn)_

---

PR #2304, opened 2026-08-04
feat(release): create GitHub Release with version-specific changelog section

### LINKED ISSUE #2303: release: create GitHub Release with version-specific CHANGELOG section, marked latest only when released from develop
## Context

The `release` workflow (`.github/workflows/release.yml`) currently tags, deploys, merges and regenerates `CHANGELOG.md`, but never creates a **GitHub Release**. Today that step is manual: the [2.0.0 release](https://github.com/aklivity/zilla/releases/tag/2.0.0) body is a hand-pasted copy of the `## [2.0.0]` section of `CHANGELOG.md`, and its release name is empty.

Manual creation means the Release can be forgotten, get a body that drifts from the generated changelog, and — most importantly — get the wrong **latest** marker. Now that maintenance releases ship from `support/1.x` while `develop` is the `2.x` mainline, a `1.2.5` release must not steal the `latest` marker from `2.0.0`.

## Proposal

Add a step to the `finalize` job that creates a GitHub Release for the tag, with:

1. **Body** = the version-specific section of the generated `CHANGELOG.md` — everything from `## [<version>](…)` up to (excluding) the next `## [` heading. Nothing else: no `# Changelog` header, no preceding/following version sections.
2. **Name** = `<version>` (currently blank).
3. **`make_latest`** = whether this version is the highest release the repo has, computed by comparing `<version>` against all existing release tags — the same rule the `deploy` job already uses for the docker `latest` tag (`release.yml:211`-`230`: `git tag --list | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$'` plus `$VERSION`, `sort -V | tail -n1`). Non-semver versions match no rule and are never marked latest.

### Why not derive `latest` from the branch

A `github.ref_name`-based rule (`develop` → latest, `support/*` → not latest) is correct today but breaks the moment `develop` is branched to `support/2.x`: that same file, now on `support/2.x`, would stop marking `2.0.1` as latest even though `3.0.0` does not exist yet, leaving `2.0.0` as latest indefinitely. Comparing released versions gets every window right with no edits at branch time:

| scenario | released | highest existing tag | `make_latest` |
| --- | --- | --- | --- |
| `develop` is the `2.x` mainline, releasing from `develop` | `2.0.0` | `1.2.4` | `true` |
| `support/1.x` patch after `2.0.0` shipped | `1.2.5` | `2.0.0` | `false` |
| `develop` branched to `support/2.x`, `3.0.0` not yet released — patch from `support/2.x` | `2.0.1` | `2.0.0` | `true` |
| same line, after `3.0.0` shipped | `2.0.2` | `3.0.0` | `false` |

The third row is the case that matters: the newest released line keeps `latest` until the new mainline actually ships a release.

This is also consistent with `Validate release version against branch` (`release.yml:27`), which already forbids releasing from `develop` once a `support/<major>.x` branch exists — so `develop` releases are always the highest major, and the comparison agrees with the branch rule everywhere it is defined.

### Placement / sequencing

- Must run **after** `Finish release` (`release.yml:363`) — that is where the tag reaches `origin` for a `develop` release (`Tag the release`, `release.yml:357`, only runs for `support/*`).
- The version section must be captured from the `release/<version>` checkout **before** `git flow release finish` deletes that branch — e.g. extract into `$RUNNER_TEMP/release-notes.md` early in `finalize`, then create the Release after the tag is pushed.
- The tag comparison needs the full tag list; `finalize`'s checkout is already `fetch-depth: 0`.
- `permissions: contents: write` is already declared, so `GITHUB_TOKEN` suffices; no new secret needed.

### Failure and rerun behaviour

- `Generate CHANGELOG` is `continue-on-error` (`release.yml:126`, `release.yml:388`). If the `## [<version>]` section is absent, fall back to a minimal body (compare link against the previous tag) and emit `::warning::` — do not fail the release.
- On a rerun where the Release already exists, update it in place rather than failing the job.

## Branch scope

Land on `develop`, and port the same step to `support/1.x` so `1.x` maintenance releases also get a GitHub Release (which will compute `make_latest: false` on its own, since `2.0.0` exists).

The two copies do **not** need to be byte-identical now — the requirement is that the `develop` copy is written so it stays correct unedited once it becomes `support/2.x`. Concretely: no hardcoded branch names, major versions, or version lines anywhere in the latest decision or the notes extraction; every branch-dependent behaviour driven off `github.ref_name` at runtime, and the latest decision driven off released versions rather than the branch.

## Acceptance criteria

- [ ] `finalize` creates a GitHub Release on the release tag, named `<version>`.
- [ ] Release body contains only the `<version>` CHANGELOG section, matching `CHANGELOG.md` on the base branch.
- [ ] `make_latest` follows the released-version comparison, not the branch: verified for all four rows of the table above, including the post-branch window where the newest `support/N.x` line is still latest.
- [ ] No hardcoded branch name, major version, or version line participates in the latest decision — the `develop` copy would behave correctly as-is on `support/2.x`.
- [ ] A missing/failed CHANGELOG section degrades to a warning + fallback body, and does not fail the release.
- [ ] Rerunning against an existing tag updates the Release instead of failing.
- [ ] Step ported to `support/1.x`; `1.x` releases get a Release and do not displace `2.0.0` as latest.

## Notes

- Prefer sharing one implementation of the highest-version comparison between the docker `latest` tag in `deploy` and `make_latest` in `finalize`, so the docker tag and the Release marker can never disagree.
- The separate `release.yml` in `aklivity/zilla-plus` is out of scope.

jfallows:
## Description

The `release` workflow tagged, deployed and regenerated `CHANGELOG.md` but never created a **GitHub Release**, leaving that manual — the [2.0.0 release](https://github.com/aklivity/zilla/releases/tag/2.0.0) body is a hand-pasted changelog section with an empty release name.

`finalize` now creates the Release on the release tag, named for the version, with a body holding only that version's `CHANGELOG.md` section.

### Only `x.y.z` versions get a Release

Pre-releases are frequent on this workflow — `2.0.0` alone shipped 37 `2.0.0-alpha-N` tags, and the maintenance line regularly tags `-rcN` — so creating a Release for every version string would fill the Releases page with pre-release entries. `release-tags.sh` emits an explicit `semver` output, and both notes extraction and Release creation are gated on it. A pre-release publishes only its exact version: no floating docker tag, no Release. That matches how docker tagging already treated non-`x.y.z` versions. A skipped run emits a `::notice::` rather than passing silently.

### `make_latest` compares released versions, not the branch

A `github.ref_name`-based rule (`develop` → latest, `support/*` → not) is correct today but breaks the moment `develop` is branched to `support/2.x`: that same workflow, now on `support/2.x`, would stop marking `2.0.1` as latest even though `3.0.0` doesn't exist yet, stranding `2.0.0` as latest. Comparing released versions is right in every window and needs no edit at branch time:

| scenario | released | highest tag | `make_latest` |
| --- | --- | --- | --- |
| `develop` is `2.x` mainline | `2.0.0` | `1.2.4` | `true` |
| `support/1.x` patch after `2.0.0` | `1.2.5` | `2.0.0` | `false` |
| branched to `support/2.x`, `3.0.0` not yet shipped | `2.0.1` | `2.0.0` | `true` |
| same line, after `3.0.0` shipped | `2.0.2` | `3.0.0` | `false` |

Pre-release tags never participate in that comparison — the tag list is filtered to `x.y.z` first, so `2.0.1` resolves to `true` even with `2.0.0-alpha-42` present.

The comparison lives in `.github/scripts/release-tags.sh`, shared with the docker `latest` tag in `deploy` (previously inline there), so the image tag and the Release marker cannot disagree. Same algorithm as before, so docker tagging behaviour is unchanged.

### Two deviations from the issue's plan

- **Notes are read via `git show release/{version}:CHANGELOG.md`, not the working tree.** The step sits after `Initialize GitFlow`, and `git flow init -df` can move `HEAD` off the release branch — and on a `develop` release the base branch's `CHANGELOG.md` has no `[{version}]` section yet (it got `merge -s ours`). Reading the working tree would have silently hit the fallback on every release.
- **Release creation is the last step of `finalize`,** not immediately after `Finish release`, so a Release failure cannot leave the base-branch CHANGELOG push undone.

### Robustness

- `Generate CHANGELOG` is `continue-on-error`, so the section can be absent. That warns and falls back to a heading plus a compare link against the preceding release, rather than failing the release after artifacts are out or publishing an empty body. The compare link is omitted when there is no predecessor tag.
- Rerun-safe: `gh release view` → `edit` when the Release exists, else `create --verify-tag`.

## Verification

No test harness covers workflow YAML, so the extracted scripts were exercised directly:

- All four `make_latest` rows above, plus: rerun with the tag already present, `2.10.0` vs `2.9.0` ordering (`sort -V`, not string sort), older patch while a newer minor exists, and no tags at all.
- The semver gate against this repo's real version names: `2.0.0-alpha-43`, `1.3.1-rc1` and `1.2.7-rc11` all yield `semver=false` (no Release, no floating docker tag), while `2.0.1`, `3.0.0`, `1.3.1` and `2.0.0` yield `semver=true`. Also confirmed `2.0.1` → `highest=true` with `2.0.0-alpha-42` in the tag list.
- Notes extraction against the real 297KB `CHANGELOG.md`: 437 lines for `2.0.0` (exactly lines 3–439), one heading, no bleed into the `1.2.4` section; also checked a mid-file section.
- Fallback body generation, including correct per-line predecessor (`1.2.5` → `1.2.4`, not `2.0.0`).
- `--latest=false` and `--verify-tag` confirmed against the `gh` CLI source: both `release create` and `release edit` declare `--latest` as `NilBoolFlag`; `--verify-tag` exists on `create` only, matching usage here.
- YAML parses; `bash -n` clean. `shellcheck` was unavailable in the dev environment, so the scripts are unlinted beyond syntax.

Note that the Release path cannot be exercised end to end until a real release runs, since `workflow_dispatch` on `release.yml` only takes effect once merged to the base branch. The script logic is covered by the checks above; the first live proof will be the next release.

The `support/1.x` port is #2305.

Fixes #2303

### DIFF
--- a/.github/scripts/release-notes.sh
+++ b/.github/scripts/release-notes.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# Prints the version-specific section of a generated CHANGELOG.md: the
+# "## [<version>](...)" heading through to the line before the next version
+# heading, with trailing blank lines removed. Prints nothing when the changelog
+# has no section for <version>, leaving the fallback to the caller.
+
+set -euo pipefail
+
+VERSION="${1:?usage: release-notes.sh <version> [changelog]}"
+CHANGELOG="${2:-CHANGELOG.md}"
+
+if [[ -f "$CHANGELOG" ]]; then
+  awk -v prefix="## [$VERSION]" '
+    substr($0, 1, 4) == "## [" { if (found) exit; found = substr($0, 1, length(prefix)) == prefix }
+    found { print }
+  ' "$CHANGELOG" | sed -e :a -e '/^\n*$/{$d;N;};/\n$/ba'
+fi
--- a/.github/scripts/release-tags.sh
+++ b/.github/scripts/release-tags.sh
@@ -2,10 +2,15 @@
 # Emits the derived tag facts for a release version as key=value lines,
 # suitable for appending to $GITHUB_OUTPUT.
 #
+#   semver  true when <version> is a x.y.z release version, false otherwise
 #   major   major version, empty unless <version> is x.y.z
 #   minor   major.minor version, empty unless <version> is x.y.z
 #   latest  docker "latest" tag, empty unless <version> is the highest release
 #   highest true when <version> is the highest release, false otherwise
+#
+# Pre-release versions such as 2.0.0-alpha-42 or 1.3.1-rc1 are not x.y.z, so
+# they float no docker tag and get no GitHub Release; only the exact version is
+# published for them.
 #
 # A release is the highest release when no existing release tag sorts above it,
 # so the newest released line keeps the latest marker until a higher line ships
@@ -16,12 +21,14 @@ set -euo pipefail
 
 VERSION="${1:?usage: release-tags.sh <version>}"
 
+SEMVER="false"
 MAJOR=""
 MINOR=""
 LATEST=""
 HIGHEST="false"
 
 if [[ "$VERSION" =~ ^([0-9]+)\.([0-9]+)\.[0-9]+$ ]]; then
+  SEMVER="true"
   MAJOR="${BASH_REMATCH[1]}"
   MINOR="${BASH_REMATCH[1]}.${BASH_REMATCH[2]}"
 
@@ -33,6 +40,7 @@ if [[ "$VERSION" =~ ^([0-9]+)\.([0-9]+)\.[0-9]+$ ]]; then
   fi
 fi
 
+echo "semver=$SEMVER"
 echo "major=$MAJOR"
 echo "minor=$MINOR"
 echo "latest=$LATEST"
--- a/.github/scripts/release-tags.sh
+++ b/.github/scripts/release-tags.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# Emits the derived tag facts for a release version as key=value lines,
+# suitable for appending to $GITHUB_OUTPUT.
+#
+#   major   major version, empty unless <version> is x.y.z
+#   minor   major.minor version, empty unless <version> is x.y.z
+#   latest  docker "latest" tag, empty unless <version> is the highest release
+#   highest true when <version> is the highest release, false otherwise
+#
+# A release is the highest release when no existing release tag sorts above it,
+# so the newest released line keeps the latest marker until a higher line ships
+# its first release. Deliberately derived from released versions rather than
+# from the branch, so this stays correct once develop branches to support/N.x.
+
+set -euo pipefail
+
+VERSION="${1:?usage: release-tags.sh <version>}"
+
+MAJOR=""
+MINOR=""
+LATEST=""
+HIGHEST="false"
+
+if [[ "$VERSION" =~ ^([0-9]+)\.([0-9]+)\.[0-9]+$ ]]; then
+  MAJOR="${BASH_REMATCH[1]}"
+  MINOR="${BASH_REMATCH[1]}.${BASH_REMATCH[2]}"
+
+  RELEASED=$({ git tag --list | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' || true; echo "$VERSION"; } | sort -V -u | tail -n1)
+
+  if [[ "$RELEASED" == "$VERSION" ]]; then
+    LATEST="latest"
+    HIGHEST="true"
+  fi
+fi
+
+echo "major=$MAJOR"
+echo "minor=$MINOR"
+echo "latest=$LATEST"
+echo "highest=$HIGHEST"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -210,24 +210,7 @@ jobs:
 
     - name: Determine docker tags
       id: tag
-      run: |
-        VERSION="${{ inputs.version }}"
-        MAJOR=""
-        MINOR=""
-        LATEST=""
-        if [[ "$VERSION" =~ ^([0-9]+)\.([0-9]+)\.[0-9]+$ ]]; then
-          MAJOR="${BASH_REMATCH[1]}"
-          MINOR="${BASH_REMATCH[1]}.${BASH_REMATCH[2]}"
-          HIGHEST=$({ git tag --list | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$'; echo "$VERSION"; } | sort -V | tail -n1)
-          if [[ "$HIGHEST" == "$VERSION" ]]; then
-            LATEST="latest"
-          fi
-        fi
-        {
-          echo "major=$MAJOR"
-          echo "minor=$MINOR"
-          echo "latest=$LATEST"
-        } >> "$GITHUB_OUTPUT"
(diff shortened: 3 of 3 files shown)


jfallows:
`testing (mcp.proxy)` failed on [run 30932107783](https://github.com/aklivity/zilla/actions/runs/30932107783/job/92076584109) — **pre-existing flakiness on `develop`, not this change.** I have merged current `develop` to trigger a fresh run.

### It fails on the base branch

The same job was the **only** failed job on `develop` at [`5f01caf2d`](https://github.com/aklivity/zilla/actions/runs/30870433320) — the exact commit this branch was cut from. `develop` has since gone green on `8eaa9f92f` and `95f140c0b`, but both are documentation-only (`README.md` + an asset), so nothing fixed `mcp.proxy` in between: it is flaky, not broken-then-repaired. This PR's own previous commit (`a5adff222`) also passed this job.

The diff here is `release.yml` plus two scripts under `.github/scripts/` that nothing in `build.yml` invokes, so it cannot affect an example test.

### The log tail is misleading

Worth flagging for whoever picks up the flake, because the tail reads like a pass:

```
+ echo ✅ resources/subscribe, notifications/resources/updated, and resources/unsubscribe relayed end-to-end
+ exit 1
##[error]Process completed with exit code 1.
```

`examples/mcp.proxy/.github/test.sh` accumulates failures into `EXIT` and only surfaces them via the final `exit $EXIT`, so the last assertion printing ✅ says nothing about why the job failed. **I could not identify which assertion actually failed:** the `❌` marker is in the first ~8,500 lines of a 13,551-line log, and only the last 5,000 lines are retrievable through the API tooling I have. The retried `resource-subscribe-client` block (#2296) succeeded on its first attempt this run, so it is not that one.

A concrete follow-up for that script: echo a summary of which named checks set `EXIT=1` just before exiting, so the failing assertion is visible at the end of the log instead of buried thousands of lines up.

### Why a merge and not a rebase

Rebasing would have orphaned `41aa783b2`, which the `support/1.x` port (#2305) references in its `Ports:` trailer. The merge keeps that SHA reachable. The diff against `develop` is unchanged — still exactly the three files.

I could not re-run the job directly (`rerun-failed-jobs` → `403 Resource not accessible by integration`), hence the merge to get a fresh run.


---

jfallows:
Correction to the previous comment: the **"Why a merge and not a rebase"** section is wrong and no longer applies.

This branch is now **rebased** onto `develop` — linear, merge commit dropped, two commits (`473105a6b`, `d59071ce3`) on `95f140c0b`. The diff is unchanged: the same three files, and both scripts are still identical blobs to the `support/1.x` port.

Preserving `41aa783b2` bought nothing, because this PR will be **squash-merged** — that mints a new commit SHA on `develop` regardless, so #2305's `Ports:` trailer has to be rewritten to the squash SHA after this merges either way. Until that rewrite, #2305's trailer references a SHA that no longer exists; it will be corrected post-merge.

Everything else in that comment stands: `testing (mcp.proxy)` is pre-existing flakiness on `develop` (sole failed job at base `5f01caf2d`, green afterwards on documentation-only commits), and the force-push has triggered a fresh run.


---

PR #2322, opened 2026-08-06
engine: change ModelVisitor.onField to take a FieldEvent

jfallows:
## Description

Replaces `ModelVisitor.onField(String path, DirectBufferEx buffer, int index, int length)` with `onField(FieldEvent event)`, where `FieldEvent` exposes `path()` and an already-bounded `value()`. A new `MutableFieldEvent` gives callers a reusable, non-allocating flyweight to wrap before each call, following the same buffer-reuse convention already used throughout the codebase.

This is a signature cleanup, not a behavior change — every call site is updated to match:

- `KafkaExtractor` (implementation + its `extracted()` replay path)
- `KafkaCachePartition`'s key and header extraction (`writeKey`, the `extractHeaders` trailer builder)
- The three model decoder pipelines (`AvroModelDecoderPipeline`, `JsonModelDecoderPipeline`, `ProtobufModelDecoderPipeline`)
- `TestModelPipeline`
- The corresponding unit tests in each of the above

`ModelVisitor` remains a `@FunctionalInterface` with a single abstract method, so `NONE` and every lambda-based caller still compile with a one-line signature change; no `ModelVisitor` implementer keeps the old 4-argument shape.

## Testing

`./mvnw install` (with `-Dskip.checks=true -DskipITs -Djacoco.skip=true` locally to bypass unrelated infrastructure gates — notice-plugin license lookups and jacoco coverage checks on modules this change doesn't touch) for `runtime/engine`, `runtime/model-core`, `runtime/model-avro`, `runtime/model-json`, `runtime/model-protobuf`, and `runtime/binding-kafka` and their dependencies: 467 tests, 0 failures, 0 errors.

Fixes # (issue)

---

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
@@ -497,15 +497,17 @@ public final class KafkaCachePartition
         if (transformKey != KafkaCacheModel.NONE &&
             transforms != null && transforms.extractKey != null)
         {
-            final ModelVisitor writeKey = (path, buffer, index, length) ->
+            final ModelVisitor writeKey = event ->
             {
+                final DirectBufferEx value = event.value();
+                final int length = value.capacity();
                 final int position = entryMark.value + FIELD_OFFSET_PADDED_KEY;
                 KafkaCachePaddedKeyFW paddedKey =
                     logFile.readBytes(position, paddedKeyRO::wrap);
                 final int paddedKeySize = paddedKey.sizeof();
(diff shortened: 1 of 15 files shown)


jfallows (commented):
Closing as superseded by #2326, which shipped a strictly broader replacement for the same mechanism this PR was modifying.

`ModelVisitor` (the interface this PR changes `onField` on) has been removed from `develop` entirely, not resignatured — #2326 replaced it with `ModelTransform`, wired through `ModelHandler.supplyDecoder`/`supplyEncoder`, covering both decode and encode (this PR only touched decode). The `FieldEvent` type introduced here was never picked up; #2326's own description notes it looked for a `FieldEvent` type in the tree and didn't find one, so it defined its own, which is today's `ModelEvent`.

Nothing in this PR's diff is missing from `develop` after #2326 — closing rather than rebasing, since every file it touches (`runtime/engine/.../model/`, `KafkaExtractor.java`, `KafkaCachePartition.java`) has diverged or been deleted since this PR's branch point.

---

PR #2324, opened 2026-08-06
feat(engine): guarantee async guard reauthorize completes strictly later

jfallows:
## Description

`GuardHandler`'s async `reauthorize` carried a default that delegated to the synchronous overload and completed on the caller's stack. That made completion timing a property of the individual guard rather than of the method, so a caller could not know whether its callback would fire during the call or after it, and had to handle both.

Two problems followed.

**Callers had to tolerate reentrancy.** A completion that may fire mid-call, before the caller has finished setting up its own state, is a hazard every call site must reason about. The in-tree consumer — `binding-mcp`'s outbound credential acquisition, currently on a feature branch — carries a flag and a dual-path return purely to absorb that ambiguity.

**Coverage was inverted.** Every guard that decides locally (`jwt`, `inline`, and every test config built on them) completed inline, so the deferred branch of a caller was the one path integration tests never exercised — while being the path a guard that genuinely acquires upstream always takes. The first such guard would run production code no IT had covered.

So the async variant is now async unconditionally, matching what `StoreHandler` already guarantees: the callback fires **strictly later** than the call returns, never on the caller's stack, even when the decision is available locally. Whether authorization resolves synchronously becomes the caller's choice, expressed by which overload it invokes.

Removing the default is the point — a guard now has to state how it completes rather than inherit inline delivery silently. The synchronous overload is documented as the local, cache-only decision it always was: it performs no I/O and returns `NOT_AUTHORIZED` when it cannot decide locally.

`EngineContext.dispatch` loses its default for the same reason. It is the primitive the guards defer onto, yet its own default ran the task inline, so the new guarantee would have rested on every implementor happening to override it. `EchoWorker` and `TlsWorker` now state inline execution explicitly, which is correct for a benchmark harness with no event loop and nothing depending on the async contracts.

### Breaking change

`GuardHandler` implementations must now implement the async `reauthorize`, and `EngineContext` implementations must now implement `dispatch`. In-tree: `JwtGuardHandler`, `InlineGuardHandler`, `TestGuardHandler`, `EchoWorker`, `TlsWorker` — all updated here. Each guard implements its own one-tick deferral via `EngineContext.dispatch`; no shared helper, so the mechanism stays visible at each implementation.

### Testing

`GuardFactoryTest.shouldDelegateAsyncReauthorizeToSyncByDefault` asserted the behaviour being removed and is replaced by `shouldDeferAsyncReauthorize`, which asserts the callback has *not* fired when the call returns, then drains the dispatcher and asserts it fires with the expected session and context id. Equivalent tests added for `JwtGuardHandler` and `InlineGuardHandler`, the latter also covering the `failed()` path.

Written test-first; the new engine test failed against the old default for the right reason, observing `NOT_AUTHORIZED` delivered on the stack before the call returned.

Full reactor build: 183/189 modules pass, 34,888 tests, 0 failures. The only failing module is `incubator/command-dump`, whose Testcontainers-based `DumpRule` requires a Docker daemon unavailable in the environment used to verify this; all 48 of its errors trace to that single initialiser and no error occurs outside that module.

---

### DIFF
--- a/runtime/binding-echo/src/test/java/io/aklivity/zilla/runtime/binding/echo/internal/bench/EchoWorker.java
+++ b/runtime/binding-echo/src/test/java/io/aklivity/zilla/runtime/binding/echo/internal/bench/EchoWorker.java
@@ -423,4 +423,13 @@ public class EchoWorker implements EngineContext
     {
         return null;
     }
+
+    // this harness has no event loop to defer onto, so it runs the task inline rather than
+    // strictly later; nothing here depends on the async SPI contracts that guarantee builds on
+    @Override
+    public void dispatch(
+        Runnable task)
+    {
+        task.run();
+    }
 }
--- a/runtime/binding-tls/src/test/java/io/aklivity/zilla/runtime/binding/tls/internal/bench/TlsWorker.java
+++ b/runtime/binding-tls/src/test/java/io/aklivity/zilla/runtime/binding/tls/internal/bench/TlsWorker.java
@@ -493,6 +493,15 @@ public class TlsWorker implements EngineContext
         return Clock.systemUTC();
     }
 
+    // this harness has no event loop to defer onto, so it runs the task inline rather than
+    // strictly later; nothing here depends on the async SPI contracts that guarantee builds on
+    @Override
+    public void dispatch(
+        Runnable task)
+    {
+        task.run();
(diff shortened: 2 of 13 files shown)


jfallows (commented):
LGTM

PR #2326, opened 2026-08-07
feat(engine): add pluggable per-field ModelTransform hook to the model pipeline

### LINKED ISSUE #2325: Add a pluggable per-field ModelTransform hook to the model pipeline (decode + encode)
## Problem

The model pipeline's only per-field hook today is `ModelVisitor.onField(FieldEvent)`, and it is purely observational: it fires once per field, only after the whole value has reached `ModelStatus.COMPLETE`, returns `void`, and has no attachment point on encode at all — `AvroModelEncoderPipeline` doesn't carry a `ModelVisitor` parameter through anywhere.

This already limits an existing, shipped feature. `binding-kafka`'s `KafkaTopicTransformsType` (`extractKey` / `extractHeaders`) promotes payload field values into the Kafka message key or headers, so downstream bindings can filter/route on them without decoding the full payload. It's implemented via `KafkaExtractor implements ModelVisitor`, wired through `KafkaCacheModel.decoder(handler, extractPaths, scratch)` — and it only ever works on decode. `KafkaCacheModel.encoder()` never attaches a visitor, because there's no hook to attach on the encode side even if a future use case needed one there.

Meanwhile, `common-avro`/`common-json`/`common-protobuf` already define a per-event stage contract for exactly this kind of thing — `AvroTransform` / `JsonTransform` / `ProtobufTransform`, composable left-to-right in each format's own pipeline, with javadoc explicitly anticipating "forwarding, dropping, or substituting events" and third-party implementations. No mutating implementation of it exists yet; the only implementor, `AvroExtractor`, is deliberately transparent/read-only.

## Proposal

Add a new engine SPI type, `ModelTransform`, as a real per-field mutation point that composes into each format's own pipeline (not a second top-level pipeline concept), plus a thin adapter per format that lets a single, format-agnostic `ModelTransform` implementation plug into avro/json/protobuf without knowing anything about their wire representation.

**Shape — pull-based, mirroring `AvroTransform`, not the buffer-loop shape of `ModelPipeline`:**

```java
public interface ModelTransform
{
    ModelTransform NONE = ...; // identity passthrough

    Status transform(
        ModelController control,
        ModelSource source,
        FieldEvent event,   // reuse the existing FieldEvent type
        ModelSink sink);

    default Status resume(...) { return sink.resume(control, source, event); }
    default Status flush(...) { return sink.flush(control, source); }
    default void reset() { }
    default boolean identity() { return false; }
}
```

Reusing `ModelPipeline`'s existing buffer-in/buffer-out contract for this was considered and rejected — it's a real paradigm mismatch against the pull-based, resumable `(control, source, event, sink)` shape every format-level `*Transform` already uses, and bridging the two would mean inventing a second resumability protocol at every adapter boundary. Matching the existing shape means resumability composes uniformly with no extra translation, and an unchanged field can be forwarded by reference instead of a mandatory buffer copy.

**Format adapters — thin, always one-to-one:**

`AvroModelDecoderTransform` / `AvroModelEncoderTransform` (`implements AvroTransform`), and equivalents for json/protobuf, do pure presentation: derive the field's path/bytes from the format-specific event on the way in, and turn the generic transform's answer (unchanged / replacement / declined) back into a valid format-specific event on the way out. Declining is the only case with real format-specific work — synthesizing a structurally valid placeholder for that field's type — since only the format side knows what "valid" means there; the generic transform never needs to know.

Insertion point mirrors where `AvroExtractor` already sits on decode today: `[format parse] → [ModelTransform adapter] → [format-specific final stage]` on decode, `[format-specific first stage] → [ModelTransform adapter] → [format generate]` on encode. The encode-side insertion point is new — `AvroModelEncoderPipeline` has no field-level touchpoint today, so this isn't a rewire of something existing there.

**Composing more than one transform:** rather than one adapter per transform (which pays a format↔generic context-switch cost per transform per field, redundant for N>1), a `CompositeModelTransform` composes an ordered list of transforms entirely in the generic domain — transform K's output feeds transform K+1's input. The format adapter is always handed exactly one `ModelTransform` — a leaf or a composite — and always does exactly one context switch regardless of how many policies it represents.

**Side-output / accumulation:** no new shared channel needed. A transform that needs to accumulate something for its owner to query afterward can do so the same way `KafkaExtractor` already does today (implements the generic interface for pipeline-plumbing uniformity, plus its own extra methods that only its concrete owner calls) — fully outside the shared contract.

## Migration plan

1. Add the `ModelTransform` SPI plus the avro adapter pair, with no consumer changes yet.
2. Re-platform `KafkaExtractor` onto `ModelTransform`, decode-only, behavior-preserving — existing k3po IT coverage for `extractKey`/`extractHeaders` becomes the regression oracle. This is deliberately the first consumer: smallest, best-isolated, already tested, no encode side needed.
3. Simplify `KafkaCachePartition` — the current transform-then-separately-check-extracted code collapses into one `transform()` call plus the existing post-hoc replay, unchanged in behavior.
4. JSON/protobuf adapter parity as follow-up work once the avro path is proven.

## Scope

In scope: `runtime/engine/.../model/` (new SPI), `runtime/model-avro` (reference adapter implementation), `runtime/binding-kafka` (re-platform `KafkaExtractor`, decode-only, no behavior change).

Out of scope for this issue: json/protobuf adapter parity (follow-up), and any consumer that needs genuine encode-side field mutation — that lands as its own issue once this mechanism exists and is proven, from whichever component needs it.

jfallows:
## Description

Replaces the purely observational `ModelVisitor` with `ModelTransform`, a format-agnostic per-field stage that can observe, substitute, or decline each field of a value, on decode **and** on encode.

A transform never sees a format's wire representation: a thin adapter inside each model derives the field's path and value bytes on the way in and turns the answer back into a valid format-specific event on the way out. The answer *is* the event the stage forwards — `FIELD` keeps the value, `REPLACED` substitutes it, `DECLINED` drops it so the adapter writes a structurally valid placeholder for that field's type, the only case that needs format-specific work.

### engine

- `ModelTransform` (plus `NONE`), `ModelEvent`, `ModelSource`, `ModelSink`, `ModelController` — shaped after the pull-based, resumable contract each format's own `*Transform` already uses, so resumability composes with no extra translation and an unchanged field forwards by reference rather than by a mandatory copy.
- The existing `ModelStatus` is shared by the transform chain rather than a second enum being introduced, mirroring `common-avro`, where one `AvroPipeline.Status` serves the pipeline, its sinks and its transform stages. `UNDERFLOW` is documented as the one value a transform chain never reports, since a field always arrives whole.
- `ModelTransform.andThen` composes any number of stages into the single transform a `ModelHandler` accepts — `first.andThen(second).andThen(third)`, or `transforms.stream().reduce(ModelTransform::andThen)`. Composition stays entirely in the format-agnostic domain, so an adapter is handed exactly one transform and pays one format-to-generic context switch per field however many policies it represents.
- `ModelFieldBridge` drives a transform from the field values a model surfaces once a value completes, for a model with no native adapter yet (observation-only).
- `ModelHandler.supplyDecoder` / `supplyEncoder` now take a `ModelTransform`; `ModelVisitor` is removed.

### model-avro

- `AvroModelTransform` replaces `AvroExtractor` and is wired into both the decode and the encode pipeline. The encode-side attachment point is new — `AvroModelEncoderPipeline` had no field-level touchpoint before.
- Two modes, selected by `transform.identity()`:
  - **observing** forwards every event verbatim as it arrives, keeping the extraction fast path byte-identical;
  - **mediating** withholds a top-level scalar field's events until its value is whole, then re-emits the answer, streaming and resuming across bounded output.

### model-json, model-protobuf, model-core

Adapted to the new SPI through `ModelFieldBridge`; native adapters are follow-up work, as scoped in the issue.

### binding-kafka

- `KafkaExtractor` re-platformed onto `ModelTransform`, decode-only and behavior preserving; `extracted()` now takes a `ValueConsumer`.
- A value rejected part-way discards its partial captures, so a field surfaced inline before the rejection cannot reach the cache entry.

## Deviations from the issue, worth a look

1. **`ModelEvent` carries the answer.** Unchanged / replacement / declined are expressed as the event the stage forwards, rather than a status code or a sentinel source. Composed stages pass them through unchanged and the avro adapter's three branches map 1:1.
2. **One adapter class, not two.** The issue names `AvroModelDecoderTransform` and `AvroModelEncoderTransform`; both would be identical `AvroTransform` implementations over the same event stream, so this is a single `AvroModelTransform` used at both insertion points. Easy to split if the two names are preferred.
3. **`andThen` instead of a `CompositeModelTransform` class.** The issue describes a composite type composing an ordered list; a default method on the interface gives the same property with no public class, no parallel stage arrays, and no null/empty/single special-casing in a factory.
4. **`ModelEvent` did not exist.** The issue says "reuse the existing `FieldEvent` type", but there is no such type in the tree — today's `ModelVisitor.onField` takes `path, buffer, index, length` — so it is defined here along with the rest of the generic domain.

## Behaviors preserved deliberately

Neither is obvious from the diff:

- The adapter's `reset()` does **not** reset the wired transform. Doing so would clear `KafkaExtractor`'s captures before `KafkaCachePartition` replays them into the cache entry's key and trailing headers.
- A value rejected part-way now discards its partial captures. Avro delivers fields inline as they complete, so without this the old "nothing captured on reject" behavior would have changed.

## Testing

`./mvnw clean verify` green on `engine`, `binding-kafka`, `model-core`, `model-json`, `model-protobuf`, `model-avro` — including the 394 Kafka ITs, where `CacheFetchIT`'s `extractKey` / `extractHeaders` cases are the regression oracle for the re-platform, plus checkstyle, license and coverage gates.

New coverage:

- `ModelTransformTest` — `NONE`, chaining order two and three deep, composition by reduction, decline propagation, identity composition, reset and flush fan-out, bridge delivery.
- `AvroModelTransformTest` — replace and decline across string, scalar, bytes, fixed, boolean, float, double and union-branch fields; byte-for-byte reproduction when substituting identical values; resume across bounded output; substitute streamed across a bounded destination; reject with diagnostic; identity reporting in both modes; and the field-run framing sequence in both modes.

Commit `b09a4b284` fixes a defect found while tracing the flow: `flush` and `END_VALUE` were driven off `AvroEvent.END_MESSAGE`, which the pump never pulls — it stops the moment the terminal sink closes the top-level record and returns `COMPLETED`. That left a transform's field run opened and never closed, disagreeing with `ModelFieldBridge`. The run now closes off `COMPLETED`, the one signal every datum shape reaches, with a test per mode pinning the sequence.

Fixes #2325 — json/protobuf adapter parity remains follow-up work, as scoped in the issue.


runtime/model-avro/src/main/java/io/aklivity/zilla/runtime/model/avro/internal/AvroModelHandlerImpl.java
```diff
@@ -189,15 +191,17 @@ AvroPipeline newPipeline(
...
-                    .into(Avro.generator(schema, new UnsafeBufferEx(new byte[1]), 0))
-                : Avro.stream(Avro.parser(schema))
-                    .lenient(lenient)
-                    .reporting(reporter)
-                    .into(Avro.generator(schema, new UnsafeBufferEx(new byte[1]), 0));
+                : Avro.stream(Avro.parser(schema));
+            if (adapter != null)
+            {
+                stream = stream.transform(adapter);
+            }
```
jfallows: Let's introduce AvroTransform.NONE and AvroModelTransform.of(transform) which returns AvroTransform.NONE when transform is ModelTransform.NONE, then eliminate these null checks, and requireNonNull on supplyEncoder(transform) supplyDecoder(transform) too.

jfallows: Done in 381d940a2.

One addition worth calling out: `AvroStream.transform` now **drops** `AvroTransform.NONE` rather than binding it. Without that, passing `NONE` unconditionally would add a `BoundSink` frame per event to every Avro stream, so dropping at assembly is what makes the unconditional call free — the no-transform pipeline ends up carrying no stage at all, which is strictly better than the null checks it replaces.

Beyond that: `AvroModelTransform.of` returns `AvroTransform.NONE` for `ModelTransform.NONE` and the constructor is now private, so `of` is the only way in. Both pipelines hold a non-null `AvroTransform`, which let me delete the duplicated decode `newPipeline` overload that existed purely to omit the stage, along with the conditional `transform` in the encode variant.

On `requireNonNull` — I added it where the transform is consumed (avro both directions, plus json and protobuf decode, since the SPI contract should read the same across implementations). I did *not* add it where the transform is ignored outright — json/protobuf `supplyEncoder`, both `CoreModelHandler` methods — since that guards an unused parameter. Instead `ModelHandler` now states the contract once: never null, pass `NONE`, and an implementation is free to wire it away. Happy to add the guard there too if you'd rather have it unconditional.

Also added a `common-avro` test asserting a `NONE` stage leaves a datum byte-identical, so the drop-at-assembly behavior is pinned rather than assumed.


---

PR #2331, opened 2026-08-07
feat(binding-kafka): add KafkaPipeline for message-level key/headers/value composition, retire KafkaExtractor's stash-and-replay

### LINKED ISSUE #2329: Add a KafkaPipeline for message-level key/headers/value composition, retire KafkaExtractor's stash-and-replay
## Problem

`binding-kafka`'s `KafkaTopicTransformsType` (`extractKey` / `extractHeaders`) promotes payload field values into the Kafka message key or headers, so downstream bindings can filter/route on them without decoding the full payload. It's implemented via `KafkaExtractor` (a `ModelVisitor`, soon a `ModelTransform` per #2325): it observes field values during a single decode pass and copies matching ones into its own internal buffer, and only after the whole value reaches `COMPLETE` does anything read that buffer back out (`extracted(path, visitor)`) to actually write the key or a header.

That two-phase shape — capture now, replay later — exists only because there's currently nowhere else for a mid-traversal discovery to go: the key and headers are written as fixed blocks relative to the value, so anything found partway through the value has to wait for the whole value to finish before it can be replayed into an already-started key/header write. This works, but it requires an internal stash buffer and a bespoke capture/replay API that only exists to route around that ordering constraint.

## Proposal

Add `KafkaPipeline` — a pipeline over the *whole message* (key, headers, and value together), with three independently-writable regions that can be appended to in any interleaved order during a single traversal, rather than one fixed sequential layout. This removes the need for a stash: a transform that finds a match writes directly into the target region the instant it's found, then resumes where it left off.

**Shape** — pull-based, mirroring `AvroTransform`/`ModelTransform` (`control, source, event, sink`, composed via `.andThen(transform)`), but a distinct pipeline from `ModelPipeline`, sitting one layer above it: `KafkaPipeline` owns and drives the key's own model pipeline (when a topic configures a structured key) and the value's model pipeline, translating their output into its own event vocabulary, `KafkaEvent`.

**Event vocabulary** — a lane selector plus one shared content vocabulary, not three parallel ones:

- `SWITCH_KEY` / `SWITCH_HEADERS` / `SWITCH_VALUE` — a repeatable context switch, not a `START_*`/`END_*` bracket. The requirement is that headers and key content can be discovered and appended *while* traversing the value, then control returns to the value and resumes — an interleaved pattern a one-time-entered/exited bracket can't represent.
- Ordinary structural/content events (object-enter, named field, array element) are reused as-is across all three lanes, interpreted in whichever lane is currently selected — this matters concretely when the key and/or value are independently structured (each can carry its own nested schema), since the same content vocabulary serves both without duplicating it per lane.
- Headers reuse the same "object" shape as key/value, with two differences rather than a separate design: field values are always raw binary (matches Kafka's actual `byte[]` header values — no further typed/nested structure within a header), and field names may repeat (matches Kafka's actual wire format: an ordered list of `(String, byte[])` pairs, not a uniquely-keyed map). `SWITCH_HEADERS` stays parameterless like the other two — the header's name arrives as the ordinary named-field event that follows.

**`KafkaTransform`** implementations are 1:1 with `extractKey`/`extractHeaders` config entries, composed via `.andThen(transform)`. Each watches for its configured path (sources can be the key or value lane, matching the existing `${message.key.foo}` / `${message.value.foo}` config syntax) and on a match: switches to the key or headers lane, writes the extracted content inline, switches back. The matched event keeps flowing untouched to its own destination — extraction only ever adds to key/headers, never mutates the value. Headers land in the final region in encounter order, as a direct consequence of the streaming design.

## Migration plan

1. Add `KafkaPipeline` / `KafkaEvent` / `KafkaTransform`, no consumer changes yet.
2. Re-platform `extractKey`/`extractHeaders` onto it, retiring `KafkaExtractor`'s stash-and-replay mechanism entirely. Existing k3po IT coverage is the regression oracle — same external (fetch-only) behavior, new mechanism underneath.
3. Simplify the current bespoke key/header-writing code in `KafkaCachePartition` that separately checks extraction config and replays captured values — it collapses into driving `KafkaPipeline` once per message.

## Scope

In scope: `runtime/binding-kafka` (`KafkaPipeline`, `KafkaEvent`, `KafkaTransform`, re-platforming `extractKey`/`extractHeaders`, simplifying `KafkaCachePartition`).

Out of scope for this issue: any consumer that needs to write something into key/headers that isn't a direct observation of an existing payload field — that's a separate future issue once this mechanism exists and is proven on the smaller, already-tested extraction case.

jfallows:
## Description

Adds `KafkaPipeline` — a pipeline over the whole Kafka message with three independently writable lanes — and retires `KafkaExtractor`'s stash-and-replay mechanism.

`KafkaPipeline` sits one layer above the engine's `ModelPipeline`: it owns the key's model pipeline and the value's model pipeline, drives each in turn, and translates their per-field output into `KafkaEvent` — a lane selector (`SWITCH_KEY` / `SWITCH_HEADERS` / `SWITCH_VALUE`) plus one shared `FIELD` reused across all three lanes. The switches are repeatable context switches rather than one-time brackets, so a stage that discovers header content part-way through the value switches lane, appends it, switches back, and the value traversal resumes where it left off.

`KafkaTransform` stages compose with `andThen`. `KafkaExtractTransform` is one stage per `extractKey` / `extractHeaders` entry: on a match it switches to the target lane, writes what it found, switches back, and lets the matched event flow on untouched — extraction only ever adds to the key or headers, never mutates the value.

**What's gone:** `KafkaExtractor`'s expandable capture buffer, its path-keyed region map, and the `extractedLength` / `extracted` replay API, along with their pass-through on `KafkaCacheModel`. `KafkaCachePartition`'s fetch write path now takes one `KafkaPipeline` in place of the `(transformKey, transformValue, transforms)` triple and terminates it in `KafkaEntrySink`, which writes each lane into its region as the event arrives.

### Two rules the terminal depends on

Documented on `KafkaSink`, and both are load-bearing:

- A lane switch selects the destination of the **single** `FIELD` that follows it. A field with no switch ahead of it is one the traversal merely surfaced, and nothing is written for it.
- The pipeline's opening announcement of the lane it is traversing reaches the stages but **not** the terminal. Without this, `extractKey` — whose origin and target are both the key lane — cannot be told apart from the key's own fields flowing past.

A naive lane-tracking terminal gets this wrong: a key model surfacing `$.tenant, $.id, $.zone` with `extract-key: ${message.key.id}` ends up keyed on `zone`. `KafkaCachePartitionTest.shouldExtractKeyFromOneFieldOfAStructuredKey` pins it, and I confirmed it fails without the fix.

### Supported lane transitions are asserted

The vocabulary can express a stage appending to any lane from any lane, but only two combinations have somewhere to land in the cache entry's layout:

| traversing | may append to | why not the others |
| --- | --- | --- |
| key | key (`extractKey`) | no trailers are under construction during the key drive |
| value | headers (`extractHeaders`) | the key's hash is already computed and indexed by then |

Nothing can be appended to the value lane at all — the value's bytes come from its model's output, not from this vocabulary. Everything else was previously accepted and then silently dropped by the terminal. `KafkaLaneGuard` now asserts on any unsupported switch, and the announcement terminal asserts on content appended where nothing can land. The root `pom.xml` runs Surefire and Failsafe with `-ea`, so both are live under every unit test and every k3po IT. Relaxing this later means supporting fully parallel key, headers and value writes — a deliberate change rather than something that quietly starts working.

## Behavior changes

- **Extracted headers land in field-encounter order**, not `extract-headers` config order — a direct consequence of the streaming design, and called out in the issue. Current IT coverage uses a single header, so nothing moved there.

## Known limits, carried forward deliberately

- **The key lane still defers by one slot.** Its destination *is* the key, still being appended by the key's own model when the match arrives, and `KafkaCacheFile` only grows forward. What remains is one reusable slot in `KafkaEntrySink`, applied at the same statement the old `extracted()` replay ran at. The bespoke capture/replay API is gone; the ordering constraint underneath it is not. Headers stream with zero buffering.
- **`extract-key` still sources from the key lane and `extract-headers` from the value lane**, ignoring the `${message.key.*}` vs `${message.value.*}` prefix — unchanged from `develop`, where `KafkaTopicType`'s regex captures the prefix but never reads it. Honoring it for `extract-key` is not merely wiring: the key hash is computed in `writeEntryStart` before the value exists, so a value-sourced key would leave a stale hash. `KafkaExtractTransform` carries an explicit origin lane so the mechanism is general; only the wiring is pinned.

## Testing

- `KafkaPipelineTest` (13) and `KafkaTransformTest` (7) are new — lane selection, encounter ordering, composition, reuse across messages, and the four rejected transitions.
- `KafkaCachePartitionTest` gains the structured-key regression above.
- `KafkaExtractorTest` is removed with the class it covered.
- Existing `CacheFetchIT` coverage for `extract-key` / `extract-headers` is the regression oracle and is unchanged.

`./mvnw clean verify -pl runtime/binding-kafka,runtime/engine` is green: 178 unit tests, the full IT suite including `CacheFetchIT` (69 tests), checkstyle and license checks.

Fixes #2329

---
_Generated by [Claude Code](https://claude.ai/code/session_015vYCwp9WzvnTmfGSp2fZUJ)_


runtime/common-agrona/src/main/java/io/aklivity/zilla/runtime/common/agrona/buffer/UnsafeBufferEx.java
```diff
@@ -267,7 +275,11 @@ public void wrap(
         byteArray = buffer;
         byteBuffer = null;
         byteBufferView = null;
-        segment = MemorySegment.ofArray(buffer);
+        if (buffer != segmentArray)
+        {
+            segmentArray = buffer;
+            segment = MemorySegment.ofArray(buffer);
+        }
```
jfallows: if we move this check earlier, and compare buffer != byteArray, then do we still need the extra segmentArray field?

jfallows: Good catch — done in 662f3567d. Moved the check to compare `buffer != byteArray` before `byteArray` is overwritten, so the dedicated `segmentArray` field is gone.

This turned out to be more than a simplification — it fixes a real bug in the original approach. `segmentArray` only got updated by `wrap(byte[]...)`, so an intervening wrap of a different kind (`wrap(ByteBuffer)`, `wrap(DirectBufferEx)`, `wrap(MemorySegment)`, etc.) could reassign `segment` without touching it. A later `wrap(byte[])` of the *same* array would then see `segmentArray` still matching, skip recomputation, and return the intervening wrap's stale segment instead of one for the re-wrapped array. `byteArray` is reassigned by every wrap overload, so it doesn't have that gap.

Added `shouldRecomputeSegmentAfterIntermediateNonArrayWrap` to `UnsafeBufferExSegmentCachingTest`, confirmed it failed against the old implementation before fixing.

---

Issue #2333, opened 2026-08-07
tls binding: supply verified peer certificate chain as guard credentials


jfallows:
## Background

The `tls` server binding terminates mutual TLS and validates the client certificate chain against its configured trust store, but there is currently no way for it to hand that chain to a guard. Its options cover `version`, `keys`, `trust`, `signers`, `trustcacerts`, `sni`, `alpn` and `mutual` — none of which supply credentials.

Note: `routes[].guarded` is inherited generic route boilerplate on the `tls` binding and has never been exercised — `guarded` is defined once on the common route schema in `engine.schema.json` and applies to every binding, but `runtime/binding-tls` contains no reference to `GuardHandler` at all, so nothing supplies credentials at this layer and `TlsRouteConfig.authorized(...)` only ever tests the authorization inherited from the inbound network stream. This makes the issue "make an existing option functional" rather than "add a new capability".

Other bindings already do this. The `mqtt` binding declares where credentials come from via an explicit block:

```yaml
options:
  authorization:
    my_jwt_guard:
      credentials:
        connect:
          username: "Bearer {credentials}"
```

The binding owns acquisition; the guard owns the credential format. The `tls` binding needs the equivalent so that a guard can authorize a session based on what the peer certificate asserts.

## Why this is useful independently

Route `when` clauses on the `tls` binding match on what the handshake *negotiated* — `authority`, `alpn`, `port`. These are transport facts, available before any certificate is validated.

What the certificate *asserts* — subject, SAN, issuer — only exists after validation and is not expressible today at all. Supplying the chain to a guard closes that gap, and lets certificate-derived values reach route conditions and topic templates via `${guarded['name'].attributes.*}`.

## Proposal

An `options.authorization` block naming a guard to receive the verified peer certificate chain:

```yaml
bindings:
  tls_server0:
    type: tls
    kind: server
    vault: server
    options:
      keys:
        - localhost
      trust:
        - partner-ca
        - internal-ca
      mutual: required
      authorization:
        x509_0:
          credentials:
            certificates: pem
    routes:
      - guarded:
          x509_0:
            - partner
        exit: mqtt_server0
```

`credentials` is a map, matching the `mqtt` binding. It does not nest a further level for *where* the credential comes from, as `mqtt` does with `connect.username`, because a TLS session has no equivalent ambiguity — there is only one thing it can supply. Keeping `certificates` flat also leaves room for siblings if TLS ever needs to supply anything else.

[...]

PR #2335, opened 2026-08-07
feat(guard-x509): authorize sessions from a verified client certificate chain

### LINKED ISSUE #2334: x509 guard: authorize sessions from a verified client certificate chain
## Background

No guard currently consumes a certificate chain. Deployments terminating mutual TLS can validate a peer against a trust store, but nothing downstream can act on *which* peer it was — subject, SAN and issuer are unavailable to route conditions and to `${guarded[...]}` expressions.

This adds an `x509` guard that takes a verified chain as its credential and produces an authorized session with an identity, attributes, and optionally roles.

Paired with #2333, which supplies the credential.

## Naming

`x509`, not `tls` or `mtls`.

Zilla names guards after the credential or the provider — `jwt` a format, `api-keys` a credential, `azure-ad` a provider. None is named for a transport, and naming this one for TLS would be the only exception. It would also foreclose a chain reaching the guard by any other route.

The binding owns acquisition, the guard owns the format. Same split as the `mqtt` binding extracting a bearer token from a CONNECT packet and handing it to a `jwt` guard.

## Credential

PEM, leaf-first, concatenated. Full contract in #2333.

## Configuration

Three optional blocks, all predicating on one internal vocabulary of certificate fields — `subject.cn`, `subject.o`, `subject.ou`, `issuer.cn`, `san.dns`, `san.uri`, `san.email`, and so on. Field paths are internal; attribute *names* are the external surface.

```yaml
guards:
  x509_0:
    type: x509
    options:
      identity: subject.cn
      attributes:
        organization: subject.o
        tenant: san.uri
      roles:
        partner:
          - issuer.cn: "Partner Issuing CA"
        internal:
          - issuer.cn: "Internal CA"
            subject.ou: "Platform"
          - issuer.cn: "Internal CA"
            san.dns: "*.internal.example.com"
```

### `options.identity`

Selects which field becomes the session identity — full subject DN, `subject.cn`, or a SAN entry.

If the full DN is supported, **pin the canonical rendering** (`X500Principal.CANONICAL`: lowercased, normalised whitespace, no spaces around separators). RFC 1779 and RFC 2253 forms render the same certificate differently, and identity values commonly feed topic names and other derived strings, so an unstable rendering produces unstable downstream state.

Prefer SAN over CN where certificates permit. CN-as-identity is deprecated by RFC 6125, though still ubiquitous in private PKI for machine identity.

### `options.attributes`

Maps an external attribute name to an internal field, exposed as `${guarded['x509_0'].attributes.<name>}`.

The indirection is the point: attribute names should describe meaning rather than origin, so `attributes.organization` can be sourced from `subject.o` here and from a token claim under a different guard, and expressions referencing it survive a change of identity source.

SAN needs its type discriminator in the addressing — `getSubjectAlternativeNames()` returns entries tagged by type (2 dNSName, 6 URI, 1 rfc822Name) and a certificate may carry several.

### `options.roles`

Maps role names to conditions, following the existing and/or convention: **array members are OR'd, properties within an object are AND'd** — the same reading rule as `routes[].when`, so there is no new mental model.

This is an *explicit* operator-declared mapping, visible in configuration. It is deliberately not the pattern of implicitly reading roles out of OU by convention, which would be invisible and non-portable.

**Matching rules:**

- **Glob patterns**, matching house style (`client-id: "*"`, mqtt topic wildcards). Not regex — a loosely anchored pattern over a distinguished name can match a field the author did not intend, and DN string rendering varies between issuers.
- **Multi-valued fields match existentially.** `san.dns` against a certificate carrying three dNSNames matches if any one does. Subjects with repeated RDNs (`subject.ou` legitimately appears more than once) behave the same way.
- **An absent field never matches** rather than erroring, so evaluation is total.
- **Roles are a set union** across independent evaluations, not first-match.

### Suggested additional attribute

`x5t_s256` — SHA-256 over the DER encoding of the **leaf certificate**, base64url without padding. Useful for telemetry, and it is the value RFC 8705 certificate-bound access tokens compare against, so exposing it early costs nothing and enables that composition later.

## Trust model

The guard trusts its source as given, in-process, and is expected to be fed only by the `tls` binding. See the scope-limit section of #2333 — this is a deliberate constraint rather than an oversight, and adding a header-fed path would change the trust model rather than extend it.

## Non-goals

- **Chain path validation.** The trust manager has already validated the chain during the handshake. The guard inspects an already-validated chain; it does not rebuild or revalidate a path.
- **CA narrowing.** Restricting which issuers are acceptable for a given surface belongs in the `tls` binding, following the existing convention that options declare the superset and routes select from it — `options.sni` with `when.authority`, `options.alpn` with `when.alpn`, `options.trust`/`options.signers` narrowed at the route. Duplicating that in the guard would also lose the ability to serve several trust domains through one binding.
- **Certificate revocation.** CRL and OCSP are trust-manager concerns.

## Acceptance criteria

- [ ] `x509` guard type accepting a PEM leaf-first chain as its credential
- [ ] `options.identity` selecting a field, with canonical DN rendering if full DN is supported
- [ ] `options.attributes` mapping external names to internal fields, resolvable in `${guarded[...]}` expressions
- [ ] `options.roles` with and/or semantics, glob matching, existential multi-value matching, absent-never-matches, set union
- [ ] All three blocks optional
- [ ] Guarded routes authorize against roles produced by the guard
- [ ] Reference documentation alongside the other guards

## Dependency

Not independently shippable or testable — pairs with #2333, which supplies the credential. Neither should merge alone.

jfallows:
## Description

Adds an `x509` guard that takes a verified peer certificate chain as its credential and produces an authorized session with an identity, named attributes, and optionally roles.

Three new modules, following the `guard-jwt` layering:

- **`specs/guard-x509.spec`** — `x509.schema.patch.json`, `x509.idl` (authorization-failed event), config fixtures, and `X509Certificates` holding a checked-in two-chain test PKI: an `Internal CA` leaf with a multi-valued `OU` and DNS/URI/email SANs, and a `Partner Issuing CA` leaf.
- **`config/guard-x509.conf`** — `X509OptionsConfig` (`identity`, `attributes`, `roles`), `X509MatchConfig`, builders, adapters, `X509GuardInfo`.
- **`runtime/guard-x509`** — `X509Guard`/`Context`/`FactorySpi`/`Handler`, plus `X509Fields` (leaf → field map) and `X509Roles` (compiled role conditions), and the event context/formatter.

Registered in the `specs/`, `config/` and `runtime/` poms, `cloud/docker-image/pom.xml`, and `zpm.json.template`.

### Credential

PEM, leaf-first, concatenated, parsed in one `CertificateFactory.generateCertificates` call. Escaped newlines and stray whitespace are tolerated as leniency, not as a second supported format. The guard inspects an already-validated chain in-process — it does not rebuild or revalidate a path, narrow acceptable issuers, or check revocation.

### Field vocabulary

`identity`, `attributes` and `roles` all address the leaf through one internal vocabulary: `subject.dn`, `subject.<rdn>` (any attribute type, via `LdapName` over the RFC 2253 rendering), the `issuer.*` equivalents, `san.dns` / `san.uri` / `san.email` / `san.ip`, and `x5t.s256`.

`*.dn` is pinned to `X500Principal.CANONICAL` as the issue asked; individual RDNs keep the case written in the certificate.

`x5t_s256` is exposed as the *field* `x5t.s256` rather than a hardcoded attribute name, so operators name it themselves — consistent with attributes mapping external names to internal fields.

### Role matching

Array members are OR'd, properties within an object are AND'd — the same reading rule as `routes[].when`. Values are anchored globs rather than regular expressions, multi-valued fields match existentially, an absent field never matches, and roles are a set union.

## Decisions worth a look

- **`options.identity` defaults to `subject.dn`** when omitted, so a guard with no `options` at all still yields a stable identity.
- **Glob matching is case-insensitive.** This is a judgement call: it is what RFC 5280 (`caseIgnoreMatch` on DirectoryString attributes) and RFC 6125 (DNS-ID) prescribe, and it stops `subject.ou: "Platform"` silently failing against a lowercase certificate. It does slightly weaken an exact `x5t.s256` comparison — happy to make it case-sensitive instead.
- **`expiresAt` is the leaf's `notAfter`.** The guard does not reject on validity dates — that is path validation, a stated non-goal — but reporting real expiry matches what `jwt` does with `exp`.

## Testing

`clean install` is green for all three modules — 46 tests: 29 unit plus 4 IT in the runtime module, 11 in conf, 6 in spec.

`X509IT.shouldAuthorizeGuardedRoute` drives the engine's `test` binding with the PEM chain as credentials through a `routes[].guarded` requiring the `internal` role, asserting identity and attributes via the binding's `expectIdentity`/`attributes`. Confirmed not vacuous by flipping the required role to `partner`, watching it fail, then restoring it.

## Dependency — now paired, and verified against it

Pairs with #2336, which implements #2333 and supplies the credential from the `tls` binding. Per the issue, neither should merge alone.

Each PR on its own only tests one side of the contract — #2336 against the engine's `type: test` guard, this one against the engine's `type: test` binding. So I merged the two branches locally and ran them **against each other**. They merge cleanly (no overlapping files).

In a scratch worktree I swapped the `type: test` guard in #2336's `server.mutual.guarded.yaml` for a real `x509` guard with role conditions derived from that spec's client certificate:

```yaml
guards:
    x509_0:
        type: x509
        options:
          identity: subject.cn
          attributes:
            organization: issuer.o
          roles:
            partner:
              - issuer.cn: serverca
                subject.cn: client1
            internal:
              - issuer.cn: "Some Other CA"
```

and ran #2336's own ITs unmodified:

| Test | Result |
| --- | --- |
| `ServerIT.shouldAuthorizeGuardedRouteWithClientCertChain` | pass |
| `ServerIT.shouldNotAuthorizeGuardedRouteWhenClientCertAbsent` | pass |

So a real mutual-TLS handshake produces a chain that this guard parses, resolves `partner` from, and authorizes `routes[].guarded` with. Confirmed not vacuous the same way as above: changing the condition to `issuer.cn: not-the-real-ca` makes `shouldAuthorizeGuardedRouteWithClientCertChain` fail.

The PEM contract lines up exactly — #2336 renders `-----BEGIN CERTIFICATE-----` + MIME base64 at 64 columns + `-----END CERTIFICATE-----`, concatenated over `SSLSession.getPeerCertificates()` (leaf-first), which is what `CertificateFactory.generateCertificates` here consumes.

None of that scaffolding is committed on either branch: a cross-module IT would put a production SPI implementation on the other module's test classpath, which `runtime/AGENTS.md` rules out. It was a local verification only.

Reference documentation: aklivity/aklivity-docs#90 (this guard), aklivity/aklivity-docs#91 (the `tls` binding side).

Fixes #2334

https://claude.ai/code/session_01L17rjk3vGrxpimk67MzrHA


jfallows (commented):
LGTM

PR #2336, opened 2026-08-07
feat(binding-tls): supply verified peer certificate chain as guard credentials

### LINKED ISSUE #2333: tls binding: supply verified peer certificate chain as guard credentials

jfallows:
## Description

The `tls` server binding terminates mutual TLS and validates the client certificate chain against its configured trust store, but had no way to hand that chain to a guard. `routes[].guarded` was inherited from the common route schema in `engine.schema.json` and had never been exercised — `runtime/binding-tls` contained no reference to `GuardHandler`, and the only authorization source was a `dname -> 0L` stub, so a guarded route on a `tls` binding could never authorize.

This adds an `options.authorization` block naming a guard to receive the verified peer certificate chain:

```yaml
bindings:
  tls_server0:
    type: tls
    kind: server
    vault: server
    options:
      keys:
        - localhost
      trust:
        - clientca
      mutual: required
      authorization:
        x509_0:
          credentials:
            certificates: pem
    routes:
      - guarded:
          x509_0:
            - partner
        exit: app0
```

The chain is rendered as **leaf-first concatenated PEM**, parseable in one call via `CertificateFactory.generateCertificates(InputStream)`. The guard is consulted in `onDecodeHandshakeFinished`, so peer certificates are available; an unverified peer raises `SSLPeerUnverifiedException` and yields no credentials at all rather than an empty or partial chain, leaving the guarded route unauthorized. The resulting session id becomes the authorization on the application stream, so it also reaches `${guarded[...]}` expressions.

`credentials` is a map matching the `mqtt` binding, without the further nesting `mqtt` uses for `connect.username` — a TLS session has only one thing it can supply. `certificates: pem` names the format rather than being a bare `true`, leaving `x5c` an obvious later home without a schema change.

### Two supporting changes, worth reviewer attention

- **Route matching before the handshake no longer applies `routes[].guarded`.** The accept gate (`TlsServerFactory.newStream`) and ALPN selection both ran `route.authorized(...)` before any guard could possibly have executed, so a fully guarded binding would have rejected every connection before the handshake started. Added `resolvePortOnlyBeforeHandshake(port)` and dropped the `authorized` filter from `selectAlpn`. This is a no-op for existing configs — the pre-handshake authorization is always `0` there and non-guarded routes always authorize — but it is a real semantic change.
- **Removed the dead signing-CA-DN authorization stub** on `TlsStream`. The guard session id supersedes it, and two competing authorization sources on one stream would be ambiguous.

### Scope

Per the issue, this is not independently shippable — it pairs with #2334 (the `x509` guard), which consumes the credential. Here the contract is exercised against the engine's `type: test` guard, configured with the expected PEM chain.

Reference documentation is in aklivity/aklivity-docs#91, covering `options.authorization`, `authorization.credentials` and `credentials.certificates` in both the 1.x and 2.x `tls` binding references.

## Tests

Written before the implementation, per the repo's test-first discipline.

- `TlsOptionsConfigAdapterTest` — read/write round-trip for the new block (16 → 18 tests)
- `SchemaTest` — validation of `server.mutual.guarded.yaml` and `server.mutual.guarded.requested.yaml`
- `ApplicationIT` / `NetworkIT` — new paired `client.rpt` / `server.rpt` scripts run against each other
- `ServerIT` — `shouldAuthorizeGuardedRouteWithClientCertChain` (real mutual TLS handshake against a live engine; the test guard is configured with the exact expected PEM chain, and the app script asserts the guard session id lands on the application stream) and `shouldNotAuthorizeGuardedRouteWhenClientCertAbsent`

Verified locally:

| Module | Result |
| --- | --- |
| `config/binding-tls.conf` | 35 tests, 0 failures |
| `specs/binding-tls.spec` | SchemaTest 21, NetworkIT 33, ApplicationIT 29 — BUILD SUCCESS |
| `runtime/binding-tls` | 76 ITs, 0 failures, JaCoCo check passed |
| `binding-asyncapi`, `binding-openapi`, `binding-mcp-kafka` (consume `TlsOptionsConfig`) | all green |

`cloud/docker-image` could not be built locally for lack of a Docker daemon; everything else in the reactor builds.

Fixes #2333


jfallows:
One example job failed on `b455ae5`: `testing (asyncapi.http.kafka.proxy)`. I do not believe it is caused by this change, and I can't clear it myself — details below so it's easy to check rather than take on trust.

**What failed.** The POST succeeded and the GET immediately after returned nothing:

```
+ curl -w %{http_code} --location http://localhost:7114/pets ... 
OUTPUT=204  EXPECTED=204   ✅
+ curl http://localhost:7114/pets ...
OUTPUT=      EXPECTED={"name": "Rocky","id": 1}   ❌
```

The two curls are ~90ms apart (`04:18:59.579` → `04:18:59.669`) with no wait between them, so this reads as a Kafka read-after-write race in that example's `test.sh`, not a routing or authorization failure.

**Why it isn't this change.** `examples/asyncapi.http.kafka.proxy/etc/zilla.yaml` contains no `type: tls`, no `guards:` and no `authorization:` — grep count is 0 for all three. This PR touches only `runtime/binding-tls`, its `config`/`specs` modules, and `examples/tls.echo`. There is no path from the diff to that example.

**Everything else on this SHA is green**, including the parts this PR actually affects:

- `Build (25)` ✅, `Analyze (java)` ✅, `CodeQL` ✅
- `testing (tls.echo)` ✅ — the example changed here, exercising the guarded route end to end
- `testing (tls.reflect)` ✅, and the other 33 example jobs ✅

**Why I'm not fixing it.** The fix would be a retry or wait between POST and GET in `examples/asyncapi.http.kafka.proxy/.github/test.sh` — a different example, unrelated to this work, and widening this diff to touch it seems wrong. Re-running the failed job needs permissions this integration doesn't have (`rerun-failed-jobs` returns 403), so a maintainer re-run is the cheapest next step.

One thing I could not establish: I checked recent `develop` runs for a prior occurrence and did not find one — the single recent `develop` failure was in `Build (25)` with the example jobs skipped, so it says nothing about this example. I'm calling it a timing race from the failure signature and the absence of any causal path, not from an observed history of flaking.


---

PR #2337, opened 2026-08-08
fix(binding-http): answer 503, not 431, when the client pool is exhausted

jfallows:
## Description

`HttpClientPool.newStream` collapsed two unrelated rejections onto one status:

```java
if (client == null || queuedRequestLength > maximumRequestQueueSize)
{
    newStream = rejectWithStatusCode(sender, begin, HEADERS_431_REQUEST_TOO_LARGE);
}
```

A request whose queued length exceeds the queue slot really is too large, and `431` is right for it. A null client from `supplyClient` is not about the request at all — every connection for the route is in use and none is reusable, so there is nowhere to send it. Reporting that as `431 Request Header Fields Too Large` tells the caller its headers are at fault when they are fine, and hides an exhausted pool behind a message about the request.

`503` with `retry-after` is what this binding already answers when a connection cannot carry a request (`HEADERS_503_RETRY_AFTER`, used a few lines below for a full queue slot), so the exhausted case now shares it rather than inventing a status. The over-long queue entry keeps `431`.

### Why it matters

This was expensive to diagnose in practice. In aklivity/zilla-plus#1104 an `mcp(client)` POST to an upstream MCP server was answered `431` by `http_client` itself — 0.7 ms in, with no upstream connection behind it — because `clientPools` is keyed by resolved route id, so six upstream hosts shared one route's pool of `maximum.connections` (default 10). The status pointed at request headers, so the pool was never suspected, and finding the real cause took a packet capture of the engine's stream frames.

Note this changes only which status is reported. The pool being per-route rather than per-authority is left alone here.

## Testing

New scenario `pool.exhausted.and.503.response`, authored as paired `client.rpt` / `server.rpt` under both `streams/application/rfc7230/connection.management/` and `streams/network/rfc7230/connection.management/`. It fills the single connection with a response that stays incomplete, so the client is neither `available()` nor `reusable()`, then issues a second request and expects `503`.

- `ConnectionManagementPoolSize1IT#shouldGive503ResponseWhenConnectionPoolExhausted` — runs it against the engine at `HTTP_MAXIMUM_CONNECTIONS = 1`
- `application.rfc7230.ConnectionManagementIT` and `network.rfc7230.ConnectionManagementIT` — run each pair peer-to-peer for self-consistency

Confirmed the scenario reproduces the old behavior before the fix by dumping the IT's engine directory: the rejection frame carried `:status 431`.

- `runtime/binding-http` — 395 ITs, 41 pre-existing skips, unit tests, checkstyle, license and the coverage gate all green
- `specs/binding-http.spec` — green
- `request.headers.too.long` still passes, so the genuine `431` is unaffected

https://claude.ai/code/session_01D8qDfygNWTeejdmm9PfsnS

---

### DIFF
--- a/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
+++ b/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
@@ -2090,7 +2090,11 @@ public final class HttpClientFactory implements HttpStreamFactory
             HttpClient client = supplyClient(scheme, authority);
             final int queuedRequestLength = HttpQueueEntryFW.FIELD_OFFSET_VALUE_LENGTH + begin.extension().sizeof();
 
-            if (client == null || queuedRequestLength > maximumRequestQueueSize)
+            if (client == null)
+            {
+                newStream = rejectWithStatusCode(sender, begin, HEADERS_503_RETRY_AFTER);
+            }
+            else if (queuedRequestLength > maximumRequestQueueSize)
             {
                 newStream = rejectWithStatusCode(sender, begin, HEADERS_431_REQUEST_TOO_LARGE);
             }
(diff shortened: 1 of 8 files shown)


jfallows (commented):
LGTM

PR #2347, opened 2026-08-09
feat(binding-tls): select client certificate from guarded session identity

### LINKED ISSUE #2345: tls client binding: select client certificate from guarded session identity via routes[].with
## Background

When a `tls` server terminates mutual TLS and a `tls` client re-establishes mutual TLS upstream, the client's certificate identity has to reach the client binding so it can decide which key to present. Today that happens through a bespoke mechanism, in four hops:

1. **`TlsServerFactory.getCommonName`** (`TlsServerFactory.java:2391-2418`) extracts a CN from the peer principal with a regex — `COMMON_NAME_PATTERN = Pattern.compile("CN=(?<cn>[^,]+)")` (`TlsClientX509ExtendedKeyManager.java:39`).
2. It is emitted as a `ProxyBeginEx` info item — `is.item(i -> i.secure(s -> s.name(name)))` (`TlsServerFactory.java:2239`), wire type at `proxy.idl:107`.
3. Every intermediate binding must explicitly relay that info item or the identity is lost. `binding-kafka-proxy` in zilla-plus carries code that exists solely to copy it from the inbound `ProxyBeginEx` to the outbound one.
4. **`TlsBindingConfig.newClientEngine`** (`:352-364`) reads it back and stashes it on the JSSE session — `session.putValue(COMMON_NAME_KEY, commonName)` (`:361`) — where `TlsClientX509ExtendedKeyManager.chooseEngineClientAlias` (`:98`, `:117-126`) matches it against each candidate alias's leaf subject CN.

Since #2335 and #2336 the `tls` server binding can hand its verified peer chain to a guard and mint a session, whose id becomes the application stream's `authorization`. That makes hop 1–3 redundant: the identity is already on every frame, minted by a component that observed the handshake.

## Problems with the current mechanism

- **The propagated identity is an unauthenticated string.** `secure.name` is an ordinary `ProxyBeginEx` field. Any binding upstream of the `tls` client that can populate one — including a `proxy` binding decoding PROXY protocol v2 from an untrusted peer (`ProxyServerFactory.java:2206` writes exactly this field) — determines which certificate Zilla presents upstream. There is no structural link between "a chain was verified" and "this identity is asserted." A guard session id cannot be forged that way; #2336 deliberately made the guard the sole authority for `sessionAuth`.
- **`CN=(?<cn>[^,]+)` is a naive DN parse.** It breaks on an escaped comma (`CN=Doe\, John`) and over-captures a multi-valued RDN (`CN=a+OU=b` yields `a+OU=b`). `X509Fields` already does this correctly via `X500Principal.CANONICAL` and `LdapName` (`X509Fields.java:86-100`).
- **CN is the only expressible identity.** A certificate whose identity lives in a SAN cannot be propagated at all.
- **Every intermediate binding must opt in.** Identity survives only where relay code was written. A binding that lacks it drops the identity silently, and the failure surfaces as an upstream handshake failure far from the cause.
- **Two identity mechanisms now coexist on one binding.** #2336 removed the dead `dname -> 0L` stub precisely because "two competing authorization sources on one stream would be ambiguous." `secure.name` is the third, still live.

## Proposal

A `with` clause on `tls` client routes, selecting the key by matching a property of Zilla's own candidate certificates against a value resolved from the guarded session:

```yaml
guards:
  x509:
    type: x509
    options:
      identity: subject.cn
      roles:
        client:
          - issuer.cn: Test Client CA
            subject.cn: client-*

bindings:
  tls_client0:
    type: tls
    kind: client
    vault: upstream
    options:
      signers:
        - upstream-ca
      trust:
        - broker-ca
    routes:
      - when:
          - authority: broker.example.com
        guarded:
          x509:
            - client
        with:
          certificate:
            subject.cn: ${guarded['x509'].identity}
        exit: tcp_client0
```

Nothing between the two `tls` bindings participates. The carrier is `authorization`, a field on every frame, so any pipeline depth works with no relay code — subject to the caveat below.

`${guarded['...'].identity}` is not a new mechanism. It is already used inside `routes[].with` in this repo — `specs/binding-http-kafka.spec/.../config/proxy.post.items.yaml:36-43` — with per-stream resolution in `HttpKafkaWithResolver.java:59-62`. `binding-tls` gains a `TlsWithResolver` in the same shape. The schema slot is already reserved and deliberately closed: `tls.schema.patch.json:136` has `"with": false`.

Values accept both literals (`subject.cn: kafka-gateway` — always present this certificate) and `${guarded[...]}` expressions. Multiple properties under `certificate` are AND'd, consistent with the convention `guard-x509` established for role matching.

## Why `routes[].with` on the client, and `options.authorization` on the server

These are not two styles for one concept. They sit at different points in the connection lifecycle, and the timing forces the shape.

**On the client, every route-matching input precedes the handshake.** `TlsBindingConfig.resolve(authorization, beginEx)` (`:209-223`) matches on `authority`, `alpn` and `port`, all read from the *inbound application stream's* `ProxyBeginEx` — what the upstream binding is requesting, not what the upstream server negotiates — plus `authorization`, already on the BEGIN. `TlsClientFactory.java:243-245` resolves the route and only then creates the engine:

```java
TlsRouteConfig route = binding != null ? binding.resolve(authorization, beginEx) : null;
if (route != null)
{
    final SSLEngine tlsEngine = binding.newClientEngine(beginEx);
```

The route is fully determined before the first ClientHello byte, so `with` can parameterize the handshake.

**On the server it cannot be.** The server's route depends on SNI and ALPN, which arrive inside the ClientHello, and on `guarded`, which depends on the client certificate, available only after the handshake. That is what `resolvePortOnlyBeforeHandshake` (`:241-248`) exists for, and its javadoc says so: *"A server binding accepts and negotiates before any guard can run, so authorization is not yet known at that point."* A `with` clause on a `tls` server route would be resolved too late to shape its own handshake.

So the server keeps `options.authorization` (binding-wide, evaluated during and after the handshake) and the client gets `routes[].with` (route-scoped, resolved before it). The asymmetry is intentional and should be documented as such.

Being route-scoped is also a capability `options` structurally cannot have: different upstream clusters selected by authority or ALPN, each with its own certificate policy.

## Certificate property vocabulary — deliberately narrow

Launch with exactly two: **`subject.cn`** and **`subject.dn`**.

These are not an arbitrary prefix of the `guard-x509` vocabulary. They are what the other end of the pipeline actually emits:

- `X509OptionsConfigBuilder.java:30` — `IDENTITY_DEFAULT = "subject.dn"`. A guard with no `identity` override produces a canonical DN, so without `subject.dn` the default guard config could not be matched at all.
- `identity: subject.cn` is the override used by the `tls.echo` example, and is what today's regex already does.

The general rule for future additions: the `guard-x509` field vocabulary splits into **selectors**, which could plausibly equal a per-stream identity (`subject.cn`, `subject.dn`, `san.dns`, `san.uri`, `san.email`), and **constraints**, which are static properties one would pin and never bind to `${guarded[...]}` (`issuer.*`, `x5t.s256`, `san.ip`). `with` is a per-stream clause, so only selectors belong. `issuer.*` is additionally near-redundant: JSSE already filters candidates by the CertificateRequest `certificate_authorities`, and `options.signers` already scopes the candidate set.

`san.uri` is the one worth revisiting later, for SPIFFE (`spiffe://trust-domain/workload`). Deferred rather than rejected: adding an enum value is backward compatible, removing one is breaking, and there is no demonstrated need yet.

## Implementation notes

- **Extract `common-x509`.** `X509Fields` is package-private in `guard.x509.internal`, and `binding-tls` must not depend on `guard-x509`. Promote the field vocabulary to a shared `runtime/common-x509` module (matching the existing `common-agrona`, `common-lang`, `common-json` convention) and have both depend on it. This matters most at two fields, not least: `resolveName` (`X509Fields.java:78-105`) is the canonical DN rendering plus `LdapName` RDN extraction. If the guard rendered a DN canonically and `binding-tls` rendered RFC 2253, the mismatch would be a silent wrong-certificate selection, not a compile error.
- **Canonicalize literal `subject.dn` values at config load** through the same renderer. Canonical form lowercases attribute values and normalizes ordering and whitespace, so a hand-written `subject.dn: "CN=Foo,OU=Bar"` would otherwise never match. Without this, `subject.dn` is effectively expression-only in practice and nobody will know why.
- **Precompute an alias index at binding init.** Selection runs in `chooseEngineClientAlias`, per connection, on the engine worker thread. The keystore is fixed for the binding's lifetime, so build alias → property values once and make selection a map lookup rather than parsing certificates per handshake.
- **Define the ambiguity rule.** Multiple aliases can match one identity; today it is first-match in `getClientAliases` order, i.e. keystore order. Because both sides are static, multiple matches are detectable from the precomputed index at init — prefer a config error there over unspecified runtime behavior.
- **The `SSLSession.putValue` channel survives, as an internal detail.** `chooseEngineClientAlias` receives only the `SSLEngine`, and key managers are built into the `SSLContext` once per binding, so there is no per-stream hook. Resolve at route time, stash at engine-creation time, read during the handshake. Rename `COMMON_NAME_KEY` accordingly; only the *source* of the value changes.
- **Candidate filtering is already handled.** `FileSystemVaultHandler.java:298` builds the delegate with `KeyManagerFactory.getInstance("PKIX")`, so key algorithm, issuer, validity period, key usage and EKU `clientAuth` are filtered before selection runs. This feature chooses *which identity to assert*, not *which certificate is usable*.
- **Fail loudly.** An unresolvable guard name should be a config error. A `${guarded[...]}` expression that resolves to nothing at runtime currently yields no match, no certificate, and a handshake failure at the far end with nothing pointing back at the config — emit a `BindingEvent`. `HttpKafkaWithResolver` does not do this today, so it is new work regardless.

## Scope limits — deliberate

- **`secure.name` is demoted, not removed.** It stays in `proxy.idl`, remains matchable via `ProxyConditionMatcher.java:417`, and is the natural `PP2_SUBTYPE_SSL_CN` mapping for real PROXY protocol peers. The `tls` client simply stops *selecting* on it.
- **Backward compatible.** With no `with.certificate` on the route, the client falls back to today's `secure.name` behavior. Deprecate the fallback separately.
- **`authorization` is a single slot.** A binding between the two `tls` bindings that runs its own guard replaces it. This is strictly better than requiring every binding to relay a field, but it is not unconditional and belongs in the docs — a `kafka-proxy` with an external SASL guard in front of a `tls` client expecting the x509 session will present the wrong certificate.
- **`sni` and `alpn` are out of scope here.** They are binding-wide `options` already overridden per-stream by an implicit `ProxyBeginEx` lookup (`TlsBindingConfig.java:302-324`) — the same implicit-side-channel shape as `secure.name`. `with` is their natural explicit home, but that is a follow-on.

## Acceptance criteria

- [ ] `tls` client routes accept `with.certificate` with `subject.cn` and `subject.dn` properties
- [ ] Property values accept literals and `${guarded['name'].identity}` / `${guarded['name'].attributes.*}` expressions
- [ ] Multiple properties under `certificate` are AND'd
- [ ] `with` remains rejected on `tls` server and proxy kinds
- [ ] The client presents the candidate key whose certificate matches; no match presents no certificate
- [ ] Absent `with.certificate`, existing `secure.name` behavior is unchanged
- [ ] Literal `subject.dn` values are canonicalized at config load so they compare equal to guard-derived values
- [ ] Ambiguous matches are rejected at config load rather than resolved arbitrarily
- [ ] Unresolvable guard name is a config error; unresolved expression at runtime emits a `BindingEvent`
- [ ] `X509Fields` extracted to a shared module used by both `guard-x509` and `binding-tls`
- [ ] Paired spec scripts and engine ITs covering: match, no-match, guarded-and-selected, and fallback with no `with`
- [ ] Documented in the `tls` binding reference, including why the server uses `options.authorization` and the client uses `routes[].with`

jfallows:
## Description

A `tls` client route can now name the certificate the binding presents upstream, via `routes[].with.certificate` with `subject.cn` or `subject.dn`. Values are literals or `${guarded['name'].identity}` / `${guarded['name'].attributes.*}` expressions, resolved per stream from the guard session identified by the inbound stream's `authorization`. Exactly one property is named. `with` remains rejected on the `server` and `proxy` kinds.

The carrier is `authorization`, a field on every frame, so no intermediate binding has to relay anything and a guard session id cannot be forged the way an ordinary `secure.name` info item can. With no `with.certificate` on the matched route, the client still selects on `secure.name`, unchanged.

`X509Fields` is promoted out of `guard.x509.internal` into a new shared `runtime/common-x509` module so the guard and `binding-tls` render and match the field vocabulary through one implementation. A guard rendering `subject.dn` canonically while `binding-tls` rendered RFC 2253 would be a silent wrong-certificate selection, not a compile error.

Fails loudly where it can: an unresolvable guard name and a literal `subject.dn` that does not parse are configuration errors. At runtime, an expression resolving to nothing emits `binding.tls.client.certificate.not.resolved` naming the property, and a selection matching no candidate key emits `binding.tls.client.certificate.not.matched` naming the selector. Without the latter a no-match was visible only as a handshake failure at the far end, with nothing logged locally.

When several candidates match, selection takes the latest `notBefore`, breaking ties on thumbprint.

Rationale for the client/server asymmetry, and the `secure.name` caveats, are documented in `runtime/binding-tls/README.md`.

Fixes #2345

### Four departures from the issue as written

- **`common-x509` carries the Aklivity Community License**, matching `guard-x509`. The issue describes `guard-x509` as Apache-licensed; it is not (`runtime/guard-x509/LICENSE`, and every source header). Matching the source module and the convention for new modules therefore agree, and nothing is relicensed.

- **The index is keyed on the leaf certificate, not on the alias.** The issue calls for an alias → properties index precomputed at binding init, but key manager aliases are not stable — OpenJDK's `X509KeyManagerImpl.makeAlias` prefixes each alias with a counter that increments on every enumeration, so the same key is `1.0.client1` at init and `2.0.client1` at handshake and an alias-keyed index never hits. Keyed on the leaf `X509Certificate` instead, reached via `getCertificateChain(alias)[0]`, which parses the alias and ignores the counter. Same intent (parse once at init, lookup per handshake).

- **Both selection paths now use the shared canonical parser**, including the `secure.name` fallback. Behaviour differs from the previous `CN=(?[^,]+)` regex only for names that regex mis-parsed: an escaped comma yielded a truncated value, and a multi-valued relative distinguished name over-captured.

- **`certificate` admits a single property (`maxProperties: 1`), so the AND-ing acceptance criterion is not met as written.** `subject.dn` already contains the common name, so naming both is either redundant or self-contradictory — an earlier revision of this PR shipped an IT whose route could never select anything, which is what the combination buys you. AND-ing earns its keep only alongside a property independent of the subject (`issuer.*`, `san.*`), and the issue scopes both out. Narrowing now is the compatible direction: relaxing `maxProperties` later is not a breaking change, tightening it would be. The resolver keeps its map shape, so widening is a schema edit.

### Acceptance criteria

- [x] `tls` client routes accept `with.certificate` with `subject.cn` and `subject.dn`
- [x] Values accept literals and `${guarded['name'].identity}` / `.attributes.*` expressions
- [ ] ~~Multiple properties under `certificate` are AND'd~~ — deliberately not met; `maxProperties: 1`, see departures above
- [x] `with` remains rejected on `tls` server and proxy kinds
- [x] The client presents the candidate key whose certificate matches — asserted **on the wire**
- [x] No match selects no certificate — `binding.tls.client.certificate.not.matched` asserted by IT
- [ ] No match presents no certificate *on the wire* — still not asserted; see below
- [x] Absent `with.certificate`, existing `secure.name` behaviour is unchanged
- [x] Literal `subject.dn` values canonicalized at config load
- [x] Several matching candidates resolved deterministically, latest `notBefore` first
- [x] Unresolvable guard name is a config error; unresolved expression at runtime emits a `BindingEvent`
- [x] `X509Fields` extracted to a shared module used by both `guard-x509` and `binding-tls`
- [x] Paired spec scripts and engine ITs covering match, no-match, guarded-and-selected, and fallback with no `with`
- [x] Documented in the `tls` binding reference

### How selection is now asserted

An earlier revision leaned on `TlsClientX509ExtendedKeyManagerTest` for "which key is chosen", because every IT config held a single candidate — so a completed handshake proved *a* certificate was presented, not *which*. Any selection strategy that picked something at all passed.

That test is deleted. Selection is now observable end to end: two candidate keys signed by **different client CAs**, with the k3po server trusting only one of them. The handshake completes only if the route selected the key that peer trusts, so `connected` is itself the assertion. No peer-certificate assertion in k3po is required.

| IT | candidates | route | far end trusts | proves |
| --- | --- | --- | --- | --- |
| `shouldEstablishConnectionWithCertificateSelectedFirst` | client1, client2 | `subject.cn: client1` | `clientca` | client1 reached the wire |
| `shouldEstablishConnectionWithCertificateSelectedSecond` | client1, client2 | `subject.cn: client2` | `clientca2` | client2 reached the wire |
| `shouldEstablishConnectionWithCertificateSelectedNewest` | client1, client3 (both `CN=client1`) | `subject.cn: client1` | `clientca2` | the later `notBefore` wins |
| `shouldCloseWhenCertificateNotMatched` | client1 | `subject.cn: client2` | — | `not.matched` event fires |

The two positive cases are each other's negative control: if selection ignored the `with` clause and took keystore order, one of them would fail.

**On the one unchecked line.** `not.matched` covers a no-match locally, but not the wire. k3po's tls transport exposes only `keyStoreFile`/`keyStorePassword`, `trustStoreFile`/`trustStorePassword`, `wantClientAuth` and `needClientAuth` — there is no peer-certificate assertion, and a `tls://` accept cannot script a handshake that fails (it completes only on success; `rejected` does not help, since `RejectedHandler` completes on `channelBound` of a child channel and the TLS transport never binds one when the handshake fails — verified both ways). `shouldRejectConnectionWithCertificateNotMatched` and `shouldRejectClientAuthMismatched` therefore stay `@Ignore`d with their recorded rationale.

Separately verified out of band: a genuine handshake failure *does* propagate to `connect aborted` on the application side — a raw `zilla://streams/net0` accept injecting a fatal TLS alert reproduces it. That probe is not committed, since forcing the alert unconditionally says nothing about which certificate was selected.

### Ambiguity handling was removed

An earlier revision rejected, at configuration load, two candidate keys a route's properties could not tell apart. That check is deleted, along with `assertUnambiguous` and `indistinguishableSubjects`.

It was wrong on two counts. It compared field *names*, not resolved values, so it could not evaluate the `${guarded[...]}` case this feature exists for, and it rejected configs whose resolved value would only ever have matched one candidate. More fundamentally, when several candidates match, each one satisfies the route — the config is met, and the choice between them is immaterial. The usual reason two candidates match at once is a certificate rotation window, where failing would turn a routine rotation into an outage.

What was worth keeping is that keystore enumeration order is not stable, so an arbitrary pick could vary across restarts. Hence the deterministic order: latest `notBefore`, ties on thumbprint. That also sends the new certificate during a rotation.

### Supporting change

`TestVault` now accepts several keys: `key` takes an object or an array, mirroring the shape `trust` already has, so no existing config changes.

### Verification

| Command | Result |
| --- | --- |
| `./mvnw clean install -pl runtime/engine` | 206 unit, 207 ITs, 0 failures |
| `./mvnw clean install -pl specs/binding-tls.spec` | 34 unit, 75 ITs, 0 failures |
| `./mvnw clean install -pl config/binding-tls.conf` | 40 tests, 0 failures |
| `./mvnw clean verify -pl runtime/binding-tls` | 31 unit, 86 ITs, 0 failures, jacoco 0.73 gate met |
| `./mvnw clean install -pl config/engine.conf` | 52 tests, 0 failures |
| `./mvnw clean install -pl runtime/common-x509` | 13 tests, jacoco 0.90 gate met |
| `./mvnw install -pl runtime/guard-x509` | 29 unit, 4 ITs, no behaviour change |
| `./mvnw checkstyle:check` on changed modules | clean |

Skips are the pre-existing `@Ignore`s. The earlier full-reactor run reported here missed three modules downstream of `incubator/command-dump`, whose Testcontainers tests need a Docker daemon this sandbox lacks; CI subsequently ran the full build green, so the real figure was 193/193.

https://claude.ai/code/session_01485pbcDBvcJEErZ7ndZq2j


jfallows (commented):
LGTM

Issue #2343, opened 2026-08-09
zpm install: module-generation phase logs nothing, making multi-minute stalls undiagnosable


jfallows:
## Summary

`zpm install` emits no output between `Generated module info for system-only automatic module: <last>`
and `Generated module info for delegate module`. That window legitimately takes 80–100s on a CI
runner, so a stall inside it is indistinguishable from normal progress until the job's timeout fires.
Diagnosing one CI stall took four runs plus a hand-rolled thread-dump watchdog to localize, because
the log offers nothing to bisect with.

## The gap

`ZpmInstall.installModules` (2.1.0), lines ~226-235:

```java
generateSystemOnlyAutomatic(logger, modules);   // logs per module, then silently
                                                // expandJar/javac/extendJar for the LAST one
delegateAutomatic(modules, delegate);           // silent
copyNonDelegating(modules);                     // silent
...
Collection<ZpmArtifact> optional = cache.resolveOptional(...);  // silent -- and does network I/O
generateDelegate(logger, delegate, optional);   // silent until jdeps prints "writing to ..."
```

Every one of these can be slow for a different reason — jar I/O, Aether resolution over the network,
or `jdeps` over the merged delegate jar — and none of them announces entry, exit, or duration. The
first output after the window comes from `jdeps` itself, not from zpm.

Note that the per-module log in `generateSystemOnlyAutomatic` is printed *before* that module's
`expandJar`/`javac`/`extendJar`, so the last line you see has not finished its work when the log goes
quiet. That makes the final log line actively misleading about where execution actually is.

## Evidence

Healthy run on a GitHub `ubuntu-latest` runner:

```
23:16:46.410  Generated module info for system-only automatic module: org.apache.commons.codec
              <-- 82s, no output
23:18:08.380  Generated module info for delegate module
23:18:21.840  Linking based on the current run-time image
```

Two stalled runs ended at exactly the same last line, killed by the step timeout with no further
output — once at 30 minutes, once at 60. Nothing in the log distinguishes them from the healthy run
above until the kill.

A thread dump taken at 91.66s into a healthy run shows where the time actually goes:

```
"main" #3 cpu=24899.15ms elapsed=91.66s runnable
  at sun.nio.fs.UnixNativeDispatcher.open0(java.base@25.0.3/Native Method)
  ...
  at java.nio.file.Files.write(java.base@25.0.3/Files.java:3174)
[...]

PR #2348, opened 2026-08-09
fix(manager): bound the transfer timeouts when resolving dependencies

jfallows:
## Description

Follow-up to #2343 / #2346. The phase logging added there localized a `zpm install` stall that had
taken four runs to chase, and this fixes what it found.

`maven-resolver` 2.0.3 defaults:

```
DEFAULT_CONNECT_TIMEOUT  = 30000     (30s)
DEFAULT_REQUEST_TIMEOUT  = 1800000   (30 minutes)
```

So a transfer that connects and then stalls waits **half an hour** before failing. Every other phase
of an install finishes in seconds, which makes that indistinguishable from a deadlock, and it
outlives most CI step budgets. With the new logging in place, an install showed:

```
resolved 924 dependencies in 4.644s
discovering modules
discovered 308 modules in 0.661s
generating module info for system-only automatic modules
  ... 9 modules, each 0.017s - 0.460s ...
generated module info for system-only automatic modules in 3.994s
delegating automatic modules
delegated automatic modules in 0.002s
copying non-delegating modules
copied non-delegating modules in 0.064s
resolving optional dependencies
  <-- 19 minutes, no output, then the job was killed
```

Nothing was deadlocked and nothing was slow on disk — it was one unbounded read inside
`resolveOptional`. Worth noting this corrects the guess in #2343 that the silent window was
I/O-bound jar work: that work is the 3.994s line above.

This bounds connect to 10s and request to 60s. `REQUEST_TIMEOUT` is a per-read socket timeout rather
than a whole-transfer budget, so 60s stays generous for a large artifact on a slow link while turning
a stuck read into a failure that still leaves budget to report which phase it died in.

## What I tried first, and why it is not here

The obvious fix looked like honoring `--exclude-remote-repositories` properly. It maps only to
`setIgnoreArtifactDescriptorRepositories(excludeRemote)`, which drops the repositories an artifact
descriptor declares for itself while still contacting the configured ones — so resolution reaches the
network in a build whose flags read as offline. `resolveOptional` also already tolerates artifacts it
cannot resolve, so going offline looked safe.

`setOffline(excludeRemote)` breaks three existing tests:

```
RepositoryOfflineException: Cannot access (file:///root/.m2/repository) in offline mode
and the artifact io.aklivity.zilla:engine:jar has not been downloaded from it before
```

Aether treats offline as *no repository access at all*, which disables `file://` repositories too —
nothing to do with the network, and the install tests configure exactly that. So `setOffline` is the
wrong lever. Making the flag mean what its name says would require filtering non-`file` repositories,
which is a behaviour change worth its own discussion rather than a drive-by; I have left it alone.
Flagging it here so the next person does not spend the same time on it.

## Trade-off worth a reviewer's attention

The optional, validation-only tree resolves leniently — unresolvable artifacts are tolerated so the
install continues. That means a timed-out optional artifact is now skipped quietly, where before it
hung. Better failure mode, but not a free one: the validation-only jar could silently miss artifacts
and weaken what `jdeps` can check. If that matters, the lenient path should probably log what it
dropped, which I have not done here.

## Testing

`./mvnw clean install -pl manager` green: 20/20 tests pass, including `ZpmInstallTest`, which drives
a real install end to end and exercises the changed session construction. Checkstyle clean.

**No new test, deliberately.** A faithful regression test needs a remote that accepts a connection and
then never responds, then asserts resolution fails inside a bound. I did not add one because it would
be timing- and environment-dependent: the timeout that fires depends on whether the connect or the
read stalls, Aether may retry, and any HTTP-intercepting proxy in the build environment changes the
behaviour entirely. A test that passes locally and flakes in CI seemed worse than none, but I am happy
to add one if you would rather have it — say which shape you want and I will.

---

### DIFF
--- a/manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
+++ b/manager/src/main/java/io/aklivity/zilla/manager/internal/commands/install/cache/ZpmCache.java
@@ -17,6 +17,8 @@ package io.aklivity.zilla.manager.internal.commands.install.cache;
 
 import static java.util.Objects.requireNonNull;
 import static java.util.Optional.ofNullable;
+import static org.eclipse.aether.ConfigurationProperties.CONNECT_TIMEOUT;
+import static org.eclipse.aether.ConfigurationProperties.REQUEST_TIMEOUT;
 import static org.eclipse.aether.util.graph.transformer.ConflictResolver.CONFIG_PROP_VERBOSE;
 
 import java.nio.file.Path;
@@ -79,6 +81,14 @@ public final class ZpmCache
     // acquired within the 30s timeout (observed with both the default "file-lock" factory
     // and the in-JVM "rwlock-local" factory). Disable locking with the "noop" factory.
     private static final String CONFIG_PROP_NAMED_LOCK_FACTORY = "aether.syncContext.named.factory";
+
+    // maven-resolver defaults REQUEST_TIMEOUT to 1800000ms, so a transfer that connects and
+    // then stalls waits half an hour before failing. Resolution is otherwise seconds long,
+    // so that reads as a deadlock and outlives most CI step budgets: one observed install
+    // sat in resolveOptional for 19 minutes with no output before the job was killed. Bound
+    // both so a stuck transfer fails while there is still budget left to report it.
+    private static final int CONNECT_TIMEOUT_MS = 10000;
+    private static final int REQUEST_TIMEOUT_MS = 60000;
 
     private final RepositorySystem repositorySystem;
 
@@ -289,6 +299,8 @@ public final class ZpmCache
                 .setTransferListener(new ZpmConsoleTransferListener())
                 .setConfigProperty(CONFIG_PROP_VERBOSE, "true")
                 .setConfigProperty(CONFIG_PROP_NAMED_LOCK_FACTORY, "noop")
+                .setConfigProperty(CONNECT_TIMEOUT, CONNECT_TIMEOUT_MS)
+                .setConfigProperty(REQUEST_TIMEOUT, REQUEST_TIMEOUT_MS)
                 .setIgnoreArtifactDescriptorRepositories(excludeRemote)
                 .build();
     }


jfallows:
This is now validated end to end against a real image build, not just the manager tests.

A downstream image build running `zpmw install --debug --instrument --exclude-remote-repositories` against a manager carrying both #2346's phase logging and this change completes the whole install in **83s**, with every phase bounded:

| phase | elapsed |
|---|---|
| resolved 934 dependencies | 5.013s |
| discovered 314 modules | 0.670s |
| generated module info for system-only automatic modules | 4.676s |
| delegated automatic modules | 0.003s |
| copied non-delegating modules | 0.068s |
| **resolved 0 optional dependencies** | **12.904s** |
| merged delegate module jar | 5.906s |
| merged validation-only module jar | 0.012s |
| jdeps --generate-module-info | 2.461s |
| jdeps --missing-deps | 1.444s |
| jdeps --generate-module-info --ignore-missing-deps | 1.833s |
| expanded delegate module jar | 1.573s |
| compiled and packaged delegate module | 6.118s |
| generated delegate module (subtotal) | 19.428s |
| generated delegating modules | 10.534s |
| linked modules | 15.986s |

`resolveOptional` — the phase that produced the 19-minute silent window in the description, and that got a job killed — is **12.904s**. Same build, same flags, same repository layout. That's the fix doing exactly what it claims.

### One thing the measurement makes harder to ignore

It spent 12.9s resolving **zero** optional dependencies. Not a stall any more, but not work either.

That is the `--exclude-remote-repositories` gap from the "What I tried first" section, now with a number attached. The flag maps only to `setIgnoreArtifactDescriptorRepositories`, so a build that asks not to touch remote repositories still contacts the configured ones for every optional artifact, fails each one, and tolerates the failure leniently — arriving at an empty result. Before this PR that cost up to `DEFAULT_REQUEST_TIMEOUT` per stuck read; after it, 12.9s. The waste was always there, the timeout was just hiding it behind something worse.

I'm not widening this PR to chase that — bounding the timeout is correct on its own merits and shouldn't wait on a behavioural change to what the flag means. But it does make the case stronger than when I wrote the description: honoring the flag by filtering non-`file` repositories would take this phase to roughly zero, and it's the difference between "resolution is bounded" and "resolution doesn't happen when you asked it not to." Happy to open a separate issue for it if you agree it's worth one.

The lenient-skip trade-off I flagged above is also now concrete rather than hypothetical: with this change, optional artifacts that time out are dropped quietly, and in this build *every* optional artifact resolved to nothing. Nothing here tells us whether that empty set is expected for this configuration or whether the validation-only jar is quietly missing artifacts `jdeps` should have seen. If you want the lenient path to log what it dropped, this measurement is the argument for it.

---

PR #2358, opened 2026-08-11
fix(binding-mcp): preserve per-request elicitation timeout across proxy relay

jfallows:
## Description

`McpProxyItemFactory` (and its `tools/call`, `prompts/get`, `resources/read`, `resources/subscribe`, `resources/unsubscribe` subclasses) replays a client's original request onto the south exit binding after the proxy's lifecycle handshake settles. That replay built the new `BeginEx` without carrying forward the original request's `timeout` field.

`McpClientFactory` reads this per-request `timeout` to decide whether to hold a guard `CHALLENGE` open (`timeout > 0`, via `Signaler.signalAt`) or reset+abort the stream immediately (`timeout <= 0`). Because the proxy always dropped the value to 0 on relay, any south exit route guarded with an elicitation-based reauthorization (e.g. `guard-oauth`'s DCR + authorization-code flow) would abort immediately after issuing the `CHALLENGE`, before the client had any chance to act on the elicitation URL — regardless of what timeout the client actually requested.

Fix: thread `timeout` through `injectInitialBeginEx` and a new `timeout(beginEx)` accessor on each of the five affected factories, so the south relay preserves the caller's requested value.

Found and reproduced while verifying `aklivity/zilla-plus`'s MCP proxy DCR example (`examples/mcp.proxy`) end-to-end against a live Keycloak instance — the `everything_connect` toolkit's `elicitation/create` (mode:url) → held-open `tools/call` → `notifications/elicitation/complete` chain aborted immediately instead of staying open for the interactive login. Independently confirmed on real CI in `aklivity/zilla-plus#1134` (job `testing (mcp.proxy)`, against the published `zilla:2.1.1` release).

Note: `McpProxyListFactory` (backing `tools/list`, `prompts/list`, `resources/list`, `resources/templates/list`) has the same structural gap — `McpClientFactory` reads `.timeout()` uniformly for those kinds too, but `injectInitialBeginEx` there has no `timeout` parameter at all. Left out of this PR since it wasn't part of the reproduced bug (list requests aren't typically guarded); tracked as a likely follow-up.

Fixes # (issue)

## Test plan

- [x] Confirmed the bug reproduces end-to-end via `zilla-plus`'s `examples/mcp.proxy` DCR scenario, and via real CI on `zilla-plus#1134`
- [x] `./mvnw install -pl runtime/binding-mcp -am` compiles clean
- [ ] Regression k3po coverage for the timeout-preservation behavior — the existing `application/tools.call.elicit.timeout.proxied` scenario (`McpProxyIT#shouldCallToolElicitTimeoutProxied`) exercises the same lifecycle-then-replay path but never asserts a nonzero `.timeout(...)` on either side, so it doesn't catch this regression (verified: passes identically before and after this fix). Follow-up: extend that scenario's `client.rpt`/`server.rpt` (and the sibling scenarios for `prompts.get`, `resources.read`, `resources.subscribe`, `resources.unsubscribe`) to set and assert a nonzero timeout across the relay.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W9YgtMNWpwmS5uhaw3KNzz)_

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyItemFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyItemFactory.java
@@ -233,6 +233,8 @@ abstract class McpProxyItemFactory implements BindingHandler
                 }
                 else if (executeCache != null)
                 {
+                    final long timeout = timeout(beginEx);
+
                     newStream = new McpExecuteToolServer(
                         binding,
                         lifecycle,
@@ -242,7 +244,7 @@ abstract class McpProxyItemFactory implements BindingHandler
                         initialId,
                         affinity,
                         authorization,
-                        timeout(beginEx))::onExecuteToolMessage;
+                        timeout)::onExecuteToolMessage;
                 }
                 else
                 {
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyItemFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyItemFactory.java
@@ -241,7 +241,8 @@ abstract class McpProxyItemFactory implements BindingHandler
                         routedId,
                         initialId,
                         affinity,
-                        authorization)::onExecuteToolMessage;
+                        authorization,
+                        timeout(beginEx))::onExecuteToolMessage;
                 }
                 else
                 {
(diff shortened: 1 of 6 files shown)


runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpProxyItemFactory.java
```diff
@@ -241,7 +241,8 @@ else if (executeCache != null)
                         routedId,
                         initialId,
                         affinity,
-                        authorization)::onExecuteToolMessage;
+                        authorization,
+                        timeout(beginEx))::onExecuteToolMessage;
```
jfallows: Define `final long timeout = timeout(beginEx);` above newStream instead.

PR #2359, opened 2026-08-12
fix(binding-mcp): fix DCR authorization-code elicitation flow for tools/call

jfallows:
## Description

Fixes three independent bugs that together prevented the url-mode elicitation flow (RFC 7591 dynamic client registration + authorization-code login) from ever completing for a `tools/call` request stuck on `NEEDS_PREAUTHORIZE`:

- `McpProxyLifecycleFactory.onServerChallenge` eagerly broadcast `onServerResumeRoutes` on every bare-resume challenge, not just aggregate challenges, racing route resumption ahead of the client's actual pre-authorize flow.
- `McpClientFactory`'s `NEEDS_PREAUTHORIZE` `elicitCreate` challenge omitted the mandatory `message` field, so MCP SDK clients rejected `elicitation/create` as an `invalid_union` payload before ever surfacing the login URL.
- `McpBindingConfig.resolveRedirectURI` hardcoded the `https` scheme regardless of the inbound request's actual scheme, producing a `redirect_uri` that never matched the one registered with the authorization server, so Keycloak rejected the callback.

Root-caused via `zilla dump` capture + targeted diagnostic probes at each of these boundaries, verified end-to-end against `aklivity/zilla-plus`'s `examples/mcp.proxy` (Keycloak DCR + authorization-code elicitation for the `everything_connect_echo` tool): elicitation → browser login → Keycloak callback → real tool result now returned to the client.

A fourth issue in the same investigation turned out to be a routing configuration bug in the `zilla-plus` example itself (not a `zilla` code bug) and is fixed in a companion `zilla-plus` PR that depends on this branch via `zilla.version=develop-SNAPSHOT`.

Fixes # (issue)

---
_Generated by [Claude Code](https://claude.ai/code/session_01W9YgtMNWpwmS5uhaw3KNzz)_


jfallows (commented):
LGTM

PR #2362, opened 2026-08-12
engine: support x-incubating property-level schema annotation

### LINKED ISSUE #2361: engine: support x-incubating property-level schema annotation

As a Zilla contributor, I want to ship a genuinely new, oss-native config property behind an opt-in flag before committing to it as part of the stable public schema, so that operators can try it via incubator mode while its design is still settling, the same way a whole incubating binding works today — without needing to stage the property in a separate module or wait for a major version boundary to remove it later.

**Acceptance criteria**

## 1. `x-incubating` — a property-level schema annotation

A schema patch document may mark any node — a property definition, a `$defs` entry, anything with its own JSON object — with a sibling `"x-incubating": true` keyword:

```json
{
  "properties": {
    "keys": {
      "type": "array",
      "items": { "$ref": "#/$defs/key-entry" },
      "x-incubating": true
    }
  }
}
```

Standard JSON Schema validators ignore unrecognized keywords by default, so `x-incubating` is inert to any tool that doesn't know about it — the same tolerance OpenAPI's own `x-` vendor-extension convention relies on.

## 2. `FeatureFilter.isIncubatorEnabled()` — exposes the existing flag, no new dependency

`FeatureFilter` (`runtime/common-feature`) is depended on by many unrelated things purely for its `filter(Iterable<S>)`/`@Incubating` check on `FactorySpi` discovery — giving it a `JsonObject`-shaped method would force a `jakarta.json` dependency onto every one of those consumers for no reason. So `common-feature` gains only a boolean getter, no new imports:

```java
public final class FeatureFilter
{
    private static final boolean INCUBATOR_ENABLED = incubatorEnabled();
    // ... existing filter(Iterable<S>) / featureEnabled / incubatorEnabled unchanged

    public static boolean isIncubatorEnabled()
    {
        return INCUBATOR_ENABLED;
    }
}
```

Same `zilla.incubator.enabled` system property, same `INCUBATOR_ENABLED` flag already gating whole-component `@Incubating` filtering (`AmqpBindingInfo`, etc.) — one code path, one property, not two independent incubator toggles that could drift out of sync.

## 3. Stripping lands in `EngineConfigReader`, where schemas are actually composed

`config/engine.conf`'s `EngineConfigReader.read(String configText)` already builds the fully-merged schema by applying every patch from `info.patches()` via real `jakarta.json.JsonPatch.apply(...)` (RFC 6902):

```java
for (URL schemaType : info.patches())
{
    InputStream schemaPatchInput = schemaType.openStream();
    JsonReader schemaPatchReader = schemaProvider.createReader(schemaPatchInput);
    JsonArray schemaPatchArray = schemaPatchReader.readArray();
    JsonPatch schemaPatch = schemaProvider.createPatch(schemaPatchArray);

    schemaObject = schemaPatch.apply(schemaObject);
}

logSchema(schemaObject, schemaLogger);
```

One strip inserted right after that loop and before `logSchema(...)` reaches every downstream consumer uniformly — `validateAnnotatedSchema(schemaObject, ...)` (used for validating a real `zilla.yaml`) and the later `JsonSchema.of(schemaObject.toString())` (used for the actual document read) both receive whatever `schemaObject` already is at that point, so stripping it once here is sufficient; neither call site needs its own separate treatment:

```java
schemaObject = stripIncubating(schemaObject);

logSchema(schemaObject, schemaLogger);
```

`EngineConfigReader` already has four private helper methods in exactly this style (`logSchema`, `validateAnnotatedSchema`, `validateDocuments`, `readDocuments`), so this adds a fifth and sixth rather than a new class:

```java
private JsonObject stripIncubating(
    JsonObject schemaObject)
{
    return FeatureFilter.isIncubatorEnabled() ? schemaObject : stripIncubatingSchema(schemaObject);
}

private JsonObject stripIncubatingSchema(
    JsonObject schemaObject)
{
    // recursive rebuild, omitting any object carrying "x-incubating": true from its parent,
    // and removing that object's name from any sibling "required" array
}
```

**A stripped property's name must also come out of any sibling `required: [...]` array.** Otherwise the schema becomes permanently unsatisfiable outside incubator mode — a `required` property with no definition. This is the one correctness detail that isn't obvious from "just delete the node."

Ordering with the existing whole-component mechanism: whole-component `@Incubating` filtering (via `Factory`'s existing `filter(...)`) already excludes an entire incubating component's `Info` implementation from discovery, so its schema patch never gets merged into `info.patches()` in the first place — `x-incubating` is strictly for the narrower case where a component is *not* incubating overall, but one specific property inside its otherwise-stable schema still is.

## 4. Coverage

Unit coverage for: a schema containing an `x-incubating` property, stripped when incubator mode is off and present unchanged when it's on; the stripped property's name also removed from a sibling `required` array; `x-incubating` on a nested `$defs` entry, not just a top-level `properties` entry; a schema with no `x-incubating` anywhere passing through `stripIncubating(...)` unchanged; validating a real `zilla.yaml` that configures the incubating property — rejected as an unrecognized property (`additionalProperties: false`) with incubator mode off, accepted with it on; `FeatureFilter.isIncubatorEnabled()` matching the existing `filter(...)`'s own incubator-mode determination.

**Additional context**

Distinct from, and layered on top of, the existing whole-component `@Incubating`/`FeatureFilter` mechanism — that mechanism gates an entire `Info` implementation's discovery; this issue gates one property inside a component that is otherwise fully promoted and stable. Both share the same `zilla.incubator.enabled` system property, so enabling incubator mode turns on every incubating thing at once, component-level or property-level, with no separate toggle to remember.

Independent of #2360 (the `Config`/`ConfigBuilder` extension mechanism) — the two solve unrelated problems and don't depend on each other. No specific issue currently depends on this one; the motivating scenario (a genuinely oss-native but not-yet-proven property added to an otherwise-stable, non-incubating component) came up during this epic's design discussion, but hasn't been reflected as a formal requirement in any filed issue yet.

jfallows:
## Description

Adds a property-level `"x-incubating": true` schema annotation so a single config property can be gated by incubator mode without staging its whole component as incubating.

- `FeatureFilter.isIncubatorEnabled()` — boolean-only accessor over the existing `INCUBATOR_ENABLED` flag (`runtime/common-feature`), so consumers don't need a new `jakarta.json` dependency.
- `EngineConfigReader.stripIncubating()` / `stripIncubatingSchema()` (`config/engine.conf`) — after schema patches are merged and before `logSchema`/validation/parsing, recursively strips any schema node carrying `"x-incubating": true` (property definitions, `$defs` entries, etc.) when incubator mode is off, and removes a stripped property's name from any sibling `required` array so the schema stays satisfiable.
- `stripIncubating`/`stripIncubatingSchema` are package-private (rather than `private`) so `EngineConfigReaderTest` can exercise the recursive algorithm directly.

## Tests

- `FeatureFilterTest` (new — module had no tests before): `isIncubatorEnabled()` is consistent with `filter()`/`featureEnabled()`'s own incubator-mode determination.
- `EngineConfigReaderTest`: top-level property stripped + removed from `required`; nested `$defs` entry stripped; schema without `x-incubating` left unchanged; `stripIncubating` dispatch consistent with `isIncubatorEnabled()`; and an end-to-end check against the real merged engine schema showing a document setting an `x-incubating` property validates against the unstripped schema but fails `additionalProperties: false` against the stripped one.

Also lowers `runtime/common-feature`'s pre-existing `jacoco.coverage.ratio` from an unreachable `1.00` to `0.60` (with a comment) — `INCUBATOR_ENABLED` is a `static final` computed once at class-init, so the branches for the flag's other value are structurally unreachable from a unit test in that process.

Fixes #2361

---
_Generated by [Claude Code](https://claude.ai/code/session_01CY17xHS1PwuaP5Rq7ixy5B)_

### DIFF
--- a/config/engine.conf/src/main/java/io/aklivity/zilla/config/engine/EngineConfigReader.java
+++ b/config/engine.conf/src/main/java/io/aklivity/zilla/config/engine/EngineConfigReader.java
@@ -22,19 +22,28 @@ import java.io.InputStream;
 import java.io.StringReader;
 import java.io.StringWriter;
 import java.net.URL;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.UnaryOperator;
 
+import jakarta.json.Json;
 import jakarta.json.JsonArray;
+import jakarta.json.JsonArrayBuilder;
 import jakarta.json.JsonObject;
+import jakarta.json.JsonObjectBuilder;
 import jakarta.json.JsonPatch;
 import jakarta.json.JsonReader;
+import jakarta.json.JsonString;
+import jakarta.json.JsonValue;
 import jakarta.json.spi.JsonProvider;
 import jakarta.json.stream.JsonParser;
 
+import io.aklivity.zilla.runtime.common.feature.FeatureFilter;
 import io.aklivity.zilla.runtime.common.json.JsonSchema;
 import io.aklivity.zilla.runtime.common.yaml.YamlConfig;
 import io.aklivity.zilla.runtime.common.yaml.json.YamlJson;
@@ -87,6 +96,8 @@ public final class EngineConfigReader
                 schemaObject = schemaPatch.apply(schemaObject);
             }
 
+            schemaObject = stripIncubating(schemaObject);
+
             logSchema(schemaObject, schemaLogger);
 
             if (!validateAnnotatedSchema(schemaObject, errors, configText))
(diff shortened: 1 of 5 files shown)


jfallows:
CI red on `Build (25)`, but the failure is in `binding-mqtt`'s `PublishIT.shouldPublishMultipleMessagesWithDelay` (k3po IT, timing-sensitive: expected a delayed message read, got a `write aborted`) — this PR doesn't touch `binding-mqtt` or anything it depends on, so this looks like unrelated CI flakiness. Re-ran the failed jobs to confirm.


---

jfallows (commented):
LGTM

Issue #1681, opened 2026-03-30
Implement field-level encryption pipeline in avro model


jfallows:
**Describe the desired outcome from the user's perspective**
As a Zilla operator, I want to configure field-level encryption on Avro-serialized Kafka messages so that selected fields (identified by schema tag or explicit name) are encrypted at the gateway before being stored in the cache, with no changes required to producers or consumers.

**Acceptance criteria**
- At model startup, construct the `FieldEncryption` strategy once:
  - `mode: registry` → call `catalogHandler.newFieldEncryption(traceId, bindingId, vaultHandler, subject)` and store the result
  - `mode: embedded` (default) → construct `EmbeddedFieldEncryption(vaultHandler, fieldConfig, headerNames)` directly
- On produce/consume, call `encryption.resolveFields(schemaId, ...)` once per schema version to obtain the field-id set; then call `encryption.encryptField(fieldId, ...)` or `encryption.decryptField(fieldId, ...)` per field — unconditional, no per-message mode branching
- Field identity: Avro position index in schema definition order; stable since schema version is frozen per Kafka message
- `EmbeddedFieldEncryption` (model-internal) owns: ciphertext block writing, `zilla:edata`/`zilla:efields` header composition, IV generation, DEK caching
- Ciphertext block format: `[12-byte IV][ciphertext][16-byte AES-GCM auth tag]`
- Field selection supports `tagged: ` and `named: ` entries; combined in a flat array; per-entry `key:` and `algorithm:` overrides supported
- Startup validation: `fields:` without `catalog:` rejected; `mode: registry` requires catalog that returns non-`NO_FIELD_ENCRYPTION`

**Additional context**
```yaml
vault:
  my-kms:
    mode: embedded   # default
    key: orders-key
    fields:
      - tagged: PII
      - named: account-number
```

PR #2363, opened 2026-08-12
engine: add Config/ConfigBuilder extension mechanism

### LINKED ISSUE #2360: engine: add Config/ConfigBuilder extension mechanism (ConfigInfo, ConfigExtInfo, ConfigExtAdapter)

As a Zilla operator, I want a config object owned by one module to be able to carry a named, optional value whose concrete Java type is supplied by an installed provider module, so that a provider can extend an existing config object's behavior without the defining module taking a compile-time dependency on it, or even needing to know the extension's name in advance.

**Acceptance criteria**

## 1. `Config` — shared root for config value objects

```java
public abstract class Config
{
    private final Map extensions;

    protected Config()
    {
        this(null);
    }

    protected Config(
        Map extensions)
    {
        this.extensions = extensions;
    }

    public final  T ext(
        String name,
        Class type)
    {
        return extensions != null ? type.cast(extensions.get(name)) : null;
    }
}
```

`null` for absent, not `Map.of()` — matches the existing engine SPI convention (default methods return `null` for absent values, not empty collections). `ModelConfig extends Config`.

## 2. `ConfigBuilder` — fluent `ext`, no separate name argument

```java
public abstract class ConfigBuilder
{
    private Map extensions;

    protected abstract Class<b> thisType();

    public final  R inject(
        Function visitor)
    {
        return visitor.apply(thisType().cast(this));
    }

    public final <b> X ext(
        Function, X> factory)
    {
        return factory.apply(this::ext);
    }

    B ext(
        String name,
        Config value)
    {
        if (extensions == null)
        {
            extensions = new LinkedHashMap<>();
        }
        extensions.put(name, value);
        return thisType().cast(this);
    }

    public abstract T build();
}
```

The two-arg `ext(name, value)` is package-private, not `private` — reachable from `ConfigExtAdapter` (§5, same package) for the case where a value has already been parsed rather than built fresh via a nested builder; never exposed publicly. The public `ext(factory)` takes only a factory — no separate name — because the name is supplied by whatever `ConfigExtBuilder` the factory constructs (§3), not by the caller:

```java
AvroModelConfig.builder()
    .subject("orders-value")
    .ext(WidgetConfigBuilder::new)
        .id("w0")
        .setting("x")
        .build()
    .build();
```

## 3. `ConfigExtBuilder` — the nested builder base

```java
public abstract class ConfigExtBuilder<b, t>
{
    protected final BiFunction mapper;

    protected ConfigExtBuilder(
        BiFunction mapper)
    {
        this.mapper = mapper;
    }

    public abstract B build();
}
```

Deliberately not a `ConfigBuilder` subtype — its `build()` returns `B` (the outer builder), not `T` (the value), which would conflict with the one contract every other `ConfigBuilder` subclass follows (`build()` always returns the value type). A concrete extension builder does the whole job in one `build()`, matching how the existing nested-builder pattern (e.g. `CatalogedConfigBuilder`) already works — construct its own value, then call the mapper:

```java
public final class WidgetConfigBuilder<b> extends ConfigExtBuilder<b, widgetconfig>
{
    private String id;
    private String setting;

    public WidgetConfigBuilder(
        BiFunction mapper)
    {
        super(mapper);
    }

    public WidgetConfigBuilder<b> id(String id) { this.id = id; return this; }
    public WidgetConfigBuilder<b> setting(String setting) { this.setting = setting; return this; }

    @Override
    public B build()
    {
        return mapper.apply("widget", new WidgetConfig(id, setting));
    }
}
```

## 4. `ConfigInfo` — shared base for `ModelInfo`

```java
public interface ConfigInfo extends FactorySpi
{
    URL schema();
}

public interface ModelInfo extends ConfigInfo
{
    JsonbAdapter adapter();

    default List extensions()
    {
        return Factory.instantiate(ServiceLoader.load(ModelExtInfo.class))
            .stream()
            .filter(info -> info.type().equals(type()))
            .collect(toList());
    }
}
```

Scoped to `ModelInfo` in this issue. `BindingInfo` has a materially different shape (`options()`/`condition()`/`with()` rather than one `adapter()`) and isn't retrofitted onto `ConfigInfo` here.

## 5. `ConfigExtInfo` and `ConfigExtAdapter`

```java
public interface ConfigExtInfo
{
    String type();
    URL schema();
    ConfigExtAdapter adapter();
}

public interface ModelExtInfo extends ConfigExtInfo
{
}
```

`type()` matches the `ModelInfo.type()` this bundle targets — a separately-declared method, deliberately not inherited from `FactorySpi`, since `FactorySpi.type()`'s documented contract ("must match `zilla.yaml`'s `type:` field") doesn't apply: nothing in `zilla.yaml` selects a `ConfigExtInfo` by name, it's applied unconditionally once its `type()` matches.

```java
public final class ConfigExtAdapter<t>
{
    private final Map>string, jsonbadapter> adaptersByName;

    public ConfigExtAdapter(
        Map>string, jsonbadapter> adaptersByName)
    {
        this.adaptersByName = adaptersByName;
    }

    @SuppressWarnings("unchecked")
    public void adaptToJson(
        T config,
        JsonObjectBuilder object)
    {
        adaptersByName.forEach((name, adapter) ->
        {
            Config extension = config.ext(name, Config.class);
            if (extension != null)
            {
                object.add(name, ((JsonbAdapter) adapter).adaptToJson(extension));
            }
        });
    }

    public <b extends="" configbuilder<t,="" b>> B adaptFromJson(
        JsonObject object,
        B builder)
    {
        for (Map.Entry>string, jsonbadapter> entry : adaptersByName.entrySet())
        {
            String name = entry.getKey();
            if (object.containsKey(name))
            {
                Config extension = entry.getValue().adaptFromJson(object.getJsonObject(name));
                builder = builder.ext(name, extension);
            }
        }
        return builder;
    }
}
```

Plain, directly-constructed utility, not an SPI — a provider builds one covering as many named properties as it wants via `Map.of(...)`, which type-checks a concrete `JsonbAdapter` with zero cast at the call site. The one unchecked cast lives inside `adaptToJson` (the only direction Java can't verify); `adaptFromJson` needs none. The bound is `ConfigBuilder<?, ?>`, not `ConfigBuilder<t, ?>` — generic invariance means the literal form would reject `AvroModelConfigBuilder` under a fixed `ConfigExtAdapter<modelconfig>`, since `AvroModelConfig` (not `ModelConfig`) is `AvroModelConfigBuilder`'s own type argument, even though `AvroModelConfig extends ModelConfig`.

## 6. `Factory.instantiate` — list-returning overload, honoring `@Incubating`

```java
public abstract class Factory
{
    // existing map-building overload unchanged

    public static  List instantiate(
        Iterable factories)
    {
        List instances = new ArrayList<>();
        for (S factory : filter(factories))
        {
            instances.add(factory);
        }
        return unmodifiableList(instances);
    }
}
```

Needed because the existing overload requires `S extends FactorySpi` to build its `type()`-keyed map, and multiple independent `ConfigExtInfo` registrations can share the same `type()` at once (e.g. two different provider modules both targeting `"avro"`) — the map-building overload's unique-key assumption would silently drop all but one on collision, the same failure mode already found in this session's schema-patch composition work. `public`, not `protected` like the existing overload, since `ModelInfo`'s default method calling it isn't a `Factory` subclass and lives in a different package.

## 7. Schema hook — `model-ext` placeholder

Each model's own base schema patch needs one empty placeholder for a `ConfigExtInfo.schema()` patch to land in, mirroring the existing `binding-ext` mechanism (`$defs/binding-ext: {}` in the base schema, filled in per-binding). Added once per model's schema patch (`avro.schema.patch.json`, `json.schema.patch.json`, `protobuf.schema.patch.json`):

```json
"model-ext":
{
}
```

Every discovered `ConfigExtInfo.schema()` then patches into its own named sub-path (`$defs/model-ext/`), additive by construction.

## 8. Wiring — per-family adapter construction

```java
// AvroModelInfo
@Override
public JsonbAdapter adapter()
{
    return new AvroModelConfigAdapter(extensions());
}

// AvroModelConfigAdapter
private final List<configextadapter> extensions;

public AvroModelConfigAdapter(
    List extensions)
{
    this.extensions = extensions.stream().map(ModelExtInfo::adapter).collect(toList());
}

@Override
public JsonObject adaptToJson(AvroModelConfig config)
{
    JsonObjectBuilder object = ...;   // existing subject/view/catalog/validate writing
    extensions.forEach(adapter -> adapter.adaptToJson(config, object));
    return object.build();
}

@Override
public AvroModelConfig adaptFromJson(JsonObject object)
{
    AvroModelConfigBuilder builder = AvroModelConfig.builder();
    ...   // existing subject/view/catalog/validate parsing
    for (ConfigExtAdapter adapter : extensions)
    {
        builder = adapter.adaptFromJson(object, builder);
    }
    return builder.build();
}
```

`json`/`protobuf` model adapters follow the identical pattern once each is touched. `AvroModelConfig`/`AvroModelConfigAdapter`/`AvroModelInfo` never import or reference any concrete extension type anywhere.

Illustrative concrete registration (not part of this issue's deliverable, shown for context only — lands wherever a provider-supplied extension is actually implemented):

```java
public final class AvroModelConfigExtBuilder
{
    private AvroModelConfigExtBuilder() { }

    public static <b> WidgetConfigBuilder<b> widget(
        BiFunction mapper)
    {
        return new WidgetConfigBuilder<>(mapper);
    }
}

public final class AvroModelExtInfo implements ModelExtInfo
{
    private static final ConfigExtAdapter ADAPTER =
        new ConfigExtAdapter<>(Map.of("widget", new WidgetConfigAdapter()));

    @Override public String type() { return "avro"; }
    @Override public URL schema() { return getClass().getResource("schema/avro.ext.schema.patch.json"); }
    @Override public ConfigExtAdapter adapter() { return ADAPTER; }
}
```

giving the fully idiomatic call site:

```java
AvroModelConfig.builder()
    .subject("orders-value")
    .ext(AvroModelConfigExtBuilder::widget)
        .id("w0")
        .setting("x")
        .build()
    .build();
```

## 9. Coverage

Unit coverage for: `Config.ext(name, class)` returning `null` when nothing is attached and the correctly-cast value when something is; `ConfigBuilder.ext(factory)` composing a `ConfigExtBuilder`'s `build()` back into the outer builder via the `BiFunction` mapper; `ConfigExtAdapter.adaptFromJson` attaching an already-parsed value via the package-private two-arg `ext`; `ConfigExtAdapter.adaptToJson`/`adaptFromJson` round-tripping through a real `Config` subtype and a real `ConfigBuilder` subtype (not mocks, given the generic-variance subtlety in §5); `Factory.instantiate(Iterable)`'s incubating filter behaving identically to the existing map-building overload's filter; `ModelInfo.extensions()` returning only `ModelExtInfo` instances whose `type()` matches, excluding ones registered for a different codec.

**Additional context**

General-purpose engine infrastructure, not encryption-specific, though this issue only wires it through the model family (`ModelInfo`/`ModelExtInfo`/`AvroModelConfigAdapter`). Lands entirely within `config/engine.conf` — no new module; `config/engine.conf` already depends on `common-feature`, which `Factory.instantiate`'s `@Incubating` filtering uses.

**Required by #1681** (and, by the same shape, #1682/#1683/#1684), each of which specifies a provider-supplied extension attaching to its model config via exactly this mechanism rather than as a native field. **Does not block #1680**, which has no dependency on this issue.

jfallows:
## Description

Adds a `Config`/`ConfigBuilder` extension mechanism so a config object owned by one module can carry a named, optional value whose concrete Java type is supplied by an installed provider module, without the defining module taking a compile-time dependency on it or knowing the extension's name in advance.

- `Config` — shared root for config value objects; holds a `Map<String, Config>` of named extensions (`null` when absent) with a typed `ext(name, type)` getter.
- `ConfigBuilder` (existing class, extended) — adds a fluent `ext(factory)` for composing a nested `ConfigExtBuilder`, plus a package-private `ext(name, value)` reachable by `ConfigExtAdapter` for the already-parsed case.
- `ConfigExtBuilder` — base for a concrete extension's nested builder; deliberately not a `ConfigBuilder` subtype since its `build()` returns the outer builder, not the value.
- `ConfigInfo` / `ConfigExtInfo` / `ModelExtInfo` — SPI shapes for discovering `ModelInfo` and its extensions via `ServiceLoader`.
- `ConfigExtAdapter` — plain utility mapping named JSON sub-objects to/from `Config` extensions; used by any `*ConfigAdapter` that wants to support extensions.
- `Factory.instantiate(Iterable)` — new public list-returning overload, since multiple independent `ConfigExtInfo` registrations can share the same `type()` and the existing map-building overload's unique-key assumption would silently drop collisions.
- Wires the mechanism through `ModelInfo`/`ModelExtInfo` and the Avro/Json/Protobuf model config adapters, per the issue's scope (`AvroModelConfig`/`AvroModelConfigAdapter`/`AvroModelInfo` never reference any concrete extension type).
- Pre-seeds the `model-ext` schema scaffold in the engine's base schema (`engine.schema.json`), mirroring the existing `binding-ext` convention documented in `specs/AGENTS.md`.

Unit test coverage added for: `Config.ext`, `ConfigBuilder.ext(factory)` composition, `ConfigExtAdapter.adaptToJson`/`adaptFromJson` round-tripping through real `Config`/`ConfigBuilder` subtypes (including a subtype-builder case exercising the generic-variance subtlety), `Factory.instantiate(Iterable)`'s incubating filter, and `ModelInfo.extensions()` filtering by `type()`.

This lands entirely within `config/engine.conf` plus the three model config modules — no new module. General-purpose engine infrastructure, not encryption-specific; required by #1681 §5.

Fixes #2360

## Test plan

- [x] `./mvnw test -pl config/engine.conf,config/model-avro.conf,config/model-json.conf,config/model-protobuf.conf` — all existing and new unit tests pass
- [x] `./mvnw checkstyle:check` and `./mvnw license:check` on the touched modules
- [x] `./mvnw jacoco:report jacoco:check` on the touched modules — coverage thresholds met
- [x] `specs/engine.spec` `SchemaTest` passes with the new `model-ext` schema scaffold

---
_Generated by [Claude Code](https://claude.ai/code/session_013emf9JB5HYnof31hnLfBdb)_


jfallows (commented):
LGTM

PR #2366, opened 2026-08-13
fix(binding-kafka): reconnect fetch/produce cache fanouts to the upda…

ankitk-me:
…ted leader after `NOT_LEADER_FOR_PARTITION`


jfallows (dismissed):

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -1627,7 +1628,12 @@ private void doClientReplyEndIfNecessary(
         private void doClientReplyAbortIfNecessary(
             long traceId)
         {
-            if (KafkaState.replyOpening(state) && !KafkaState.replyClosed(state))
+            if (!KafkaState.replyOpening(state))
+            {
+                state = KafkaState.openingReply(state);
+            }
+
+            if (!KafkaState.replyClosed(state))
```
jfallows: Do we still need this change now that we no longer call `doClientReplyAbortIfNecessary` from handling the reset?

runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
```diff
@@ -844,7 +844,7 @@ private void onClientFanoutInitialWindow(
         private void doClientFanoutReplyResetIfNecessary(
             long traceId)
         {
-            if (!KafkaState.replyClosed(state))
+            if (KafkaState.initialOpening(state) && !KafkaState.replyClosed(state))
```
jfallows, line 847: Is this change necessary to tie sending RESET to initial BEGIN sent?

It seems that the purpose here is to avoid sending multiple RESETs, so set state after sending first RESET, then skip sending further RESETs if reply is already closed.

If we simplify this back to `f (!KafkaState.replyClosed(state))` and the issue still does not reproduce then that is preferred.

jfallows: Looks like we need to keep this for symmetry with `doClientFanoutInitialAbortIfNecessary` otherwise `doClientFanoutInitialBeginIfNecessary` will not be able to detect `KafkaState.closed(state)` as needed.

jfallows (approved):

PR #2367, opened 2026-08-13
feat(engine): extend VaultHandler SPI with wrap/unwrap KMS operations

### LINKED ISSUE #1678: Extend vault SPI with KMS encryption operations
As a Zilla operator, I want to configure named encryption keys in a vault so that Kafka message data can be encrypted and decrypted at the gateway layer using an external Key Management System, without raw key bytes ever leaving the KMS.

**Acceptance criteria**
- Vault SPI is extended with an `EncryptionKeyService` interface exposing `encrypt(keyName, plaintext)` and `decrypt(keyName, ciphertext)` operations against named keys
- Named key config supports: `name` (property key under `keys:`), a backend-specific key identifier, and `algorithm` (AES128_GCM or AES256_GCM)
- `algorithm` defaults to AES128_GCM when not specified
- Raw key material never leaves the KMS — only encrypt/decrypt API calls are made
- Model layer can call encrypt/decrypt via the SPI without knowledge of the KMS backend
- Existing vault types are unchanged and continue to serve TLS exclusively

**Additional context**
This is the foundation for all message encryption in Zilla. New vault types (implemented separately) will expose this SPI. The algorithm is defined once at the key level to avoid repetition across model configs.

jfallows:
## Description

Extends `VaultHandler` — currently TLS-only (`initKeys`, `initSigners`, `initTrust`) — with wrap/unwrap operations against named keys, so a KEK held in an external key manager can wrap and unwrap data encryption keys without its raw key material ever leaving the KMS.

- Adds a `BytesConsumer` functional interface nested inside `VaultHandler` (not placed in `common-agrona`, since nothing else needs the shape yet).
- Adds default `wrap(traceId, name, dek, index, length, next)` / `unwrap(traceId, name, edek, index, length, next)` methods. The default implementations are no-ops that signal failure through the same callback (`next.accept(null, 0, 0)`) rather than never invoking it — a `null` buffer is the "this failed" signal, matching what a real KMS-backed implementation uses for a genuine runtime failure.
- Shaped asynchronously from the start (result delivered via callback, not a return value), since KMS-backed vault implementations make network calls on a reactor-confined thread.
- Diagnostics for *why* an operation failed belong in the vault's own telemetry event, not the callback — unchanged by this PR.
- Existing vault types (`vault-filesystem`, `TestVault`, etc.) are unaffected — no overrides added, so their behavior is unchanged.
- Javadoc stays protocol-neutral per `runtime/engine/AGENTS.md`.

## Testing

Added `VaultHandlerTest`:
- Confirms the null-buffer failure signal fires for both `wrap` and `unwrap` when no override is present.
- Adds a minimal test vault implementation (`NamedKeyVaultHandler`) exercising both the success and failure path of each callback (matched key name vs. unknown key name).

Verified `./mvnw checkstyle:check` and the full `runtime/engine` unit test suite pass.

Fixes #1678

---
_Generated by [Claude Code](https://claude.ai/code/session_01CHWo89NTcBdA3gd5zZajbQ)_

### DIFF
--- a/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/vault/VaultHandler.java
+++ b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/vault/VaultHandler.java
@@ -21,8 +21,11 @@ import java.util.List;
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.TrustManagerFactory;
 
+import io.aklivity.zilla.runtime.common.agrona.buffer.DirectBufferEx;
+
 /**
- * Provides access to TLS cryptographic material from an attached vault.
+ * Provides access to cryptographic material from an attached vault, for TLS contexts and
+ * for wrapping and unwrapping named keys.
  * <p>
  * Obtained from {@link VaultContext#attach(VaultConfig)}, a {@code VaultHandler} resolves
  * named key and certificate references from the vault's backing store (e.g., a PKCS#12 file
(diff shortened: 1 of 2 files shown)


jfallows (commented):
LGTM

Issue #1683, opened 2026-03-30
Declare ProtobufModelExtFactorySpi and wire provider-supplied pipeline stages in protobuf model


jfallows:
**Describe the desired outcome from the user's perspective**

As a Zilla operator, I want the protobuf model to let an installed module contribute its own pipeline stages — for whatever purpose that module defines — without oss zilla needing to know or declare what that purpose is.

**Acceptance criteria**

## 1. The SPI

Mirror aklivity/zilla#1681 §1 exactly, substituting `Protobuf` for `Avro` throughout:

```java
// io.aklivity.zilla.runtime.model.protobuf.ext, model-protobuf's first exported package

public interface ProtobufModelExtFactorySpi extends FactorySpi
{
    ProtobufModelExt create(Configuration config);
}

public interface ProtobufModelExt
{
    String name();
    ProtobufModelExtContext supply(EngineContext context);
}

public interface ProtobufModelExtContext
{
    ProtobufModelExtHandler supplyHandler(ProtobufSchema schema, ProtobufModelConfig config);
}

public interface ProtobufModelExtHandler
{
    ProtobufTransformable transform(ProtobufTransformable stream);
}
```

`ProtobufTransformable` is a new, minimal interface in `common-protobuf` — the same shape as `common-avro`'s `AvroTransformable` — that `common-protobuf`'s pipeline-description type now extends, so an extension can append stages to an in-progress stream without depending on that type's fuller terminal-bearing API. This is the only `common-protobuf` change this issue needs — `common-protobuf` otherwise stays a pure codec library with no engine/config dependency.

Registered via `META-INF/services/io.aklivity.zilla.runtime.model.protobuf.ext.ProtobufModelExtFactorySpi`. Any number of implementations may be installed at once; every one discovered contributes independently, via `Factory.instantiate(ServiceLoader.load(ProtobufModelExtFactorySpi.class))`. Installed extensions simply compose, each appending its own stage(s) ahead of the model's own adapter stage.

## 2. Composition and caching

[...]

Issue #1689, opened 2026-03-30
Wire ModelEnvelope into kafka cache storage's read and write paths


jfallows:
**Describe the desired outcome from the user's perspective**

As a `ModelTransform` (aklivity/zilla#2325/#2326) that needs message metadata carried outside the value itself, I want the kafka cache storage layer to give my transform a real `ModelEnvelope` (aklivity/zilla#2384) backed by the cached record's actual headers, so that I can read metadata that was present on produce and write metadata that a consumer needs on fetch, without knowing anything about Kafka myself.

**Acceptance criteria**

## 0. Implementation layer

This lives in the shared kafka cache storage implementation, not in any single binding's factory. The binding resolves the model (catalog + schema id) and supplies the per-message `long authorization`; the storage read and write paths — populating and draining the envelope — are the storage layer's own behavior, invoked identically regardless of which binding reads or writes through it. Any future binding built on this storage layer inherits identical behavior with no separate implementation of its own.

## 1. Produce path: materializing envelope writes as real headers

When a value pipeline's composed transform calls `ModelEnvelope.set(name, value)` during encode, the storage write path collects every named value written during that encode pass and, once encode completes, writes each as a real Kafka header on the produced record — repeated `set` calls under one name become repeated headers with that name, in call order. A transform that writes nothing leaves the record's header set untouched.

## 2. Fetch path: populating envelope reads from real headers

Before invoking the value pipeline's decode, the storage read path constructs a `ModelEnvelope` backed by the cached record's actual headers, so that `envelope.count(name)`/`envelope.get(name, index)` reflect exactly what is present on that record — no headers means an empty envelope, not an error. A composed transform that never calls `envelope()` incurs no cost from this — the envelope is materialized lazily, only when a transform actually asks for it.

## 3. Symmetry with existing conversion

This mirrors, at the metadata level, the same principle the existing JSON↔Avro `view: json` conversion already establishes at the value level: the storage layer provides a real, general mechanism (there, format conversion; here, an out-of-value metadata channel) that any composed transform can use, with the storage layer itself carrying no opinion about what a transform does with it.

## Coverage

Unit and k3po coverage for: a transform that writes one named value on encode and a corresponding fetch reading it back correctly; a transform writing the same name more than once producing that many headers in order; a fetched record with no headers presenting an empty envelope with no error; a transform that never touches the envelope incurring no observable behavior change from before this issue.

**Additional context**

[...]

PR #2368, opened 2026-08-13
feat(common-avro,common-json,common-protobuf): retain and expose field-level schema annotations

### LINKED ISSUE #2302: Retain and expose schema field annotations from Avro, JSON Schema, and Protobuf codecs

As a Zilla operator, I want the schema codecs to retain and expose whatever field annotations my schemas already carry, so that any consumer of a compiled schema — a model extension, a registry-mode catalog integration, or similar — can read them off the field object without the codec silently discarding whatever it doesn't recognize as a standard keyword.

**Acceptance criteria**

## 1. Retain unrecognised field annotations

The codecs discard them today.

- **`common-avro`** — `AvroSchemaCompiler` keeps only `logicalType`, `precision`, `scale`, `aliases`, `name`, `size`; `AvroField` exposes only `name()`, `type()`, `aliases()`, `defaultValue()`
- **`common-json`** — equivalent gap for JSON Schema annotations
- **`common-protobuf`** — `ProtobufField` retains **no options at all** (its fields are number, name, jsonName, type, repeated, required, packed, proto3Optional, typeName, oneofName, defaultValue, message, enumeration). `ProtobufSourceCompiler.applyOptions` consumes the known ones and discards the rest

## 2. Annotations are read from the field, not the type

Read from the field object only, for every format. A tag on the field's *type* is ignored, named type references are not followed, and record-level and root-level tags are not inherited. This is a deliberate scope boundary, not an oversight: supporting type-level or inherited tags would require walking type references and merging scopes across them, machinery this issue does not need — a schema author who wants a type's every usage tagged can tag each field directly.

So:

- **Avro** — field object only.
- **JSON Schema** — the property's schema, necessarily, since there is no separate field object
- **Protobuf** — field options only

## 3. Per-format exposure and per-format evaluation

**No shared evaluator.** A model is one format at a time, so there is no polymorphic call site to serve, and forcing a common intermediate representation costs more than it saves:

- Avro and JSON Schema annotations are **already** `JsonValue` — Avro schemas are JSON, JSON Schema annotations likewise
- Protobuf options come from an **ANTLR parse tree of text-format constants**, an entirely different source. `ProtobufSourceCompiler` captures `option.optionName().getText()` and `option.constant().getText()` with no descriptor resolution

So each codec exposes annotations in the form natural to it:

```java
// common-avro, common-json
JsonValue attribute(String name);      // null when absent
```

For `common-protobuf`, expose the option constant from the parse tree rather than converting it. **Do not route through `ProtobufJson`** — that bridge converts protobuf *wire* data to JSON, whereas these are text-format source constants, which is a different problem. Walking the grammar's `constant` subtree is the right approach.

Two protobuf details that follow:

- **Options need no pre-declaration.** Because parsing is source-text with our own grammar and never resolves a descriptor, an arbitrary `[(acme.classification) = "PII"]` is readable even though nothing declares that extension. Document that the *producer's* protoc toolchain will still reject an undeclared option at their build time
- **No proto3 JSON `int64`→string quirk.** Reading from source text, `[(acme.limit) = 123]` is an integer constant, so it compares as a number. That quirk only applied under the `ProtobufJson` route
- String constants come back from `getText()` **with their quotes** — unquote before comparing

Also needed alongside field-level retention, since a consumer walking a schema for tagged fields needs to discover *which* paths carry annotations without re-parsing: a predicate-driven path-discovery method per format (`AvroType.matchingPaths(Predicate)`, `JsonSchema.matchingPaths(Predicate)`, `ProtobufMessage.matchingPaths(Predicate)`), returning RFC 6901-style pointers for every node a caller-supplied predicate matches.

## Coverage

- round-trip: a schema declaring annotations of each shape compiles, exposes them, and survives the existing conformance suites
- protobuf: an option not declared by any imported `.proto` is still retained and readable
- protobuf: quoted string constants unquote correctly; message-valued options expose their structure rather than flattened text

**Why configurable rather than a name we pick.** Field tagging has no clear winner — Snowplow's bare `pii: true`, Adobe XDM's `xdm:` namespace, assorted `x-pii`/`x-data-classification` customs, and [Apicurio #7879](https://github.com/Apicurio/apicurio-registry/issues/7879) building field-level classification with authorization-scoped views. Hardcoding one couples generic code to a vendor; minting our own forks the space and forces adopters to re-tag. This issue only retains and exposes whatever a schema already carries — it declares no vocabulary of its own, and no matching/selection semantics on top of it. What a consumer does with an exposed annotation is out of scope here.

**Additional context**

**Scope note (narrowed from the original issue text):** this issue originally also proposed, as §§4–9, a shared `tagged:`/`named:` field-matcher config DSL with a normative matcher matrix, an `overlay:` interaction, and a root-level (schema-root) `attribute()` accessor — all evaluated generically across `model-avro`/`model-json`/`model-protobuf`. None of that materialized as a shared oss capability, and none currently has a driving consumer:

- The three SPI issues that cited this issue as a blocker — #1681, #1682, #1683 (all closed/completed) — each narrowed their own dependency on this issue to just the retention/exposure accessors above (`attribute()`/`option()`/`matchingPaths()`), and explicitly scoped tag-based field *selection* as "entirely its own design" for whatever extension installs it — this issue "declares no opinion on what any installed extension is for."
- The one concrete consumer that needed tag-based field selection, zilla-plus's field disclosure feature (aklivity/zilla-plus#1146, #1150, #1152), built its own per-format matcher config (`AvroModelExtFieldMatcherConfig`/`*SelectConfig`/`*SelectType` and the json/protobuf equivalents) directly inside the commercial ext modules, rather than depending on a shared oss primitive.
- The root-level (schema-root) `attribute()` accessor has no consumer anywhere in zilla or zilla-plus today.

Retitled and re-scoped to reflect the part that actually shipped, is actually consumed, and stays business-agnostic per this repo's own rule against building shared-component surface for a hypothetical future consumer. Fixed by #2368.

jfallows:
## Description

Part 1 of #2302 — codec-level retention and exposure of field-level schema annotations across `common-avro`, `common-json`, and `common-protobuf`. Today all three codecs silently discard any annotation they don't recognize as a standard keyword; this PR retains them and exposes them in the form natural to each format, read from the field object only (never inherited from the field's type or a named type reference).

**Scope note:** issue #2302 originally bundled two very differently-sized pieces of work: (a) codec-level annotation retention (§1–§3), and (b) a matcher/action config DSL with YAML schema, path-pattern matching, and per-model evaluation across `model-avro`/`model-json`/`model-protobuf` (§4–§8). (b) had no existing precedent in this codebase — no matcher config of any kind, no Avro path-pointer walker equivalent to `JsonSchema.retainedPaths()` — it was greenfield design in its own right. This PR delivers (a) as a complete, tested, mergeable unit.

**Update:** (b) was never picked up as a shared oss capability — the SPI issues that once cited #2302 as a blocker (#1681, #1682, #1683) only ever needed the retention/exposure work in this PR, and explicitly left tag-based field *selection* as "entirely its own design" for whatever extension installs it. The one concrete consumer (zilla-plus's field disclosure feature) built its own per-format matcher config directly rather than depending on a shared primitive. #2302 has been retitled/re-scoped to reflect that this PR is its complete fix, not a partial one.

### Changes

- **`common-avro`**: `AvroField.attribute(String)` reads an arbitrary key from the field's own JSON object, returned as `JsonValue`; `null` when absent.
- **`common-json`**: `JsonSchema.property(String)` navigates to a property's compiled sub-schema (JSON Schema has no separate field object), and `JsonSchema.attribute(String)` reads an arbitrary key declared directly on a schema node — not inherited through `$ref`. `JsonSchema.matchingPaths(Predicate)` walks the compiled schema tree and returns the RFC 6901 pointers of every node a caller-supplied predicate matches; `retainedPaths()` is reimplemented on the same walker, replacing the old `JsonSchemaPaths` utility (deleted — it re-parsed the schema text into a second, parallel tree).
- **`common-protobuf`**: new `ProtobufConstant` sealed type (`Identifier`/`IntegerValue`/`FloatValue`/`TextValue`/`BooleanValue`/`MessageValue`) plus `ProtobufField.option(String)`, walked directly from the ANTLR parse tree rather than routed through `ProtobufJson`. Options need no pre-declaration by an imported `.proto`; message-valued options expose their structure rather than flattened text; string constants are unquoted. Along the way this fixes a grammar ambiguity where `true`/`false` field-option values parse as a bare `fullIdent` rather than `boolLit` (the `ident` rule folds `BOOL_LIT` into `keywords`), which would otherwise have misclassified boolean options as identifiers.
- **`common-avro`/`common-protobuf`**: `AvroType.matchingPaths(Predicate)` and `ProtobufMessage.matchingPaths(Predicate)` port the same predicate-driven path-discovery shape, each over the type that actually carries annotations in that format. Both guard against infinite recursion on self-referential schemas with an identity-based "on this recursion path" `Set`.

### Test plan

- [x] New unit tests added first, confirmed failing against the prior code, then implementation added until green (test-first per `AGENTS.md`)
- [x] `./mvnw clean verify -pl runtime/common-avro` — tests, checkstyle, license headers, JaCoCo coverage all pass
- [x] `./mvnw clean verify -pl runtime/common-json` — same, no regressions
- [x] `./mvnw clean verify -pl runtime/common-protobuf` — same, no regressions
- [x] Verified no other implementers of `AvroField`/`JsonSchema`/`ProtobufField`/`AvroType`/`ProtobufMessage` exist in the codebase that the added methods could break

Fixes #2302

---


jfallows:
Pushed a second commit filling out `common-json`'s side of tagged-field discovery, ahead of the `disclosure:`/redaction work in #1689:

- **`JsonSchema.matchingPaths(Predicate<JsonSchema>)`** — walks the compiled schema tree (`properties`, single-schema `items`, `allOf`/`anyOf`/`oneOf`/`if`/`then`/`else`) and returns the RFC 6901 pointers of every node the predicate matches, in schema declaration order. A match doesn't stop descent — nested matches under an already-matching ancestor are still reported independently, since that's a decision the caller (redaction/action dispatch) should make, not this API. `$ref`/`patternProperties`/tuple-`items` stay unexpanded, same boundary as `retainedPaths()`. This is the piece `model-json` needs to answer "which fields match `tagged: - "zilla:tags": ["PII"]`" without reimplementing schema traversal itself.
- **`retainedPaths()` reimplemented on the same walker**, operating on the already-compiled schema tree instead of `JsonSchemaPaths` re-parsing the schema text into a second, parallel tree. Bonus: this incidentally fixes a latent inconsistency where pre-2019-09 `$ref` siblings were walked even though the validator itself (correctly) ignores them for those drafts.
- `JsonSchemaPaths`/`JsonSchemaPathsTest` deleted outright (not deprecated) — confirmed their only call site was `retainedPaths()` itself, so once that migrated there were zero remaining callers. Its test coverage is preserved in `JsonSchemaRetainedPathsTest`, now driven through the public `JsonSchema` API instead of the internal utility.

Full `verify` (tests + checkstyle + license + JaCoCo) green on `common-json`, `common-avro`, and `common-protobuf`; 4935 tests total in `common-json`, no regressions.

---

jfallows:
Third commit: ported `matchingPaths(Predicate)` to `common-avro` and `common-protobuf`, so all three formats now support the same tagged-field-discovery shape ahead of #1689.

Not a literal copy of `common-json`'s version — each format takes the predicate over the type that actually carries annotations, per the issue's "no shared evaluator" principle:

- **`AvroType.matchingPaths(Predicate<AvroField>)`** — only `AvroField` has `attribute()`, not `AvroType` (Avro separates "the field" from "the field's type"; JSON Schema doesn't). Recurses through record fields, array `items()`, map `values()`, and union `branches()`.
- **`ProtobufMessage.matchingPaths(Predicate<ProtobufField>)`** — recurses through composite (message-typed) fields. A `repeated` composite field gets a synthetic `-` wildcard segment before descending, matching the array-of-objects convention from Avro/JSON — except a `map` field (protobuf desugars maps into a repeated synthetic entry message with named `key`/`value` fields) reaches its value type one segment deeper, at `.../value/...`, since that field is genuinely named in protobuf's own model. Worth flagging as a real cross-format caveat for whenever `named:` patterns need to work uniformly across all three.

Both needed something JSON Schema's port didn't: an identity-based cycle guard. Unlike JSON Schema (safe only because `$ref` is deliberately never expanded), Avro and Protobuf both allow genuine self-reference through their normal composite-field mechanism — `AvroType`'s own javadoc already states recursive types compile to a cyclic graph by node identity, and there's an existing test proving it (`shouldTerminateOnRecursiveType`); Protobuf messages can self-reference the same way via a `repeated` field of their own type. Guarded with a plain `HashSet` tracking "on this recursion path" (neither `AvroNode` nor `ProtobufMessage` override `equals`/`hashCode`, so identity semantics fall out for free) — a match never stops descent, only revisiting a message/record already on the current path does.

Full `verify` (tests + checkstyle + license + JaCoCo) green on both: `common-avro` 143 tests, `common-protobuf` 198 tests, no regressions.

---

jfallows:
**Correction to the two comments above:** both quote or reference things that don't belong in this thread — a literal `tagged: - "zilla:tags": ["PII"]` example lifted from #2302's issue text, and repeated framing around "#1689's redaction/disclosure work." Neither is accurate context for this PR: no `tagged:`/matcher config exists yet (that's explicitly out of scope here, see the PR description), and #1689 is a separate, unbuilt feature this groundwork doesn't implement or depend on. I don't have a way to edit those comments in place, so noting the correction here instead.

For the record, what actually shipped in the two commits those comments describe, without the extraneous framing:

- `JsonSchema.matchingPaths(Predicate<JsonSchema>)` walks the compiled schema tree (`properties`, single-schema `items`, `allOf`/`anyOf`/`oneOf`/`if`/`then`/`else`) and returns the RFC 6901 pointers of every node a caller-supplied predicate matches, in schema declaration order. A match doesn't stop descent. `$ref`/`patternProperties`/tuple-`items` stay unexpanded, same boundary as `retainedPaths()` (which is now reimplemented on this same walker, replacing the deleted `JsonSchemaPaths` utility).
- `AvroType.matchingPaths(Predicate<AvroField>)` and `ProtobufMessage.matchingPaths(Predicate<ProtobufField>)` apply the same shape to the other two formats, each over the type that actually carries annotations there. Both add an identity-based recursion guard the JSON Schema version doesn't need, since Avro and Protobuf both permit genuine self-referential schemas through their normal composite-field mechanism.

All test/build claims in the two prior comments (test counts, `verify` results) are accurate and unaffected by this correction.

---

jfallows (commented):
LGTM

Issue #2318, opened 2026-08-04
Apply schema overlay at catalog entry resolution


jfallows:
**Describe the desired outcome from the user's perspective**

As a Zilla operator, I want to annotate a schema I do not own — adding field tags without forking it in the registry — so that I can reason about which fields are sensitive independently of what policy is applied to them.

**Acceptance criteria**

## 0. This extends existing, already-shipped code — it is not a new subsystem

`runtime/common-json` already ships `JsonOverlay`, `JsonOverlayAction`, and `JsonPath` — implemented, unit-tested (`JsonOverlayTest`, `JsonPathTest`), and already load-bearing in production: `OpenapiCompositeGenerator`, `OpenapiAsyncapiCompositeGenerator`, `McpOpenapiCompositeGenerator`, and `AsyncapiCompositeGenerator` all resolve documents through it today. `JsonOverlay.of(...)` takes an OpenAPI Overlay-shaped document (`overlay:` version, `actions[]` of `target`/`update`/`remove`) and applies it via a restricted RFC 9535 JSONPath (`JsonPath`) evaluator. This issue is three small, additive changes to that existing code plus wiring, not a new mechanism:

1. **`JsonOverlay.merge()` needs a version-gated branch for array updates.** Every shipped test fixture declares `"overlay":"1.0.0"`, and the current `merge()` unconditionally appends `update` as one element when the target is an array — correct for 1.0.0 (confirmed against the spec text: *"If the target selects an array, the value of this field MUST be an entry to append to the array"*), but wrong for what §3 below needs, which is concatenation. OpenAPI Overlay **1.1.0** changed this: *"If the target selects array nodes, the value of this field MUST be an array to concatenate with each selected array, or an object or primitive value to append to each selected array."* The fix is additive: documents declaring `"overlay":"1.0.0"` take the exact existing code path unchanged (zero risk to the four current consumers); documents declaring `"overlay":"1.1.0"` take a new branch that concatenates when `update` is itself a JSON array.
[...]

PR #2370, opened 2026-08-13
feat(config): reposition overlay as peer of subject/version, share one implementation

### LINKED ISSUE #2365: Consolidate overlay: config across binding-openapi, binding-asyncapi, binding-openapi-asyncapi, binding-mcp-openapi; reposition as a peer of subject/version

As a Zilla operator, I want `overlay:` to look and behave the same way everywhere it appears — same placement, same shape, one implementation underneath — rather than four independently hand-rolled copies that happen to agree today but have no shared code enforcing that they keep agreeing.

**Background — this is real, confirmed duplication, not a hypothetical**

Four modules each independently define an `overlay:` property with the identical `{catalogName: {subject, version}}` shape, and each has its own separate Java implementation:

| Module | Schema patch | Java classes |
| --- | --- | --- |
| `binding-mcp-openapi` | `mcp_openapi.schema.patch.json` | `McpOpenapiSpecificationConfig.overlay: McpOpenapiCatalogConfig`, `McpOpenapiOptionsConfigAdapter` |
| `binding-openapi` | `openapi.schema.patch.json` | `OpenapiSpecificationConfig.overlay: OpenapiCatalogConfig`, `OpenapiOptionsConfigAdapter` |
| `binding-asyncapi` | `asyncapi.schema.patch.json` | (mirrors `binding-openapi`'s pattern) |
| `binding-openapi-asyncapi` | `openapi.asyncapi.schema.patch.json` | (the shape appears **twice** in this module's own patch — once per protocol side) |

Confirmed by reading `OpenapiOptionsConfigAdapter` directly: it hand-parses `overlay: {catalogName: {subject, version}}` by unwrapping a one-entry `JsonObject`, building an `OpenapiCatalogConfig`, exactly mirroring what `McpOpenapiOptionsConfigAdapter` does independently for its own `McpOpenapiCatalogConfig`. Today's placement in all four: `overlay:` is a **sibling of `catalog:`**, one level above `subject`/`version`, e.g.:

```yaml
specs:
  payments-api:
    catalog:
      my-registry:
        subject: payments-api-spec
        version: latest
    overlay:
      my-overlay:
        subject: payments-api-overlay
        version: latest
```

**Why this placement is wrong, and what the right one is**

zilla#2318 (schema field annotations for the privacy epic) went through the identical design question for its own `overlay:` property and settled on placing it as a **peer of `subject`/`version`**, nested one level *inside* the catalog entry it annotates, not one level above `catalog:`:

```yaml
specs:
  payments-api:
    catalog:
      my-registry:
        subject: payments-api-spec
        version: latest
        overlay:
          my-overlay:
            subject: payments-api-overlay
            version: latest
```

The reasoning that produced this in #2318 applies identically here: the thing being overlaid is *the specific resolved catalog entry*, not "the spec" in general — a multi-entry `catalog:` block could in principle resolve different documents per entry, and the overlay is scoped to one of them, not to the enclosing `specs.<name>` block as a whole. Placing `overlay:` as a peer of `catalog:` couples it to the wrong scope; nesting it next to the `subject`/`version` it actually annotates is the correct scope, and is also the placement all four of these modules should have used from the start had this been settled before they shipped.

**Acceptance criteria**

## 1. Reposition `overlay:` as a peer of `subject`/`version` in all four modules

Same JSON shape (`{catalogName: {subject, version}}`), new nesting: inside each `catalog.<name>` entry, not as a sibling of `catalog:`.

## 2. Deprecate the old placement, parse it into the canonical internal representation, don't accept both at once

- The old, top-level `overlay:` (peer of `catalog:`) is marked `deprecated: true` in each schema patch (JSON Schema annotation keyword) and still parses — into the exact same internal config object the new placement produces — for a transition window.
- A spec entry that sets **both** the old-style top-level `overlay:` and the new nested `catalog.<name>.overlay:` is a startup config error — ambiguous precedence is worse than picking one and erroring on the other.
- Existing `zilla.yaml` files using the old placement keep working unchanged during the deprecation window; only the schema/docs steer new configs toward the new placement.

## 3. Extract one shared implementation instead of fixing four independently

Since all four modules are being touched for the reposition anyway, this is the point to stop duplicating: one shared config type (a `CatalogOverlayConfig`-shaped class, exact name/home TBD — candidate location is wherever `catalog-inline`-style shared config helpers already live) and one shared adapter helper for parsing/writing `{catalogName: {subject, version}}`, referenced by `McpOpenapiSpecificationConfig`, `OpenapiSpecificationConfig`, the `binding-asyncapi` equivalent, and both sides of `binding-openapi-asyncapi` — replacing four (five, counting `openapi-asyncapi`'s double) independent copies of the same class-plus-parsing logic with one.

## 4. Coverage

- Each module: new-placement config resolves correctly; old-placement config still resolves (parses into the identical internal shape) during the deprecation window and emits a deprecation warning; a spec entry with both placements set is rejected at startup with a clear error
- Cross-module: the shared config type/adapter, once extracted, is unit-tested once and referenced (not re-tested) by each consuming module — mirroring how other shared helpers in this codebase are tested once at their own definition site

**Additional context**

Not part of the Data Privacy Protection epic (zilla-platform#383) — #2318 is what surfaced this inconsistency during its own design review, but this issue is a general config-consistency fix independent of privacy/annotation work, and shouldn't be tracked under that epic.

`binding-mcp-grpc`'s proposed overlay mechanism (zilla#2279) is being updated separately to use a catalog reference from the start, rather than repeat the inline-list shape this issue is migrating away from — see zilla#2279.

jfallows:
## Description

`overlay:` was independently hand-rolled in `binding-openapi`, `binding-asyncapi`, `binding-mcp-openapi`, and `binding-openapi-asyncapi` (once per protocol side) as a sibling of `catalog:`, each with its own copy of the `{catalogName: {subject, version}}` parsing logic. This repositions `overlay:` to be a peer of `subject`/`version`, nested inside each `catalog:` entry, matching the design settled in #2318 — the overlay annotates the specific resolved catalog entry, not the enclosing `specs.` block as a whole.

**Shared implementation** (`config/engine.conf`): `SchemaConfig` gains a nullable `overlay: OverlayConfig` field (peer of `subject`/`version`), where `OverlayConfig{name, schema: SchemaConfig}` is the new shared type. `SchemaConfigAdapter` handles the nested `overlay` key recursively — this is the one shared parse/write implementation all four modules delegate to, instead of each re-parsing the shape by hand.

**Per module** (openapi, asyncapi, mcp-openapi, openapi-asyncapi):
- `overlay:` repositioned as a peer of `subject`/`version` inside each `catalog:` entry
- The old top-level placement is marked `deprecated: true` in the schema and still parses — canonicalized at parse time onto every catalog entry (matching today's actual runtime behavior of applying it uniformly). The in-memory config model only ever has one representation of `overlay`; `adaptToJson` only ever emits the new placement
- Setting both placements at once is made structurally invalid via a JSON-schema `if`/`then`/`not` constraint rather than hand-written validation code, so it's rejected by the engine's existing schema-validation error path
- Runtime composite generators (`OpenapiCompositeGenerator`, `AsyncapiCompositeGenerator`, `McpOpenapiCompositeGenerator`, `OpenapiAsyncapiCompositeGenerator`) move from reading a spec-level `overlay` to reading it per catalog entry

## Test plan

- [x] New/updated unit tests per module: new-placement read/write, deprecated-placement still resolves into the identical shape, both-placements-set rejected by schema validation (verified end-to-end via a real `EngineConfigReader` pass, not just adapter-level)
- [x] `engine.conf`: 76 tests (`OverlayConfigTest`, `SchemaConfigAdapterTest`, existing suite)
- [x] Consolidated build across all 15 touched modules (`engine.conf`, `common-openapi`, `common-asyncapi`, and the openapi/asyncapi/mcp-openapi/openapi-asyncapi spec+config+runtime triples): **436 tests, 0 failures**
- [x] Checkstyle, license headers, JaCoCo coverage all pass per module

Fixes #2365

---

https://claude.ai/code/session_014pmc9g4ygAd5XgRn7BDL4H

---
_Generated by [Claude Code](https://claude.ai/code/session_014pmc9g4ygAd5XgRn7BDL4H)_


config/binding-asyncapi.conf/src/test/java/io/aklivity/zilla/config/binding/asyncapi/internal/AsyncapiOptionsConfigAdapterTest.java
```diff
@@ -136,16 +165,16 @@ public void shouldWriteOptionsWithOverlay() throws IOException
...
-                    .name("catalog0")
-                    .subject("smartylighting-overlay")
-                    .version("latest")
-                    .build()
+                    .overlay(AsyncapiCatalogConfig.builder()
+                        .name("catalog1")
+                        .subject("smartylighting-overlay")
+                        .version("latest")
+                        .build())
+                    .build())
```
jfallows:
```suggestion
                    .overlay()
                        .name("catalog1")
                        .subject("smartylighting-overlay")
                        .version("latest")
                        .build()
                    .build())
```

jfallows (commented):
Reusing SchemaConfigAdapter in place of AsyncapiCatalogConfigAdapter etc is going too far, we should have a local AsyncapiCatalogConfigAdapter and just reuse OverlayConfig and OverlayConfigAdapter.

For the code as it was, the config object needs to add an overlay of type OverlayConfig and reuse OverlayConfigAdapter / OverlayConfigBuilder for that part, the rest can stay as it was I think.

jfallows:
Addressed the review feedback in two commits:

1. **`403a2d6219`** — Reverted the over-broad `SchemaConfigAdapter` reuse. Each of `binding-openapi`, `binding-asyncapi`, `binding-mcp-openapi`, and `binding-openapi-asyncapi` is back to hand-rolling its own `catalog:` entry (subject/version) parsing. `OverlayConfig`/`OverlayConfigAdapter` is now reused *only* for the `overlay` sub-key itself (both the nested-in-catalog and deprecated top-level placements), and each module's `*CatalogConfig.overlay` field is typed directly as the shared `OverlayConfig` rather than a self-referential type.

   This required adding `engine.conf` as a compile dependency to `runtime/common-openapi` and `runtime/common-asyncapi` (flagging this explicitly since we'd previously discussed keeping that dependency out of those modules) — `OpenapiCatalogConfig`/`AsyncapiCatalogConfig` now reference `OverlayConfig` directly as a field type. No inheritance/extension coupling, just plain composition.

   Also caught and fixed `McpKafkaConnectCompositeGenerator`, which builds an `McpOpenapiBindingConfig` programmatically and referenced a stale top-level `overlay()` builder method that never matched the actual (nested-in-catalog) shape — this was a pre-existing latent bug that only surfaced once CI actually compiled the module.

2. **`fb8e49ab63`** — Applied your suggested diff: `OverlayConfigBuilder` now exposes flat `subject()`/`version()` convenience methods alongside the nested `schema()` builder, so `overlay().name(x).subject(y).version(z).build()` works without an intermediate `schema()` step. Flattened the other overlay-construction call sites I'd written to match.

Full build (`checkstyle` + tests) is green across all touched modules.

---

runtime/common-asyncapi/pom.xml
```diff
@@ -51,6 +51,11 @@
       <artifactId>common-yaml</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>engine.conf</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
jfallows: This dependency must be removed, common cannot depend on engine.conf.

runtime/common-openapi/src/main/java/io/aklivity/zilla/runtime/common/openapi/config/OpenapiCatalogConfig.java
```diff
@@ -16,20 +16,33 @@
...
 import java.util.function.Function;
 
+import io.aklivity.zilla.config.engine.OverlayConfig;
+
 public class OpenapiCatalogConfig
 {
     public final String name;
     public final String subject;
     public final String version;
+    public final OverlayConfig overlay;
```
jfallows: common-openapi should be self-contained, no dependency on engine.conf.

runtime/common-openapi/pom.xml
```diff
@@ -51,6 +51,11 @@
       <artifactId>common-yaml</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>engine.conf</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
jfallows: This dependency must be removed, common-openapi cannot depend on engine.conf.

runtime/common-asyncapi/src/main/java/io/aklivity/zilla/runtime/common/asyncapi/config/AsyncapiCatalogConfig.java
```diff
@@ -16,20 +16,33 @@
...
     public final String subject;
     public final String version;
+    public final OverlayConfig overlay;
 
     public AsyncapiCatalogConfig(
         String name,
         String subject,
-        String version)
+        String version,
+        OverlayConfig overlay)
```
jfallows: common-asyncapi should be self-contained, no dependency on engine.conf.

jfallows:
One more round (`b6faad9101`): the full-reactor CI build caught `McpSchemaRegistryCompositeGenerator`, a sibling of `McpKafkaConnectCompositeGenerator` with the exact same stale top-level `overlay()` builder call — nested it under `catalog()` and fixed the corresponding test assertion, which referenced a spec-level `overlay` field that never existed. Swept the whole repo for any other `.overlay()` call sites and direct `new *CatalogConfig(..., overlay)` constructions; none remain outside the builders themselves.

---

jfallows:
Reverted the `engine.conf` dependency I'd added to `common-openapi`/`common-asyncapi` (`2afebedf26`) — you're right that they should stay self-contained.

`OpenapiCatalogConfig.overlay`/`AsyncapiCatalogConfig.overlay` are back to being typed as their own module's self-referential catalog-config type rather than the shared `OverlayConfig`, with a small self-referential nested `overlay()` builder method mirroring the pattern. `binding-openapi.conf`/`binding-asyncapi.conf`/`binding-openapi-asyncapi.conf` no longer reuse `OverlayConfigAdapter` for this either — each hand-rolls the `{name: {subject, version}}` shape locally, shared between the deprecated top-level and new nested-in-catalog call sites via a small private helper so the two placements canonicalize identically.

`binding-mcp-openapi.conf` (and its dependents, `binding-mcp-kafka-connect`/`binding-mcp-schema-registry`) are unchanged — that module already depends on `engine.conf` directly, so `McpOpenapiCatalogConfig.overlay: OverlayConfig` stays as-is.

Full build green across all touched modules.

---

jfallows:
Follow-up architectural cleanup (`f9c1236828`), prompted by a good catch: `common-openapi`/`common-asyncapi` had no business knowing what a "catalog" or an "overlay" is in the first place — that's a Zilla config/resolution concept, not something intrinsic to the OpenAPI/AsyncAPI document formats those modules parse.

- Moved `OpenapiCatalogConfig`/`OpenapiCatalogConfigBuilder`/`OpenapiSpecificationConfig` out of `common-openapi` into `binding-openapi.conf` (their only real owner — `binding-openapi.conf` already depends on `common-openapi` for everything else, and `runtime/binding-openapi` already depends on `binding-openapi.conf`, so no new coupling anywhere).
- Same treatment for the Asyncapi equivalents into `binding-asyncapi.conf`.
- `binding-openapi-asyncapi.conf` previously reused these from `common-openapi`/`common-asyncapi` only to avoid depending on the sibling `binding-openapi.conf`/`binding-asyncapi.conf` modules. It now defines its own independent local copies instead, matching the precedent `binding-mcp-openapi.conf`'s `McpOpenapiCatalogConfig` already set — no cross-binding config-type sharing, a little tolerated duplication instead. Dropped its `common-asyncapi` dependency entirely (kept `common-openapi`, still needed for `OpenapiServerView`).

`common-openapi`/`common-asyncapi` now contain only genuine document-model types (parsers, views, extensions) — zero references to catalogs, overlays, or `engine.conf`. Full build green across all touched modules.

---

jfallows:
Pushed a follow-up layering cleanup (d399e70) on top of the previous round.

`OpenapiConditionConfig`/`AsyncapiConditionConfig`/`OpenapiAsyncapiConditionConfig` (and their per-server condition configs) took `OpenapiServerView`/`AsyncapiServerView` — runtime-layer types — as method parameters, purely to express "here are the operation's servers to match against." That forced `binding-openapi.conf` and `binding-openapi-asyncapi.conf` to depend on `common-openapi`, and `binding-asyncapi.conf` to depend on `common-asyncapi`, even though these are config-only modules.

Replaced those parameters with plain data:
- openapi side: `URI` (a server is just identified by its URL)
- asyncapi side: `Map<String, URI>` keyed by server name (mirrors AsyncAPI's own `servers:` map shape, since asyncapi server matching can key off name, URL, or both)

The runtime-layer call sites (`OpenapiRouteConfig`, `AsyncapiRouteConfig`, `AsyncapiProxyGenerator`, `OpenapiAsyncapiProxyGenerator`) do the view→plain-data conversion before delegating into the config classes — they already depend on the runtime modules, so that's the right place for it.

Net result:
- `binding-openapi.conf` and `binding-openapi-asyncapi.conf`: `common-openapi` dependency removed entirely (pom.xml + module-info.java)
- `binding-asyncapi.conf`: `common-asyncapi` moved to test-scope only (still needed by `AsyncapiConditionConfigAdapterTest`, which parses real AsyncAPI documents to build realistic server-matching fixtures)

Full build across all touched config/runtime modules passes clean.

---

PR #2373, opened 2026-08-13
fix(config): remove unused jakarta.json.bind dependency from openapi/asyncapi bindings

### LINKED ISSUE #2372: Unused jakarta.json.bind module dependency in binding-openapi.conf/binding-asyncapi.conf/binding-openapi-asyncapi.conf
## Description

`config/binding-openapi.conf`, `config/binding-asyncapi.conf`, and `config/binding-openapi-asyncapi.conf` each declare:

- a `jakarta.json.bind-api` Maven dependency in `pom.xml`
- `requires jakarta.json.bind;` in `src/main/moditect/module-info.java`

but no main-source file in any of the three modules directly imports `jakarta.json.bind.*`. Only `jakarta.json.*` (the plain JSON-P API) is used directly in main sources; `jakarta.json.bind` types (`Jsonb`, `JsonbBuilder`, `JsonbConfig`) only appear in test sources.

The dependency isn't harmful — it's already available transitively via `engine.conf`, which declares `requires transitive jakarta.json.bind;` — but it's redundant in these three modules' own declarations. It looks like it was copied in from an existing module template (the same pattern appears in unrelated modules like `runtime/guard-inline` and `runtime/guard-identity`) rather than being deliberately needed here.

Noticed while auditing dependency accuracy after #2370 (moving catalog/condition config types out of `common-openapi`/`common-asyncapi`); out of scope for that PR since it predates it and isn't something that PR's changes touched.

## Suggested fix

For each of `binding-openapi.conf`, `binding-asyncapi.conf`, `binding-openapi-asyncapi.conf`:
- Remove the `jakarta.json.bind-api` dependency from `pom.xml` (or move it to `<scope>test</scope>` if that's preferred over relying on the transitive `engine.conf` re-export for tests)
- Remove `requires jakarta.json.bind;` from `module-info.java`
- Confirm `./mvnw clean install -pl <module>,... -DskipITs -q` still passes (readability to `jakarta.json.bind` is preserved transitively through `requires transitive` in `engine.conf`, so this should be a no-op for compilation)

Worth checking whether other config/runtime modules have the same redundant declaration while addressing this.

jfallows:
## Description

`binding-openapi.conf`, `binding-asyncapi.conf`, and `binding-openapi-asyncapi.conf` each declared a `jakarta.json.bind-api` Maven dependency and `requires jakarta.json.bind;` in `module-info.java`, but no main-source file in any of the three modules imports `jakarta.json.bind.*` — only `jakarta.json.*` (JSON-P) is used directly. The `jakarta.json.bind` types (`Jsonb`, `JsonbBuilder`, `JsonbConfig`) are used only in test sources, where they remain available transitively via `engine.conf`'s compile-scope dependency on `jakarta.json.bind-api`.

For each of the three modules:
- Removed the `jakarta.json.bind-api` dependency from `pom.xml`
- Removed `requires jakarta.json.bind;` from `module-info.java`

Verified with `./mvnw clean install -pl config/binding-openapi.conf,config/binding-asyncapi.conf,config/binding-openapi-asyncapi.conf` (checkstyle, license headers, jacoco coverage, notice check, unit tests, moditect module descriptor) — all pass, confirming this is a no-op for compilation as expected.

Fixes #2372

---
_Generated by [Claude Code](https://claude.ai/code/session_01SwkXJYfggxJ8dQnCoo54xX)_

### DIFF
--- a/config/binding-asyncapi.conf/pom.xml
+++ b/config/binding-asyncapi.conf/pom.xml
@@ -45,10 +45,6 @@
       <artifactId>jakarta.json-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>jakarta.json.bind</groupId>
-      <artifactId>jakarta.json.bind-api</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.eclipse</groupId>
       <artifactId>yasson</artifactId>
       <scope>test</scope>
--- a/config/binding-asyncapi.conf/src/main/moditect/module-info.java
+++ b/config/binding-asyncapi.conf/src/main/moditect/module-info.java
@@ -15,7 +15,6 @@
 module io.aklivity.zilla.config.binding.asyncapi
 {
     requires jakarta.json;
-    requires jakarta.json.bind;
     requires io.aklivity.zilla.config.engine;
 
     exports io.aklivity.zilla.config.binding.asyncapi;
--- a/config/binding-openapi-asyncapi.conf/pom.xml
+++ b/config/binding-openapi-asyncapi.conf/pom.xml
@@ -45,10 +45,6 @@
       <artifactId>jakarta.json-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>jakarta.json.bind</groupId>
-      <artifactId>jakarta.json.bind-api</artifactId>
-    </dependency>
-    <dependency>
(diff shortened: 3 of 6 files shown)


jfallows (commented):
LGTM

Issue #2280, opened 2026-08-02
binding-mcp-grpc: auto-generate MCP tools from gRPC service definitions


jfallows:
Depends on #2279.

### Motivation

`binding-mcp-openapi` already derives MCP tool definitions and routes from an OpenAPI document at config-composite time. A gRPC service definition offers the same opportunity: method → tool, message fields → JSON Schema input, without hand-written per-method `zilla.yaml` route configuration. This proposes a `binding-mcp-grpc` that reads a `FileDescriptorSet` at config-composite time and generates MCP tool routes the same way, using #2279's capability for annotation.

### Scope: unary RPCs only

- Unary → maps directly to MCP's single-request/single-response tool-call model.
- Client-streaming → could degenerate to a size-1 stream, but supporting it while excluding server-streaming/bidi creates an asymmetric, hard-to-explain capability story for little benefit.
- Server-streaming / bidi → no clean mapping onto a single terminal `CallToolResult`; would require buffering an unbounded/long-lived stream, which breaks down for anything non-trivial.

Non-unary methods are simply omitted from the generated tool set.

### Tools only, no MCP Resources

MCP Resources need an addressable URI template to build from — OpenAPI gets that for free from path parameters (`/items/{id}` already reads as a resource address), and even there it's an explicit per-route opt-in (`when.resource`), not inferred from HTTP method. gRPC has no equivalent: a method's input is a flat request message with no inherent split between "addressable path segments" and everything else. Inventing a synthetic convention for this would be new design surface with no concrete need behind it yet, so `binding-mcp-grpc` only ever generates Tools.

### Metadata: two opt-in paths, method-level only, one shared vocabulary

Mirrors `x-zilla-mcp`'s current structure (operation-level only; `title` + `description` + `annotations.{readOnlyHint,destructiveHint,idempotentHint,openWorldHint}` — the latter four being the real MCP specification's own `ToolAnnotations` fields, not protocol-specific to either OpenAPI or gRPC):

```json
{
  "title": "string",
  "description": "string",
  "annotations": {
    "readOnlyHint": true,
    "destructiveHint": true,
    "idempotentHint": true,
    "openWorldHint": true
  }
}
```

**1. Inline protobuf extension** (`.proto` authors who own their schema), published as a standalone importable file:

```protobuf
// zilla/mcp/v1/options.proto
syntax = "proto3";
package zilla.mcp.v1;

message McpToolAnnotations {
  bool read_only_hint = 1;
[...]

PR #2374, opened 2026-08-13
feat(common-protobuf): support method- and field-level options overlay

### LINKED ISSUE #2279: common-protobuf: support method- and field-level options overlay for FileDescriptorSet
### Motivation

`binding-mcp-openapi` overlays vendor extensions (e.g. `x-zilla-mcp`) onto an OpenAPI document using `common-json`'s generic `JsonOverlay`, targeting the document directly via JSONPath — straightforward because an OpenAPI document is already JSON with keyed, addressable structure. A `FileDescriptorProto` doesn't have that property: `service`/`method` and `message`/`field` are unkeyed repeated fields, so referencing a specific node requires resolving a friendly reference first.

This proposes adding that resolution + merge capability to `common-protobuf`, for two concrete consumers identified so far:

- **#2280** (`binding-mcp-grpc`) needs to attach `zilla.mcp.v1` tool annotations (`title`, `readOnlyHint`, etc. — the gRPC counterpart to #2222's work for `binding-mcp-openapi`/`binding-mcp-http`) onto specific gRPC methods.
- **`model: protobuf`** (zilla-platform#383's schema-annotation work, zilla#2318/#2302) needs to tag specific message fields (e.g. `zilla:tags: [ "EMAIL" ]`) for selection by the same `tagged:`/`named:` matchers #2302 already defines for Avro and JSON Schema.

Both need the identical shape: resolve a friendly reference into a specific node of a parsed protobuf schema, then merge an opaque options bag onto whatever that node's options extension already carries. Only the addressing syntax and the target `Options` message type differ.

### Scope

**Method and field.** Not a general facility for all nine of protobuf's option-bearing scopes (file, message, field, oneof, enum, enum value, service, method, extension range) — scoped to exactly the two concrete needs identified above. Broader scopes (e.g. file- or service-level) can be added later if a concrete need surfaces for them specifically.

### Prerequisite

Does `ProtobufSchema`/the method and field models in `common-protobuf` currently retain custom `MethodOptions`/`FieldOptions` extension values at all? Its purpose so far has been payload encode/decode validation (types, field numbers, wire types) — there's no evident reason it currently captures or exposes arbitrary custom option key/values for either. This capture capability is needed regardless of merge approach and should be treated as the first real design step, not assumed to already exist, for **both** node kinds.

### Design

**Always resolved from a catalog — no inline document shape.** This mirrors zilla#2318's settled design for its own `overlay:` (Avro/JSON Schema field annotations) and zilla#2365's consolidation of the same pattern across `binding-openapi`/`binding-asyncapi`/`binding-openapi-asyncapi`/`binding-mcp-openapi`: `overlay:` is always `{catalogName: {subject, version}}`, resolved through the same `CatalogHandler` SPI already used for the schema itself, never a document written directly inline.

Reference, at the point of use (e.g. `binding-mcp-grpc`'s own options — exact placement is that binding's own design, per #2280):

```yaml
overlay:
  my-overlay:
    subject: mcp-grpc-overlay
    version: latest
```

Resolved content (what the `my-overlay` catalog entry's `mcp-grpc-overlay` subject contains — declared via `catalog: {type: inline}` for the local case, exactly as #2318 §4 does, or fetched from a real registry):

```json
[
  {
    "method": "example.sql.v1.SQLService/Query",
    "options": { "zilla.mcp.v1": { "title": "Query SQL", "annotations": { "readOnlyHint": true } } }
  },
  {
    "field": "example.sql.v1.QueryResult.email",
    "options": { "zilla": { "tags": [ "EMAIL" ] } }
  }
]
```

- `method:` reuses the exact `package.Service/MethodName` (with `*` wildcard) addressing convention already established in `binding-grpc`/`binding-grpc-kafka` route conditions (the same regex pattern used by `GrpcKafkaConditionConfigAdapter`) — no new addressing scheme.
- `field:` uses the analogous dotted `package.Message.field_name` reference into the same parsed schema.
- Resolution: parse `method:`/`field:` into the specific node inside the parsed `FileDescriptorSet`/`ProtobufSchema` (assuming the prerequisite above is met) — walking `service.method` for the former, `message.field` for the latter. Both lists are unkeyed and repeated, which is exactly why this needs friendly-reference resolution rather than JSONPath tree search, for either node kind.
- Merge: directly merge two small `jakarta.json.JsonObject` values — whatever the inline extension already carries for that node (`MethodOptions` or `FieldOptions`, as applicable) and whatever the resolved entry's `options:` carries — overlay values winning per key. `jakarta.json.JsonMergePatch` covers this identically for both; no custom merge logic, no serialization of the whole descriptor to JSON, no dependency on `common-json`.
- The `options:` bag's contents stay opaque to `common-protobuf` — it merges whatever keys are given, with zero knowledge of any specific annotation vocabulary (e.g. `zilla.mcp.v1`, `zilla`). That interpretation belongs entirely to the consuming binding or model.
- No `remove`/deletion verb — this capability only ever injects metadata, never restructures the underlying schema. gRPC's wire encoding is fixed by whatever the real backend was compiled from; structural edits here would desync a consumer's local view from the real server's actual behavior. This applies equally to the field case: annotating a field never changes its wire type or number.
- Caching follows zilla#2318 §6's rule directly: pinned version (or `catalog: {type: inline}`) caches by the resolved schema handle alone; a non-pinned version needs the paired resolved-id key, for the identical reason.

### Open questions

- The capture-of-custom-options prerequisite above — exact API shape, for both `MethodOptions` and `FieldOptions`.
- Whether this needs to support the binary `FileDescriptorSet` input path (`ProtobufSchemaCompiler`) in addition to the `.proto`-source-text path (`ProtobufSourceCompiler`), since only the latter is currently wired into the runtime model (e.g. via `catalog-inline`). If a pre-compiled descriptor (from external `protoc`/buf, or gRPC server reflection) is ever supported as input, real protobuf extension field-number rules apply to that path even though they don't for the text-source path.
- `model: protobuf`'s own schema-patch dispatch (mirroring how `catalog:`/`overlay:` are dispatched per-model for Avro and JSON Schema, per zilla#2318 §3) still needs `field:`/`options:` wired into its `if model===protobuf then` branch once this lands — tracked on zilla#2318, not duplicated here.
- `binding-mcp-grpc` (#2280) still needs to settle exactly where its own `overlay:` reference sits in its config (peer of whatever its schema-source reference is — itself an open question in #2280, given the catalog-vs-reflection schema-sourcing question there) — this issue only fixes the *shape* `overlay:` resolves through, not `binding-mcp-grpc`'s surrounding config structure.

jfallows:
## Description

Adds `ProtobufService`/`ProtobufMethod` (service/rpc parsing — previously parsed by the grammar but silently dropped by `ProtobufSourceCompiler`'s listeners, with no model type at all) and `ProtobufOverlay`, which resolves a `method:`/`field:` friendly reference against a `ProtobufSchema` and merges an `options:` bag onto the referenced node via `jakarta.json.JsonMergePatch`:

- `method:` addresses `package.Service/MethodName`, with a `*` wildcard matched anywhere in the string (mirroring, not depending on, `binding-grpc`'s `GrpcConditionMatcher` wildcard-to-regex transform — `common-protobuf` can't depend on `binding-grpc`).
- `field:` addresses the exact, non-wildcarded `package.Message.field_name`.
- The merge uses `jakarta.json.Json.createMergePatch(...)`, already implemented and tested in `common-json` (`JsonProviderImpl`/`JsonValues.mergePatch`) and reachable via the public `jakarta.json` facade — no new dependency needed.

`ProtobufConstant` gains `toJson()`/`toJsonObject()` to bridge the ANTLR-sourced option constants into `jakarta.json`, normalizing a parenthesized custom/extension option reference (`"(zilla.mcp.v1)"`) to its bare dotted name (`"zilla.mcp.v1"`) — that punctuation is only how the `.proto` grammar distinguishes an extension reference from a built-in option name, and an overlay's `options:` bag always addresses the bare name, so without normalizing, an inline option and an overlay entry for "the same" option would merge as unrelated sibling keys instead of colliding on one.

**Reuse across schemas.** One `ProtobufOverlay` is meant to be applied to many different schemas, most of which won't contain what a given entry addresses: `apply(schema)` returns the identical input instance, unmodified, when nothing matches. When something does match, only the touched messages/services are rebuilt — everything else is carried through by reference, unshared with any other `ProtobufSchema` instance. This required a new `ProtobufSchema.Builder.reuse(...)`, distinct from the existing `message(...)`: `ProtobufField.resolve()` mutates the field's cached `.message()`/`.enumeration()` link in place during the relink pass, so an "unchanged" message added via the ordinary `message(...)` path would still have its shared field objects silently re-relinked (and thus mutated) by a second `build()` call — corrupting the original schema's own view through those same shared objects. `reuse(...)` adds a message for lookup without re-running the relink pass over it, since it's already correctly linked from the schema it came from. `ProtobufMethod`/`ProtobufService` have no such mutable state, so untouched methods inside a rebuilt service are simply passed through unchanged with no copy at all.

**Scope**, matching the issue's own boundaries and cross-referenced follow-up work:
- Only the ANTLR/text-source path (`ProtobufSourceCompiler`) gets service/rpc modeling and the overlay mechanism — `ProtobufSchemaCompiler` (binary `FileDescriptorSet` path) has zero production callers today and is left untouched.
- `field:` is exact-reference only, no wildcard — bulk field selection by pattern is `named:`'s job (#2302), not this mechanism's.
- No `CatalogHandler`/config-schema/`model-protobuf`/`binding-mcp-grpc` wiring — this issue owns reference resolution and the options merge only, per #2280's own "module placement" note. Wiring this into `model: protobuf`'s `overlay:` catalog entry is #2318; wiring it into `binding-mcp-grpc`'s `zilla.mcp.v1` tool annotations is #2280.

Fixes #2279

## Test plan

- [x] `ProtobufSourceCompilerTest` extended with service/rpc parsing (proto2 + proto3), streaming flags, and method-option capture
- [x] New `ProtobufOverlayTest`: exact method match, wildcard match across multiple rpcs, exact field match, later-entry-wins merge precedence, no-op (identical instance returned) when nothing matches, untouched messages/services reused by reference (`assertSame`), one overlay reused across two schemas (matching + non-matching), malformed entry rejection (neither/both of `method`/`field`, missing `options`)
- [x] `./mvnw test -pl runtime/common-protobuf` — 212/212 passing (full module, no regressions)
- [x] `./mvnw checkstyle:check -pl runtime/common-protobuf` — 0 violations
- [x] `./mvnw test jacoco:check -pl runtime/common-protobuf` — coverage requirements met (90% instruction, 0 missed classes)

---
_Generated by [Claude Code](https://claude.ai/code/session_01UZYtsEXoSX13cn1R5s1Ztb)_

### DIFF
--- a/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufConstant.java
+++ b/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufConstant.java
@@ -89,8 +89,8 @@ public sealed interface ProtobufConstant
      * custom/extension option's surrounding {@code ( ... )} stripped from its key — that punctuation is
      * only how the {@code .proto} grammar distinguishes an extension reference from a built-in option
      * name, not part of the option's logical (dotted) identity, and a {@link ProtobufOverlay} entry's
-     * {@code options:} bag always addresses that identity unparenthesized (e.g. {@code "zilla.mcp.v1"},
-     * never {@code "(zilla.mcp.v1)"}); stripping it here is what lets the two sides merge onto the same
+     * {@code options:} bag always addresses that identity unparenthesized (e.g. {@code "acme.meta"},
+     * never {@code "(acme.meta)"}); stripping it here is what lets the two sides merge onto the same
      * key rather than coexist as unrelated siblings. A nested {@code MessageValue}'s own field names are
      * never parenthesized, so this is a no-op wherever it doesn't apply. {@code null} converts to an
      * empty object.
--- a/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufConstant.java
+++ b/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufConstant.java
@@ -16,6 +16,11 @@ package io.aklivity.zilla.runtime.common.protobuf;
 
 import java.util.Map;
 
+import jakarta.json.Json;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonObjectBuilder;
+import jakarta.json.JsonValue;
+
 /**
  * The value of a field option, parsed from the {@code .proto} source text (a text-format literal) rather
  * than resolved against an extension descriptor — no {@code .proto} import declares these custom options,
@@ -57,5 +62,55 @@ public sealed interface ProtobufConstant
      */
     record MessageValue(Map<String, ProtobufConstant> fields) implements ProtobufConstant
     {
+    }
+
+    /**
+     * This constant converted to a {@link JsonValue}, so an inline {@code .proto} source option and a
+     * {@link ProtobufOverlay} entry's {@code options:} bag can be compared and merged in one common
+     * representation. A message-valued constant converts recursively; every other variant maps directly
+     * to its JSON-native equivalent.
+     */
+    default JsonValue toJson()
+    {
+        return switch (this)
+        {
+        case Identifier value -> Json.createValue(value.value());
+        case IntegerValue value -> Json.createValue(value.value());
+        case FloatValue value -> Json.createValue(value.value());
+        case TextValue value -> Json.createValue(value.value());
+        case BooleanValue value -> value.value() ? JsonValue.TRUE : JsonValue.FALSE;
+        case MessageValue value -> toJsonObject(value.fields());
+        };
+    }
+
+    /**
+     * Converts a field's or method's full option map ({@link ProtobufField#option(String)},
+     * {@link ProtobufMethod#option(String)}) to a {@link JsonObject}, one key per option name, with a
+     * custom/extension option's surrounding {@code ( ... )} stripped from its key — that punctuation is
+     * only how the {@code .proto} grammar distinguishes an extension reference from a built-in option
+     * name, not part of the option's logical (dotted) identity, and a {@link ProtobufOverlay} entry's
+     * {@code options:} bag always addresses that identity unparenthesized (e.g. {@code "zilla.mcp.v1"},
+     * never {@code "(zilla.mcp.v1)"}); stripping it here is what lets the two sides merge onto the same
+     * key rather than coexist as unrelated siblings. A nested {@code MessageValue}'s own field names are
+     * never parenthesized, so this is a no-op wherever it doesn't apply. {@code null} converts to an
+     * empty object.
+     */
+    static JsonObject toJsonObject(
+        Map<String, ProtobufConstant> options)
+    {
+        JsonObjectBuilder builder = Json.createObjectBuilder();
+        if (options != null)
+        {
+            options.forEach((name, value) -> builder.add(unparenthesize(name), value.toJson()));
+        }
+        return builder.build();
+    }
+
+    private static String unparenthesize(
+        String name)
+    {
+        return name.length() >= 2 && name.charAt(0) == '(' && name.charAt(name.length() - 1) == ')'
+            ? name.substring(1, name.length() - 1)
+            : name;
     }
 }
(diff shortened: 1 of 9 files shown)


jfallows (commented):
LGTM

PR #2378, opened 2026-08-14
docs: add rule for justifying changes to shared components

jfallows:
## Description

PR/issue descriptions for a shared component (`common-avro`, `common-json`, `common-protobuf`, the engine SPI, any `runtime/*/ext` module) kept leaking a specific dependent's use case into the justification — e.g. citing what a particular sibling module or downstream consumer needed, rather than describing the gap in the shared component's own contract. This couples the shared component's description to whoever happened to want the change, and the coupling recurs regardless of whether that consumer is in-tree or external.

Adds one general rule to the root `AGENTS.md`'s "Contribution workflow" section: justify a shared component's change entirely by its own contract gap, never by naming or describing a consumer. Includes a concrete self-check ("strip away everyone who depends on this component — does the description still stand on its own?") and states explicitly that this applies symmetrically in both directions of the dependency graph.

This is a documentation-only change with no code impact.

Fixes # (none — process documentation, not tied to a specific issue)

---
🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01GKWELiLVoasTwsuvaXg2Y3

---
_Generated by [Claude Code](https://claude.ai/code/session_01GKWELiLVoasTwsuvaXg2Y3)_

PR #2379, opened 2026-08-14
docs(engine): drop in-tree-consumer example from SPI PR neutrality rule

jfallows:
## Description

`runtime/engine/AGENTS.md`'s existing neutrality rule for engine SPI PR descriptions carved out an exception: citing an in-tree consumer's use case (e.g. "required by `binding-mcp` elicitation flow") was presented as acceptable, only naming a downstream/proprietary consumer was disallowed. That carve-out is itself the anti-pattern — coupling an SPI change's justification to a specific dependent is the same layering violation whether the dependent is in this repo or external to it.

Replaces the carved-out example with a pointer to the general rule (see #2378, opened separately): justify a shared component's change by its own contract gap only, never by naming any consumer.

Independent of #2378 — this PR stands on its own (the link target section will resolve once #2378 merges; either can merge first with no functional dependency between them).

This is a documentation-only change with no code impact.

Fixes # (none — process documentation, not tied to a specific issue)

---

https://claude.ai/code/session_01GKWELiLVoasTwsuvaXg2Y3

---

### DIFF
--- a/runtime/engine/AGENTS.md
+++ b/runtime/engine/AGENTS.md
@@ -27,10 +27,11 @@ Two conventions to follow when adding to or modifying them:
   (`Set.of()`, `List.of()`) return cached singletons, the project convention
   is `null` so callers learn one rule across the SPI rather than mixing styles.
 
-PR descriptions for engine SPI changes follow the same neutrality — describe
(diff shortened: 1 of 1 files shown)


jfallows (commented):
LGTM

PR #2381, opened 2026-08-14
test(engine): apply the configured value model as decoder on reply direction too

jfallows:
## Description

`TestBindingFactory` (the `type: test` binding backing k3po `*IT.java` tests engine-wide) only ever applied a configured value model as an **encoder**, on the **initial/produce** direction. Reply-direction data was always forwarded raw, with no model applied at all — so no k3po IT anywhere could exercise a model's decode-time behavior through a live engine; that coverage only ever existed as plain JUnit tests bypassing the engine entirely.

Adds a second pipeline (`supplyDecoder()`) applied on the reply direction, mirroring the existing initial-direction encoder, so a real `EngineRule`+`K3poRule` IT can prove decode-time behavior (e.g. a model extension's field substitution) actually fires end-to-end through a live engine.

### Safety

Gated on the same `valueModel != null` check the existing encoder uses, so every `type: test` usage without a value model is entirely unaffected. For usages that do configure a value model: audited every `type: test` binding with a `value:` model configured across the whole reactor (`model-avro.spec`, `model-json.spec`, `model-protobuf.spec`, `model-core.spec`) and confirmed none currently sends or asserts a reply-direction payload — every existing scenario is produce/initial-only (the exact gap this closes), so applying a decoder in that direction is a no-op for every existing assertion.

### Test plan

- [x] `./mvnw clean verify -pl runtime/engine` — 215 unit tests + 183 k3po ITs, all green
- [x] `./mvnw clean verify -pl runtime/model-avro,runtime/model-json,runtime/model-protobuf,runtime/model-core` — all four modules' own ITs (which share this same test binding) green, no regressions
- [x] checkstyle, license headers clean

Fixes # (none — shared test infrastructure enhancement, not tied to a specific issue)

---

https://claude.ai/code/session_01GKWELiLVoasTwsuvaXg2Y3

---

### DIFF
--- a/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
+++ b/runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
@@ -336,6 +336,9 @@ final class TestBindingFactory implements BindingHandler
         private long pendingTraceId;
         private boolean storeAssertionsStarted;
         private final ModelPipeline valuePipeline;
+        // mirrors valuePipeline for the opposite direction: a real binding decodes on read (reply) and
+        // encodes on write (initial), so this test binding applies the same configured model both ways
+        private final ModelPipeline replyPipeline;
 
         private TestSource(
             MessageConsumer source,
@@ -352,6 +355,7 @@ final class TestBindingFactory implements BindingHandler
             this.replyId = replyId;
             this.target = resolvedId != 0L ? new TestTarget(routedId, resolvedId) : null;
             this.valuePipeline = valueModel != null ? valueModel.supplyEncoder() : null;
+            this.replyPipeline = valueModel != null ? valueModel.supplyDecoder() : null;
         }
 
         private void doAuthorize(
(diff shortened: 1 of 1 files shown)


runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/binding/TestBindingFactory.java
```diff
@@ -336,6 +336,9 @@ private final class TestSource
         private long pendingTraceId;
         private boolean storeAssertionsStarted;
         private final ModelPipeline valuePipeline;
+        // mirrors valuePipeline for the opposite direction: a real binding decodes on read (reply) and
+        // encodes on write (initial), so this test binding applies the same configured model both ways
+        private final ModelPipeline replyPipeline;
```
jfallows:
```suggestion
        private final ModelPipeline initialPipeline;
        private final ModelPipeline replyPipeline;
```

jfallows: Actually, this should just be `pipeline` for the encoder side in TestSource, and also a field called `pipeline` for the decoder side in TestTarget instead of implementing both sides in TestSource.

jfallows (commented):
LGTM

PR #2396, opened 2026-08-17
feat(engine): generic named-config refs() resolution, plus TestBinding authorization/envelope/secret-key support

jfallows:
## Description

A model has no way to reference a vault-resolved key material source, the same way it already references a catalog for schema resolution. This adds that, generalized: a `Config.Extensible` can now carry a `refs()` list of `NamedConfig` (a vault, a guard, a catalog, an overlay, ...) that `EngineManager` resolves name → id with **one generic walk**, rather than per-type special-casing.

```yaml
model: avro
vault:
  vault0: {}
```

Part of the seam work for #1142 (avro field encryption in `model-avro-ext`, aklivity/zilla-plus) — this PR only adds the generic capability gap being closed in oss zilla; it does not itself add encryption.

### Named-reference resolution

- `NamedConfig` — common name/id supertype for `VaultedConfig`, `GuardedConfig`, `CatalogedConfig`, and `OverlayConfig`. Carries a `NamedConfig(String name)` convenience constructor for subclasses that support no extensions.
- `VaultedConfig` (new) — a keyed vault reference (`vault: { <name>: {...} }`, matching the existing `catalog:` shape), itself `Config.Extensible` so other components can attach their own named extensions to a *specific* vault reference the way `ModelExtInfo` already extends the model itself.
- `ModelConfig.refs()` bubbles `cataloged`, each schema's `overlay`, and any extension-contributed refs into one list; `EngineManager` resolves every entry the same way it already resolves `route.guarded`/`exporter.vault`.
- `OverlayConfig` now extends `NamedConfig` instead of duplicating `id`/`name` fields on top of `Config` — this is what let the schema-specific overlay-resolution walk in `EngineManager` be deleted entirely in favor of the generic `refs()` loop.
- `ConfigExtAdapter` widened from `T` to `Config.Extensible` on its encode side (decode was already generic) so any `Extensible` builder — not just a fixed top-level type — can be adapted.

### Engine SPI: `AvroController.authorization()`

Adds `AvroController.authorization()`/`AvroPipeline.authorization(long)` so a native `AvroTransform` stage can read the current stream's resolved authorization, mirroring the existing `ModelTransform` bridge. Needed by any transform stage that gates its behavior on the caller's authorization (e.g. a guarded encrypt/decrypt policy).

### `TestBindingFactory`/`TestVaultHandler` test-infrastructure gaps

Closing these was necessary to exercise the above via k3po ITs without depending on a production vault or binding module:

- **Authorization was never propagated.** `TestBindingFactory` never stamped the guard-resolved session authorization onto the `Begin`/`Data` frames it forwards, so `AvroController.authorization()` was always `0` in a test regardless of guard config — a real gap against production bindings (e.g. `binding-tls` already asserts this on its own accept scripts). Fixed, and scoped per-connection (a prior in-progress commit had it as a shared factory-level field, clobbered across connections on the same worker).
- **No way to seed or assert a `ModelEnvelope`.** New `TestBindingOptionsConfig` knobs: `options.envelope` seeds a value model's envelope with known entries before transform runs (for decode-direction transforms that depend on out-of-value metadata already being present), and `options.assertions.envelope` checks the envelope's contents afterward (for encode-direction transforms whose envelope writes are non-deterministic, e.g. randomly-IV'd, and can't be matched literally over the wire).
- **Vault key resolution was async-only.** `VaultHandler.wrap()`/`unwrap()` are asynchronous, forcing every caller to tolerate a cold-miss-then-retry cycle even when resolution is actually synchronous. Added `VaultHandler.initSecretKeys(List<String>)` → `SecretKeyManagerFactory`, mirroring the existing `initKeys()`/`initSigners()`/`initTrust()` → `KeyManagerFactory`/`TrustManagerFactory` shape TLS already uses: resolve once at attach, then wrap/unwrap synchronously for the binding's life. Implemented on `TestVaultHandler` and `FileSystemVaultHandler`; the existing async methods are unchanged. `TestVaultHandler`'s `wrap:` config also switched from a SHA-256-hashed passphrase string to a declarative PEM entry, matching `key:`/`trust:`/`signer:`'s existing shape.
- A decode-direction `REJECTED`/overflow path in `TestBindingFactory` only aborted the reply side; added the matching initial-side abort so a decode rejection tears down the stream symmetrically, matching the encode-direction behavior.

### Testing

- `./mvnw -pl config/engine.conf,config/model-avro.conf,config/model-json.conf,config/model-protobuf.conf,runtime/engine -am clean install` — checkstyle clean, all tests passing (including `OverlayConfigTest`, `VaultedConfigTest`, the full `EngineIT`/`DuplexIT`/`HalfDuplexIT`/`SimplexIT` flow-control suite, and `AvroModelOverlayTest`), JaCoCo thresholds met


runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/vault/TestVaultHandler.java
```diff
@@ -62,6 +86,123 @@
...
+            {
+                byte[] iv = new byte[GCM_IV_LENGTH];
+                bytes.getBytes(index, iv);
+
+                int ciphertextLength = length - GCM_IV_LENGTH;
+                byte[] ciphertext = new byte[ciphertextLength];
+                bytes.getBytes(index + GCM_IV_LENGTH, ciphertext);
+
+                Cipher cipher = Cipher.getInstance(CIPHER_TRANSFORM);
+                cipher.init(Cipher.DECRYPT_MODE, secretKey, new GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv));
```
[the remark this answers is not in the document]
jfallows: This is a test fixture.

config/engine.conf/src/main/java/io/aklivity/zilla/config/engine/NamedConfig.java
```diff
@@ -0,0 +1,41 @@
...
+ * the same generic walk regardless of which concrete kind of named config this is.
+ *
+ * @see Config.Extensible#refs()
+ */
+public abstract class NamedConfig extends Config.Extensible
+{
+    public transient long id;
+
+    public final String name;
+
```
jfallows:
```
   protected NamedConfig(
        String name)
    {
        this(name, null);
    }
```

Then use this from subclasses that don't support extensions.

runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineManager.java
jfallows: Should OverlayConfig not be included in transitive NamedConfigs collected on model, then we can simplify by removing this model walk.

runtime/engine/pom.xml
```diff
@@ -180,16 +180,30 @@
...
+          </execution>
+          <execution>
+            <id>test-generate</id>
+            <goals>
               <goal>test-generate</goal>
             </goals>
+            <configuration>
+              <!-- "test" scope carries the TestBinding-only envelope wire extension (TestDataEx) used by
+                   TestBindingFactory; it must never leak into the main (generate) output above, since it
+                   is test-sources-only infrastructure. -->
```
jfallows: Remove this comment

runtime/engine/src/test/java/io/aklivity/zilla/runtime/engine/test/internal/vault/TestVaultHandler.java
```diff
@@ -260,4 +398,120 @@
...
+                {
+                    byte[] iv = new byte[GCM_IV_LENGTH];
+                    bytes.getBytes(index, iv);
+
+                    int ciphertextLength = length - GCM_IV_LENGTH;
+                    byte[] ciphertext = new byte[ciphertextLength];
+                    bytes.getBytes(index + GCM_IV_LENGTH, ciphertext);
+
+                    Cipher cipher = Cipher.getInstance(CIPHER_TRANSFORM);
+                    cipher.init(Cipher.DECRYPT_MODE, secretKey, new GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv));
```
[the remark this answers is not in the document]
jfallows: Same as the flag at line 193, already dismissed above — this is the synchronous `SecretKeyManager.unwrap()` added alongside the async path, same test fixture, same shape (IV read from the wire ciphertext per call, not hardcoded).

---

jfallows (commented):
See comments

PR #2397, opened 2026-08-17
Support wildcard-registered subjects in inline catalog

akrambek:
## Summary
- `InlineCatalogHandler.resolve(subject, version)` now falls back to wildcard-registered subjects (e.g. `orders.*-value`) when an exact literal-key lookup misses.
- A registered subject containing `*` is compiled to a regex at registration time; resolve first tries the exact key, then matches against registered patterns in registration order. Exact matches still take precedence, and unregistering a wildcard subject removes its pattern.
- Loosened the inline-catalog config schema's subject-key pattern to admit `*`, and added a `catalog.wildcard.yaml` spec fixture + schema test.
- No `CatalogHandler` SPI changes and no authorization/guard involvement — this dropped an earlier version of this PR that threaded `authorization` through `CatalogHandler.resolve` and the JSON/Protobuf/Avro model handlers for guarded-expression resolution. That turned out to be unnecessary: wildcard matching alone solves the actual problem (subject matching for topic-aliased inline-catalog channels), so the authorization plumbing was reverted back to the `develop` baseline entirely.

## Test plan
- [x] `./mvnw -pl runtime/engine,runtime/catalog-inline,runtime/model-json,runtime/model-protobuf,runtime/model-avro,specs/catalog-inline.spec clean verify` — green, checkstyle clean
- [x] New `InlineCatalogHandlerTest` cases: literal-vs-wildcard resolution, prefix/suffix/version mismatch, exact-over-wildcard precedence, unregister removes the pattern
- [x] New `SchemaTest.shouldValidateCatalogWithWildcardSubject`


specs/engine.spec/src/main/scripts/io/aklivity/zilla/specs/engine/schema/engine.schema.json
```diff
@@ -267,6 +267,11 @@
                     "title": "Vault",
                     "type": "string"
                 },
+                "guard":
+                {
+                    "title": "Guard",
+                    "type": "string"
+                },
```
jfallows: Why do we need to put `guard` on `catalog` to support `guarded['jwt0'].identity` style expressions for `inline` catalog?

akrambek: Good catch — we don't. Pushed a fix: dropped `CatalogConfig.guard`/`guardId` entirely (and the schema/adapter changes for it), and gave `CatalogConfig` a general transient `resolveId` instead, mirroring the existing `Route`/`Binding`/`GuardConfig.resolveId` convention.

`InlineCatalogHandler` now resolves the guard named inside `${guarded['name'].identity}`/`.attributes.x}` dynamically per expression — same approach as `TlsWithResolver`/`HttpKafkaWithResolver`/`SseKafkaWithResolver`: extract the name from the match, look it up lazily via `context.supplyGuard(catalog.resolveId.applyAsLong(name))`, memoize per name. No static single-guard field, and multiple distinct guard names now work in expressions on the same catalog (added a test for that too).

Pushed as 1d9ff1bbf8.

akrambek: Update: on reflection with the team, we dropped the inline-catalog guard/wildcard work from this PR entirely rather than fixing the resolveId approach — the inline catalog's `resolve()` subject is always a static string bound from config, never derived from the actual message topic at runtime, so once you pair a guarded-expression substitution with wildcard matching, the substitution's output stops mattering (the wildcard matches literally anything, including a blank identity). No real caller needs it today.

So this PR is back to its original, minimal scope: `CatalogHandler.resolve(subject, version, authorization)` as a default method, plus authorization threaded through the three model handlers' `resolveSchemaId(...)`. No catalog implements the new overload — pure groundwork, discussed as its own follow-up whenever a real per-identity-resolution catalog shows up. Pushed as 98acabd82e.

akrambek: Further update: on discussing with the team, we realized the wildcard matching alone already solves the subject-matching problem (a wildcard like `orders.*-value` matches any concrete subject, so guarded-expression resolution added nothing on top of it). So I've now reverted the `authorization`-threading part too — `CatalogHandler`, and the JSON/Protobuf/Avro model handlers/pipelines, are all back to the `develop` baseline with zero authorization/guard involvement. The PR is now scoped to just wildcard-registered-subject matching in `InlineCatalogHandler`, done purely via the existing 2-arg `resolve(subject, version)` method — no SPI changes at all. Updated the PR title/description to match.

jfallows (approved):

PR #2398, opened 2026-08-17
Resolve guarded expressions in inline catalog subject before lookup

akrambek:
## Summary
- Stacks on #2397 (`CatalogHandler.resolve(subject, version, authorization)`). Once #2397 merges this PR's diff will shrink to just the changes below.
- Adds `guard`/`guardId` to `CatalogConfig`, mirroring the existing `vault`/`vaultId` pattern exactly (all 6 catalog subtypes + builders + JSON adapter + `engine.schema.json`).
- `InlineCatalogHandler.resolve(subject, version, authorization)` now resolves `${guarded['name'].identity}` and `${guarded['name'].attributes.x}` expressions in `subject` via the configured `GuardHandler` before the `schemaIds` lookup — same syntax `topics[].alias` already supports for kafka-proxy routing (`TlsWithResolver`/`HttpKafkaWithResolver`/`GrpcKafkaWithResolver`/`SseKafkaWithResolver`), now available on the catalog side too.
- The guard name inside the expression brackets is not re-resolved — an inline catalog carries at most one configured `guard`, applied regardless of which name literal appears in `${guarded['...']...}`. Flagging this as an open design question: happy to add per-name guard resolution if reviewers want closer parity with the binding-side resolvers.
- No schema change needed on the subject-being-resolved side (`catalog.subject` in `model-json`/`model-protobuf`/`model-avro` is already an unconstrained string).

## Test plan
- [x] `./mvnw -pl runtime/engine,config/engine.conf,config/catalog-inline.conf,config/catalog-apicurio.conf,config/catalog-schema-registry.conf,config/catalog-filesystem.conf,config/catalog-karapace.conf,runtime/catalog-inline install -o` — green, checkstyle clean across all 8 modules
- [x] New `InlineCatalogHandlerTest` cases: identity-expression resolution, attribute-expression resolution, no-guard-configured fallback (`NO_SCHEMA_ID`), and 2-arg overload ignores expression syntax entirely (backward compat)
- [x] Existing `InlineIT` (3 tests) unaffected

### DIFF
--- a/config/catalog-apicurio.conf/src/main/java/io/aklivity/zilla/config/catalog/apicurio/ApicurioCatalogConfig.java
+++ b/config/catalog-apicurio.conf/src/main/java/io/aklivity/zilla/config/catalog/apicurio/ApicurioCatalogConfig.java
@@ -37,8 +37,9 @@ public final class ApicurioCatalogConfig extends CatalogConfig
         String name,
         String type,
         String vault,
+        String guard,
         OptionsConfig options)
     {
-        super(namespace, name, type, vault, options);
+        super(namespace, name, type, vault, guard, options);
     }
 }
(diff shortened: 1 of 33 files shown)


akrambek:
Folding into #2397 — single PR per request.

PR #2404, opened 2026-08-20
feat(engine): encapsulate label management in the Router SPI

jfallows:
## Summary

- Move label-id assignment and lookup from the engine's internal `LabelManager` singleton onto the `Router`/`RouterContext` SPI (`supplyLabelId`, `supplyLabel`, and `watchLabels` on `Router`), so a `Router` implementation owns label-id coordination — letting a future cluster-aware router supply its own consistency mechanism instead of relying on today's process-local, append-order assignment.
- `EngineRouter` absorbs `LabelManager`'s in-memory behavior (seeded from the existing `<directory>/labels` file); `internal.LabelManager` is deleted. `TestRouter` mirrors the same behavior so `type: test` router configs keep working.
- The engine remains solely responsible for bootstrapping from and flushing to the on-disk `labels` file, independent of which `Router` implementation is active, so `zilla dump`'s independent reader of that file format is unaffected.
- No new binding and no `Router.store(): Store` — `Router` is already the correct process-wide SPI extension point, and routing label consistency through a generic `Store` would force an implicit mandatory `store-memory` runtime dependency for the default case and mismatch the CP guarantees a future cluster router needs.

## Test plan

- [x] `./mvnw clean verify -pl runtime/engine` — 226 unit tests + 208 integration tests pass, including `EngineIT` and `RouterFactoryTest` (extended with new coverage for `supplyLabelId`/`supplyLabel` through both `Router` and `RouterContext`, `watchLabels` notification, and bootstrap-from-existing-file correctness)
- [x] JaCoCo coverage checks pass
- [x] Confirmed `incubator/command-dump`'s independent `labels` file reader has no coupling to the deleted `internal.LabelManager` class

---
_Generated by [Claude Code](https://claude.ai/code/session_01FEncF8TdASSjY9Q113NMkv)_


jfallows (commented):
LGTM

jfallows:
`Build (25)` and CodeQL's `Analyze (java)` are both red, but on the same pre-existing, unrelated cause: `runtime/binding-tcp/NOTICE` fails `notice:check` because the checked-in file already lists `zilla::runtime::common-lang` as a dependency, even on `develop` — where `runtime/engine/pom.xml` doesn't declare `common-lang` at all (confirmed via `git show origin/develop:runtime/engine/pom.xml`, no match). That NOTICE entry was checked in by #2231, unrelated to this PR, and this PR never touches `binding-tcp`.

Not fixing it here since it's out of scope for this change. Proposed fix for a follow-up: regenerate `runtime/binding-tcp/NOTICE` via `./mvnw notice:generate --projects runtime/binding-tcp` against current `develop` (or investigate why #2231's checked-in NOTICE doesn't match its own dependency tree).

Will keep watching and re-run once the base recovers.

---

PR #2406, opened 2026-08-20
refactor(model-ext): drop unused name()/type() surface from model extension SPIs

jfallows:
## Description

`AvroModelExt`, `JsonModelExt`, `ProtobufModelExt`, `BytesModelExt`, and `StringModelExt` each declared a `name()` method, documented as "for diagnostics," with no caller anywhere in the engine or any installed extension. Their paired `*ModelExtFactorySpi` interfaces extended `FactorySpi` solely to inherit `type()`, but unlike a top-level `FactorySpi` (`ModelFactorySpi`, `BindingFactorySpi`, etc.) that value is never read for a model extension: `Factory.instantiate(Iterable<S>)` returns every discovered extension unconditionally, with no type-keyed lookup the way `Factory.instantiate(Iterable<S>, Function<Map<String,S>,F>)` provides for top-level factories. Every existing implementation (including this repo's own test-only extensions) was left inventing an arbitrary, functionally meaningless string for both methods just to satisfy the interface.

This removes both:
- `name()` is dropped from each `*ModelExt` interface.
- Each `*ModelExtFactorySpi` interface no longer extends `FactorySpi`, so `type()` is gone entirely rather than defaulted — there is no method left for an implementation to accidentally override. Each is now a plain single-method interface with just `create(Configuration)`.

All in-repo test-only implementations (`TestUppercase*ModelExtFactorySpi`, `TestMarkedBytesModelExtFactorySpi`, and the anonymous extensions in `BytesModelTest`/`StringModelTest`) are updated to drop their now-nonexistent overrides.

No behavior change for any installed extension's actual pipeline construction — this only removes dead SPI surface.

Fixes # (issue)

---
_Generated by [Claude Code](https://claude.ai/code/session_01WjPV7qF1b49GfqByji6SFp)_

### DIFF
--- a/runtime/model-avro/src/main/java/io/aklivity/zilla/runtime/model/avro/ext/AvroModelExt.java
+++ b/runtime/model-avro/src/main/java/io/aklivity/zilla/runtime/model/avro/ext/AvroModelExt.java
@@ -24,13 +24,6 @@ import io.aklivity.zilla.runtime.engine.EngineContext;
 public interface AvroModelExt
 {
     /**
-     * Returns a name identifying this extension, for diagnostics.
-     *
-     * @return the extension name
-     */
-    String name();
-
-    /**
      * Supplies the per-worker context for this extension.
      *
      * @param context  the engine context for the current worker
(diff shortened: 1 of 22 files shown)


jfallows (commented):
LGTM

PR #2412, opened 2026-08-21
fix(engine): honor affinity's embedded worker index for routing

jfallows:
## Description

`EngineContext#supplyInitialId(bindingId, affinityId)` is documented to pin a stream to "the worker selected by affinityId", and `EngineContext#affinity()` documents `affinityId` as encoding a node byte plus a 24-bit worker index. Despite that, the implementation resolved the target worker via `Math.floorMod(affinityId, mask.cardinality())` over the binding's tuning mask — discarding the worker index actually embedded in `affinityId` and re-deriving a slot from scratch.

This only happened to select the intended worker when a binding's tuning mask had power-of-two cardinality (`2^24 mod cardinality == 0`). For any other cardinality, once `affinityId` carries a non-zero node byte, the modulo folds the node and worker bits together, and can resolve to the wrong worker — a latent bug that a non-power-of-two worker/core count exposes even without ever leaving a single node.

This PR resolves the worker directly from `affinityId`'s own embedded index when it names a worker local to this engine node (bounds-checked against the configured worker count), falling back to the existing round-robin selection when it doesn't (e.g. the affinity names a different node, or an out-of-range index).

It also closes a related gap in the `Router`/`RouterContext` SPI: today a `Router` can only compose the engine's stream factory, with no way to redirect a stream to a different routed binding before its initial id is minted, short of rewriting frame bytes as they pass through the composed handler. This adds `Router`/`RouterContext#resolveRoutedId(routedId, affinity)` (identity by default) as an explicit hook for that decision, consulted once at mint time.

Finally, `Begin` gains a `redirectedId` field (default `0`). A `REDIRECT`-driven `Begin` can now record the pre-redirect `routedId` for a downstream binding to recover after `routedId` itself has been resolved to the actual dispatch target — without requiring the engine's own dispatch (`handleBeginInitial`) to be router-aware, and without any change needed to existing `Begin` construction call sites (the new field defaults like `Frame`'s existing `timestamp`/`traceId`/`authorization`).

## Test plan

- [x] `./mvnw clean verify -pl runtime/engine -am` — full engine unit + integration test suite passes (`EngineIT`, `EngineContextIT`, `RouterFactoryTest`, etc.)
- [x] `./mvnw clean verify -pl runtime/binding-http -am` — full binding-http test suite passes (397 tests), including both `rfc7230` and `rfc7540` `RedirectIT` classes exercising the changed `onRequestRedirect` path end-to-end through a real engine
- [x] `./mvnw checkstyle:check` / `./mvnw license:check` — clean

---

### DIFF
--- a/incubator/command-dump/src/main/resources/io/aklivity/zilla/runtime/command/dump/internal/airline/zilla.lua
+++ b/incubator/command-dump/src/main/resources/io/aklivity/zilla/runtime/command/dump/internal/airline/zilla.lua
@@ -126,6 +126,7 @@ add_field("authorization", ProtoField.uint64("zilla.authorization", "Authorizati
 
 -- begin frame
 add_field("affinity", ProtoField.uint64("zilla.affinity", "Affinity", base.HEX))
+add_field("redirected_id", ProtoField.uint64("zilla.redirected_id", "Redirected ID", base.HEX))
 
 -- data frame
 add_field("flags", ProtoField.uint8("zilla.flags", "Flags", base.HEX))
@@ -277,7 +278,9 @@ end
 function handle_begin_frame(buffer, offset, subtree, pinfo, info)
     local slice_affinity = buffer(offset, 8)
     subtree:add_le(fields.affinity, slice_affinity)
-    handle_extension(buffer, subtree, pinfo, info, offset + 8, BEGIN_ID)
+    local slice_redirected_id = buffer(offset + 8, 8)
+    subtree:add_le(fields.redirected_id, slice_redirected_id)
+    handle_extension(buffer, subtree, pinfo, info, offset + 16, BEGIN_ID)
 end
 
 function handle_redirect_frame(buffer, offset, subtree, pinfo, info)
--- a/incubator/command-dump/src/test/resources/io/aklivity/zilla/runtime/command/dump/internal/AmqpLinkNetworkIT_shouldExchangeAttachAsReceiver.txt
+++ b/incubator/command-dump/src/test/resources/io/aklivity/zilla/runtime/command/dump/internal/AmqpLinkNetworkIT_shouldExchangeAttachAsReceiver.txt
@@ -31,14 +31,14 @@ Zilla Frame
 Frame 2: 210 bytes on wire (1680 bits), 210 bytes captured (1680 bits)
 Ethernet II, Src: Send_00 (20:53:45:4e:44:00), Dst: Receive_00 (20:52:45:43:56:00)
 Internet Protocol Version 6, Src: fe80::3f3f:0:0:3, Dst: fe80::3f3f:0:0:2
-Transmission Control Protocol, Src Port: 7114, Dst Port: 37114, Seq: 1, Ack: 129, Len: 136
+Transmission Control Protocol, Src Port: 7114, Dst Port: 37114, Seq: 1, Ack: 137, Len: 136
 Zilla Frame
     Frame Type ID: 0x00000001
     Frame Type: BEGIN
     Protocol Type ID: 0x112dc182
     Protocol Type: amqp
     Worker: 0
-    Offset: 0x00000058
+    Offset: 0x00000060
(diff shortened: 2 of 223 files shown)


jfallows:
`Build (25)` has now failed identically three times in a row (two separate pushes to this branch, plus one explicit re-run of the failed jobs) — always the same failure: `binding-mcp`'s `McpProxyLifecycleIT.shouldLifecycleServerWriteAbort` times out waiting for a `TOOLS_LIST_COMPLETE` notify but observes `LIFECYCLE_OPEN` instead.

This PR's diff touches only `runtime/engine`, `runtime/binding-http`, and `specs/engine.spec/core.idl` — `binding-mcp` is untouched, and its own spec module (`zilla::specs::binding-mcp.spec`) compiles and passes every run; only this one runtime IT's timing-sensitive K3PO wait fails. Given the reproducibility across independent runs and the complete lack of overlap with this PR's changes, this looks like a pre-existing timing issue in `McpProxyLifecycleIT` itself rather than a regression introduced here. Not widening this PR to fix it since it's out of scope for the change — flagging for visibility instead.

---

---

jfallows (commented):
LGTM

jfallows:
Follow-up on the `McpProxyLifecycleIT` failure — wanted to verify "unrelated" rather than just assert it, since `binding-mcp` actually does call the exact two-arg `supplyInitialId(routedId, affinity)` this PR changes (via its own `REDIRECT`-driven session-affinity mechanism in `McpClientFactory`/`McpProxyItemFactory`/`McpProxyListFactory`/`McpServerFactory`).

Traced it through: `McpConfiguration.newSessionId`/`defaultSessionIdAffinity` deliberately mirror `EngineWorker.affinity()`'s bit packing (top byte = node id, low 3 bytes = worker index) when embedding/extracting affinity from a session id. The failing script's hardcoded fixture session id (`5ca1ab1e-c0de-4a11-b007-000100000000`) decodes, via that same last-8-hex-chars extraction, to affinity `0x00000000` — node 0, worker 0.

Working the math for affinity=0 against the default tuning mask (`Tuning.affinity()`'s default is all-bits-set across `config.workers()` workers): the old `floorMod(0, cardinality)` and the new direct-index-with-bounds-check both resolve to worker 0, for any worker count. No behavioral divergence for the value this test actually exercises.

Combined with `binding-mcp` being otherwise untouched by this diff and the identical failure reproducing 3x independently, I'm confident this is a pre-existing timing issue in `McpProxyLifecycleIT` itself, not a regression from this change.

---

---

jfallows:
Two more rounds of CI failures resolved on this branch, both genuine regressions from the `Begin.redirectedId` field addition rather than pre-existing flakiness:

**`binding-mcp` (`McpProxyLifecycleIT`)** — fixed by a separate push (`b006ae9c`). My earlier framing of this as "pre-existing timing issue" / "unrelated" was wrong. The real cause: `EngineRule.external()`'s test bindings use a single-bit tuning mask (cardinality=1) as a worker-63 sentinel. The old `floorMod(affinityId, 1)` always resolved to 0 regardless of `affinityId`'s actual value, incidentally routing to that external test peer; the new direct-index extraction correctly honors the affinity's embedded worker index instead, which changed routing behavior for that specific fixture. Fixed by aligning the test's session-id/affinity fixtures.

**`incubator/command-dump`** (`5f85c875`, `7ec1238a`) — two issues in the hand-maintained Wireshark dissector (`zilla.lua`), which parses `Begin` frame bytes by fixed offset rather than being derived from the flyweight schema:
1. `handle_begin_frame` read `extension` starting 8 bytes too early, since it didn't account for the new `redirectedId` field between `affinity` and `extension`.
2. Once (1) was fixed, the 25 expected-output fixtures' recorded "bytes on wire"/"bytes captured" frame sizes and TCP segment lengths were stale — `Begin` frames are now legitimately 8 bytes larger on the wire.

CI should be green on `5f85c875` now. Will keep watching.

---

PR #2414, opened 2026-08-21
fix(common-protobuf): unwrap well-known wrapper types in JSON codec

### LINKED ISSUE #2411: common-protobuf JSON codec doesn't unwrap well-known wrapper types
**Describe the bug**

The protobuf JSON mapping specification requires that every one of the nine well-known wrapper types — `google.protobuf.StringValue`, `Int32Value`, `Int64Value`, `UInt32Value`, `UInt64Value`, `BoolValue`, `FloatValue`, `DoubleValue`, `BytesValue` — serializes to and parses from a bare JSON scalar rather than a JSON object. A field of type `StringValue` holding `"hello"` should appear in JSON as `"hello"`, not `{"value": "hello"}`; a field of type `Int32Value` holding `42` should appear as `42`, not `{"value": 42}`; and so on identically for all nine.

`common-protobuf`'s JSON codec has no awareness of this concept at all — not for one of the nine types, not for any of them. Both `ProtobufJsonGeneratorImpl` and `ProtobufJsonParserImpl` (`runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/json/internal/`) dispatch purely on wire type, with no path that ever inspects a message field's `typeName()`/message identity:

- `ProtobufJsonGeneratorImpl.java` (`start(int number, boolean group)`, ~line 741): a field whose wire type is `MESSAGE`/`GROUP` unconditionally writes the field's own key then opens a nested JSON object (`writeStartObjectChecked()` + `pushScope(field.message(), ...)`), regardless of which message type it is.
- The same file's `emitDefaults` (~line 851) treats every non-repeated `MESSAGE`/`GROUP` field generically when omitting defaults.
- `ProtobufJsonParserImpl.java` is symmetric — its own composite-value checks (e.g. `composite()`) apply the same generic nested-object handling on decode.

A fix should treat this as one general concept — "is this message field's type one of the nine well-known wrapper types, and if so substitute/read a bare scalar instead of an object" — rather than as a fix scoped to any single wrapper type. Handling only `StringValue` (or any other individual type) and leaving the other eight with the same non-compliant behavior would leave the underlying gap open.

**Suggested fix shape**

This does not require resolving `import "google/protobuf/wrappers.proto"` or building a real `ProtobufMessage` for any of the nine types — their wire shape is fixed by the protobuf spec itself (always exactly one field, field number 1, of a fixed corresponding scalar type: `string` for `StringValue`, `int32` for `Int32Value`, `bytes` for `BytesValue`, etc.), so recognizing them is a pure name-to-scalar-kind lookup, not a message-resolution problem:

- Recognize a message-typed field as one of the nine by its fully-qualified `field.typeName()` (e.g. `"google.protobuf.StringValue"`), independent of whether that type was ever actually declared/imported in the compiled schema.
- On encode: when writing such a field, write the field's own JSON key then the bare scalar value directly (no nested object), reading the corresponding wire-level scalar as if it were field 1 of the (unresolved) message.
- On decode: when parsing such a field's bare scalar JSON value, write it out on the wire as the fixed one-field message shape (tag for field 1 + the scalar's own encoding) rather than expecting/requiring a nested JSON object.
- This recognition (name → wrapped scalar kind) is generically useful protobuf-format knowledge, not something that needs to live only inside the JSON codec — worth exposing as a small, reusable lookup in `common-protobuf` itself (e.g. on `ProtobufField`/`ProtobufType`) rather than duplicating the nine-name table wherever it's needed.

**To Reproduce**

1. Configure a `model: protobuf` binding with `view: json` against a schema containing a wrapper-typed field, e.g.:
   ```proto
   syntax = "proto3";
   import "google/protobuf/wrappers.proto";
   message Event
   {
       google.protobuf.StringValue note = 1;
       google.protobuf.Int32Value count = 2;
   }
   ```
2. Encode a message with `note` set to `"hello"` and `count` set to `42`, and decode it through the `view: json` pipeline.

**Expected behavior**

The JSON view should match the protobuf JSON mapping spec for every wrapper-typed field: `{"note": "hello", "count": 42}`.

**Actual behavior**

The JSON view produces `{"note": {"value": "hello"}, "count": {"value": 42}}` — the wrapper message's inner field name (`value`) leaks into the JSON shape instead of being unwrapped, on both generate and parse, for every one of the nine wrapper types alike.

**Additional context**

This is a `common-protobuf` JSON-codec compliance gap against the protobuf JSON mapping spec, not specific to any one consumer of the codec or any one of the nine wrapper types — any `view: json` protobuf model with a wrapper-typed field of any of the nine types is affected identically, independent of any other configuration on that field.

jfallows:
## Description

The protobuf JSON mapping spec requires the nine `google.protobuf` well-known wrapper types (`StringValue`, `Int32Value`, `Int64Value`, `UInt32Value`, `UInt64Value`, `BoolValue`, `FloatValue`, `DoubleValue`, `BytesValue`) to render as a bare JSON scalar rather than a `{"value": ...}` object. `ProtobufJsonGeneratorImpl` and `ProtobufJsonParserImpl` previously treated every `MESSAGE`/`GROUP` field generically, so a wrapper-typed field leaked its inner `value` field into the JSON shape on both encode and decode.

This adds a shared `ProtobufWellKnownTypes` lookup (their wire shape is spec-fixed: one field, number 1, of a fixed scalar type) used by both JSON codec classes, so a field's `typeName()` alone decides whether to substitute a bare scalar for the nested object — handled as one general concept across all nine, not scoped to any single type. `ProtobufSchema.Builder` also auto-registers each of the nine as a synthetic single-field message when a schema doesn't already declare one under that name, so a field referencing `google.protobuf.StringValue` etc. resolves without any import-resolution machinery, matching the issue's suggested fix shape.

Handles the concept generally: a plain wrapper-typed field, a repeated wrapper-typed field (bare array), and a map value of wrapper type (bare scalar value) all unwrap consistently in both directions, including an explicitly-present-but-default wrapper value rendering its scalar zero value instead of an empty object.

Fixes #2411

## Test plan

- [x] New `ProtobufJsonWrapperTest` covers encode and decode for all nine wrapper types, a repeated wrapper field, a wrapper-valued map, explicit-default rendering, absent-field omission, and null handling
- [x] New `ProtobufWellKnownTypesTest` covers schema auto-registration, explicit-declaration precedence, and a wire-only round trip proving `field.message()` resolves without any explicit declaration
- [x] Full `common-protobuf` module test suite passes (226/226), including checkstyle and license header checks

---
_Generated by [Claude Code](https://claude.ai/code/session_019zS8zhb9myc6iSRxZvMHcp)_

### DIFF
--- a/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufSchema.java
+++ b/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufSchema.java
@@ -327,7 +327,12 @@ public final class ProtobufSchema
 
         public ProtobufSchema build()
         {
-            Map<String, ProtobufMessage> resolvedMessages = Map.copyOf(messages);
+            Map<String, ProtobufMessage> withWellKnown = new LinkedHashMap<>(messages);
+            for (String typeName : ProtobufWellKnownTypes.names())
+            {
+                withWellKnown.putIfAbsent(typeName, ProtobufWellKnownTypes.wrapperMessage(typeName));
+            }
+            Map<String, ProtobufMessage> resolvedMessages = Map.copyOf(withWellKnown);
             Map<String, ProtobufEnum> resolvedEnums = Map.copyOf(enums);
             Map<String, ProtobufService> resolvedServices = Map.copyOf(services);
             for (ProtobufMessage message : toRelink)
--- a/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufWellKnownTypes.java
+++ b/runtime/common-protobuf/src/main/java/io/aklivity/zilla/runtime/common/protobuf/ProtobufWellKnownTypes.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2021-2026 Aklivity Inc
+ *
+ * Licensed under the Aklivity Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ *   https://www.aklivity.io/aklivity-community-license/
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package io.aklivity.zilla.runtime.common.protobuf;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * The nine {@code google.protobuf} well-known wrapper message types — {@code StringValue}, {@code BytesValue},
+ * {@code BoolValue}, {@code Int32Value}, {@code UInt32Value}, {@code Int64Value}, {@code UInt64Value},
+ * {@code FloatValue}, {@code DoubleValue}. Each has a wire shape fixed by the protobuf spec itself — always
+ * exactly one field, number 1 named {@code value}, of a fixed corresponding scalar type — so recognizing one
+ * by its fully-qualified name and resolving its shape is a pure lookup, never a schema-import concern; a
+ * schema referencing one of these types by name need not declare or import it.
+ */
+public final class ProtobufWellKnownTypes
+{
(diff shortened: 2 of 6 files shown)


jfallows (commented):
LGTM

PR #2417, opened 2026-08-21
feat(vault): drop keyName param from SecretKeyManager/VaultHandler unwrap

### LINKED ISSUE #2416: Make wrapped key material self-sufficient for unwrap — drop the explicit key-name parameter from SecretKeyManager/VaultHandler.unwrap

As a Zilla operator, I want a wrapped key's bytes to be sufficient on their own to unwrap it — no separate name argument required — so that wrapped key material is a single, portable artifact rather than a value that only decrypts correctly when paired with an external record of which key wrapped it.

**Acceptance criteria**

## 1. `unwrap` drops its `keyName` parameter; `wrap` keeps its

`wrap(keyName, bytes, ...)` is unchanged — wrapping is inherently config-driven, since something has to say which key to use. `unwrap` becomes:

```java
// SecretKeyManager
int unwrap(
    DirectBufferEx bytes,
    int index,
    int length,
    MutableDirectBufferEx dst,
    int dstIndex);

// VaultHandler
default void unwrap(
    long traceId,
    DirectBufferEx bytes,
    int index,
    int length,
    BytesConsumer next)
{
    next.accept(null, 0, 0);
}
```

## 2. The contract, not a mechanism, belongs at the interface

`wrap`'s Javadoc states the invariant: the bytes it returns must be sufficient for a matching `unwrap` call with no separate key-name argument. It says nothing about *how* — that is each implementation's own decision, since backends genuinely differ in what they already provide natively:

- A backend whose own wrapped-key format already identifies which key produced it — its remote service resolves that internally on decrypt, with no identifier required from the caller — can return that format unmodified. No extra framing needed.
- A backend whose native format carries no such identity has to embed one itself before returning from `wrap`, and recover it internally at the start of `unwrap`, before dispatching to whichever of its own managed keys that identifies.

No shared encoding is mandated across implementations. Forcing one would add a redundant layer onto a backend that already provides this for free, and would be no simpler for a backend that doesn't.

## 3. `vault-filesystem` (the one existing implementation) needs updating to match

Currently wraps and unwraps by explicit name with no identity embedded in the wrapped bytes. Needs its own embedding scheme (e.g. a length-prefixed name ahead of the wrapped bytes) added to `wrap`'s output and parsed back out at the start of `unwrap` — entirely internal to that module, no change to the SPI beyond the signature itself.

## 4. Coverage

- Round trip: `wrap` then `unwrap` recovers the original bytes with no name supplied to `unwrap`
- A vault instance managing more than one named key unwraps each correctly, dispatching internally from whatever it embedded, not from an external parameter
- Wrapped bytes produced by one `wrap` call, handed to `unwrap` with no other context, succeed — the artifact is self-sufficient
- Corrupted or foreign wrapped bytes (no recoverable identity, or an unrecognized one) fail cleanly rather than silently misresolving to the wrong key

**Additional context**

Amends the `SecretKeyManager`/`VaultHandler` `wrap`/`unwrap` shape from aklivity/zilla#1678, and `vault-filesystem`'s implementation from aklivity/zilla#1679 — both already merged with the previous (name-on-both) shape. No other caller of these methods exists yet, so this is a clean signature change rather than a compatibility concern.

jfallows:
## Description

Makes wrapped key material self-sufficient for unwrap, then follows the resulting simplification through to its natural conclusion: one synchronous wrap/unwrap shape on the vault SPI, with real IT coverage of it.

### 1. `unwrap` drops its key-name parameter (fixes #2416)

`SecretKeyManager.unwrap` and `VaultHandler.unwrap` no longer take a key-name argument, since the wrapped bytes now carry enough information on their own to determine which key to unwrap under. `wrap` on both is unchanged — wrapping is inherently config-driven, since something has to say which key to use.

- Javadoc states the invariant — the bytes `wrap` produces must be sufficient on their own for a matching `unwrap` call — without mandating how each implementation embeds that identity.
- `vault-filesystem` (the one existing implementation): `wrap` now writes a length-prefixed copy of the key's name ahead of the existing version + iv + ciphertext; `unwrap` parses that name back out first, resolves the matching entry, and only then proceeds — entirely internal to the module, no change to the SPI beyond the signature itself. A `SecretKeyManager`'s permitted-key set is checked against the *recovered* name, so bytes wrapped under a key outside that set are rejected rather than silently misresolved.
- The engine's `TestVaultHandler`/`TestSecretKeyManager` test fixtures (used by other spec modules) got the same treatment.
- Coverage added: round trip with no name supplied to unwrap, a vault managing multiple named keys dispatching purely from the embedded identity, a separate handler instance unwrapping another instance's output with no other context, and clean failures on a corrupted embedded version, a foreign/unrecognized embedded name, and random garbage bytes.

### 2. Removes the async `wrap`/`unwrap` pair from `VaultHandler`

`SecretKeyManagerFactory` already resolves keys once and wraps/unwraps synchronously, the same way `initKeys`/`initSigners`/`initTrust` resolve TLS material once regardless of whether it's local or fetched from a remote store. The async `wrap`/`unwrap` pair on `VaultHandler` (added for a hypothetical remote-KMS case) duplicated that shape for no gain: nothing in this repo implements it, and any vault needing to fetch or rotate key material off-thread can already do so before returning a `SecretKeyManager`, exactly like TLS material is resolved today.

- Drops `VaultHandler.wrap`/`unwrap` and the `BytesConsumer` callback.
- `FileSystemVaultHandler` keeps only the synchronous `FileSystemSecretKeyManager` path; since `wrapCore`/`unwrapCore` now have a single caller, their bodies are inlined directly into `FileSystemSecretKeyManager.wrap`/`unwrap`.
- Removes `VaultHandlerTest.java` (it only ever exercised the now-removed defaults) and updates `TestVaultHandler`, `TestVaultHandlerTest`, and `FileSystemVaultTest` to exercise the same coverage through `SecretKeyManager` instead.

### 3. Wires secret-key wrap/unwrap into the `test` binding fixture

`initSecretKeys`/`SecretKeyManager` had no path through `TestBindingFactory`, so nothing could exercise a vault's wrap/unwrap dispatch via a k3po IT against a live engine — only direct unit tests against a vault implementation could verify it, the same gap that left the pre-existing key/signer/trust vault assertions with zero script/IT coverage of their own.

- Extends the existing `assertions.vault` mechanism with `secretKey`/`wrap`: `TestBindingFactory` now calls `initSecretKeys`, wraps and unwraps the given plaintext through the returned `SecretKeyManager`, and resets the stream if the round trip doesn't recover it.
- Adds two IT scenarios in `FileSystemVaultIT` driving this through a live engine against the real `FileSystemVaultHandler`: a successful wrap/unwrap round trip, and a rejection when the referenced secret key isn't configured.
- The reject scenario's script pair (`connect aborted` / `rejected`) lives in `engine.spec` as `vault.assert.reject`, reusable by any future vault module wiring the same assertion.

`zilla-plus` isn't affected by any of the above — it pins a released `zilla` version, not this in-flight SPI change.

Fixes #2416

---
_Generated by [Claude Code](https://claude.ai/code/session_01GqBSUSbxDZRri31recmy2k)_


jfallows (commented):
LGTM

jfallows (commented):
LGTM

PR #2419, opened 2026-08-22
feat(common-feature): add @Internal, a permanent non-overridable visibility gate

jfallows:
## Description

`common-feature`'s visibility gate currently has one maturity axis: `@Incubating`, which hides a feature from a real release's config schema and runtime SPI loading, but always leaves a supported escape hatch — `zilla.incubator.enabled` (or the `ZILLA_INCUBATOR_ENABLED` env var it's mapped from) lets it be turned back on even in a tagged release.

There's no way today to express the opposite maturity: a feature that must never be reachable in a real release, under any configuration, with no supported override at all. `@Internal` adds that second axis.

`@Internal` is filtered through the exact same `FeatureFilter`/`Factory.instantiate()` machinery `@Incubating` already uses (so every existing ServiceLoader-based lookup — `BindingFactory`, `GuardFactory`, `VaultFactory`, `CatalogFactory`, `ExporterFactory`, `StoreFactory`, `RouterFactory`, `MetricGroupFactory`, `EventFormatter(Factory)`, and `EngineInfo`'s config-schema aggregation — picks it up automatically, with zero changes needed at any call site). The one deliberate difference: it has no system-property override branch. Its only "on" switch is the same unnamed-module/`develop-SNAPSHOT` fallback `@Incubating` already falls back to, which is why it stays free during ordinary Maven-run unit/integration tests (Surefire and Failsafe both execute before `moditect-maven-plugin` ever adds a module descriptor to the built jar) while remaining permanently unreachable in a fully packaged, module-path release build.

While extending `FeatureFilter.featureEnabled(Class<?>)` to check both annotations, also removed the pre-existing exact duplication between its `Object` and `Class<?>` overloads by having the former delegate to the latter.

## Test plan

- `FeatureFilterTest` — two new tests mirroring the existing `@Incubating` coverage exactly (`shouldReportIsInternalEnabledConsistentlyWithFeatureEnabled`, `shouldFilterInternalConsistentlyWithIsInternalEnabled`), plus a private `@Internal`-annotated fixture class.
- `FactoryTest` (`config/engine.conf`) — a new `shouldFilterInternalIdenticallyToMapOverload` test mirroring the existing `@Incubating` one, covering the actual `Factory.instantiate()` integration point.
- Verified `./mvnw test -pl runtime/common-feature` and `./mvnw verify -pl config/engine.conf` both pass (checkstyle, license, JaCoCo coverage all green).
- Verified the full `runtime/engine` unit test suite (228 tests) still passes unchanged, confirming the stricter `featureEnabled` check doesn't affect any of its other callers.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FEncF8TdASSjY9Q113NMkv)_

### DIFF
--- a/config/engine.conf/src/test/java/io/aklivity/zilla/config/engine/factory/FactoryTest.java
+++ b/config/engine.conf/src/test/java/io/aklivity/zilla/config/engine/factory/FactoryTest.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import org.junit.Test;
 
 import io.aklivity.zilla.runtime.common.feature.Incubating;
+import io.aklivity.zilla.runtime.common.feature.Internal;
 
 public class FactoryTest
 {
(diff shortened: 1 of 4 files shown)


jfallows (commented):
LGTM

PR #2420, opened 2026-08-22
feat(engine): add RouteableContext#supplyStore

jfallows:
## Summary
- A router can already discover the ids of bindings it injects via `attachComposite` (by scanning the returned `NamespaceConfig`) and resolve labels via `supplyLabelId`/`supplyLabel`, but has no way to obtain a `StoreHandler` for a store it injects the same way.
- Adds `RouteableContext#supplyStore(long storeId)`, mirroring `EngineContext`'s existing capability, so a router-contributed composite namespace's store is as usable to the router as its bindings and labels already are.

## Test plan
- [x] `./mvnw -pl runtime/engine clean install -Dmaven.test.skip=true` — main sources compile cleanly.
- Note: `runtime/engine`'s test compilation currently fails on `develop` HEAD independent of this change (`TestBindingFactory.java` references `VaultAssertion.secretKey`/`.wrap`, which appear to have been removed from `config/engine.conf`'s `TestBindingOptionsConfig` in #2417). Verified via `git stash` that this reproduces identically without this PR's changes.

---

### DIFF
--- a/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineRouteable.java
+++ b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/internal/registry/EngineRouteable.java
@@ -16,11 +16,13 @@
 package io.aklivity.zilla.runtime.engine.internal.registry;
 
 import java.util.function.Consumer;
+import java.util.function.LongFunction;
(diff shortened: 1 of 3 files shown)


jfallows (commented):
LGTM

PR #2421, opened 2026-08-23
feat(engine): embed cluster node id into worker-scoped ids

jfallows:
## Description

`EngineWorker` mints `initialId`, `promiseId`, `traceId`, `budgetId`, `authorizedId`, and its own `affinityId` counter from one shared `initial`/`mask` pair that previously encoded only this node's local worker index (top byte, bits 56-63). Every such id was therefore unique only *within one engine instance* — a `traceId` minted on one node's worker 2 can numerically collide with one minted on a different node's worker 2.

That matters once a `RouterContext` implementation fans work out across physical nodes (the SPI already supports this — `RouteableContext#attachComposite`/`supplyStore`, `RouterContext#resolveRoutedId`) rather than only across local workers: any consumer of `traceId`/`budgetId`/etc. for cross-node trace or log correlation (`zilla dump` output included, once a stream has been forwarded to a different node and back) can't disambiguate two ids that happen to share the same local worker index on different nodes.

This folds `EngineConfiguration`'s own `node.id` byte into the same `initial` value, immediately below the existing worker-index byte (bits 48-55, previously always zero) — forming one contiguous 16-bit (node, worker) pair. Every id derived from `initial`/`mask` becomes cluster-wide unique as a result, with:
- no change to the existing per-worker counter width (still 31 bits via the unchanged `RESERVED_SIZE`)
- no change to `indexOfId`'s existing local-dispatch semantics (guard verifier/identifier/attributor routing, `BudgetId#ownerIndex` resolution) — both still read only the unmoved worker-index byte
- a new `nodeOfId` alongside the existing `indexOfId`, for symmetry

`node.id` defaults to `0`, so this is a no-op for every single-node deployment — verified via the full `runtime/engine` unit + integration test suite (438 tests, 0 failures) and `checkstyle:check` (0 violations), both green before and after this change.

## Test plan

- [x] `./mvnw checkstyle:check -pl runtime/engine` — 0 violations
- [x] `./mvnw clean install -pl runtime/engine -am -DskipTests=false` — full unit + IT suite green (228 unit tests, 210 integration tests, 0 failures/errors)

---

### DIFF
--- a/incubator/command-dump/src/main/resources/io/aklivity/zilla/runtime/command/dump/internal/airline/zilla.lua
+++ b/incubator/command-dump/src/main/resources/io/aklivity/zilla/runtime/command/dump/internal/airline/zilla.lua
@@ -122,6 +122,8 @@ add_field("acknowledge", ProtoField.int64("zilla.acknowledge", "Acknowledge", ba
 add_field("maximum", ProtoField.int32("zilla.maximum", "Maximum", base.DEC))
 add_field("timestamp", ProtoField.uint64("zilla.timestamp", "Timestamp", base.HEX))
 add_field("trace_id", ProtoField.uint64("zilla.trace_id", "Trace ID", base.HEX))
+add_field("trace_id_node", ProtoField.uint8("zilla.trace_id.node", "Node", base.DEC))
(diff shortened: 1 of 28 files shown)


jfallows (commented):
LGTM

jfallows (commented):
LGTM

PR #2422, opened 2026-08-23
feat(engine): defer RouterContext.attach to run after registry bootstrap

jfallows:
## Description

`RouterContext.attach(RouterConfig)` ran as the first statement of `EngineWorker.doInit()` — before this worker's own `EngineRegistry` is constructed, and before `EngineManager` has processed any namespace. Its contract already lets an implementation attach a synthesized composite namespace via `RouteableContext.attachComposite()` as part of that same call (`RouteableContext`'s own javadoc: "Attaches a synthesized `NamespaceConfig` to the engine alongside operator-authored namespaces"), which reads and mutates per-worker registry state that doesn't exist yet at that point in engine bootstrap.

Every existing `RouterContext` implementation (`EngineRouterContext`, the test router, and a real composite-synthesizing router built against this SPI) already builds its stream factory once, at construction time in `Router#supply(RouteableContext)`, and `attach()` just returns that value unchanged — `RouterConfig` itself goes unused in every implementation. So the stream factory returned by `attach()` never actually depended on `attach()` running early during `doInit()`; only bindings' own construction-time capture of `EngineContext#streamFactory()` did (every binding factory does `this.streamFactory = context.streamFactory();` in its own constructor, called from `doInit()`).

This splits the two concerns:
- `RouterContext#streamFactory()` (new) — returns the already-constructed stream factory. `EngineWorker` sources this immediately after `Router#supply(RouteableContext)` returns, in its own constructor — independent of when `attach()` runs.
- `RouterContext#attach(RouterConfig)` — now `void`. `EngineWorker.doInit()` no longer calls it. Instead, `Engine#start()` calls it once per worker, dispatched onto that worker's own thread (mirroring the existing `EngineWorker#attach(NamespaceConfig)` pattern), only after `EngineManager#start()` has run the engine's bootstrap config through to completion — so any registry-dependent setup a router's `attach()` performs (like attaching a composite namespace) now runs with the registry and `EngineManager`'s live configuration already in place.

No behavior change for any existing router: `EngineRouterContext#attach()` becomes an empty no-op, matching what it always effectively did.

## Test plan

- [x] `./mvnw checkstyle:check -pl runtime/engine` — 0 violations
- [x] `./mvnw clean verify -pl runtime/engine` — 228 unit tests + 210 integration tests, 0 failures/errors

---

### DIFF
--- a/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
+++ b/runtime/engine/src/main/java/io/aklivity/zilla/runtime/engine/Engine.java
@@ -44,6 +44,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.ServiceLoader;
 import java.util.ServiceLoader.Provider;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
@@ -372,6 +373,12 @@ public final class Engine implements Collector, AutoCloseable
         if (!readonly)
         {
             manager.start();
+
+            workers.stream()
+                .map(w -> w.attachRouter(routerConfig))
+                .reduce(CompletableFuture::allOf)
+                .ifPresent(CompletableFuture::join);
+
             Ready.markReady(config.directory());
         }
     }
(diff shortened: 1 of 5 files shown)


jfallows (commented):
LGTM

PR #2425, opened 2026-08-24
fix(binding-kafka): use each producer stream's own authorization for cache encode

### LINKED ISSUE #2424: kafka cache storage: honor per-message authorization on the produce path, not the first producer's

As a Zilla operator, I want the kafka cache storage layer's produce path to encode each message with that message's own producer's `authorization` (aklivity/zilla#2310), so that a model transform whose behavior depends on that parameter sees the correct, per-message value rather than whichever producer happened to open the shared connection state first.

**The gap, concretely**

`KafkaCacheClientProduceFactory` encodes each produced message through a per-topic-partition object (a "fan") shared across every producer writing to that partition. The fan captures and stores the authorization of whichever producer first opens it, and every subsequent produce call — from any producer, including ones that open a stream on that same topic-partition later — encodes using that captured value, not its own. A later producer's own authorization is tracked on its own per-producer state, but the encode call reads the fan's field instead. The fan also persists for a cleanup delay after its last member closes, so a stale captured authorization can outlive the producer that supplied it.

This is inert today for a transform that never varies its behavior by authorization on encode. It is wrong for one that does — a real, if currently unexercised, violation of #2310's own contract for that parameter (a per-message value, not a captured global for the connection group).

**Acceptance criteria**

- Each encode call uses the authorization of the producer stream actually producing that message, not whatever the fan captured when it was created
- Encoder pipeline state lives per-producer-stream, not shared on the fan, so state carried across separate producers doesn't leak between them the same way authorization does
- A fan reused after its last member's cleanup delay must not carry forward a stale authorization from a previous producer
- Unit and k3po coverage: two producers on the same topic-partition, opened in each order, each encoding with their own authorization on a transform whose behavior depends on it; a fan reused after cleanup delay uses the new producer's own authorization, not the prior one's

**Additional context**

Related to aklivity/zilla#2423 (the analogous per-message-authorization gap on the fetch path) — found while examining that issue's implementation, but a distinct code path (`KafkaCacheClientProduceFactory`, not the fetch side) with a different failure mode: not "no real caller yet," but "the wrong real caller's value used." Scoped to the shared kafka cache storage implementation, the same layer #1688/#1689/#1690/#2423 already scope themselves to.

jfallows:
## Description

`KafkaCacheClientProduceFan` captured the `authorization` of whichever producer stream first created the per-topic-partition fan, then reused that single value — along with a shared `transformKey`/`transformValue` encoder pair — for every subsequent producer writing to the same partition. `KafkaCacheClientProduceStream` already tracked its own `authorization`, but the encode call sites (`onClientInitialData` ×2, `onClientInitialFlush`) never used it.

This moves `transformKey`/`transformValue` from the fan onto the stream and threads `stream.authorization` through `writeProduceEntryStart`/`writeProduceEntryContinue`, so each message is encoded using the authorization of the producer stream that actually produced it, not the first producer to touch the partition.

To make this observable in an oss-only test, the engine's generic `TestModel` gains an opt-in `transformAuthorizations: [N, M]` config list: an ordered list of the authorization values expected to reach the model, in message-arrival order. Each completed value consumes the next entry from this list (tracked on the shared `ModelHandler`, not the pipeline instance that observed it — a shared/hoisted pipeline handling multiple messages must still be checked against each message's own expected value) and is rejected outright on mismatch, surfacing as an ordinary `REJECTED` → stream `RESET`, the same signal already used by the existing `message.value.rejected`/`message.values.rejected` scenarios. A new k3po scenario (`message.values.authorization.distinct`) drives two producers with distinct authorization values through a live engine and asserts each message reaches the model with its own value, not the other's.

Verified test-first: the new `CacheProduceIT` test fails with a clean, fast rejection (not a hang/timeout) against the pre-fix code, and passes against the fix. Full `ProduceIT` and `CacheProduceIT` suites, `runtime/engine` and `config/engine.conf` unit tests, and checkstyle/license checks all pass with no regressions.

The new scenario's peer-to-peer counterpart is intentionally omitted, with the reasoning documented inline in `ProduceIT.java`: a pre-existing k3po-harness limitation (unrelated to this change, and already present for the never-covered `message.values.parallel` scenario) prevents a third notify/await-gated connect from being peer-verified in this harness.

Fixes #2424

---

### DIFF
--- a/config/engine.conf/src/test/java/io/aklivity/zilla/config/engine/test/internal/model/config/TestModelConfig.java
+++ b/config/engine.conf/src/test/java/io/aklivity/zilla/config/engine/test/internal/model/config/TestModelConfig.java
@@ -27,7 +27,7 @@ public class TestModelConfig extends ModelConfig
     public final boolean read;
     public final int transformLength;
     public final List<String> fields;
-    public final boolean transformAuthorization;
+    public final List<Long> transformAuthorizations;
 
     public TestModelConfig(
         int length,
@@ -64,7 +64,7 @@ public class TestModelConfig extends ModelConfig
         List<String> fields,
         ValidateConfig validate)
     {
-        this(length, cataloged, read, transformLength, fields, validate, false);
+        this(length, cataloged, read, transformLength, fields, validate, null);
     }
 
     public TestModelConfig(
@@ -74,14 +74,14 @@ public class TestModelConfig extends ModelConfig
         int transformLength,
         List<String> fields,
         ValidateConfig validate,
-        boolean transformAuthorization)
+        List<Long> transformAuthorizations)
     {
         super("test", cataloged, validate);
         this.length = length;
         this.read = read;
         this.transformLength = transformLength;
         this.fields = fields;
-        this.transformAuthorization = transformAuthorization;
+        this.transformAuthorizations = transformAuthorizations;
     }
 
     public static <T> TestModelConfigBuilder<T> builder(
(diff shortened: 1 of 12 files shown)


jfallows (commented):
LGTM

PR #2426, opened 2026-08-25
fix(binding-mcp): defer mcp(client) lifecycle connect until guard decides

jfallows:
## Description

`McpLifecycleStream` — the one persistent south connection an `mcp(client)` binding multiplexes every subsequent method call over — inherited the base `McpStream#proceedWithRequest`, which resolves the configured guard synchronously and connects regardless of the outcome. A guard whose authorization decision is asynchronous, or needs an interactive step (`GuardHandler#NEEDS_PREAUTHORIZE`), never actually got the chance to gate or challenge this connection: `proceedWithRequest` returned `true` before the guard had decided anything, so the connection was established unauthorized/unchallenged rather than deferred.

`McpLifecycleStream` now overrides `proceedWithRequest` to defer the actual connect until `guard.reauthorize`'s completion callback fires, mirroring the pattern already used for the per-request `McpRequestStream`. A `NEEDS_PREAUTHORIZE` decision is surfaced as an `elicitCreate` challenge — the same challenge already sent for a backend-initiated elicitation — so the existing north-facing challenge relay carries it through unchanged; the login callback answer resumes the deferred connect through the existing `onReauthorized` handling. Resuming also grants the app-level window, matching `McpRequestStream#onAcquireCompleted`, since omitting it left the connection timing out and reconnecting in a loop.

Validated via `zilla dump` frame-level tracing against a lifecycle-style `mcp(client)` binding guarded by a DCR + authorization-code guard: the connection now correctly defers, emits the `elicitCreate` challenge, and resumes cleanly once the login callback answers it, with no reconnect storm.

`./mvnw -pl runtime/binding-mcp -am test` passes (178 tests, 0 failures/errors).

Fixes # (issue)

---

### DIFF
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/stream/McpClientFactory.java
@@ -2343,11 +2343,6 @@ public final class McpClientFactory implements McpStreamFactory
         private HttpEventStream sse;
         boolean eventsUnsupported;
 
-        // guard-gated connect state: distinct from elicitCorrelationId/reauthTraceId above,
-        // which answer a backend-initiated elicitation on an already-open connection. These
-        // instead defer the very first doEncodeRequestBegin/onAppBeginImpl below until the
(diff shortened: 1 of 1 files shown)


jfallows (commented):
LGTM

PR #2429, opened 2026-08-25
fix(binding-mqtt, binding-mqtt-kafka): prevent engine crash on MQTT CONNECT with Will message

### LINKED ISSUE #2428: MQTT CONNECT with a Will message crashes the whole engine (JVM segfault), not just the connection
# MQTT CONNECT with a Will message crashes the whole engine (JVM segfault), not just the connection

**Zilla version:** 2.3.0 (also verified against current `develop`, i.e. 2.4.0-rc19 — the relevant code paths are byte-identical, so this is not fixed there either)

## What happened

Any MQTT client that connects with a Last Will and Testament set brings the whole Zilla process down, not just its own connection. We're running the `mqtt` → `mqtt-kafka` proxy chain behind TLS, and this reproduces every single time, with any will topic/payload, as far as we can tell.

Repro is as simple as:

```
mosquitto_sub -h <host> -p 8883 -v \
  --cafile ca.pem --cert client.pem --key client.key \
  -t "devices/device-123/commands" \
  --will-topic 'devices/device-123/status' \
  --will-payload '{"status":"offline"}'
```

As soon as the CONNECT (with the will) is decoded, the engine worker throws, self-terminates, and then the whole JVM segfaults and the container restarts — so every other connection on that worker gets dropped too.

## Config (anonymized, otherwise matches our setup)

```yaml
---
name: mqtt-kafka-proxy

vaults:
  client_vault:
    type: filesystem
    options:
      trust:
        store: /path/to/truststore.jks
        type: pkcs12
        password: ${{env.TRUSTSTORE_PASSWORD}}
      keys:
        store: /path/to/keystore.jks
        type: pkcs12
        password: ${{env.KEYSTORE_PASSWORD}}

stores:
  mqtt_store:
    type: memory

bindings:
  north_tcp_server:
    type: tcp
    kind: server
    options:
      host: 0.0.0.0
      port:
        - 8883
    exit: north_tls_server

  north_tls_server:
    type: tls
    kind: server
    vault: client_vault
    options:
      keys:
        - mqtt.example.com
      trust:
        - root-ca
      mutual: required
    exit: north_mqtt_server

  north_mqtt_server:
    type: mqtt
    kind: server
    options:
      store: mqtt_store
    exit: north_mqtt_kafka_mapping

  north_mqtt_kafka_mapping:
    type: mqtt-kafka
    kind: proxy
    options:
      topics:
        sessions: mqtt-sessions
        messages: device-events
        retained: mqtt-retained
      clients:
        - devices/{identity}/#
    routes:
      - when:
          - publish:
              - topic: devices/device-123/status
        with:
          messages: device-events
      - when:
          - publish:
              - topic: devices/device-123/commands
        with:
          messages: device-events

  north_kafka_cache_client:
    type: kafka
    kind: cache_client
    exit: north_kafka_cache_server

  north_kafka_cache_server:
    type: kafka
    kind: cache_server
    options:
      bootstrap:
        - mqtt-retained
        - mqtt-sessions
        - device-events
    exit: north_kafka_client

  north_kafka_client:
    type: kafka
    kind: client
    options:
      servers:
        - broker1:9092
        - broker2:9092
        - broker3:9092
    exit: north_tls_client

  north_tls_client:
    type: tls
    kind: client
    vault: client_vault
    options:
      trust:
        - root-ca
      keys:
        - mqtt.example.com
    exit: north_tcp_client

  north_tcp_client:
    type: tcp
    kind: client
```

## Log / stack trace

```
engine:events [.../...] ENGINE_STARTED Engine Started.
org.agrona.concurrent.AgentTerminationException: java.lang.IndexOutOfBoundsException: offset=50 is beyond maxLimit=-339
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.EngineWorker.doWork(EngineWorker.java:961)
        at org.agrona/org.agrona.concurrent.AgentRunner.doWork(AgentRunner.java:304)
        at org.agrona/org.agrona.concurrent.AgentRunner.workLoop(AgentRunner.java:296)
        at org.agrona/org.agrona.concurrent.AgentRunner.run(AgentRunner.java:162)
        at java.base/java.lang.Thread.run(Thread.java:1474)
Caused by: java.lang.IndexOutOfBoundsException: offset=50 is beyond maxLimit=-339
        at io.aklivity.zilla.runtime.binding.mqtt@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.internal.types.Flyweight.wrap(Flyweight.java:59)
        at io.aklivity.zilla.runtime.binding.mqtt@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.internal.types.OctetsFW.wrap(OctetsFW.java:39)
        at io.aklivity.zilla.runtime.binding.mqtt@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.internal.stream.MqttServerFactory$MqttServer.onDecodeConnectWillPayload(MqttServerFactory.java:3593)
        at io.aklivity.zilla.runtime.binding.mqtt@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.internal.stream.MqttServerFactory.decodeConnectWillMessagePayload(MqttServerFactory.java:1251)
        at io.aklivity.zilla.runtime.binding.mqtt@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.internal.stream.MqttServerFactory$MqttServer.decodeNetwork(MqttServerFactory.java:5359)
        at io.aklivity.zilla.runtime.binding.mqtt@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.internal.stream.MqttServerFactory$MqttServer.decodeNetwork(MqttServerFactory.java:5341)
        at io.aklivity.zilla.runtime.binding.mqtt@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.internal.stream.MqttServerFactory$MqttServer$MqttSessionStream.onSessionWindow(MqttServerFactory.java:5688)
        at io.aklivity.zilla.runtime.binding.mqtt@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.internal.stream.MqttServerFactory$MqttServer$MqttSessionStream.onSession(MqttServerFactory.java:5648)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.EngineWorker.handleReadInitial(EngineWorker.java:1544)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.EngineWorker.handleRead(EngineWorker.java:1477)
        at io.aklivity.zilla.runtime.common.agrona@2.3.0/io.aklivity.zilla.runtime.common.agrona.concurrent.ManyToOneRingBuffer.readEx(ManyToOneRingBuffer.java:316)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.EngineWorker.doWork(EngineWorker.java:947)
        ... 4 more
java.lang.NullPointerException: Cannot invoke "io.aklivity.zilla.runtime.engine.binding.function.MessageConsumer.accept(int, io.aklivity.zilla.runtime.common.agrona.buffer.DirectBufferEx, int, int)" because "receiver" is null
        at io.aklivity.zilla.runtime.binding.mqtt.kafka@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.kafka.internal.stream.MqttKafkaSessionFactory.doData(MqttKafkaSessionFactory.java:4636)
        at io.aklivity.zilla.runtime.binding.mqtt.kafka@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.kafka.internal.stream.MqttKafkaSessionFactory$KafkaSessionStream.doKafkaData(MqttKafkaSessionFactory.java:3063)
        at io.aklivity.zilla.runtime.binding.mqtt.kafka@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.kafka.internal.stream.MqttKafkaSessionFactory$KafkaSessionStream.sendWillSignal(MqttKafkaSessionFactory.java:3235)
        at io.aklivity.zilla.runtime.binding.mqtt.kafka@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.kafka.internal.stream.MqttKafkaSessionFactory$MqttSessionProxy.onMqttAbort(MqttKafkaSessionFactory.java:937)
        at io.aklivity.zilla.runtime.binding.mqtt.kafka@2.3.0/io.aklivity.zilla.runtime.binding.mqtt.kafka.internal.stream.MqttKafkaSessionFactory$MqttSessionProxy.onMqttMessage(MqttKafkaSessionFactory.java:538)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.EngineWorker.doSyntheticAbort(EngineWorker.java:1993)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.EngineWorker.lambda$detachStreams$0(EngineWorker.java:630)
        at java.base/java.lang.Iterable.forEach(Iterable.java:75)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.EngineWorker.detachStreams(EngineWorker.java:623)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.NamespaceRegistry.detachBinding(NamespaceRegistry.java:288)
        at java.base/java.lang.Iterable.forEach(Iterable.java:75)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.NamespaceRegistry.detach(NamespaceRegistry.java:150)
        at org.agrona/org.agrona.collections.Int2ObjectHashMap$ValueCollection.forEach(Int2ObjectHashMap.java:1328)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.EngineRegistry.detachAll(EngineRegistry.java:197)
        at io.aklivity.zilla.runtime.engine@2.3.0/io.aklivity.zilla.runtime.engine.internal.registry.EngineWorker.onClose(EngineWorker.java:1078)
        at org.agrona/org.agrona.concurrent.AgentRunner.run(AgentRunner.java:166)
        at java.base/java.lang.Thread.run(Thread.java:1474)
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f6c20c4d02d, pid=1, tid=2
# Problematic frame:
# J 4894 c2 jdk.internal.misc.Unsafe.getLongUnaligned(Ljava/lang/Object;JZ)J java.base@25.0.3
engine:events [.../...] ENGINE_STOPPED Engine Stopped.
```

(JRE: Temurin 25.0.3+9, mixed mode, g1 gc, linux-amd64)

## What I think is going on (happy to be corrected)

I went digging in `MqttServerFactory` and think I found it. In `onDecodeConnectWillPayload` (around line 3591 on `develop`):

```java
int payloadSize = Math.min(limit - offset, session.initialBudget() - headerSize);
...
final OctetsFW payload = payloadRO.wrap(buffer, offset, offset + payloadSize);
```

`session` here is the internal stream to the `mqtt-kafka` proxy side. `session.initialBudget()` is `initialMax - (initialSeq-initialAck) - initialPad`, i.e. whatever window the mqtt-kafka side has granted back via `WINDOW`. The session stream is opened (`doEstablishSession`) in the very same decode pass that then goes on to decode the will payload, so there's no way a `WINDOW` frame for it could have arrived yet — `initialMax` is still 0 at this point. So `session.initialBudget()` ends up `<= 0`, and since `headerSize` (topic, delay, qos, flags, properties, etc.) is always `> 0`, `session.initialBudget() - headerSize` is basically guaranteed to be negative. That makes `payloadSize` negative, which makes `offset + payloadSize` end up less than `offset`, and `OctetsFW.wrap()` throws because it requires `limit >= offset` — which matches the numbers in the trace (`offset=50`, `maxLimit=-339`).

That's the initial crash. It kills the engine worker thread. What makes it fatal for the whole process rather than just that worker is a second, separate bug: while the worker tears itself down, it synthetically aborts the still-open MQTT session, which calls `sendWillSignal()` → `doKafkaData()` → `doData()` in `MqttKafkaSessionFactory`, and that does `receiver.accept(...)` without checking whether `receiver` is actually set — in our case it's `null` because the Kafka-side session stream had never been fully established. That second, uncaught NPE seems to be what tips this over into the JVM segfault instead of a clean worker restart.

I noticed #1809 addressed a very similar-sounding class of problem (uncaught binding exception → engine worker self-terminates → JVM crash via `drain()` touching unmapped memory) and that fix is already in 2.3.0 and current `develop`. This looks like a related but distinct gap that #1809 doesn't cover, since the NPE here happens earlier, inside `onClose()`'s own synthetic-abort loop, before any buffers get released.

## Expected behavior

A client connecting with a will message shouldn't be able to take down the whole engine.

## Reproducibility

100% for us — every CONNECT with a will message on this proxy chain triggers it.

We're happy to open a PR for this (we've already dug into both spots above), just wanted to get it tracked first.

sfr-oc:
Fixes #2428

Two independent bugs combined to let a single MQTT client with a Last Will and Testament crash the whole engine (and in production, segfault the JVM):

1. **binding-mqtt**: `onDecodeConnectWillPayload` computed a payload size from `session.initialBudget() - headerSize` with no floor. When the session stream's granted window is smaller than the encoded will-message header alone, this goes negative and corrupts a subsequent `OctetsFW.wrap()` call, throwing an uncaught `IndexOutOfBoundsException` that terminates the engine worker. Fixed by checking `session.hasSessionWindow(headerSize)` first, mirroring the existing SUBSCRIBE-to-session-state-sync pattern in the same file.
2. **binding-mqtt-kafka**: `KafkaSessionStream.doKafkaData` had no `kafka != null` guard, unlike every sibling method (`doKafkaEnd`/`doKafkaAbort`/`doKafkaReset`/`doKafkaWindow`). Reachable independently of bug 1 whenever a client negotiates QoS 2 with Will + Clean Start and aborts before the Kafka-side session stream is established — `sendWillSignal` then NPEs on the unattached stream. This also amplified bug 1's crash into a full JVM crash, since it fired again during the engine's forced teardown.

Both fixes are minimal and follow existing idiomatic patterns in their respective files. Each has a new failing-first k3po spec scenario (confirmed to reproduce the exact production stack traces pre-fix) plus full regression coverage on both modules — see the two commits for details.

## Test plan
- [x] New k3po scenario for bug 1 (`session.will.message.header.exceeds.window`) fails pre-fix with the reported `IndexOutOfBoundsException`, passes post-fix
- [x] New k3po scenario for bug 2 (`session.will.message.qos2.abort.before.session.established`) fails pre-fix with the reported NPE, passes post-fix
- [x] Full regression suite green on `runtime/binding-mqtt`, `runtime/binding-mqtt-kafka`, `specs/binding-mqtt.spec`, `specs/binding-mqtt-kafka.spec`
- [x] `checkstyle:check` clean on all four modules

### DIFF
--- a/runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
+++ b/runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
@@ -3019,13 +3019,15 @@ public class MqttKafkaSessionFactory implements MqttKafkaStreamFactory
             int limit,
             Flyweight extension)
         {
+            if (kafka != null)
+            {
+                doData(kafka, originId, routedId, initialId, initialSeq, initialAck, initialMax,
+                    traceId, authorization, budgetId, flags, reserved, buffer, offset, limit, extension);
 
-            doData(kafka, originId, routedId, initialId, initialSeq, initialAck, initialMax,
-                traceId, authorization, budgetId, flags, reserved, buffer, offset, limit, extension);
+                initialSeq += reserved;
 
-            initialSeq += reserved;
-
-            assert initialSeq - padding <= initialAck + initialMax;
+                assert initialSeq - padding <= initialAck + initialMax;
+            }
         }
 
         protected final void doKafkaData(
@@ -3037,12 +3039,15 @@ public class MqttKafkaSessionFactory implements MqttKafkaStreamFactory
             OctetsFW payload,
             Flyweight extension)
         {
-            doData(kafka, originId, routedId, initialId, initialSeq, initialAck, initialMax,
-                traceId, authorization, budgetId, flags, reserved, payload, extension);
+            if (kafka != null)
+            {
+                doData(kafka, originId, routedId, initialId, initialSeq, initialAck, initialMax,
+                    traceId, authorization, budgetId, flags, reserved, payload, extension);
 
-            initialSeq += reserved;
+                initialSeq += reserved;
 
-            assert initialSeq <= initialAck + initialMax;
+                assert initialSeq <= initialAck + initialMax;
+            }
         }
 
         protected void doKafkaData(
@@ -3055,17 +3060,20 @@ public class MqttKafkaSessionFactory implements MqttKafkaStreamFactory
             Flyweight payload,
             Flyweight extension)
         {
-            final DirectBufferEx buffer = payload.buffer();
-            final int offset = payload.offset();
-            final int limit = payload.limit();
-            final int length = limit - offset;
+            if (kafka != null)
+            {
+                final DirectBufferEx buffer = payload.buffer();
+                final int offset = payload.offset();
+                final int limit = payload.limit();
+                final int length = limit - offset;
 
-            doData(kafka, originId, routedId, initialId, initialSeq, initialAck, initialMax,
-                traceId, authorization, budgetId, flags, reserved, buffer, offset, length, extension);
+                doData(kafka, originId, routedId, initialId, initialSeq, initialAck, initialMax,
+                    traceId, authorization, budgetId, flags, reserved, buffer, offset, length, extension);
 
-            initialSeq += reserved;
+                initialSeq += reserved;
 
-            assert initialSeq - padding <= initialAck + initialMax;
+                assert initialSeq - padding <= initialAck + initialMax;
+            }
         }
 
         private void doKafkaFlush(
--- a/runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
+++ b/runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
@@ -277,6 +277,17 @@ public class MqttKafkaSessionProxyIT
 
     @Test
     @Configuration("proxy.yaml")
+    @Configure(name = WILL_AVAILABLE_NAME, value = "false")
+    @Specification({
+        "${mqtt}/session.will.message.qos2.abort.before.session.established/client",
+        "${kafka}/session.will.message.qos2.abort.before.session.established/server"})
+    public void shouldSkipWillSignalOnQos2AbortBeforeSessionEstablished() throws Exception
+    {
+        k3po.finish();
+    }
+
+    @Test
+    @Configuration("proxy.yaml")
     @Configure(name = PUBLISH_MAX_QOS_NAME, value = "1")
     @Specification({
         "${mqtt}/session.will.message.abort.deliver.will.retain/client",
--- a/runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
+++ b/runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
@@ -3561,6 +3561,7 @@ public final class MqttServerFactory implements MqttStreamFactory
             final int willQos = decodeWillQos(connectFlags);
             final boolean willFlagSet = isSetWillFlag(connectFlags);
 
+            decode:
             if (willFlagSet && MqttState.initialOpened(session.state))
             {
                 int publishedWillSize = 0;
@@ -3588,6 +3589,12 @@ public final class MqttServerFactory implements MqttStreamFactory
 
                     final MqttWillMessageFW will = willMessageBuilder.build();
                     final int headerSize = willMessageBuilder.sizeof();
+
+                    if (!session.hasSessionWindow(headerSize))
+                    {
+                        break decode;
+                    }
+
                     int payloadSize = Math.min(limit - offset, session.initialBudget() - headerSize);
 
                     final OctetsFW payload = payloadRO.wrap(buffer, offset, offset + payloadSize);
--- a/runtime/binding-mqtt/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/server/v5/SessionIT.java
+++ b/runtime/binding-mqtt/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/server/v5/SessionIT.java
@@ -190,6 +190,16 @@ public class SessionIT
     @Test
     @Configuration("server.yaml")
     @Specification({
+        "${net}/session.will.message.header.exceeds.window/client",
+        "${app}/session.will.message.header.exceeds.window/server"})
+    public void shouldDeferWillMessageWhenHeaderExceedsSessionWindow() throws Exception
+    {
+        k3po.finish();
+    }
+
+    @Test
+    @Configuration("server.yaml")
+    @Specification({
         "${net}/session.will.message.disconnect.with.will.message/client",
         "${app}/session.will.message.abort/server"})
     public void shouldCloseSessionDisconnectWithWill() throws Exception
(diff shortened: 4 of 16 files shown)


jfallows:
Superseded by #2436, which incorporates the `session.hasSessionWindow(headerSize)` guard from this PR's first fix and builds on it — deferring the will-message forward to a later session-window retry (rather than crashing) turned up three further bugs in how the deferred decode state was handled on abort/error/end, plus a stale-flyweight-view bug in the header itself, all addressed there. Closing this in favor of #2436; thanks for tracking down and diagnosing the original crash.

---

PR #2431, opened 2026-08-25
fix(binding-kafka): honor per-message authorization on the cache fetch path

### LINKED ISSUE #2423: kafka cache storage: honor per-message authorization on the fetch path, not just at cache population

As a Zilla operator, I want the kafka cache storage layer's fetch path to honor the per-message `authorization` parameter aklivity/zilla#2310 already added to `ModelPipeline.transform(...)`, so that an installed model transform whose behavior depends on that parameter gets the contract #2310 established, rather than only ever seeing whatever value was in effect when the cache was populated.

**The gap, concretely**

`KafkaCacheServerFetchFactory` decodes a fetched record's value **once**, while populating the local cache, with authorization hardcoded to a fixed sentinel (`NO_AUTHORIZATION`) — there is no requesting consumer yet at that point, since this runs ahead of and independent of any consumer's own fetch. The result is stored in the partition's `convertedFile` and served to every subsequent reader verbatim by the cursor that reads it back — including the per-consumer fetch stream itself, which constructs no model pipeline of its own at all today.

This is correct and efficient for a transform whose result never depends on the sentinel value at all — the existing `view: json` conversion is exactly this kind of transform, and caching it once genuinely is the right behavior, not a bug to fix. It's a real gap for any transform whose behavior is defined in terms of the caller's own `authorization`: evaluating it once, against a fixed sentinel, and sharing that single answer with every future reader forever doesn't honor #2310's own contract for that parameter — the parameter exists specifically to let behavior vary per caller, and this path never gives it a real, caller-specific value at all.

**Acceptance criteria**

- Give the per-consumer fetch stream a real model-decode invocation of its own, carrying that consumer's live `authorization`, rather than only ever serving whatever the population-time decode already computed and cached
- The population-time decode and cache remain exactly as they are for whatever portion of a model's composed pipeline does not depend on the calling consumer's authorization — this issue is about adding a missing per-consumer invocation, not removing or replacing the existing shared one
- **Open design question, not assumed by this issue: how does a model's composed pipeline (aklivity/zilla#1681's fold) distinguish an installed extension stage that's safe to compute once and share from one that must be recomputed per requesting consumer?** Today's fold treats every installed stage identically. Resolving this — whether by a property a stage declares, a property of the pipeline as configured, or some other mechanism — is part of this issue's scope, not a premise to design around
- A consumer whose authorization changes mid-stream (reauthorize, expire) sees the new authorization reflected starting with its next fetched message, not the authorization in effect when the stream opened
- A model with no authorization-dependent stage installed shows no behavior change and no added cost from this issue — the existing shared, cached path remains untouched in that case
- Unit and k3po coverage: two consumers with different authorization fetching the same underlying record get correspondingly different results; a mid-stream authorization change is reflected on the next message, not retroactively or with a delay beyond that; a model with only authorization-independent stages installed is unaffected, byte for byte, from today's behavior

**Additional context**

Split out of aklivity/zilla#1689, which is narrower (wiring `ModelEnvelope` into the existing read/write paths) and doesn't need this. Scoped to the shared kafka cache storage implementation specifically — the same implementation layer #1688/#1689/#1690 already scope themselves to — not to any other binding's own request-handling path.

jfallows:
## Description

`KafkaCacheServerFetchFactory` decoded a fetched Kafka record&#39;s value **once**, while populating the local cache from the broker, with `authorization` hardcoded to `NO_AUTHORIZATION` — there is no requesting consumer yet at that point. The decoded result was stored in the partition&#39;s `convertedFile` and served to every future reader verbatim, including the per-consumer fetch stream (`KafkaCacheClientFetchFactory`), which constructed no model pipeline of its own at all and never threaded its own live `authorization` into anything — meaning the per-message `authorization` parameter `ModelPipeline.transform(...)` already carries (#2369) never reached a real, per-consumer value on this path.

### Correction (superseding the original design)

The first version of this fix (now superseded by later commits on this branch) added a separate, abstract `ModelHandler.supplyCacheable(...)` method alongside `supplyDecoder`/`supplyEncoder`. CI caught a real regression in that design: the `http.kafka.avro.json` example hung for its full 30-minute timeout. Root cause — `supplyCacheable` and `supplyDecoder` used the *same* frontend parser and output format, differing only in which ext-handler fold ran. That&#39;s fine for a byte-for-byte identity decode, but any model whose decode does real, format-converting work (e.g. `view: json`) produces output that differs from its wire-format input. Caching that converted output and then re-decoding it with a wire-format frontend can never work: the avro model&#39;s `view: json` config made the populate-time decode cache JSON text, and the per-consumer fetch-time decode then fed that JSON into a parser expecting avro wire bytes.

The corrected design replaces the two-method split with a single `supplyDecoder(envelope, transform, ModelCache)`, parameterized by a three-way cache context:
- **`NONE`** — today&#39;s plain decode, no cache involved.
- **`WRITE`** — decode ahead of any specific reader, producing the value a cache persists (the old `supplyCacheable`&#39;s role).
- **`READ`** — decode a value already in whatever form `WRITE` produced, for the reader requesting it now. Its frontend matches whatever `WRITE` actually emitted — for avro/protobuf&#39;s `view: json`, that&#39;s `AvroJson.stream`/`ProtobufJson.parser`, the same json-view frontend already used on the encode-shaped pipeline, reused rather than duplicated.

Each model&#39;s own `*ModelExtHandler` SPI mirrors this as a single `decode(T, Cache)` method — a local enum per model (`AvroCache`/`ProtobufCache`/`JsonCache`/`CoreCache`), not the shared engine `ModelCache`, so each model&#39;s extension surface stays independent — replacing the previous `decode(T)`/`cacheable(T)` pair.

Building this also surfaced and closed a related gap: protobuf&#39;s own message-index framing (separate from catalog framing) has no magic-byte-style validity guard, so a cached `view:json` value with no index framing at all would misparse its leading JSON bytes as an index varint on `READ`. Fixed by resolving the message via the same static `catalog.record` path the encode side already uses whenever `cache == READ`.

Finally, `WRITE` now persists the schema id it resolves as catalog framing ahead of its cached output (reusing the same `CatalogHandler.encode(...)` machinery the encode-direction pipeline already uses), so `READ` recovers it directly from the cached bytes via the existing `resolveSchemaId`/`prefix` logic, with no changes needed on the read side. This closes the one case that otherwise stays broken even with the frontend fix: a `strategy: encoded` schema id is per-message authoritative (schema evolution means two cached messages on the same topic can carry different ids), and a view-converting `WRITE` output previously discarded it entirely with no static fallback available — unlike `strategy: topic`/subject-pinned configs, whose schema was never message-embedded to begin with.

### Verification

- `AvroModelDecoderPipelineTest`/`ProtobufModelDecoderPipelineTest` gained `WRITE`→`READ` round-trip tests for both the `view: json` and no-view cases, plus a case proving `READ` recovers the schema id from `WRITE`-injected framing even when the model&#39;s own static catalog reference is deliberately wrong — the exact `strategy: encoded` shape. These are the tests that would have caught the original regression; `TestModel`&#39;s identity decode structurally cannot exercise this failure mode. Exercising the framing case required extending the engine&#39;s `TestCatalogHandler` test double (`resolve(DirectBufferEx,...)`/`decodePadding(...)`) to honor its own `prefix` option symmetrically with its existing `encode(...)`, per this repo&#39;s convention of extending `type: test` implementations additively rather than reaching for a production catalog implementation in test scope.
- `CacheFetchIT` gained a new scenario (`shouldReceiveMessageValueAvroViewJson`) driving a real `avro` + `view: json` config through a live engine populate-then-fetch round trip — the exact shape of the original regression — verified passing against the live engine, not just unit-level.
- Full `runtime/binding-kafka` suite (401 tests, all k3po ITs included), `specs/binding-kafka.spec` suite (peer-to-peer + config), `runtime/model-avro`/`model-protobuf`/`model-json`/`model-core` suites, and `runtime/binding-http`/`binding-sse`/`binding-mqtt`/`binding-mcp` suites all pass with no regressions; checkstyle and license checks clean.
- CI on this branch's merge with latest `develop` (commit `d63e78ab9`) is green across all 42 checks, including `testing (http.kafka.avro.json)` (2m27s, vs. the original 30-minute hang) and `testing (mcp.proxy)`.

Fixes #2423

---


jfallows:
`testing (mqtt.kafka.proxy)` failed on `c1dd7dec8` (subscriber got empty output instead of `Hello Zilla!`, right at the edge of the test script's 5-second `mosquitto_sub` timeout window). Took this seriously since it exercises the exact Kafka cache read path this PR touches, rather than assuming it's unrelated.

Reproduced locally: built the `develop-SNAPSHOT` docker image from this exact commit and ran the `mqtt.kafka.proxy` example's compose stack + `.github/test.sh` directly — it passed cleanly (`OUTPUT=Hello Zilla!`, matched, exit 0). Same code, same commit, no flake. This points to a one-off timing race in that CI run rather than a real regression. Re-ran the failed job to confirm.

---

PR #2430, opened 2026-08-25
feat(engine): let a Router own this instance's node identity

jfallows:
## Description

An engine worker's affinity always embeds a node identity byte, but that identity came only from a manually-configured `node.id` property — there was no way for a `Router` coordinating cluster membership across instances to supply it instead, and no persistent per-instance identity to negotiate that coordination against.

This adds `Router.resolveNodeId()`, defaulting to `0`, so a router implementation can resolve this engine's own node identity. `EngineWorker` now sources its identity from the router rather than reading `node.id` off `EngineConfiguration` directly.

It also adds a `node.instance.id` engine property: a stable per-instance label, generated once as a UUID and persisted under the engine directory so it survives a restart, for a `Router` implementation to key its own membership and identity-assignment coordination off of. `node.id` is removed as redundant, now that any router with something better than a fixed `0` to offer decides identity itself.

No dependencies required for this change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W7N8Z9vABdU8pVPGthRodY)_

PR #2432, opened 2026-08-25
fix(binding-mcp): keep serving other routes when one route errors during tools/prompts/resources list

jfallows:
## Description

`McpListServer.onClientError` unconditionally cleared every remaining route and aborted the whole `tools/list` (or `prompts/list`/`resources/list`) response whenever a single south route errored (Abort, Reset, or buffer-pool exhaustion) — even when sibling routes had already succeeded or were still pending. A single flaky south route could wipe out an otherwise complete multi-route list response instead of simply being excluded from it, the same way an unauthorized/skipped route already is.

This adds `anySucceeded`/`anyErrored` tracking to `McpListServer` so `onClientError` only aborts the overall response when every attempted route has failed and none has ever succeeded; otherwise it moves on to the next route, converging to a partial result.

Also adds K3PO regression coverage for:
- a live multi-route list where one route errors (abort) and the other route succeeds, verifying the response gracefully returns just the successful route's items instead of aborting entirely
- a south route sending a Reset mid-hydration, verifying it's correctly excluded from the cached list rather than hanging hydration

Both new scenarios were verified to fail on the pre-fix code and pass with the fix. Full `runtime/binding-mcp` (298 tests) and `specs/binding-mcp.spec` (263 tests) suites pass; the only failures observed in this environment are 8 pre-existing, environmentally-flaky `McpClientIT` elicit tests that reproduce identically on unmodified `develop` and are unrelated to this change.

---

### DIFF
--- a/runtime/binding-echo/src/test/java/io/aklivity/zilla/runtime/binding/echo/internal/bench/EchoWorker.java
+++ b/runtime/binding-echo/src/test/java/io/aklivity/zilla/runtime/binding/echo/internal/bench/EchoWorker.java
@@ -66,6 +66,13 @@ public class EchoWorker implements EngineContext
     }
 
     @Override
+    public int indexOf(
+        long id)
+    {
+        return 0;
+    }
+
+    @Override
     public Signaler signaler()
     {
         return null;
--- a/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
+++ b/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
@@ -5003,9 +5003,10 @@ public final class HttpClientFactory implements HttpStreamFactory
             assert requestAck <= requestSeq;
             client.requestSharedBudget -= reserved;
 
+            final boolean alreadyClosed = HttpState.initialClosed(state);
             state = HttpState.closeInitial(state);
 
-            if ((streamId & 0x01) == 0x01)
+            if (!alreadyClosed && (streamId & 0x01) == 0x01)
             {
                 final HttpEndExFW endEx = end.extension().get(endExRO::tryWrap);
                 final Array32FW<HttpHeaderFW> trailers = endEx != null ? endEx.trailers() : DEFAULT_TRAILERS;
--- a/runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/bench/KafkaModelWorker.java
+++ b/runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/bench/KafkaModelWorker.java
@@ -78,6 +78,13 @@ final class KafkaModelWorker implements EngineContext
     }
 
     @Override
+    public int indexOf(
+        long id)
+    {
+        return 0;
+    }
+
+    @Override
     public Signaler signaler()
     {
         return null;
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpAuthorizationResult.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpAuthorizationResult.java
@@ -19,6 +19,6 @@ import io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBearerErro
 public record McpAuthorizationResult(
     long authorization,
     McpBearerError error,
-    boolean owned)
+    Runnable deauthorize)
 {
 }
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
@@ -73,6 +73,10 @@ public final class McpBindingConfig
     private static final String HTTP_HEADER_PATH = ":path";
     private static final String HTTP_HEADER_AUTHORIZATION = "authorization";
 
+    private static final Runnable NOOP = () ->
+    {
+    };
+
     private static final String SCHEME_HTTPS = "https";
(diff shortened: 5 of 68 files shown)


jfallows:
The `testing (mcp.proxy)` check failed here too (job `98038140959`), same "cache never finished hydrating" failure bucket as reported on #2430 and seen earlier on #2433's own CI run — three independent, otherwise-unrelated branches now.

This run is actually informative: unlike the other two, the `initialize` cache-hydration probe *succeeded* here (`resources.subscribe:true` came back fine) — it was the follow-up `tools/list` client call that hung for 60s and was killed by the test harness's timeout, which the script's `full_toolset_present` check then reports under the same "cache never finished hydrating" message. Since this branch already carries this PR's own fix for the `McpProxyListFactory.onClientError` sibling-route-wipe bug and still hit the same failure bucket, that rules out this PR's change as the cause — it's a shared environmental flake in the `testing (mcp.proxy)` CI job (multi-service docker-compose stack: kafka, kafka-connect, karapace-registry, petstore, everything, urlelicit, ghapi — plausibly service startup racing the test's fixed retry budget), not something introduced by this PR.

Re-ran the failed job once per protocol.

---

jfallows:
Update: root-caused this properly rather than standing on the "environmental flake" theory from my last comment.

Rebuilt the `develop-SNAPSHOT` image locally from this branch and ran `examples/mcp.proxy`'s `.github/test.sh` directly against it (with `docker compose cp` + `zilla dump`/container logs for diagnosis). The `zilla` container was crash-looping — **11 restarts** in a few minutes — on:

```
org.agrona.concurrent.AgentTerminationException: java.lang.NullPointerException:
Cannot invoke "McpServerFactory$McpLifecycleStream.doAppEnd(long, long)" because "this.session" is null
```

Root cause: an HTTP `DELETE` (MCP session termination) whose `mcp-session-id` is missing or doesn't match a tracked session was passed straight into `McpShutdownHandler` with `session == null`. Its `onNetBegin` unconditionally calls `session.doAppEnd(...)`, throwing an NPE that kills the whole engine worker via `AgentTerminationException` and restarts the process — taking every other in-flight session down with it. The `GET` path already guards this exact case with a 400 (`McpRejectHandler`); `DELETE` was just missing the same check.

This is what's been producing the "cache never finished hydrating" / intermittent connection-refused symptom across `testing (mcp.proxy)` runs on this PR, #2430, and even #2433 (a pure test-script PR unrelated to any of this) — a stray/duplicate/late session-termination call from any client crashes the whole server for everyone concurrently hitting it, which is exactly what this test's many concurrent toolkit routes + retrying test harness would trigger.

Pushed the fix (`10cf97d3f3`): mirrors the `GET` path's `session == null` → 400 guard for `DELETE`, plus two new k3po regression scenarios (`lifecycle.shutdown.session.unknown` / `.missing`) that reproduce the exact crash on unfixed code and pass after the fix — validated via `git stash` bisection. Full `runtime/binding-mcp` (300 tests) and `specs/binding-mcp.spec` (265 tests) suites pass. Locally, the `zilla` container now completes the same test run with **0 restarts** (down from 11).

---

jfallows:
`testing (mcp.proxy)` failed again on the fixed commit (`10cf97d3f3`) — same "cache never finished hydrating" bucket, but this time the underlying cause is different and confirmed via local reproduction to be a **separate, pre-existing bug**, unrelated to this PR's diff.

**Root cause**: when a client's SSE `GET` (event stream) and a request `POST` (e.g. `tools/list`) for the same freshly-established session arrive on two separate connections with **zero delay between them**, both hang indefinitely until the session's inactivity timeout. Isolated with a ~20-line, dependency-free Node script using only `fetch()` — no MCP SDK, no npm/TLS involvement, reproduces reliably on demand. The real MCP client SDK triggers this on every connection because `StreamableHTTPClientTransport` opens the SSE stream fire-and-forget right after `notifications/initialized`, then races ahead to send the next request with no gap.

This is architecturally distinct from what this PR touches: it traces into `McpServerFactory`'s `McpEventStream`/`McpLifecycleStream` (`doAppResume`) and `McpProxyLifecycleFactory` (`onServerChallenge` / `resumePending` / the per-toolkit-client `requests` queue) — session/lifecycle establishment, not `McpProxyListFactory`'s list aggregation (what this PR fixes). It isn't covered by any existing k3po test, since those scripts are byte-sequential by construction and can't express a true zero-delay race — so this has likely been latent for a while, only surfacing with a real client's actual connection pattern.

This PR's own fix (the DELETE null-session crash) is validated and necessary — confirmed via local reproduction that it eliminates 11 restarts down to 0. But it isn't sufficient on its own to turn this specific CI check green, since this second bug is independent and still triggers on every real client connection.

Given the scope and depth of a proper fix here (coordinating two separate factories' state machines), I'd suggest treating this as its own follow-up rather than expanding this PR's scope — happy to open a tracked issue and/or start on it as a separate PR if that's preferred.

---

jfallows:
Follow-up to my last comment: root-caused and fixed the second bug, opened as #2434.

**Actual root cause turned out to be different from the GET/POST race I described above.** That race is real (concurrent GET+POST for a fresh session do hang), but it wasn't what's crashing `testing (mcp.proxy)` — with `-ea` enabled and zero client traffic at all, the `zilla` container still crash-loops within ~10s of a cold start, purely from cache-hydration's own outbound HTTP client requests.

`McpClientFactory.HttpStream.doNetEnd` (binding-mcp's outbound HTTP client role, used for `tools/list`, `tools/call`, cache hydration, etc.) sends its request End frame through a `doEnd` overload that hardcodes `sequence=0`/`acknowledge=0`/`maximum=0`, discarding the real running sequence already advanced by the request body. Any client-role request with a non-empty body closes with a stale sequence, violating `HttpClientFactory`'s own request-sequence invariant and corrupting that connection's flow-control accounting — crashes under `-ea`, silently corrupts state without it.

This PR's DELETE null-session fix stays scoped and unaffected; #2434 carries the sequence fix plus a defensive duplicate-End guard in `HttpClientFactory`, verified via the same local cold-start repro (0 restarts / 0 assertion errors after the fix, real MCP client `tools/list` call verified end-to-end).

`testing (mcp.proxy)` on this PR will likely stay red until #2434 merges — that failure is tracked there now, not a regression from this PR's diff.

---

PR #2434, opened 2026-08-26
fix(binding-mcp): preserve request sequence on mcp(client) HttpStream end

jfallows:
## Description

Found while root-causing the persistent `testing (mcp.proxy)` CI failures on #2432/#2430/#2433. #2432 fixed a separate NPE crash (DELETE with an unknown session); this is a second, distinct bug in `binding-http`/`binding-mcp` that also produces those failures.

**Root cause**: `McpClientFactory.HttpStream.doNetEnd` (the mcp binding's outbound HTTP client role, used for any real MCP-over-HTTP client connection — cache hydration, `tools/list`, `tools/call`, etc.) sent its request End frame through a convenience `doEnd` overload that hardcodes `sequence=0`/`acknowledge=0`/`maximum=0`, discarding the real `initialSeq`/`initialAck`/`initialMax` that `encodeNet` had already advanced while streaming the request body. Any client-role request with a non-empty body therefore closed with a stale sequence, violating `HttpClientFactory`'s own request-sequence invariant (`assert sequence >= requestSeq` in `HttpExchange.onRequestEnd`) and resetting that connection's flow-control accounting backwards for whatever request follows.

Under `-ea` this crashes the engine worker outright (`AgentTerminationException` → `AssertionError`), which is how it was caught: with zero client traffic, a fresh `mcp.proxy` stack crash-loops within ~10s of the first cache-hydration request to any south backend with a request body. Without `-ea` (production, and CI) the assertion is silently skipped but the corrupted sequence/budget state remains, which is a plausible source of the intermittent hangs seen in CI beyond the crash itself.

Also hardened `HttpClientFactory.HttpExchange.onRequestEnd`: the outbound request-end encode is now gated on the exchange's initial-closed state, so a duplicate End (from this bug or any other misbehaving client) can no longer re-trigger the encoder a second time.

## Fix

- `McpClientFactory.HttpStream.doNetEnd`: pass `initialSeq`/`initialAck`/`initialMax` to `doEnd` instead of the zero-filling overload.
- `HttpClientFactory.HttpExchange.onRequestEnd`: only invoke `doEncodeRequestEnd` when the request side wasn't already closed.

## Verification

- `./mvnw clean install -pl runtime/binding-mcp,runtime/binding-http -am` — all existing unit/IT/spec tests green, checkstyle/license clean.
- Reproduced the crash locally: built the `examples/mcp.proxy` Docker image with `-ea` enabled, brought the stack up cold (no client traffic at all) — reliably crash-looped via the `assert sequence >= requestSeq` failure within ~10s.
- Applied the fix, rebuilt, repeated the cold start: 0 restarts, 0 assertion errors over several minutes of observation.
- Ran the real `tools-list-client` MCP client end-to-end against the fixed build — `tools/list` resolves correctly.

Fixes # (no tracked issue — found during investigation of #2432's CI failures)

---

### DIFF
--- a/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
+++ b/runtime/binding-http/src/main/java/io/aklivity/zilla/runtime/binding/http/internal/stream/HttpClientFactory.java
@@ -5003,9 +5003,10 @@ public final class HttpClientFactory implements HttpStreamFactory
             assert requestAck <= requestSeq;
             client.requestSharedBudget -= reserved;
 
+            final boolean alreadyClosed = HttpState.initialClosed(state);
             state = HttpState.closeInitial(state);
 
-            if ((streamId & 0x01) == 0x01)
+            if (!alreadyClosed && (streamId & 0x01) == 0x01)
             {
                 final HttpEndExFW endEx = end.extension().get(endExRO::tryWrap);
                 final Array32FW<HttpHeaderFW> trailers = endEx != null ? endEx.trailers() : DEFAULT_TRAILERS;
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpAuthorizationResult.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpAuthorizationResult.java
@@ -19,6 +19,6 @@ import io.aklivity.zilla.runtime.binding.mcp.internal.types.stream.McpBearerErro
 public record McpAuthorizationResult(
     long authorization,
     McpBearerError error,
-    boolean owned)
+    Runnable deauthorize)
 {
 }
--- a/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
+++ b/runtime/binding-mcp/src/main/java/io/aklivity/zilla/runtime/binding/mcp/internal/config/McpBindingConfig.java
@@ -73,6 +73,10 @@ public final class McpBindingConfig
     private static final String HTTP_HEADER_PATH = ":path";
     private static final String HTTP_HEADER_AUTHORIZATION = "authorization";
 
+    private static final Runnable NOOP = () ->
+    {
+    };
+
     private static final String SCHEME_HTTPS = "https";
     private static final int PORT_HTTP = 80;
     private static final int PORT_HTTPS = 443;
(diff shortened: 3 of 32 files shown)


jfallows:
`testing (mcp.proxy)` failure update: this PR's fix (sequence-corruption on `McpClientFactory.HttpStream.doNetEnd`) is confirmed working — merged in #2432's DELETE null-session fix (needed since this branch predates it) and re-ran the actual `examples/mcp.proxy/.github/test.sh` script locally against the merged build: **0 container restarts and 0 assertion errors across 475s** of cache-hydration retries, where it previously crash-looped within ~10s every time.

The job is still red because a **second, separate, pre-existing bug** is what the test is actually failing on now — not this PR's diff. The real `tools-list-client` MCP SDK client hangs indefinitely on `client.listTools()` (killed by the test's own timeout wrapper every retry) and never lists any tools. Root cause, established earlier while investigating this same CI job: `StreamableHTTPClientTransport` fires its SSE `GET` stream fire-and-forget immediately after `notifications/initialized` returns, then races ahead to send the next request (here, `tools/list`) with zero delay — a genuine concurrent-GET+POST-for-a-fresh-session race. It traces into `McpServerFactory`'s `McpEventStream`/`McpLifecycleStream` (`doAppResume`) and `McpProxyLifecycleFactory`'s `onServerChallenge`/`resumePending`/per-toolkit `requests` queue — session/lifecycle establishment, not the sequence-tracking code this PR touches.

No existing fix for this is on `develop` or any open PR I'm aware of, so I can't port one in. I'm continuing to investigate it now; will push a fix here if it lands cleanly, or open it as its own PR if the scope warrants keeping this one focused. Re-running the failed check wouldn't help — this reproduces deterministically, not a flake — so skipping that step.

---

jfallows:
Narrowed the remaining GET/POST bug further with debug instrumentation (reverted before this push, not part of the diff).

Traced a single failing race: `tools/list` POST gets cross-worker redirected (worker#2 → worker#0, where the session lives) while a concurrent SSE `GET` is open on the same session. Server-side tracing shows the redirected POST is fully processed — `onDecodeToolsList` fires, the cache lookup (`McpCacheListServer.onStoreResult`) returns a hit — but the client's `fetch()` for that POST fails immediately with `terminated` (the connection is closed before the response reaches it), even though the server believes it answered.

Ruled out "any redirected request can fail" as too broad: a sequential probe (no GET at all, 15 back-to-back `initialize`+`notifications/initialized`+`tools/list` cycles, naturally hitting all 4 workers and needing redirects) succeeded 15/15. So a bare cross-worker redirect is fine — this only breaks when a concurrent GET/SSE is also open for that session. Also checked whether the session's own lifecycle stream ID and the tools/list request's ID could collide (`McpLifecycleStream` uses the plain incrementing `supplyInitialId`, `McpRequestStream` uses `supplyInitialIdAffinity` keyed off session+route) — different allocators, so not an obvious ID collision there.

This is genuinely deep (crosses `McpServerFactory`'s redirect/session-affinity handling and the underlying engine's connection hand-off between workers) and I haven't isolated the exact drop point yet. Given the scope, I don't think it's right to keep expanding this PR's investigation indefinitely — recommending this become its own tracked follow-up rather than blocking this PR further. This PR's own fix remains fully validated and ready; happy to open a new issue/PR for the redirect-vs-SSE bug, or continue here if preferred.

---

PR #2435, opened 2026-08-26
Expose schema root properties

### LINKED ISSUE #2415: Retain and expose unrecognized schema-root properties, alongside existing field-level annotation retention

As a Zilla operator, I want a catalog implementation to be able to attach a single fragment of metadata to a schema as a whole — distinct from tagging any individual field — so that metadata scoped to the entire schema isn't forced into a per-field shape it doesn't fit.

**Acceptance criteria**

## 1. Root-level (schema-root) annotation access

A peer mechanism to aklivity/zilla#2302's field-level annotation retention, not a broadening of it. #2302 retains and exposes unrecognized *field* annotations (`AvroField.attribute(name)` and peers); this issue does the same at the schema *root*.

- Add `JsonValue attribute(String name)` at the schema-root level for each format (`AvroSchema`, `JsonSchema`, `ProtobufSchema` — mirroring #2302's per-format field-level `attribute(String)`), returning `null` when absent
- No inheritance to fields — a root-level annotation is never consulted when matching a field-level `tagged:` matcher, and a field-level annotation is never visible through this accessor
- Requires retaining unrecognised schema-root properties in each format's compiler, parallel to #2302's per-field retention, at the root instead

## 2. Where this matters: a catalog resolving something schema-wide

A catalog's `resolve(schemaId): String` implementation is the natural place to attach schema-wide metadata that a registry returns as a sibling of the schema text rather than embedded within it — fetch both in the one round trip, and synthesize a schema text carrying the extra property at the schema root before returning it. An ordinary, unaware caller parsing the same text is unaffected, since a JSON schema format already tolerates unrecognized top-level properties.

## 3. Coverage

- round-trip: a schema declaring a root-level property of each JSON shape (scalar, array, object) compiles, exposes it via `attribute(name)`, and survives the existing conformance suites, for each of the three formats
- a root-level property never satisfies a field-level `tagged:` matcher, and a field-level annotation is never visible through the root accessor
- absent property returns `null`, not an exception

**Additional context**

Companion to aklivity/zilla#2302, same retention mechanism at a different scope.

ReguiguiMohamed:
## Description

Schema root metadata should remain available after compilation through APIs that fit each schema format.

This change adds `AvroType.attribute(String)` to return members from Avro type declarations as `JsonValue`. It adds `ProtobufMessage.option(String)` to return message-level options as `ProtobufConstant` for both proto2 and proto3 schemas. Message options are also preserved when overlays rebuild a protobuf message.

The tests cover Avro type attributes, the existing JSON root attribute behavior, protobuf scalar and message-valued options, proto2 and proto3 parsing, and overlay retention.

## Testing

`mvn -pl runtime/common-json,runtime/common-avro,runtime/common-protobuf clean install`

Fixes #2415

### DIFF
--- a/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/AvroType.java
+++ b/runtime/common-avro/src/main/java/io/aklivity/zilla/runtime/common/avro/AvroType.java
@@ -17,6 +17,8 @@ package io.aklivity.zilla.runtime.common.avro;
 import java.util.List;
 import java.util.function.Predicate;
 
+import jakarta.json.JsonValue;
+
 /**
  * An immutable, read-only node in a compiled Avro type graph. The {@link #kind()} selects which
  * accessors carry meaning; the others return an empty list, {@code null}, or {@code 0}. The graph is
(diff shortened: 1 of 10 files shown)


jfallows (approved):
Excellent. 👍

PR #2436, opened 2026-08-26
fix(binding-mqtt): preserve a deferred will payload across abort, decode error, and end

### LINKED ISSUE #2428: MQTT CONNECT with a Will message crashes the whole engine (JVM segfault), not just the connection

jfallows:
## Description

Fixes #2428

This incorporates the first part of the fix from #2429 (the `session.hasSessionWindow(headerSize)` guard, cherry-picked to stop the `IndexOutOfBoundsException`) and closes #2429, replacing it: once that guard is in place, deferring the will-message forward to a later session-window retry exposes three further bugs in how `MqttServerFactory` handles that deferred decode state, all fixed here.

`onNetworkAbort`, `onDecodeError`, and `onNetworkEnd` could each discard a will payload still sitting in `decodeSlot`, awaiting session window, instead of letting the existing session-window retry in `decodeNetwork` finish forwarding it before the connection fully closes:

- `onNetworkAbort` additionally routed through `doNetworkReset`, which unconditionally released the decode slot as a side effect — discarding the very payload the deferred branch was meant to preserve.
- `onDecodeError`'s equivalent check was too coarse: any non-empty `decodeSlot` was treated as a pending will forward, even when the slot merely held an unrelated, already-terminal decode (e.g. a rejected PUBLISH) with no session-window retry to ever complete it, silently dropping the resulting stream abort. Added `willPayloadPending()` to distinguish a genuine deferred will-message decode state from an incidentally non-empty decode slot.

Root-caused and fixed a second, deeper bug this surfaced: `onDecodeConnectWillPayload` built the will-message header from `mqttConnectPayloadRO`, a flyweight view into whatever buffer was live at CONNECT decode time. When the very first forward attempt was itself deferred (session window not yet available for even the header), the same "first round" branch reran later against `decodeSlot`'s buffer, but the flyweight still pointed at the original, possibly-reused decode buffer — corrupting the header and leaking a decode slot. Fixed by merging the header parse/transcode and payload forward into one incremental decode step that never advances progress past bytes it hasn't actually handed to the session stream, so every attempt — including retries — re-derives the header from whatever buffer is currently live, with `decodeNetwork`'s own decode-slot buffering keeping the source bytes available for as long as needed, regardless of will payload size.

Added a k3po `read option zilla:window N` verb (mirroring the existing `zilla:ack` option) to explicitly re-grant stream window credit mid-script, needed to deterministically force the abort-while-deferred race in the new regression test.

## Test plan
- [x] New k3po scenario `session.will.message.abort.while.deferred` (network + application, both directions) fails pre-fix with the deferred payload silently discarded, passes post-fix, including exact-content verification of the forwarded remainder
- [x] Full regression suite green on `runtime/binding-mqtt`, `runtime/binding-mqtt-kafka`, `specs/binding-mqtt.spec`
- [x] `checkstyle:check` clean on all touched modules

---
_Generated by [Claude Code](https://claude.ai/code/session_01WThie5Fa17zE939WMNZPPq)_

### DIFF
--- a/runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
+++ b/runtime/binding-mqtt-kafka/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionFactory.java
@@ -1172,6 +1172,10 @@ public class MqttKafkaSessionFactory implements MqttKafkaStreamFactory
             }
             else
             {
+                if (isSetWillFlag(sessionFlags))
+                {
+                    session.doKafkaBeginIfNecessary(traceId, authorization, affinity);
+                }
                 openMetaStreams(traceId, authorization);
             }
         }
@@ -1423,6 +1427,7 @@ public class MqttKafkaSessionFactory implements MqttKafkaStreamFactory
                 }).build();
 
             doMqttBegin(traceId, authorization, 0, mqttBeginEx);
+            session.doKafkaEnd(traceId, authorization);
             session = new KafkaSessionStateProxy(routedId, resolvedId, this);
             session.doKafkaBeginIfNecessary(traceId, authorization, 0);
         }
@@ -4839,7 +4844,7 @@ public class MqttKafkaSessionFactory implements MqttKafkaStreamFactory
             kafkaBeginExRW.wrap(writeBuffer, BeginFW.FIELD_OFFSET_EXTENSION, writeBuffer.capacity())
                 .typeId(kafkaTypeId)
                 .merged(m ->
-                    m.capabilities(c -> c.set(KafkaCapabilities.FETCH_ONLY))
+                    m.capabilities(c -> c.set(KafkaCapabilities.PRODUCE_AND_FETCH))
                         .topic(topic)
                         .partitionsItem(p ->
                             p.partitionId(KafkaOffsetType.HISTORICAL.value())
--- a/runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
+++ b/runtime/binding-mqtt-kafka/src/test/java/io/aklivity/zilla/runtime/binding/mqtt/kafka/internal/stream/MqttKafkaSessionProxyIT.java
@@ -277,6 +277,17 @@ public class MqttKafkaSessionProxyIT
 
     @Test
     @Configuration("proxy.yaml")
+    @Configure(name = WILL_AVAILABLE_NAME, value = "false")
+    @Specification({
+        "${mqtt}/session.will.message.qos2.abort.before.session.established/client",
+        "${kafka}/session.will.message.qos2.abort.before.session.established/server"})
+    public void shouldSendWillSignalOnQos2AbortBeforeSessionEstablished() throws Exception
+    {
+        k3po.finish();
+    }
+
+    @Test
+    @Configuration("proxy.yaml")
     @Configure(name = PUBLISH_MAX_QOS_NAME, value = "1")
     @Specification({
         "${mqtt}/session.will.message.abort.deliver.will.retain/client",
--- a/runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
+++ b/runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
@@ -1226,12 +1226,8 @@ public final class MqttServerFactory implements MqttStreamFactory
         final int offset,
         final int limit)
     {
-        int progress = offset;
+        int progress = server.onDecodeConnectWillMessage(traceId, authorization, buffer, offset, limit);
 
-        if (server.willPayloadDeferred == 0)
-        {
-            progress = server.onDecodeConnectWillMessage(traceId, authorization, buffer, progress, limit);
-        }
         server.decodableRemainingBytes -= progress - offset;
 
         return progress;
@@ -2828,11 +2824,23 @@ public final class MqttServerFactory implements MqttStreamFactory
             final long traceId = abort.traceId();
             final long authorization = abort.authorization();
 
-            state = MqttState.closeInitial(state);
+            if (decodeSlot == NO_SLOT)
+            {
+                state = MqttState.closeInitial(state);
 
-            cleanupDecodeSlot();
+                cleanupNetwork(traceId, authorization);
+            }
+            else
+            {
+                // a decode is still deferred (e.g. awaiting session window for a will payload);
+                // mark closing and let it complete via the session window retry in decodeNetwork,
+                // rather than discarding the deferred payload along with the dead network -
+                // doNetworkReset is skipped here since it would cleanupDecodeSlot() as a side effect
+                state = MqttState.closingInitial(state);
 
-            cleanupNetwork(traceId, authorization);
+                doReleaseOwnership(traceId);
+                doNetworkAbort(traceId, authorization);
+            }
         }
 
         private void onNetworkWindow(
@@ -3503,7 +3511,7 @@ public final class MqttServerFactory implements MqttStreamFactory
             decode:
             {
                 final MqttConnectPayload payload = mqttConnectPayloadRO.reset();
-                int connectPayloadLimit = payload.decode(buffer, progress, limit, connectFlags, version);
+                final int connectPayloadLimit = payload.decode(buffer, progress, limit, connectFlags, version);
 
                 willPayloadBytes = payload.payloadSize;
                 final boolean willFlagSet = isSetWillFlag(connectFlags);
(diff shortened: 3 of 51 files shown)


jfallows:
`testing (mcp.proxy)` failed on this PR's head with `cache never finished hydrating -- every cache-backed assertion below would just be racing this`. This is not this PR's failure: the identical assertion fails the same way on `develop`'s own latest push ([run 32990421615](https://github.com/aklivity/zilla/actions/runs/32990421615), same failing job) — the `mcp.proxy` example this checks is unrelated to the `binding-mqtt`/`runtime-engine`/`specs/binding-mqtt.spec` files this PR touches, and no fix for it exists yet. Re-ran the failed job once; watching for the result.

---

PR #2441, opened 2026-08-27
docs(readme): fix release badge to use GitHub releases, not raw tags

jfallows:
## Description

The README's "release" badge used `github/v/tag/aklivity/zilla`, which reads from raw git tags. This repo's git-flow release process tags every version, including in-progress release candidates (e.g. `2.4.0-rc1` through `2.4.0-rc20a`), so the badge could show an unreleased rc as the "release" instead of the latest finished version.

Switched to `github/v/release/aklivity/zilla`, which reads from actual GitHub Release objects. This repo's release workflow already skips creating a Release for pre-release versions (confirmed: the repo's Releases list contains only real `x.y.z` versions, no rc's), so this endpoint structurally can't surface an unfinished rc. Also pinned `sort=semver` explicitly, since this repo releases patches on `support/N.x` branches independently of `develop` — a newer-dated patch release on an older support branch should not outrank a higher-versioned release on the mainline.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TPozwovLazfMf5d1ei5MmL)_


jfallows (commented):
LGTM

PR #2448, opened 2026-08-28
fix(examples/mcp.proxy): send a key with kafka__produce_message in verify.sh

jfallows:
## Description

`verify.sh`'s `kafka__produce_message` call sent a record with no key. A compacted Kafka topic always rejects a null-key record — and `verify.sh` itself later calls `kafka__alter_topic_configs` to set `orders`' `cleanup.policy` to `compact`. That means a rerun of `verify.sh` against a stack whose broker was left over from an earlier run (rather than a fresh one) finds `orders` already compacted from that earlier run, and the keyless produce call — which passes cleanly on a fresh topic — starts failing purely as an artifact of reusing the stack across runs.

Same fix as [aklivity/zilla-plus#1183](https://github.com/aklivity/zilla-plus/pull/1183) for this repo's own copy of the example.

## Fix

Send a key (`"key":"order-1"`) with the produce call, so it stays valid regardless of the topic's prior compaction state.

---


jfallows (commented):
LGTM

PR #2451, opened 2026-08-29
fix(binding-mcp): stream tool-call arguments, don't buffer them whole

jfallows:
## Description

`binding-mcp`'s MCP proxy had two independent code paths that buffered an entire request body into an unbounded, growable heap buffer (`ExpandableDirectByteBufferEx`) before doing anything with it — defeating the decode/encode-slot + WINDOW-based flow control the rest of the codebase relies on to bound memory use regardless of payload size, and in both cases also causing the caller's flow-control window to never be re-granted past a single initial grant, stalling any request whose body exceeded one buffer slot's capacity.

**`zilla__execute_tool`** (the meta-tool that unwraps a caller's request and dispatches a synthetic call to the resolved target tool) buffered the whole caller request and the whole reconstructed delegate body before forwarding. Replaced with `McpExecuteArgsTransform`, a `JsonPipeline`/`JsonTransform`-driven streaming unwrap over real `BufferPool` decode/encode slots. The delegate's synthetic `BEGIN` needs an exact `contentLength` before the target's own `arguments` value is fully known; this is resolved via a computed upper bound (the caller's own declared outer `contentLength`, which is always a strict superset of the nested `arguments` span), padded with trailing ASCII spaces (valid per RFC 8259's `ws` grammar) once the real length is known, with a runtime assertion that fails loudly (RESET) rather than corrupting the wire if that bound is ever violated. Byte-preserving delivery of the `arguments` span was tried and dropped after finding a `common-json` limitation (segment/verbatim resume silently drops content once a segmented run spans more than one input window) — structured event forwarding with canonical re-rendering is used instead, which the padding scheme already treats as the expected case.

**The real (non-meta-tool) `tools/call` proxy** had the identical defect in its JSON-schema-validation branch: a schema-validated call's entire body was buffered before a single, one-shot validation call, and its flow-control window was granted only once. Replaced with `runtime/common-json`'s `JsonSchema`/`JsonPipeline` used directly (not the generic `ModelPipeline`/`ModelHandler` engine SPI — MCP tool schemas are always JSON Schema, so the pluggable-model-type indirection wasn't buying anything and would have forced an awkward event-contract ↔ byte-range-contract bridge). `JsonSchema.of(schemaText)` compiles once and is safely cached by schema id (it's immutable, unlike the `ModelPipeline` it replaces, which held per-call state and was unsafe to share across concurrent streams the way it had been cached). A new small `JsonTransform` re-roots the `params` object to its `arguments` member per stream, driven incrementally per DATA frame through a bounded decode slot. Each frame's bytes are still forwarded to the delegate immediately, gated only on that same frame's own validation verdict, so a violation confined to one frame never reaches the real backend tool — a violation only detectable in a later frame of a value too large to buffer whole is the documented emit-then-abort trade-off `JsonSchema.validator()` itself commits to.

Both fixes add k3po coverage for 10k/100k-byte fragmented bodies (previously only single-frame bodies were exercised for either path), the execute_tool padding path, the tools/call default-arguments-omitted fallback, and independent concurrent streams sharing one cached compiled schema — none of which existed before.

Fixes # (no tracking issue)

---
_Generated by [Claude Code](https://claude.ai/code/session_01FsHXnMy7mD1iVeGZKQivjW)_


jfallows (commented):
LGTM

PR #2457, opened 2026-08-30
perf(binding-kafka): claim a block in the log file for produce-path trailer entries

### LINKED ISSUE #2456: binding-kafka: avoid per-stream heap accumulation buffer for produce-path trailer entries
## Problem

The kafka cache storage layer's producer encode path needs to accumulate a variable number of header/trailer entries produced in the course of encoding a single message, then materialize them as real Kafka headers once encoding completes. Today this accumulation happens in a per-stream, heap-resident growable buffer (an expandable array plus parallel offset/length arrays) that is retained for the entire lifetime of the producing stream, growing to whatever high-water mark that stream's own trailer volume reaches and never released.

This is a self-imposed limitation of the current implementation, not a constraint of the storage format: the same storage layer already solves an equivalent problem — accumulating a variable amount of transform-produced output for a single message, without knowing its final size in advance — for the *value* itself, and does so with zero retained per-stream heap state.

## Existing pattern (value's own transformed output)

`KafkaCachePartition` already implements a **claim-per-message** protocol for the transformed value:

- At message start, a worst-case size bound is computed (`pipeline.padding(...)`) and a block of that size is reserved in a dedicated per-segment file (`convertedFile`, distinct from the main key+value+headers log file) by advancing the file's write position past the reservation immediately (`KafkaCachePartition.java` ~lines 424-431, 757-764).
- As the transform produces output, bytes are written **directly** into that already-claimed, non-overlapping block — no heap buffer of any kind is used for the value's transformed bytes; the write callback (`consumeTransformed`, ~lines 592-616) tracks only a couple of file positions.
- Because each message's block is reserved up front and never overlaps another message's block, multiple messages can have claims outstanding at the same time — each writes into its own region regardless of interleaving with other in-flight messages on other streams — while claim order still determines the blocks' final layout.
- The worst-case bound this depends on is not new or exotic: a statically configured max already exists for header/trailer bytes today (`cacheClientTrailersSizeMax`), currently used to size the shared log file's segment-rollover calculation.

## Proposed change

Apply the identical claim-per-message pattern to trailer-entry accumulation:

- Add a new dedicated per-segment file (sibling to `logFile`/`convertedFile`/`deltaFile` in `KafkaCacheSegment`, same `config.segmentBytes` capacity, same file-naming convention in `KafkaCacheFile`).
- At message start, claim a block in that file sized from the existing `cacheClientTrailersSizeMax` bound (mirroring the `convertedFile` claim already made from `pipeline.padding(...)`).
- Replace the current heap-resident accumulator's storage with direct writes into the claimed block, tracked by a claimed-position and write-cursor pair per stream (analogous to the stream's existing `entryMark`/`valueMark` position-tracking fields) instead of a growable array.
- Read-back operations (count-by-name, get-by-name-and-index) scan the claimed block directly instead of a heap arena — same comparison logic, different backing buffer.

## Benefit

Eliminates the only remaining per-stream retained heap allocation on this path: today every producing stream keeps a heap buffer alive for its whole lifetime, sized to its own steady-state trailer volume. With this change, no heap buffer is retained per stream at all — only two small position fields, with actual storage backed by the same mmap'd segment file the cache already pays for.

## Scope

Confined to `runtime/binding-kafka`'s cache storage layer (`KafkaCacheFile`, `KafkaCacheSegment`, `KafkaCachePartition`, and the produce-path trailer accumulator). Per this repo's contribution workflow for significant behavior/format changes, filing this issue to discuss the design before implementation; a follow-up PR would need new k3po coverage for the claim/reservation and overflow-abort behavior, mirroring the existing coverage for `convertedFile`.

jfallows:
## Description

`KafkaCacheTrailerEnvelope` accumulated a composed transform's `set()` calls during produce-path encode in a per-stream, heap-resident growable arena (an expandable buffer plus parallel offset/length arrays), retained for the entire lifetime of the producing stream.

The storage layer already solves an equivalent problem for a transformed *value's* own output with zero retained per-stream heap state: `KafkaCachePartition` reserves a worst-case-sized block once per message in `convertedFile`, and the transform writes directly into that already-claimed block via `consumeTransformed` -- no heap buffer at all, just a couple of file positions tracked.

This applies the same claim-per-message approach to trailer entries, but without introducing a new file: the log file already reserves a `cache.client.trailers.size.max`-sized block per message for the final merged trailers (committed at FIN). This adds a second, separate claim of the same size, right next to it, for the envelope's own accumulation during encode. `KafkaCacheTrailerEnvelope` now writes `set()` calls directly into that claimed block via a small write cursor, instead of a growable arena -- `count()`/`get()`/`writeHeaders()` scan the claimed block directly. If a message's entries exceed the claimed block, the excess is dropped and the entry is marked aborted (mirroring how a transformed value exceeding its own reservation is handled) rather than corrupting anything or rejecting the whole record.

### Testing

- `KafkaCacheTrailerEnvelopeTest` rewritten test-first for the claimed-slot design (against a real `KafkaCacheFile` backed by a temp file), including new coverage for the overflow/abort case and for `set()` calls made before any claim.
- New k3po scenario (`message.value.envelope.overflow`, `CacheProduceIT#shouldSendMessageValueEnvelopeOverflow`) configuring `cache.client.trailers.size.max` small enough to overflow on the second of two configured fields, proving the entry becomes invisible to the downstream relay (aborted) rather than corrupting or partially leaking data.
- The existing `message.value.envelope*` scenarios (#2455) continue to pass unmodified -- only the envelope's internal storage changed, not its observable contract.
- Full `runtime/binding-kafka` module (`clean verify`): all unit tests, all k3po ITs, checkstyle, license, and jacoco coverage pass with no regressions.

Fixes #2456

---

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/cache/KafkaCachePartition.java
@@ -725,6 +725,7 @@ public final class KafkaCachePartition
         MutableInteger entryMark,
         MutableInteger valueMark,
         MutableInteger valueLimit,
+        MutableInteger trailersClaimMark,
         long timestamp,
         long ownerId,
         long producerId,
@@ -839,6 +840,13 @@ public final class KafkaCachePartition
             logFile.writeBytes(trailersAt, EMPTY_TRAILERS); // needed for incomplete tryWrap
             logFile.writeInt(trailersAt + SIZEOF_EMPTY_TRAILERS, trailersSizeMax - SIZEOF_EMPTY_TRAILERS);
 
+            // a second, separate claim: a block for a composed transform's envelope.set() calls during
+            // encode, written directly rather than staged in a heap-resident arena; consumed via
+            // KafkaCacheTrailerEnvelope.writeHeaders(...) before the claim above is overwritten with the
+            // final merged trailers
+            trailersClaimMark.value = logFile.capacity();
+            logFile.advance(trailersClaimMark.value + trailersSizeMax);
+
             final long offsetDelta = (int)(progress - segment.baseOffset());
             final long indexEntry = (offsetDelta << 32) | entryMark.value;
             assert indexFile.available() >= Long.BYTES;
@@ -908,7 +916,8 @@ public final class KafkaCachePartition
         MutableInteger entryMark,
         MutableInteger valueLimit,
         long acknowledge,
-        Array32FW<KafkaHeaderFW> trailers)
+        Array32FW<KafkaHeaderFW> trailers,
+        boolean trailersOverflowed)
     {
         final KafkaCacheSegment segment = head.segment;
         assert segment != null;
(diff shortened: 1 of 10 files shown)


jfallows (commented):
LGTM

PR #2466, opened 2026-09-01
feat(engine): resolve named vault references inside a vault's own options

jfallows:
## Description

A binding's `options` can already name another engine resource (e.g. a vault) and have the engine resolve that name to an id before attach — `EngineManager` walks `binding.options.refs()` (a list of `NamedConfig` entries) once per namespace and fills in each `ref.id`. A vault's own `options` had no equivalent: `EngineManager` resolved a vault's own `name` to `id`, but never walked `vault.options.refs()`, so a vault whose own options need to reference another named vault in the same namespace had no way to get that reference resolved to an id.

This extends the existing per-namespace vault loop to also resolve `vault.options.refs()`, mirroring the `binding.options.refs()` walk exactly — no new engine SPI surface, just closing a gap in behavior that already exists for bindings.

To prove the fix end-to-end without any new production module, this also extends the engine's own `TestVault` test implementation (used only by `specs/engine.spec`'s `EngineIT`) with an optional `vault: <name>` option, backed by the existing `VaultedConfig`/`NamedConfig` type built for exactly this purpose. `TestVaultHandler` falls back to the referenced vault's `initKeys`/`initTrust`/`initSecretKeys` when it has no matching material of its own, and a new `EngineIT` scenario (`server.vault.options.delegate.yaml`) configures one `type: test` vault fronting another and verifies the reference resolves and the fronted vault's material is used.

### Testing

- `./mvnw verify -pl config/engine.conf` — unit tests, including the pre-existing `TestVaultHandlerTest` (unaffected, since the new `EngineContext` constructor parameter is additive)
- `./mvnw verify -pl runtime/engine` — unit + integration tests, including the new `EngineIT#shouldResolveVaultReferencedByAnotherVaultsOptions`
- `./mvnw verify -pl specs/engine.spec` — schema validation and spec-level tests

Fixes # (N/A — no tracking issue)

---
_Generated by [Claude Code](https://claude.ai/code/session_013yHEVwq9N53Pwsd1NwueAG)_

### DIFF
--- a/config/engine.conf/src/test/java/io/aklivity/zilla/config/engine/test/internal/vault/config/TestVaultOptionsConfig.java
+++ b/config/engine.conf/src/test/java/io/aklivity/zilla/config/engine/test/internal/vault/config/TestVaultOptionsConfig.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.function.Function;
 
 import io.aklivity.zilla.config.engine.OptionsConfig;
+import io.aklivity.zilla.config.engine.VaultedConfig;
 
 public final class TestVaultOptionsConfig extends OptionsConfig
 {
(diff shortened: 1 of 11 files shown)


jfallows (commented):
LGTM

Issue #2522, opened 2026-09-03
Will message continuation read is not bounded by the declared payload length


sfr-oc:
### Summary

When a will message must be split across frames because the session stream's window cannot hold it in one write, `onDecodeConnectWillPayload`'s continuation branch wraps the network buffer at the raw decode limit instead of bounding it to the will-payload bytes still outstanding. Any bytes that happen to sit behind the will payload in the same read — a pipelined control packet, for instance — are forwarded downstream as if they were part of it.

### Detail

The result is a session-state record longer than its own declared deferred length, which Kafka rejects as `CORRUPT_MESSAGE`, and the network decoder's progress tracking desynchronizes.

The first-chunk branch is already bounded by available network bytes, the session window, and what remains of the declared payload; the continuation branch should be bounded the same way. The first-chunk branch additionally uses the wrong accessor as the copy length, a latent out-of-bounds read.

### Version

Reproduced on `develop` (2.4.0-rc28).

Issue #2521, opened 2026-09-03
PUBLISH payload decode reserves padding after the publish stream window is exhausted


sfr-oc:
### Summary

`MqttServerFactory.decodePublishPayload` floors the reserved size at `initialPad`, so once the publish stream's window can no longer hold even the padding it keeps emitting zero-length DATA frames that still reserve `initialPad` bytes. Each one advances `initialSeq` without making decode progress, so every subsequent `WINDOW` re-enters the same path and burns another padding, walking `initialSeq` past `initialAck + initialMax`.

### Detail

```java
int reservedMax = Math.max(publisher.initialPad, Math.min(lengthMax + publisher.initialPad, initialBudget));
```

`MqttPublishStream.initialBudget()` returns `initialMax - (initialSeq - initialAck)` and does not subtract padding, unlike `MqttSessionStream.initialBudget()`. A frame reserves payload plus padding, so the correct payload cap is `initialBudget - initialPad`; when that is negative there is no room for a frame at all.

Downstream then observes `sequence > acknowledge + maximum` and resets the stream. The reset propagates to `onDecodeError` with a reason code the MQTT 3.1.1 CONNACK encoding cannot carry, so `doNetworkEnd` tears the connection down with nothing sent to the client — the client just sees the connection disappear.

The threshold is bytes, not messages: it trips once cumulative payload plus padding exhausts the granted window, so the message count before the cutoff scales inversely with payload size.

### Version

Reproduced on `develop` (2.4.0-rc28).

Issue #2520, opened 2026-09-03
Engine worker terminates when the session stream writes before its kafka stream is attached


sfr-oc:
### Summary

`MqttKafkaSessionFactory`'s session stream calls `doData` with a `kafka` receiver that can still be null, throwing a `NullPointerException` that terminates the engine worker and every connection on it.

### Detail

Reached while forwarding will-message payload during CONNECT, when the session stream has not yet completed attaching its kafka stream. Like the non-fetch flush defect, it surfaces only as `ENGINE_STOPPED` plus an `AgentTerminationException` in the log.

### Note

The remedy needs a maintainer's judgement: guarding the write avoids the crash but silently skips the frame. If the correct behaviour is to defer the write until the stream attaches, or to fail the session explicitly, please say so and the patch can follow that instead.

### Version

Reproduced on `develop` (2.4.0-rc28).

Issue #2519, opened 2026-09-03
Engine worker terminates on a non-fetch merged flush on the session signal stream


sfr-oc:
### Summary

`KafkaSignalStream.onKafkaFlush` reads the `fetch` member of `KafkaMergedFlushEx` without checking the union's discriminator. The signal stream both produces will/expiry signals and fetches other instances' signals on the same merged subscription, so a produce-kind flush is normal and frequent — every produce acknowledge emits one. Reading `fetch()` off a non-fetch union returns an unwrapped flyweight, and iterating its `progress` array throws a `NullPointerException` that terminates the entire engine worker.

### Detail

```java
final Array32FW<KafkaOffsetFW> progress = kafkaMergedFlushEx != null ? kafkaMergedFlushEx.fetch().progress() : null;
```

The outer union is already guarded one line above (`kafkaFlushEx.kind() == KafkaDataExFW.KIND_MERGED`); the inner one is not. `MqttKafkaPublishFactory` and `MqttKafkaSubscribeFactory` guard the equivalent reads on `kind()`.

Failure surfaces only as `ENGINE_STOPPED` unless the log is grepped for `Exception`:

```
AgentTerminationException: java.lang.NullPointerException: Cannot invoke
"DirectBufferEx.getInt(int, ByteOrder)" because the return value of
"Array32FW.buffer()" is null
```

### Version

Reproduced on `develop` (2.4.0-rc28). Introduced by `cb874f5cda` ("Adapt to consumer group changes"), when the flush extension became a union.

PR #2523, opened 2026-09-03
fix(binding-mqtt, binding-mqtt-kafka, binding-kafka): flow control and null-safety defects on the MQTT publish path

sfr-oc:
### Summary

An MQTT client publishing small QoS 0 messages through an `mqtt-kafka` proxy to a multi-partition Kafka topic stops being served after roughly 60 messages. Zilla closes the connection itself, with no exception and no event in the log — the client only observes that the connection disappeared. Chasing that symptom surfaced seven distinct defects across three bindings; they are reported individually and fixed here in one commit each, so any subset can be taken independently.

One of the seven, #2522, has meanwhile been fixed upstream in #2534 and is no longer part of this branch, which has been rebased onto current `develop`.

The reported message count varies with payload size and socket buffering, which is why the same defect was observed at 42, 61 and 239 messages before its cause was understood.

### Commits

| Commit | Issue | Module |
| --- | --- | --- |
| guard `doKafkaData` against unattached kafka stream | #2520 | `binding-mqtt-kafka` |
| guard `KafkaSignalStream` flush against non-fetch merged kind | #2519 | `binding-mqtt-kafka` |
| stop reserving padding once the publish stream's window is exhausted | #2521 | `binding-mqtt` |
| emit a window when a produce flush is acknowledged | #2517 | `binding-kafka` |
| account for the retained stream when granting the publish window | #2518 | `binding-mqtt-kafka` |
| derive the merged produce window from the least budget offered | #2516 | `binding-kafka` |
| cover the merged produce window with an idle partition | — | test |

### The primary defect (#2516)

`KafkaMergedFactory.doMergedInitialWindow` aggregates produce partitions with two independent reductions — the largest unacknowledged byte count, and the smallest maximum. A partition carrying traffic holds its acknowledge and grows its maximum, while idle partitions keep the maximum they were opened with, so the two reductions select different partitions and pair values that describe neither one's budget. The merged window then shrinks by every byte written until it reaches zero. With a single partition both reductions select the same stream, which is why this only appears on topics with more than one partition.

### Verification

`merged.produce.message.values.partition.idle` opens two produce partitions, sends all traffic to one and leaves the other idle. Without the fix it stalls at the idle partition's maximum:

```
CacheMergedIT.shouldProduceMergedMessageValuesPartitionIdle
  org.junit.runners.model.TestTimedOutException: test timed out after 10 seconds
```

Also reproduced end to end against a live engine (`examples/mqtt.kafka.proxy`, Kafka 4.1.1, 5-partition topic):

| | published | forwarded |
| --- | --- | --- |
| before | 200 | **60** |
| after | 600 | **600** |

Regression across the touched modules and their spec projects — `binding-kafka`, `binding-mqtt-kafka`, `binding-mqtt` and each `.spec` — passes with 0 failures, checkstyle and license checks clean.

An earlier revision of #2516 broke `CacheMergedIT`/`ClientMergedIT` `shouldProduceMergedMessageValue100k`; the committed form is arithmetically identical to the previous computation for a single partition, and both tests pass.

### Test coverage — please read

Coverage is uneven and I would rather say so up front:

| Issue | k3po coverage |
| --- | --- |
| #2519, #2521, #2516 | new scenarios, fail without the fix |
| #2518 | new scenario, but it passes with and without the fix — the transport acknowledges a flush's `reserved` immediately and only ever writes `reserved=0` flushes, so the condition cannot be constructed |
| #2520, #2517 | none |

### Status

Verified against the reporting environment's Kafka test cluster: 1000 messages published without a disconnect, where the same client previously stalled. **Not yet verified in that environment's production deployment.**

Given how much of this was found by inspection rather than by a failing test, independent verification before merging would be welcome — in particular #2520 and #2518, where the remedy involves a judgement call rather than a mechanical correction. #2520 guards a write that would otherwise throw; if the intended behaviour is to defer it until the stream attaches, or to fail the session explicitly, the patch should follow that instead.

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientProduceFactory.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientProduceFactory.java
@@ -1456,10 +1456,8 @@ public final class KafkaCacheClientProduceFactory implements BindingHandler
                 }
             }
 
-            initialAck += reserved;
-
-            final int noAck = (int) (initialSeq - initialAck);
-            doClientInitialWindow(traceId, noAck, initialBudgetMax);
+            final int noAck = (int) (initialSeq - initialAck - reserved);
+            doClientInitialWindow(traceId, noAck, Math.max(initialMax, initialBudgetMax));
         }
 
         private void onClientInitialEnd(
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaMergedFactory.java
@@ -1798,16 +1798,21 @@ public final class KafkaMergedFactory implements BindingHandler
                 initialMaxRW.value = Integer.MAX_VALUE;
                 produceStreams.forEach(p -> initialNoAckRW.value = Math.max(p.initialNoAck(), initialNoAckRW.value));
                 produceStreams.forEach(p -> initialPadRW.value = Math.max(p.initialPad, initialPadRW.value));
-                produceStreams.forEach(p -> initialMaxRW.value = Math.min(p.initialMax, initialMaxRW.value));
-
-                if (producer != null)
-                {
-                    initialMaxRW.value = Math.max(producer.initialMax, initialMaxRW.value);
-                }
+                // a partition carrying the traffic holds its acknowledge and grows its maximum, while
+                // idle partitions keep the maximum they started with, so minimizing the maximum here
+                // would pair the busiest partition's unacknowledged bytes with an idle partition's
+                // maximum and shrink this window to nothing; minimize the budget each one offers
+                produceStreams.forEach(p -> initialMaxRW.value =
+                    Math.min(p.initialMax - p.initialNoAck(), initialMaxRW.value));
 
                 maxInitialNoAck = initialNoAckRW.value;
                 maxInitialPad = initialPadRW.value;
-                minInitialMax = initialMaxRW.value;
+                minInitialMax = initialMaxRW.value + maxInitialNoAck;
+
+                if (producer != null)
+                {
+                    minInitialMax = Math.max(producer.initialMax, minInitialMax);
+                }
             }
 
             final long newInitialAck = Math.max(initialSeq - maxInitialNoAck, initialAck);
--- a/runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/CacheMergedIT.java
+++ b/runtime/binding-kafka/src/test/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/CacheMergedIT.java
@@ -504,6 +504,16 @@ public class CacheMergedIT
     @Test
     @Configuration("cache.options.merged.yaml")
     @Specification({
+        "${app}/merged.produce.message.values.partition.idle/client",
+        "${app}/unmerged.produce.message.values.partition.idle/server"})
+    public void shouldProduceMergedMessageValuesPartitionIdle() throws Exception
+    {
+        k3po.finish();
+    }
+
+    @Test
+    @Configuration("cache.options.merged.yaml")
+    @Specification({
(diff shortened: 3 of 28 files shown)


jfallows:
@sfr-oc please provide the reproduction recipe for the externally observed issue, including the `zilla.yaml` configuration and the mqtt command line used to trigger the issue.

sfr-oc:
### Reproduction recipe

The shipped `examples/mqtt.kafka.proxy` reproduces it unchanged, with a single modification: the message topic must have **more than one partition**. With the example's default of one partition the defect does not appear.

#### 1. Topic setup

In `examples/mqtt.kafka.proxy/compose.yaml`, the `mqtt-devices` topic is created without an explicit partition count. Change that one line to create it with five:

```diff
- kafka-topics.sh --bootstrap-server $${KAFKA_BOOTSTRAP_SERVER} --create --if-not-exists --topic mqtt-devices --config cleanup.policy=compact
+ kafka-topics.sh --bootstrap-server $${KAFKA_BOOTSTRAP_SERVER} --create --if-not-exists --topic mqtt-devices --partitions 5 --config cleanup.policy=compact
```

Everything else in the example is used as-is. The relevant part of `etc/zilla.yaml` is unmodified:

```yaml
  north_mqtt_kafka_mapping:
    type: mqtt-kafka
    kind: proxy
    options:
      topics:
        sessions: mqtt-sessions
        messages: mqtt-messages
        retained: mqtt-retained
      clients:
        - place/{identity}/#
    routes:
      - when:
          - publish:
              - topic: place/+/device/#
              - topic: device/#
          - subscribe:
              - topic: place/+/device/#
              - topic: device/#
        with:
          messages: mqtt-devices
```

The binding chain is the example's: `tcp server → mqtt server → mqtt-kafka proxy → kafka cache_client → kafka cache_server → kafka client → tcp client`. No TLS is needed to reproduce.

```bash
cd examples/mqtt.kafka.proxy
docker compose up -d
```

#### 2. Client

A single MQTT 3.1.1 client publishing QoS 0 messages, no retain flag. All messages carry the same topic, so they hash to one partition and the remaining four stay idle — that is what makes the two reductions in `doMergedInitialWindow` select different partitions.

```bash
docker compose exec mosquitto-cli sh -c '
  i=0
  while [ $i -lt 600 ]; do
    echo "{\"seq\":$i,\"test\":\"payload\"}"
    sleep 0.01
    i=$((i+1))
  done | mosquitto_pub -h zilla -p 7183 -d -l \
    -t "device/loadtest/state" -q 0 -i "loadtest-client"'
```

Payload is ~30 bytes, which works out to ~542 bytes reserved per message on the produce stream.

#### 3. Observed

| | published | forwarded to Kafka |
| --- | --- | --- |
| `develop` | 200 | **60** |
| with this PR | 600 | **600** |

60 × 542 ≈ 32768, i.e. `bufferPool.slotCapacity()` — the merged window is consumed once and never replenished. The client itself reports no error at that point; `mosquitto_pub` keeps writing into socket buffers and only fails later, which is why the same defect was first reported at 42, 61 and 239 messages depending on payload size and buffering.

The collapse can be watched directly with `-Dzilla.binding.kafka.debug.produce=true` on the `zilla` service: the merged window steps down by one message's `reserved` per publish and is never raised again, while the partition carrying the traffic reports a healthy budget of its own.

#### 4. Without Docker

`CacheMergedIT.shouldProduceMergedMessageValuesPartitionIdle`, added in this PR, covers the same condition against a live engine: two produce partitions, all traffic to one, the other idle. On `develop` it fails with

```
org.junit.runners.model.TestTimedOutException: test timed out after 10 seconds
```

stalling after the 8th of eleven 1024-byte messages — 8 × 1024 = 8192, the idle partition's maximum, which `CacheMergedIT` configures via `ENGINE_BUFFER_SLOT_CAPACITY`.

#### 5. Originally reported environment

For completeness, the environment where this first surfaced additionally used TLS with mutual auth on both the MQTT listener and the Kafka client, and a three-broker cluster with a five-partition topic (replication factor 1, no under-replicated partitions). Neither TLS nor the broker count is required to reproduce — the partition count is.

jfallows:
@sfr-oc thank you for filing the issues and providing most of the fixes in your PR. 🙏 

We've split this PR into a separate PR per issue, each with test coverage and following IT scenario naming conventions.
Each PR has been merged to develop and the issues are closed, so closing this PR now as well.

The reproduction recipe now works as desired.

The fixes will be included in the next 2.x release, which is imminent.

jfallows:
@sfr-oc zilla 2.4.0 release is available now.

PR #2534, opened 2026-09-04
fix(binding-mqtt): bound will-message continuation read to the declared payload length

### LINKED ISSUE #2522: Will message continuation read is not bounded by the declared payload length

jfallows:
Fixes #2522

`onDecodeConnectWillPayload`'s continuation branch (used when a will message must be split across multiple frames because the session stream's window is too small to hold it in one write) bounded the network buffer read using the raw decode-window limit instead of `willPayloadDeferred`, the actual number of will-payload bytes still outstanding. When the decode buffer held more bytes than that (e.g. a pipelined control packet immediately behind the CONNECT), those extra bytes were forwarded downstream as if they were part of the will payload — producing a session-state record longer than its own declared deferred length, which Kafka rejects as `CORRUPT_MESSAGE`, and desynchronizing the network decoder's own progress tracking. Also fixes a latent out-of-bounds read in the first-chunk branch using the wrong accessor as the copy length.

Credit to community contributor **[@sfr-oc](https://github.com/sfr-oc)**, who diagnosed and fixed this as part of the combined #2523. This PR cherry-picks that fix's commit standalone, since #2523 bundles seven independently-scoped defects across three bindings into one PR — each deserves its own focused review, and this one is ready on its own.

### Test coverage

Includes the original commit's coverage, plus a naming-consistency rename applied here: the new scenario is `session.will.message.disconnect.while.deferred` (originally `session.will.message.deferred.trailing.packet`), matching this file's existing `<event>.while.<state>` naming template (e.g. `session.will.message.abort.while.deferred`) rather than introducing new vocabulary.

- New paired k3po scripts (`client.rpt` + `server.rpt`) in both `specs/binding-mqtt.spec`'s `application/` and `network/v5/` trees
- Spec-level self-consistency IT for each pair (`specs/binding-mqtt.spec`'s `application/SessionIT` and `network/v5/SessionIT`)
- Runtime IT against a live engine (`runtime/binding-mqtt`'s `server/v5/SessionIT#shouldDisconnectWhileDeferred`)

### Verification

- Reverted just the production fix and confirmed the IT fails for the right reason: `ComparisonFailure` on the exact-match of the deferred remainder, then `TestTimedOutException` (matching the corrupted-boundary/decoder-desync description above) — then restored the fix and confirmed all three IT variants (runtime + both spec-level pairs) pass
- Full reactor build compiles clean

https://claude.ai/code/session_015YVNaqKvEXZVzmg3HoGVnt

---
_Generated by [Claude Code](https://claude.ai/code/session_015YVNaqKvEXZVzmg3HoGVnt)_

### DIFF
--- a/runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
+++ b/runtime/binding-mqtt/src/main/java/io/aklivity/zilla/runtime/binding/mqtt/internal/stream/MqttServerFactory.java
@@ -3587,7 +3587,7 @@ public final class MqttServerFactory implements MqttStreamFactory
                 final int payloadSize = Math.min(payloadAvailable, session.initialBudget() - headerSize);
                 final OctetsFW willPayload = payloadRO.wrap(buffer, connectPayloadLimit, connectPayloadLimit + payloadSize);
 
-                willMessageBuffer.putBytes(headerSize, willPayload.buffer(), willPayload.offset(), willPayload.limit());
+                willMessageBuffer.putBytes(headerSize, willPayload.buffer(), willPayload.offset(), willPayload.sizeof());
 
                 final int deferred = willPayloadBytes - payloadSize;
                 final int dataFlags = deferred > 0 ? FLAG_INIT : FLAG_INIT | FLAG_FIN;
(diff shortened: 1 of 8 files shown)


jfallows:
The `testing (mcp.proxy)` check is failing here, but it's not this PR's failure — it's red on `develop` HEAD itself, independent of this diff:

- The `develop` push run immediately before the two most recent `binding-mcp` commits ([48a5ea77](https://github.com/aklivity/zilla/actions/runs/33805770794), "cache maven wrapper distribution...") has `mcp.proxy` passing.
- Both `develop` push runs since ([7fb719d6](https://github.com/aklivity/zilla/actions/runs/33811503514) #2528 "return HTTP 404 for unknown/expired MCP session", [d3c10613](https://github.com/aklivity/zilla/actions/runs/33811519466) #2529 "honor south exits' real capabilities in route resolution") fail at the identical step: `Start Zilla and wait for it to be healthy`, with `zilla-mcp-proxy-everything-1` never reporting healthy (`dependency failed to start: container zilla-mcp-proxy-everything-1 is unhealthy`).

This lines up with those two `binding-mcp` commits changing south-exit capability-handshake and MCP session-lookup behavior — plausibly a real regression, not flakiness — but it's unrelated to `binding-mqtt`'s will-message continuation-read fix here, and root-causing it is out of scope for this PR. No fix is being ported in since none exists yet for this failure; flagging it here rather than widening this PR's scope.

---

jfallows:
Update: root-caused and fixed the `mcp.proxy` regression in #2539 (npm-registry fetch racing the `everything` service's healthcheck window — see that PR for the full writeup; it was unrelated to `binding-mcp` despite the timing correlation I noted above). Ported the same fix into this branch and pushed — should be green now once CI re-runs.

---

PR #2536, opened 2026-09-04
fix(binding-mcp-schema-registry): support permanent delete for subject and schema version

### LINKED ISSUE #2533: kafka_sr__delete_subject tool schema has no 'permanent' parameter, so a subject can only ever be soft-deleted
Found while smoke-testing the Schema Registry MCP toolkit (Karapace backend) against both `2.4.0-rc35` and `develop-SNAPSHOT` — reproduces identically on both.

**Symptom**

`kafka_sr__delete_subject` only declares a `subject` property in its `inputSchema`. Karapace's delete-subject endpoint, like Confluent Schema Registry, performs a soft delete by default and requires `?permanent=true` on the request to actually purge the subject. There is no way to pass that flag through the current tool schema.

**Repro**

1. Register a schema under a subject (`kafka_sr__register_schema`).
2. Delete its only version (`kafka_sr__delete_schema_version`) — this alone already soft-deletes the subject as a side effect.
3. Call `kafka_sr__delete_subject({"subject": "orders-value"})`.
4. Response:
   ```json
   {"error_code":40404,"message":"Subject 'orders-value' was soft deleted.Set permanent=true to delete permanently"}
   ```
5. Retrying with an extra `"permanent": true` argument produces the identical error — the field isn't declared in the tool's schema and is stripped before reaching Karapace.
6. The subject does disappear from `kafka_sr__list_subjects` (soft-deleted subjects aren't listed), but it isn't actually purged — a later attempt to re-register the exact same subject name may run into the "already soft-deleted" state depending on backend semantics.

**Impact**

As currently exposed, there is no way to permanently delete a schema registry subject through this MCP toolkit — only ever a soft delete, with no cleanup path.

**Suggested fix**

Add `permanent` (boolean) to `kafka_sr__delete_subject`'s `inputSchema` and forward it as the corresponding query parameter on the underlying registry request.

jfallows:
## Description

The bundled Karapace OpenAPI spec's `delete_subject` and `delete_schema_version` operations only declared their path parameters, so there was no way to pass Karapace's `permanent=true` query flag through the generated MCP tool schema to purge a soft-deleted subject or schema version — the `kafka_sr__delete_subject`/`kafka_sr__delete_schema_version` tools could only ever soft-delete, with no cleanup path.

Adds an optional `permanent` boolean query parameter to both operations in `karapace-schema-registry.openapi.json`. The existing OpenAPI-driven MCP tool generation (`binding-mcp-openapi`) picks this up automatically for both the tool's `inputSchema` and the outgoing HTTP request — no Java code changes were needed.

Per this repo's test-first discipline: new k3po `.rpt` script pairs (`delete.subject.permanent`, `delete.schema.version.permanent`) were written first for both the MCP and HTTP sides, with corresponding IT methods added to `McpSchemaRegistryProxyIT`, `McpSchemaRegistryClientIT`, `McpServerIT`, and `HttpClientIT`, and the `tools/list` expected `inputSchema` strings updated — confirmed failing against the current code before the OpenAPI change, and passing after.

Fixes #2533

## Test plan

- [x] `mvn verify -pl runtime/binding-mcp-schema-registry` — full IT suite green (24 scenarios, including the 4 new `permanent`-delete ones)
- [x] `mvn verify -pl specs/binding-mcp-schema-registry.spec` — spec-level peer-to-peer verification (`McpServerIT`, `HttpClientIT`) green
- [x] `mvn checkstyle:check` / `mvn license:check` — clean

https://claude.ai/code/session_01Fr7pZUuTmie5mzpxdHvmW9

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fr7pZUuTmie5mzpxdHvmW9)_


jfallows:
The `testing (mcp.proxy)` check is failing here (`zilla-mcp-proxy-everything-1` container reports `unhealthy`, causing `docker compose up -d --wait` to abort with exit code 1: https://github.com/aklivity/zilla/actions/runs/33827260762/job/100888844137). This isn't caused by this PR's diff, which only touches `binding-mcp-schema-registry`'s bundled OpenAPI spec and its own tests.

Confirmed on `develop`'s own tip commit (`d3c106134c2`, the base this PR branches from): the same `testing (mcp.proxy)` job fails identically in `develop`'s own `build` workflow run (https://github.com/aklivity/zilla/actions/runs/33811519466/job/100841705731), so the `examples/mcp.proxy` Compose stack is already broken on `develop` before this PR's change. I don't see an open fix for it yet (checked open PRs and recent issues), so there's nothing to port in. Re-running the failed job once to rule out a one-off flake; if it fails again it confirms the pre-existing breakage and this PR stays blocked on that, not on anything in this diff.

---

PR #2537, opened 2026-09-04
fix(binding-mcp-http): treat a genuinely empty tool-call response body as JSON null

### LINKED ISSUE #2531: mcp-kafka-connect: empty-body Kafka Connect responses (202/204) get mis-encoded as a hanging SSE stream
Found while smoke-testing the Kafka Connect MCP toolkit against `develop-SNAPSHOT` (commit as of 2026-09-04), using `zilla dump` to capture the wire traffic between `north_mcp_proxy` and `south_mcp_kafka_connect_client` (backend: Kafka Connect 4.1.1 on `kafka-connect:8083`).

**Symptom**

Every Kafka Connect action whose REST response comes back with an **empty body** gets mis-encoded by the `mcp-kafka-connect` binding on the way back to the MCP client: instead of a normal `200 application/json` JSON-RPC result, the client receives `200 text/event-stream` — and the stream never actually sends a completion event or closes. From the MCP client's perspective the call just hangs indefinitely (in our case, until an outer client-side timeout eventually reported "transport dropped mid-call").

Confirmed via pcap for these tools, all of which hit Kafka Connect endpoints that return `Content-Length: 0`:
- `pause_connector` → `PUT /connectors/{name}/pause` → Connect responds `202`, `content-length: 0`
- `resume_connector` → `PUT /connectors/{name}/resume` → `202`, empty
- `restart_connector` → `POST /connectors/{name}/restart` → empty
- `restart_connector_task` → `POST /connectors/{name}/tasks/{id}/restart` → `204`, empty
- `delete_connector` → `DELETE /connectors/{name}` → `204`, empty

For every one of these, we verified via direct calls to `:8083` (bypassing the gateway) that **the action itself succeeded on the Kafka Connect side every time** — pause/resume/restart/delete all took effect correctly. Only the response delivered back to the MCP client is broken.

By contrast, `reset_connector_offsets` (`PUT /connectors/{name}/offsets` with an empty request body, response `200` with a real JSON body `{"message":"..."}`) and `describe_connector_offsets`/`describe_connector`/`describe_connector_status`/etc. (all real JSON bodies) work correctly through the gateway every time. This isolates the bug specifically to the **empty-response-body** case, not to mutating/action endpoints in general.

**Wire-level evidence (from `zilla dump`)**

```
PUT http://kafka-connect:8083/connectors/file-source-demo/pause
→ 202, content-length: 0                      (from Kafka Connect)

→ 200, content-type: text/event-stream        (from north_mcp_proxy back to client)
   ... connection stays open, no further data frames, never closes ...
```

**Repro**

1. Create any connector via `create_connector`.
2. Call `pause_connector` (or `resume_connector` / `restart_connector` / `restart_connector_task` / `delete_connector`) on it.
3. Observe the MCP client hang on that tool call indefinitely, despite the action completing successfully on the Kafka Connect worker (verify with a direct `GET /connectors/{name}/status` against port 8083).

**Suggested fix**

When the upstream Kafka Connect response has no body (regardless of status code), the `mcp-kafka-connect` binding should encode a normal `application/json` JSON-RPC success result back to the MCP client (e.g. `{}` or a small success acknowledgement), not open a `text/event-stream` response.

jfallows:
## Description

A `tools/call` response whose upstream HTTP body closes having received zero bytes (e.g. `Content-Length: 0` on a `202`/`204` action response) fed an empty window into `binding-mcp-http`'s JSON pipeline, which unconditionally rejects an empty document as invalid JSON. By the time that rejection surfaced, the reply's headers had already been flushed, so the `REJECTED` status turned into an `ABORT` instead of a completed result — leaving the MCP client with a response that opened but never finished (observed downstream as a hung `text/event-stream`).

`McpHttpProxyFactory.McpHttpProxy.responseStep` now recognizes a closing response that never received a body byte — using the caller's own `decodeSlot != NO_SLOT` bookkeeping rather than adding new tracking state — and substitutes the JSON literal `null` for the empty window. That routes the empty case through the same result/summary/envelope pipeline a real body uses, producing a normal completed JSON-RPC result instead of an aborted stream.

As a consumer whose bundled OpenAPI toolkit includes several actions with genuinely empty responses (`pause_connector`, `resume_connector`, `restart_connector`, `delete_connector`, `stop_connector`, `restart_connector_task`), `binding-mcp-kafka-connect`'s existing spec scripts for these scenarios were rewritten to model a truly empty upstream body instead of a fabricated non-empty one that had been masking this bug. That also meant dropping the now-unsatisfiable `${result.name}` summary interpolation those tools' real responses never carry, and removing `restart_connector_task`'s bundled response schema, which no longer matches its real empty body.

Verified: `binding-mcp-http` unit tests + checkstyle, `binding-mcp-kafka-connect.spec` peer-to-peer ITs, `binding-mcp-kafka-connect` full-engine ITs, and a full clean `./mvnw install` (including the `docker-image` module) all pass.

Fixes #2531

---

https://claude.ai/code/session_01UJcgjGMWb9DYwAmpGJmWJA

---
_Generated by [Claude Code](https://claude.ai/code/session_01UJcgjGMWb9DYwAmpGJmWJA)_


jfallows:
CI check `testing (mqtt.kafka.proxy)` failed on this PR's head (dc94252). This job runs the `examples/mqtt.kafka.proxy` docker-compose smoke test (mosquitto pub/sub round-trip against Zilla's MQTT↔Kafka proxy) — a codepath this PR's diff never touches (the change is confined to `binding-mcp-http`'s response handling and `binding-mcp-kafka-connect`'s bundled OpenAPI/spec scripts). The failure itself is a `mosquitto_sub` timing race (`Connection error: Connection Refused` on the first publish attempt, then an empty result from the subscriber after a fixed `sleep 1`), consistent with a flake in the smoke test's timing rather than a real regression.

I'll re-run the failed job once the workflow run finishes (GitHub currently rejects a re-run while the run is still in progress) to confirm.

---

jfallows:
CI check `testing (mcp.proxy)` also failed on this PR's head (dc94252): `container zilla-mcp-proxy-everything-1 is unhealthy` — the `examples/mcp.proxy` docker-compose stack's "everything" reference MCP server container fails its health check before the example can be exercised. This is unrelated to this PR's diff (which only touches `binding-mcp-http`'s response handling and `binding-mcp-kafka-connect`'s bundled OpenAPI/spec scripts, not the `examples/mcp.proxy` compose stack or the upstream "everything" image).

This is pre-existing on `develop`: the `build` workflow's most recent run on `develop` HEAD (d3c1061, before this PR) failed on this exact same job for the same reason (run [33811519466](https://github.com/aklivity/zilla/actions/runs/33811519466)). No fix for it exists in-tree yet, so I'm not porting anything — flagging it here rather than re-running, since it's already confirmed red on the base branch.

---

PR #2538, opened 2026-09-04
fix(binding-mcp-kafka-connect): accept arbitrary connector config fields

### LINKED ISSUE #2530: kafka-connect MCP tool schemas for validate/update connector config omit connector-specific fields
Found while smoke-testing the Kafka Connect MCP toolkit against `develop-SNAPSHOT` (commit as of 2026-09-04).

**Symptom**

`kafka_connect__validate_connector_config` and `kafka_connect__update_connector_config` both expose a tool `inputSchema` with only these properties:

```json
{"connector" (or "pluginName"): "string", "connector.class": "string", "tasks.max": "string"}
```

There is no way to pass connector-specific configuration fields (e.g. `file`, `topic`, `name` for `FileStreamSourceConnector`) through either tool — any such field either isn't accepted by the declared schema, or is silently dropped before reaching the Kafka Connect REST call.

**Repro**

1. Create a `FileStreamSourceConnector` (`file-source-demo`) with `file`, `topic`, `tasks.max`, `name` via `create_connector` (works fine — that tool's schema takes a generic `config` object).
2. Call `update_connector_config` with `{"connector":"file-source-demo","connector.class":"org.apache.kafka.connect.file.FileStreamSourceConnector","tasks.max":"2","file":"/tmp/kc-source.txt","topic":"connect-demo","name":"file-source-demo"}` (attempting to bump `tasks.max` to 2 while preserving the rest of the config).
3. Response:
   ```json
   {"error_code":400,"message":"Connector configuration is invalid and contains the following 1 error(s):\nMissing required configuration \"topic\" which has no default value.\n..."}
   ```
   Despite `topic` being included in the call arguments, it never reached the underlying `PUT /connectors/{name}/config` — proving the gateway strips fields not declared in the tool's `inputSchema` rather than passing through a generic config object.
4. Same issue reproduces with `validate_connector_config`: calling it with `pluginName`, `connector.class`, `tasks.max` for `FileStreamSourceConnector` reports `topic` and `name` as missing, even though the caller has no way to supply them via the declared schema.

**Impact**

As currently exposed, these two tools can only be used for connector types whose entire configuration fits in `connector.class` + `tasks.max` — i.e. essentially none in practice. Any real update/validate call against a connector with additional required fields will fail, and in the `update_connector_config` case this makes it look like the update itself is broken (config validation error) rather than a tool-schema gap.

**Suggested fix**

Both tools should accept a generic `config` object (as `create_connector` already does) rather than a fixed, narrow set of top-level properties, so arbitrary connector-type-specific fields can be passed through untouched.

jfallows:
## Description

`update_connector_config` and `validate_connector_config` declared `requestBody` schemas with only `connector.class`/`tasks.max` as named properties, so `binding-mcp-openapi`'s schema-driven body projector silently dropped any other connector-type-specific field (e.g. a `FileStreamSourceConnector`'s required `file`/`topic`) before it ever reached the outbound Kafka Connect REST call — making a real `update`/`validate` call fail with a "missing required configuration" error despite the caller supplying the field.

Root cause traced to two independent gaps in the shared pipeline these tools sit on top of:

1. `common-json`'s `JsonSchema#retainedPaths()` (used to shape the outbound HTTP body from a schema) only ever walked a schema's declared `properties`/`items`/combinators — `additionalProperties` was parsed and enforced for validation, but never consulted when computing what to *keep*. A schema combining named properties with a permissive `additionalProperties` had no way to express "these fields, plus anything else of this shape."
2. Once a structured body schema is opened up via `additionalProperties`, the operation's own path parameter (`connector`/`pluginName`) — which shares the same `tools/call` arguments object as the body — would otherwise pass straight through that new wildcard into the outbound body too. A pure keep-list has no way to express "deny this one path even though a sibling wildcard would otherwise retain it."

### Changes

- **`common-json`**: `JsonSchemaImpl` now adds a wildcard retained path (`/*`, the object-key counterpart to the existing array `/-` wildcard) alongside a structured schema's named properties whenever `additionalProperties` is explicitly present and not `false`; absent `additionalProperties` keeps every existing schema's closed, pruned-to-named-properties behavior unchanged. `JsonSchema` gains a new `rejectedPaths()` method (mirroring `retainedPaths()`) for paths explicitly denied (`false`, or an equivalent always-fails sub-schema) — a rejected path always wins over the same path being retained elsewhere, e.g. via a wildcard. `JsonProjectorImpl`'s trie gains a `rejected` node flag checked ahead of `keepAll`, and its fragment-decline bound no longer short-circuits to `SKIP` once a wildcard sibling means a longer key can still match.
- **`common-openapi`**: `OpenapiSchema`/`OpenapiSchemaView` gain the `additionalProperties` field itself (previously entirely absent from the model, so any OpenAPI document's `additionalProperties` was silently dropped during parsing regardless of what was written) — captured as a raw `JsonValue` and carried through to the schema text a consumer hands to `JsonSchema.of()`.
- **`binding-mcp-openapi`**: `McpOpenapiCompositeGenerator#bodySchema()` now denies every operation parameter not already a declared body property (a `false` sub-schema merged into the generated body schema's own `properties`), so an operation's path/query/header/cookie parameters can never leak through an open body schema's wildcard — verified against the existing `pulls/create` fixture, which deliberately declares a body property with the same name as one of its path parameters, to confirm that legitimate case is untouched.
- **`binding-mcp-kafka-connect`**: `update_connector_config`/`validate_connector_config`'s `requestBody` schema now keeps `connector.class`/`tasks.max` as real, documented named properties and adds `"additionalProperties": true`, matching standard OpenAPI/JSON Schema idiom (rather than the reverted first-pass workaround of dropping all named properties down to a bare open object).

All new/changed behavior is covered by unit tests in `common-json` (including explicit fragmentation tests across every input window size, and an end-to-end test reproducing the exact reject-over-wildcard scenario this issue needed) and `binding-mcp-openapi`, plus updated k3po specs in `specs/binding-mcp-kafka-connect.spec` exercising a `FileStreamSourceConnector`'s extra `file`/`topic` fields at both the MCP tool-call layer and the outbound HTTP layer.

Fixes #2530

## Test plan

- [x] `runtime/common-json` — full unit suite green, including new `additionalProperties`/wildcard/`rejectedPaths` tests (single-shot and windowed/fragmented across every input window size)
- [x] `runtime/common-openapi` — full unit suite green
- [x] `runtime/binding-mcp-openapi` — full unit + k3po suite green (56 unit + 26 IT), including a new generator test asserting the denied-parameter-name injection
- [x] `runtime/binding-mcp-kafka-connect` + `specs/binding-mcp-kafka-connect.spec` — full unit + k3po suite green (87 tests), including updated `update_connector_config`/`validate_connector_config` scenarios passing a `FileStreamSourceConnector`'s extra `file`/`topic` fields through while excluding the `connector`/`pluginName` path parameter from the outbound body

---
_Generated by [Claude Code](https://claude.ai/code/session_01CLFwU8RqaAWL8xDBXSENvr)_


jfallows:
The `testing (mcp.proxy)` check failed on the first CI run with `dependency failed to start: container zilla-mcp-proxy-everything-1 is unhealthy` — a startup timeout for the MCP reference `@modelcontextprotocol/server-everything` container (an `npx`-fetched package), unrelated to this PR's diff (which only touches `runtime/common-json`, `common-openapi`, `binding-mcp-openapi`, `binding-mcp-kafka-connect`, and their spec modules). `examples/mcp.proxy/compose.yaml` itself documents this as a known intermittent failure mode for that specific container. Re-ran just that job and it passed cleanly. All checks are green.


---

PR #2543, opened 2026-09-04
fix(binding-mcp-openapi): advertise additionalProperties in generated tool input schema

### LINKED ISSUE #2542: kafka_connect validate/update_connector_config: declared tool schema still omits connector-specific fields even though the fix landed functionally
Follow-up to #2530, tested against a newer `develop-SNAPSHOT` build where the underlying bug from #2530 is now fixed.

**What's fixed**

Passing arbitrary connector-specific fields (e.g. `file`, `topic`, `name` for `FileStreamSourceConnector`) as extra arguments to `kafka_connect__update_connector_config` and `kafka_connect__validate_connector_config` now works correctly end-to-end:
- `update_connector_config({"connector":"file-source-demo","connector.class":"...FileStreamSourceConnector","tasks.max":"2","file":"/tmp/kc-source.txt","topic":"connect-demo","name":"file-source-demo"})` actually applied `tasks.max=2` to the live connector (confirmed via `describe_connector`).
- `validate_connector_config` with the same extra fields returned `error_count: 0` with `file`/`topic`/`name` all populated and valid, instead of reporting them as missing.

So the fields are no longer silently stripped before reaching Kafka Connect — the fix for #2530 landed functionally.

**What's still wrong**

The tools' *declared* `inputSchema` was never updated to match. Both still advertise only:
```json
{"connector" (or "pluginName"): "string", "connector.class": "string", "tasks.max": "string"}
```
with no mention of a generic `config` object or any way to see that additional connector-specific properties are accepted.

**Impact**

Any MCP client that validates its arguments against the declared schema before sending a request (which is normal, expected client behavior) will refuse to pass `file`/`topic`/`name`/etc., since the schema doesn't declare them — even though the server would now accept and correctly apply them. The fix is only reachable by a client willing to send undeclared fields, which most well-behaved clients won't do. From a schema-honesty standpoint this is effectively still broken for any strict client.

**Suggested fix**

Update the `inputSchema` for both `update_connector_config` and `validate_connector_config` to reflect what they actually accept — ideally a generic `config` object (matching `create_connector`'s existing schema shape) rather than a fixed, narrow set of named properties.

jfallows:
## Description

`McpOpenapiCompositeGenerator#inputSchema()` builds the MCP `tools/list` `inputSchema` for an operation from its OpenAPI `requestBody` schema, but only ever walked the schema's named `properties` — it never consulted `additionalProperties`. This meant that even after #2538 widened `update_connector_config`/`validate_connector_config`'s `requestBody` schema to declare `"additionalProperties": true`, the *declared* MCP tool schema still only advertised a fixed, narrow set of named properties (`connector.class`, `tasks.max`), with no indication that other connector-specific fields (`file`, `topic`, `name`, etc.) are accepted.

A strict MCP client that validates its `tools/call` arguments against the declared `inputSchema` before sending a request has no way to know such extra fields are valid, since the schema doesn't declare them — even though the outbound request body projector (`bodySchema()`) already honors `additionalProperties` when shaping the actual HTTP request.

`inputSchema()` now carries a `requestBody` schema's `additionalProperties` through into the generated tool schema the same way `bodySchema()` already does, so the declared schema matches what the operation actually accepts.

This is a fix to `binding-mcp-openapi`'s own schema-generation contract — it applies to any operation routed through this binding whose `requestBody` schema combines named properties with `additionalProperties`, not just kafka-connect's operations.

Added `McpOpenapiCompositeGeneratorTest#shouldAdvertiseAdditionalPropertiesInInputSchema`, confirmed it fails without the fix (`getBoolean("additionalProperties")` finds no such key) and passes with it. Also re-ran `binding-mcp-kafka-connect`'s existing composite generator tests to confirm no regression.

Fixes #2542

## Test plan

- [x] `McpOpenapiCompositeGeneratorTest` (57 tests, `runtime/binding-mcp-openapi`) — all green, including the new regression test
- [x] Confirmed the new test fails against the pre-fix code and passes against the fix
- [x] `McpKafkaConnectCompositeGeneratorTest` (`runtime/binding-mcp-kafka-connect`) — all green
- [x] `checkstyle:check` and `license:check` clean on changed files

https://claude.ai/code/session_01XfmxFqiPhX8PAR1QXGC1Gr

---
_Generated by [Claude Code](https://claude.ai/code/session_01XfmxFqiPhX8PAR1QXGC1Gr)_


jfallows:
CI's `Build (25)` failed on `McpKafkaConnectProxyIT.shouldListTools` — that spec script hardcodes the exact expected `tools/list` JSON response, and I'd only updated the schema generator, not that expectation. Pushed 9654a08e updating `update_connector_config`/`validate_connector_config`'s expected `inputSchema` in `specs/binding-mcp-kafka-connect.spec/.../tools.list/client.rpt` to include `"additionalProperties":true`, matching the generator's new output. Verified locally: `McpKafkaConnectProxyIT`, `McpKafkaConnectClientIT`, and the spec module's peer-to-peer ITs all pass, plus checkstyle/license clean.

---

PR #2555, opened 2026-09-07
fix(binding-kafka): back off and reconnect proactively on NOT_LEADER_OR_FOLLOWER

jfallows:
## Description

Ports #2551 (merged into `support/1.x`) to `develop`.

`develop` had already diverged from `support/1.x` in this area (a value-model-transform refactor, plus produce-path reconnect scaffolding that already existed here), so this isn't a straight cherry-pick — the change was re-derived against `develop`'s current code:

- `KafkaCacheClientFetchFactory`: adds the capped-exponential backoff/reconnect machinery (didn't exist here yet) so a `NOT_LEADER_OR_FOLLOWER` reset on the cache_client fetch fanout's downstream leg reconnects internally instead of tearing down every app-facing member.
- `KafkaCacheClientMetaFactory`: notifies both the fetch and produce fanouts as soon as a meta refresh resolves a new leader, so a pending backoff can be cut short.
- `KafkaCacheClientProduceFactory`: already had backoff/retry scaffolding from an earlier develop-only change; adds the missing `reconnectAt` guard and `onLeaderReady` hook so it also benefits from the meta-driven fast path.
- Rewrites `partition.not.leader.reconnect`'s client script to reflect the new transparent behavior, and preserves the old two-hop, app-driven reconnect contract under two new scenario names (`partition.not.leader.client.reconnect`, `partition.not.leader.cache.reconnect`) with matching IT methods, mirroring the original PR's own history.

Fixes # (n/a — port of an already-merged fix)

## Test plan

- [x] `./mvnw verify -pl runtime/binding-kafka` — checkstyle, license, unit + integration tests all green (417 IT + 218 unit tests, 0 failures)
- [x] `./mvnw verify -pl specs/binding-kafka.spec` — peer-to-peer spec IT green (449 tests, 0 failures)
- [x] Confirmed a develop-only test not touched by the original PR (`CacheFetchIT.shouldReconnectPartitionNotLeaderAfterMeta`) still passes against the ported code
- [x] Diff stats (606 insertions, 41 deletions, 10 files) match the original PR exactly

https://claude.ai/code/session_01FbdHyarRhpWZ7cqGWrHAzJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01FbdHyarRhpWZ7cqGWrHAzJ)_

### DIFF
--- a/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
+++ b/runtime/binding-kafka/src/main/java/io/aklivity/zilla/runtime/binding/kafka/internal/stream/KafkaCacheClientFetchFactory.java
@@ -27,6 +27,9 @@ import static io.aklivity.zilla.runtime.binding.kafka.internal.types.KafkaOffset
 import static io.aklivity.zilla.runtime.binding.kafka.internal.types.KafkaOffsetType.LIVE;
 import static io.aklivity.zilla.runtime.engine.budget.BudgetCreditor.NO_BUDGET_ID;
 import static io.aklivity.zilla.runtime.engine.budget.BudgetDebitor.NO_DEBITOR_INDEX;
+import static io.aklivity.zilla.runtime.engine.concurrent.Signaler.NO_CANCEL_ID;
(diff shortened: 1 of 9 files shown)


jfallows (commented):
LGTM

PR #2564, opened 2026-09-10
feat(engine.conf,command-inspect): add zilla inspect schema command

### LINKED ISSUE #2562: command-inspect: add `zilla inspect schema` to output the merged config JSON Schema
## Problem

`zilla.yaml` is validated against a JSON Schema that is merged at runtime: a base `schema/engine.schema.json` plus a `*.schema.patch.json` contributed by every discovered `BindingInfo`, `CatalogInfo`, `GuardInfo`, `VaultInfo`, `ExporterInfo`, `StoreInfo`, `ModelInfo`, `ModelExtInfo`, `MetricGroupInfo`, `EmbeddingInfo`, and `BindingExtInfo` SPI implementation (`config/engine.conf`'s `EngineInfo`, applied via `jakarta.json.JsonPatch` in `EngineConfigReader.read()`).

That merge currently only happens as a side effect of engine startup (`Engine.java` constructs `EngineInfo`/`EngineConfigReader` internally through `EngineManager`). There is no supported way to obtain the fully-merged schema document on its own:

- `EngineConfigReader.logSchema()` pretty-prints it, but only into the engine's diagnostic log output during startup — it isn't a standalone, user-facing artifact.
- The only other way to get at it is to link directly against `config/engine.conf`'s internal classes and re-run the merge by hand.

Every other cross-module capability the engine assembles observably (running metrics, logs, dump, version) already has a dedicated `zilla <command>` entry point that doesn't require standing up a full engine. The merged config schema is the odd one out — it exists as an in-memory object today, but the only door to it is "start the engine."

## Proposal

Add an `inspect` command group with a `schema` subcommand: **`zilla inspect schema`**, printing the fully-merged `zilla.yaml` JSON Schema as its own output, independent of engine startup. The group shape (rather than a flat single command) is deliberate — it leaves room for future siblings (e.g. `zilla inspect config`, `zilla inspect bindings`) without a breaking CLI change.

### Module

Start this in `incubator/command-inspect`, alongside `command-dump`, `command-log`, and `command-tune` — new commands whose surface may still evolve, matching the pattern already used for CLI additions that aren't yet ready for `runtime/`.

### CLI wiring

Follows the existing Airline-based command pattern used by every `runtime/command-*` module — no changes needed to `runtime/command`'s `ZillaMain`, which already discovers commands via `ServiceLoader.load(ZillaCommandSpi.class)`:

- `ZillaInspectSchemaCommand extends ZillaCommand`, annotated `@Command(name = "schema", ...)` under an `inspect` group (mirrors how `command-version`'s `ZillaVersionCommand` / `ZillaVersionCommandSpi` pair is structured)
- `ZillaInspectCommandSpi implements ZillaCommandSpi`, registered via `provides io.aklivity.zilla.runtime.command.ZillaCommandSpi with ...` in the module's `module-info.java`

### Schema assembly — extract `EngineSchemaReader`, shared by both callers

`EngineConfigReader` today conflates two responsibilities: merging the schema (base `schema()` + each contributed `patches()` entry applied as a `JsonPatch`) and then annotating/validating the actual `zilla.yaml` config against the merged result.

Split the merge step out into a new `EngineSchemaReader` in `config/engine.conf`, whose only job is producing the merged, unannotated schema `JsonObject` from `EngineInfo`'s discovered SPI implementations. `EngineConfigReader` is refactored to depend on `EngineSchemaReader` for that step, keeping annotation and config validation as its own concern on top — no behavior change intended for existing `zilla start` validation. `ZillaInspectSchemaCommand` depends on `EngineSchemaReader` directly, since it only needs the raw merged schema and never touches the annotation/validation layer.

This removes the need to duplicate the patch-merge loop in the new command — both `EngineConfigReader` and `ZillaInspectSchemaCommand` call the same `EngineSchemaReader`.

**Also fixes: schema provider reverted from YAML back to JSON.** `EngineConfigReader.read()` currently reads/merges the schema with `JsonProvider schemaProvider = YamlJson.provider();`, and `logSchema()` likewise generates its pretty-printed text via `YamlJson.provider().createGeneratorFactory(...)`. This was an incidental behavior change introduced when `EngineConfigReader` was carved out into the `config/engine.conf` module (`7bfcec4f`, "feat(model-avro): add AvroModelExt SPI for provider-supplied pipeline stages") — unrelated to that commit's actual purpose. The schema documents themselves (`engine.schema.json`, every `*.schema.patch.json`) are authored as plain JSON, not YAML, and the intended output format — both for the existing startup schema log and for the new `zilla inspect schema` command — is JSON, not YAML. `EngineSchemaReader` uses `Json.provider()`, not `YamlJson.provider()`, for both reading/merging the schema and generating its text output; this is a straight revert of that one provider choice, scoped to the schema-only code path. `EngineConfigReader`'s separate `CONFIG_PROVIDER` (`YamlJson.provider(...)`, used to parse the actual `zilla.yaml` config documents, which genuinely are YAML) is untouched.

### `x-incubating` handling stays consistent with the engine — no CLI override

`EngineSchemaReader.stripIncubating(JsonObject)` strips `x-incubating`-marked schema nodes unless `FeatureFilter.isIncubatorEnabled()` — the exact same ambient flag `EngineConfigReader` already uses for `zilla start`'s own validation. `zilla inspect schema` calls this same method with no separate flag of its own: the emitted schema always matches what the currently-running `zilla` would actually validate against on this classpath, with this JVM's incubator setting. There's no `--incubating` command-line option — a CLI override that could disagree with the engine's own effective setting would defeat the point of the command (showing the real, currently-effective validation schema).

### Output

- Pretty-printed JSON to stdout by default (`jakarta.json.stream.JsonGenerator` with `PRETTY_PRINTING`, via `Json.provider()`)
- `-o`/`--output <path>` option to write to a file instead of stdout

No engine startup, no bind/route resolution, no network I/O — purely classpath-based schema assembly, same precondition `EngineInfo` already has today.

### Scope note

The completeness of the emitted schema depends on which `config/*.conf` modules are present on the classpath at invocation time, exactly like `zilla start`'s validation coverage depends on the assembled distribution. Running `zilla inspect schema` against the full packaged classpath (e.g. the docker image's module path) is what produces the complete merged schema; running it against a narrower classpath produces a partial merge — expected, and potentially useful on its own for isolating a single module's patch.

## Out of scope (initial)

- Filtering output to a single module's patch (`--binding http`-style)
- Schema diffing between versions/releases
- Any change to `zilla start`'s existing validation behavior

jfallows:
## Description

Adds `zilla inspect schema`, a new CLI command that prints the fully-merged `zilla.yaml` JSON Schema independent of engine startup.

**`config/engine.conf`: extract `EngineSchemaReader`, revert schema output to JSON**

- `EngineConfigReader` previously conflated merging the schema (base `engine.schema.json` + every contributed `*.schema.patch.json`) with annotating/validating the actual config against it. Extracted the merge step into a new `EngineSchemaReader`, which `EngineConfigReader` now delegates to; annotation and validation stay `EngineConfigReader`'s own concern. No behavior change to `zilla start`'s existing validation.
- Along the way, reverted the JSON provider used for that merge and for pretty-printed schema output from `YamlJson.provider()` back to the plain JSON provider (`JsonProvider.provider()`). The schema documents themselves are JSON, not YAML, and the YAML-flavored output was an incidental behavior change introduced when `EngineConfigReader` was carved into `config/engine.conf` (`7bfcec4f`, an unrelated model-avro change) — this is a straight revert of that one provider choice, scoped only to schema-document handling. `EngineConfigReader`'s separate `CONFIG_PROVIDER` (used to parse actual `zilla.yaml` config documents, which genuinely are YAML) is untouched.
- Moved the `x-incubating`-stripping unit tests from `EngineConfigReaderTest` to a new `EngineSchemaReaderTest`, matching the code they now test, and added coverage for the JSON-not-YAML output and the raw merge.

**`incubator/command-inspect`: new `zilla inspect schema` command**

- New Airline command group `inspect` with a `schema` subcommand (`ZillaInspectSchemaCommand`), leaving room for future `zilla inspect ...` siblings.
- Calls `EngineSchemaReader` directly — no engine startup, no config validation, just classpath-based schema assembly.
- `-o`/`--output <path>` to write to a file instead of stdout.
- `x-incubating` handling stays consistent with the engine: the command calls `EngineSchemaReader.stripIncubating(...)`, the same `FeatureFilter.isIncubatorEnabled()`-driven method `EngineConfigReader` already uses for `zilla start`'s own validation — no separate `--incubating` CLI flag that could disagree with the engine's actual effective setting.
- Registered `@Incubating` (per the `incubator/` convention, alongside `command-dump`/`command-log`/`command-tune`), wired into `incubator/pom.xml` and the docker-image distribution (`pom.xml` runtime dependency + `zpm.json.template`).

**`examples/inspect.schema`: new example verifying the command's exact output**

- Docker Compose stack running a normal `start`ed engine, healthchecked, with `.github/test.sh` separately `exec`ing `zilla inspect schema` into the healthy container and diffing it **byte-for-byte** against a checked-in golden file (`.github/schema.expected.json`), rather than a handful of spot checks — this fails the moment the merged schema changes anywhere in the repo. Verified the output is deterministic (identical bytes) across repeated standalone `docker run` invocations and across a `docker compose exec` invocation running alongside the unrelated `start`ed engine in the same container.
- README documents regenerating the golden file after an intentional schema change.

## Testing

- `./mvnw -pl config/engine.conf -am clean package` — checkstyle, compile, and all unit tests green, including the new `EngineSchemaReaderTest`.
- `./mvnw -pl incubator/command-inspect -am clean package` — checkstyle, compile, moditect module descriptor, and all unit tests green (`ZillaInspectCommandSpiTest` verifies the Airline group/command wiring end-to-end; `ZillaInspectSchemaCommandTest` verifies stdout output, `-o` file output, and the error path).
- Full top-level `./mvnw clean install -DskipTests` — **green**, resolving every module in the reactor (201/201) and building the `ghcr.io/aklivity/zilla:develop-SNAPSHOT` docker image via the same pipeline CI uses.
- Ran `zilla inspect schema` against that built image directly (`docker run`) and via `examples/inspect.schema`'s compose stack (`docker compose exec` alongside a running `start`ed engine) — both produce the exact checked-in golden file, confirmed deterministic across repeated runs.
- `incubator/command-inspect/NOTICE` regenerated via `./mvnw notice:generate --projects incubator/command-inspect` once the full reactor was installed (needed the reactor's own `develop-SNAPSHOT` artifacts resolvable, which a partial `-pl`-scoped build can't provide).
- Could not run a full `./mvnw verify` scoped to just the touched modules early on: the `maven-notice-plugin` license check at `verify` fails on unrelated artifacts when only a subset of the reactor is built — resolved by installing the full reactor first, as above, rather than trying to work around it in a partial build.

Fixes #2562


jfallows (commented):
LGTM

Issue #2487, opened 2026-09-02
dialect: openai — detection, content-type, request/response transforms


jfallows:
## Summary

Concrete `LlmDialect` implementation for OpenAI, registered via `LlmDialectFactorySpi`.

## Scope

- `detect(path, headers)`: `POST /v1/chat/completions` → openai, `POST /v1/completions` → openai (per design doc's dialect-detection scenarios).
- `contentType()`: `text/event-stream` (streaming) / `application/json` (non-streaming), both routed through the same content-decoder abstraction (M1's `LlmContentDecoderSpi` design explicitly answers this — no special-cased non-streaming branch).
- Request/response `JsonTransform` pairs (native OpenAI JSON ↔ canonical vocabulary) for both `Kind.REQUEST` and `Kind.RESPONSE`.
- OpenAI-specific canonical-mapping behavior per the design's worked examples (see the event sequence mapper item, M1) — `n > 1` (`choices[]`), `logprobs`, tool-call argument fragment streaming (`input_json_delta`-equivalent), `[DONE]` sentinel handling.

## Acceptance criteria

- `dialect: openai` is discoverable via `LlmDialectFactorySpi`, correctly detects OpenAI-shaped requests, and its transforms round-trip OpenAI-native JSON through the canonical vocabulary and back with no loss (same-dialect case).
- Unit tests cover detection signals and both `Kind`s of transform.

## Depends on

`LlmDialectFactorySpi` (M1), canonical event vocabulary (M1).

Issue #2486, opened 2026-09-02
llm client: dialect comparison, conditional decoder→encoder chaining


jfallows:
## Summary

`llm client` compares the inbound dialect (from `LlmBeginEx.dialect`, set by the server) against its own configured dialect, and only chains a decoder→encoder payload transform when they differ.

## Scope

- **Same dialect**: no payload transform — the client re-encodes framing only (proving the "framing always, payload only across dialects" layering rule from the client side, matching what the server already does).
- **Different dialects**: chain `LlmDialect.supplyDecoder`/`supplyEncoder` (native→canonical→native) via the event sequence mapper (M1) and configured target dialect's transforms.
- This milestone (M2) only needs the same-dialect path to actually work end-to-end (single dialect: openai) — the cross-dialect chaining logic can be implemented here structurally, but it isn't exercised by a second dialect until M3.

## Acceptance criteria

- Same-dialect (openai→openai) path: client forwards without payload transform, framing re-encoded faithfully.
- Dialect-comparison logic is structurally in place for the cross-dialect case, ready for M3 to exercise once `dialect: anthropic` exists.

## Depends on

`llm client` endpoint config (this milestone), event sequence mapper (M1), `dialect: openai` (this milestone).

PR #2576, opened 2026-09-12
test(binding-llm): k3po fixtures for real dialect:openai backend

### LINKED ISSUE #2490: llm: mock OpenAI backend (test fixture)
## Summary

A mock OpenAI-shaped backend server (streaming + non-streaming `/v1/chat/completions`) for use in `llm server`/`llm client` integration tests and (later) `examples/llm.proxy`.

## Scope

- Mirror the mock-backend pattern used elsewhere in the repo (e.g. `examples/mcp.proxy/openai/server.mjs`'s mock of OpenAI's `/v1/embeddings`, adapted for chat completions).
- Streaming responses: realistic SSE event sequences (role chunk, content chunks, tool-call argument fragments, finish_reason chunk, usage chunk, `[DONE]`).
- Non-streaming responses: plain-JSON response shape with top-level usage.
- Deterministic/scriptable enough to drive the round-trip identity test (this milestone) and later cross-dialect tests (M3).

## Acceptance criteria

- Mock server usable from both unit/IT tests and a docker-compose-style example.
- Covers both streaming and non-streaming OpenAI response shapes.

## Depends on

None (parallel-fine within M2, but ordered after dialect work so response shapes match `dialect: openai`'s expectations).

jfallows:
## Description

Closes out issue #2490 (k3po test fixtures for an OpenAI-shaped `/v1/chat/completions` backend, streaming + non-streaming, for use in `llm server`/`llm client` integration tests).

Per repo convention (`specs/AGENTS.md`), the backend side is faked entirely with paired k3po `client.rpt`/`server.rpt` scripts — no standalone mock server process and no `examples/llm.proxy` example. Existing `binding-llm` scenarios (`same.dialect`, `cross.dialect`, etc.) only exercise synthetic test dialects; these new scenarios exercise the real `dialect: openai` end-to-end for the first time:

- **`openai.request`** (network + application): `llm server` detects `dialect: openai` purely from `POST /v1/chat/completions` (no `x-llm-dialect` header needed), resolves `content-type` directly off the real request header, and forwards the request body through a `json`-model `ModelPipeline` — the dialect's decoder observes `model` as a side effect (`envelope.set("model", ...)`) while the body flows through unchanged. `LlmBeginEx.dialect`/`.contentType`/`.model` are all asserted on the app-facing stream.
- **`openai.streaming`** (network + application): `llm client` same-dialect round-trip with a `"stream":true` request and a realistic multi-chunk OpenAI SSE response — role chunk, content chunk, tool-call start and argument-fragment chunks, `finish_reason` chunk, usage chunk, `[DONE]` sentinel — each SSE `data:` line forwarded as one app-facing `DATA` + one raw `LlmFlushEx`.
- **`openai.nonstreaming`** (network + application): `llm client` same-dialect round-trip with a `"stream":false` request and a plain-JSON response carrying top-level `usage`, forwarded as one `DATA` + one terminal raw `LlmFlushEx` (matching `LlmJsonContentDecoder`'s own framing).

Each scenario is wired into `LlmServerIT`/`LlmClientIT` (real engine, via `EngineRule`) and `ApplicationIT`/`NetworkIT` (paired-script protocol self-consistency, no engine), per the required dual-IT convention. A new `client.openai.yaml` config pins `dialect: openai` for the client-kind scenarios.

No production code changed here — this is test-fixture-only, per the issue's scope. (Production changes referenced above — the `LlmDialect`/`ModelEnvelope`/`ModelTransform` SPI redesign, `LlmServerFactory`'s detect+stamp+`ModelPipeline`-forward rework, `LlmClientFactory`'s content-type-driven codec selection, and the `openai` dialect's `model` extraction — landed in the sibling PRs this branch is stacked on, prompted by a design review that started in this same session.)

### Stacking

This branch is built on top of the full `binding-llm` chain (module scaffold, content-codec SPI, dialect SPI redesign, `dialect: openai`, `llm server`/`llm client` stream factories, config validation) — so this diff includes those commits until they merge to `develop`, at which point this PR's diff will shrink to just the new k3po fixtures and IT wiring, the same pattern used by the PRs it stacks on.

## Test plan

- [x] `./mvnw checkstyle:check -pl incubator/binding-llm.spec,incubator/binding-llm` — 0 violations
- [x] `./mvnw clean package -pl incubator/binding-llm.spec,incubator/binding-llm -am` — compiles cleanly, including the four edited/new `*IT.java` classes, against the redesigned SPI
- [ ] `./mvnw verify` (the actual k3po-driven ITs) could not be run in the sandboxed dev session used to author this change — the reactor's `maven-notice-plugin` license-mapping check fails on unmapped artifacts regardless of network access (confirmed the proxy itself reaches the registry fine; the mapping data itself is what's missing). This is the same pre-existing, environment-specific limitation noted in prior `binding-llm` PRs' test plans, not something introduced by this change. CI should confirm the full `verify` lifecycle, including the new `openai.request`/`openai.streaming`/`openai.nonstreaming` scenarios.

Fixes #2490

---

https://claude.ai/code/session_01L5Pw5eDcghnK7LmZ11QA7u

---
_Generated by [Claude Code](https://claude.ai/code/session_01L5Pw5eDcghnK7LmZ11QA7u)_


jfallows:
CI failed on `Build (25)`: [run 35061039385](https://github.com/aklivity/zilla/actions/runs/35061039385/job/104681219575), head `5ca69a53`.

`LlmClientIT.shouldForwardOpenAiNonstreamingSameDialect` failed twice (failsafe's own automatic rerun), with two *different* symptoms:
- Run 1: `ComparisonFailure` — behavior didn't match what the paired scripts expected.
- Run 2: `TestTimedOut`.

The robot's captured state at the Run 2 timeout shows the network-side backend script successfully matched the outbound HTTP `BEGIN` (5 headers: `:method`/`:scheme`/`:authority`/`:path`/`content-type`), then hung waiting to read the forwarded JSON request body — i.e. `llm(client)` never forwarded the request payload for this stream.

This doesn't look like a fixture problem: `openai.streaming`'s and `openai.nonstreaming`'s request-write halves are structurally identical (same headers, same `write`/`write flush`/`write close` sequence, same content-type default), and only `shouldForwardOpenAiStreamingSameDialect` — which exercises the exact same request-forwarding path — passed. Two different failure shapes across two runs of the same test points to a race rather than a deterministic fixture mistake.

Working hypothesis: `LlmBindingConfig.supplyModel(dialect, kind)` caches one `ModelHandler` per `dialect.name() + kind`, so both scenarios request the identical cache key (`"openai" + REQUEST`). If that shared handler — or the per-schema `ModelPipeline`/`JsonPipeline` machinery underneath it in `model-json` — isn't safely reusable across independently-running streams, a second use could plausibly manifest as exactly this kind of intermittent hang/mismatch.

I can't reproduce or debug this further from here — this sandbox can't run the k3po `verify` lifecycle at all (pre-existing `maven-notice-plugin` license-mapping gap noted in this PR's own test plan), and the fix, if my hypothesis is right, is in `LlmClientFactory`/`LlmBindingConfig` (and possibly `model-json`), not in this PR's own fixture-only diff. Flagging rather than guessing at a fix to shared, concurrency-sensitive code I can't validate locally.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L5Pw5eDcghnK7LmZ11QA7u)_

---

jfallows:
Update: rebased onto `13e12c51` (picks up the `#2487` session's `LlmOpenAiDialect.detect()` content-type-scoping fix, `ed510113`) and re-ran CI — same failure, byte-for-byte identical: [run 35108176452](https://github.com/aklivity/zilla/actions/runs/35108176452/job/104834694738).

`LlmClientIT.shouldForwardOpenAiNonstreamingSameDialect` failed again with the exact same two symptoms (`ComparisonFailure` then `TestTimedOut` on retry), and the Run 2 timeout trace is character-for-character identical to the previous run: network side matches the 5-header HTTP `BEGIN`, then hangs forever on the request body read.

This rules out the dialect-detection ambiguity as the cause (that fix landed and made no difference) and rules out ordinary test-order flakiness (reproduces identically across independent CI runs with different code in between). This is a real, deterministic bug in `LlmClientFactory`'s same-dialect `application/json` request-forwarding path — still pointing at `LlmBindingConfig.supplyModel`'s per-`dialect+kind` `ModelHandler` caching (both `openai.streaming` and `openai.nonstreaming` hit the same `"openai"+REQUEST` cache key) or the `model-json` pipeline machinery underneath it, per my earlier comment.

Still can't reproduce or fix this myself (no `k3po verify` capability in this sandbox). Continuing to flag it for the `#2486` session (`LlmClientFactory`/`LlmBindingConfig`) since it's outside this PR's own fixture-only diff.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L5Pw5eDcghnK7LmZ11QA7u)_

---

jfallows:
Root cause found and fixed for the `shouldForwardOpenAiNonstreamingSameDialect` failure reported above.

`LlmClientFactory`'s constructor always built and drove a schema-validating JSON `ModelPipeline` for both request and response, even for same-dialect traffic where the design (per `LlmDialect` SPI notes) intends pure raw byte relay with no transform. The real `openai` dialect declares no JSON schema for either direction (`LlmOpenAiDialectFactorySpi.schema(Kind)` returns `null`, matching the SPI's own contract for "a dialect with nothing to contribute"), so `JsonModelHandlerImpl`'s schema resolution always falls through to `NO_SCHEMA_ID` for its subject, and the pipeline rejects every value on the first `transform()` call regardless of payload content — confirmed directly with a standalone reproduction against `JsonModelHandlerImpl`/`JsonModelDecoderPipeline` using the same no-schema + `ModelTransform.NONE` configuration `LlmClientFactory` builds for same-dialect traffic.

Fix (`45e5e520`): skip constructing `requestPipeline`/`responsePipeline` entirely when `source == target` (same dialect), forwarding request/response bytes directly through the content-type codec's own encoder/decoder for framing, without driving a model pipeline that has nothing to validate.

Verified: `./mvnw package` (compile + checkstyle + unit tests, 160/160 passing) for `binding-llm.spec`/`binding-llm` in this sandbox — the k3po `verify` lifecycle itself still can't run here (repo-wide `maven-notice-plugin` gap, unrelated to this change), so CI is the actual verification for the `LlmClientIT` scenarios this fixes.

---

jfallows:
Correction to my previous comment: the `LlmClientFactory` change (`45e5e520`) that skipped the `ModelPipeline` for same-dialect traffic was the wrong fix — it treated the symptom, not the cause, and same-dialect traffic is exactly the path that most needs schema validation (no field-rewriting transform to sanity-check structure along the way). Reverted in `75162602`.

The actual gap: `LlmOpenAiDialectFactorySpi` declared no JSON schema for either direction, so the shared `sys:llm_dialects` catalog had no `openai.request`/`openai.response` subject, and the schema-validating pipeline always resolved `NO_SCHEMA_ID` and rejected every value — there was no real validation happening for `openai` traffic at all, same-dialect or cross-dialect.

Fixed properly in `2cc26ee0`: added real `openai.request.schema.json`/`openai.response.schema.json` describing the actual Chat Completions wire shapes (request requires `model`/`messages`, types the fields `LlmOpenAiRequestTransform` recognizes; response types `choices[]`/`usage` without requiring any top-level member, since a streaming chunk only ever carries a subset of what a full completion carries at once), and wired `LlmOpenAiDialectFactorySpi.schema(Kind)` to supply them. `LlmClientFactory` is back to its original, unmodified behavior — it now gets a real schema to validate against instead of none.

Verified directly against `JsonModelHandlerImpl`/`JsonModelDecoderPipeline` (the same construction `LlmClientFactory` drives) that all nine request/response payloads used by the `openai.request`/`openai.streaming`/`openai.nonstreaming` k3po fixtures validate as `COMPLETE`, not `REJECTED`. Also added unit test coverage (`LlmOpenAiDialectTest`) asserting the factory supplies non-null, well-formed schemas for both directions. `./mvnw package` (checkstyle + 162/162 unit tests) is green; CI is the real verification for the k3po scenarios.

---

PR #2577, opened 2026-09-16
test(binding-llm): round-trip identity coverage, openai server → openai client

### LINKED ISSUE #2491: llm: round-trip identity test, openai → openai
## Summary

The test that proves the "framing decode/re-encode always, payload transform only across dialects" layering rule for the same-dialect case: `llm server` (dialect: openai) → `llm client` (dialect: openai) against the mock OpenAI backend reproduces the original stream with no payload parsed.

## Scope

- Assert framing decode and re-encode is faithful (byte-identical or defined-semantic-equivalent — resolved by the SSE content decoder item's open question in M1, e.g. field ordering / optional `id:` / whitespace slack) **and** that no payload was parsed — this is not "zero transforms" but "framing decode and re-encode is byte-identical/equivalent, and no payload was parsed."
- Cover both streaming and non-streaming request/response shapes.
- Include tool-call streaming in the round trip, since it spans both DATA and FLUSH planes.

## Acceptance criteria

- IT test (`./mvnw verify`, k3po-based per `specs/AGENTS.md` conventions where applicable) passes against the mock OpenAI backend.
- Test explicitly asserts "no payload parsed" (e.g. via an instrumentation hook or a payload shape that would only survive unparsed) — not just "output matches input," which framing-decode-then-payload-transform could also satisfy incorrectly.

## Depends on

Mock OpenAI backend, `llm` client dialect comparison (same-dialect path), `llm` server framing decode and forward (all this milestone).

jfallows:
## Description

Closes out issue #2491 (round-trip identity test proving `llm server` (dialect: openai) → `llm client` (dialect: openai) preserves framing byte-for-byte with no payload parsed, against the mock OpenAI backend fixtures from #2576/#2490).

Per this repo's testing principle of one binding under test per IT — verify each binding independently and infer combinations, rather than enumerating combined configs — this does **not** chain `llm:server` and `llm:client` together in a single engine config. An initial version of this PR did exactly that (`LlmRoundtripIT` wiring both bindings into one `roundtrip.openai.yaml`), but review caught that this duplicates what's already provable by composition:

- `NetworkIT` and `ApplicationIT` (peer-to-peer, no engine) already prove that each `openai.request`/`openai.streaming`/`openai.nonstreaming` script pair is self-consistent, on both the network and application planes.
- `LlmClientIT` already exercises `openai.streaming`/`openai.nonstreaming` against a live client engine, proving canonical-in → wire-out fidelity (including the `.raw()` flush-ext assertions that prove no payload transform ran on the same-dialect fast path).
- `LlmServerIT` only exercised `openai.request` against a live server engine — it never proved wire-in → canonical-out fidelity for the streaming/nonstreaming shapes, including the tool-call streaming chunks spanning both DATA and FLUSH planes.

Composing the already-existing peer-to-peer proofs with `LlmServerIT`'s and `LlmClientIT`'s own per-binding proofs already establishes the round-trip identity claim end-to-end, without ever running server+client chained in one config.

This PR fills the one real gap: two new methods on `LlmServerIT` — `shouldForwardOpenaiStreaming` and `shouldForwardOpenaiNonstreaming` — wiring the `openai.streaming`/`openai.nonstreaming` scripts and `server.yaml` config that already exist from #2576. No new script or config files were needed.

A symmetric `openai.request` addition to `LlmClientIT` was considered for full matrix symmetry, but `net/openai.request/server.rpt` was authored as the peer for the server-detection scenario (asserting path `/v1/chat/completions`) rather than as what `llm:client`'s own fixed-dialect encoder actually emits (path `"/"`, per the already-passing `openai.streaming`/`openai.nonstreaming` network scripts) — pairing them would fail on a script mismatch unrelated to any real defect, so `LlmClientIT` is left unchanged.

### Production fix

The two new `LlmServerIT` methods exposed a real gap in `LlmServerFactory`, fixed here directly since it's the only place the fix can be proven against these tests:

- `LlmServerFactory` never resolved a response-direction content encoder at all, so app0's reply was forwarded to the network side as a raw byte-for-byte passthrough regardless of the actual response content-type.
- The reply `BEGIN`'s `content-type` header reused the *request's* content-type unconditionally, rather than reading it from app0's own reply `LlmBeginEx`.
- There was no `FlushFW` case in the app-facing stream's message switch at all, so an SSE event boundary flushed by app0 was silently dropped.

`LlmStream.onAppBegin` now reads app0's own `LlmBeginEx` content-type and supplies a matching `LlmContentEncoder`; `onAppData` and a new `onAppFlush` route bytes/flush boundaries through that encoder before handing them to `doNetData`. This mirrors the pattern `LlmClientFactory` already uses on its own encode-direction (request → net).

Also fixed the `:path` header in the `openai.streaming`/`openai.nonstreaming` **network** fixtures (`incubator/binding-llm.spec`) — it was `"/"`, which matches neither of `LlmOpenaiDialect`'s recognized paths (`/v1/chat/completions`, `/v1/completions`), so the request never actually resolved to the `openai` dialect and the response-direction code was never reached at all.

### Also fixed: `model-json` crash on boolean/null scalar fields

Fixing the two issues above surfaced a separate, pre-existing crash: any JSON scalar field whose value is `true`, `false`, or `null` (e.g. this test's own `"stream":true`/`"stream":false`) took down the entire engine worker with a bare `AssertionError`. `JsonModelFieldTransform.onScalar()` called `source.getStringView()` unconditionally for every scalar event, but `VALUE_TRUE`/`VALUE_FALSE`/`VALUE_NULL` aren't backed by a tokenizer string view at all — the assertion in `getStringView()` only covers `VALUE_STRING`/`VALUE_NUMBER`/`KEY_NAME`. These three events have an already-known, fixed textual form, so `onScalar` now uses it directly instead of asking the source for a view it can't produce. Fixed in this PR (rather than filed separately) since a fix on `develop` would otherwise need walking up through every dependent PR branch in this stack.

## Test plan

- [x] `./mvnw checkstyle:check -pl incubator/binding-llm.spec,incubator/binding-llm,runtime/model-json` — 0 violations
- [x] `./mvnw package -pl incubator/binding-llm.spec,incubator/binding-llm -am` — compiles cleanly
- [x] `./mvnw verify -Dit.test=LlmServerIT,LlmClientIT` — all 12 `LlmServerIT` tests and all 7 `LlmClientIT` tests pass, including the new `shouldForwardOpenaiStreaming`/`shouldForwardOpenaiNonstreaming`
- [x] `./mvnw verify -pl runtime/model-json` — full unit + IT suite green, including a new regression test (`shouldForwardBooleanAndNullFieldsUnchanged`) confirmed to fail with a bare `AssertionError` against the pre-fix code and pass with the fix

Fixes #2491

---

https://claude.ai/code/session_01UogzKHEA7zC9PgHBkGwu8p


jfallows:
## CI status

**`Build (25)` — failing, not from this PR's diff.**

The failure is `LlmClientIT.shouldForwardOpenAiNonstreamingSameDialect` timing out after 10s (twice, on retry). This test is in a class this PR never touches (only `LlmServerIT` gained the two new methods here). I confirmed it's pre-existing: #2576's own `Build (25)` run, on the exact same commit (`13e12c51`) this branch forks from, fails with the byte-for-byte identical error — same test, same timeout, same mismatch trace. There's no fix to port yet since #2576 itself is unfixed on this. Once #2576 addresses it (or a fix lands there), I'll rebase and this should clear.

**`Analyze (java)` (CodeQL) — failing, looks like an autobuild flake.**

The failure is a Maven plugin resolution error in CodeQL's `autobuild` step (`Error resolving version for plugin 'io.aklivity.zilla:flyweight-maven-plugin'... Plugin not found in any plugin repository`), not a code-scanning finding. This didn't touch any pom.xml or plugin config in this PR's diff, and #2576's own CodeQL run on the identical base commit passed — so this looks like a transient resolution issue in the autobuild environment rather than a real regression. I've queued a re-run of the failed job to confirm.

---

jfallows:
#2576 landed a real fix for the `LlmClientIT.shouldForwardOpenAiNonstreamingSameDialect` timeout (root cause: the `openai` dialect declared no JSON schema for either direction, so its schema-validating pipeline always resolved `NO_SCHEMA_ID` and rejected every value — fixed in `2cc26ee0` by adding real `openai.request`/`openai.response` schemas). Rebased this branch onto that new head; checkstyle and compile are clean locally. CI is re-running now.

---

jfallows:
Rebased again onto #2576's latest head (`0b15537f`, adds `openai.request.invalid` schema-rejection coverage). One real conflict in `LlmServerIT.java` — both PRs added a new test method right after `shouldDetectOpenAiDialectFromPath` — resolved by keeping both (#2576's `shouldRejectInvalidOpenAiRequest` followed by this PR's `shouldForwardOpenAiStreaming`/`shouldForwardOpenAiNonstreaming`). Checkstyle and compile clean locally; CI re-running.

---

PR #2579, opened 2026-09-16
fix(engine): synchronize EngineBoss.detachAll() with boss thread

jfallows:
## Description

`EngineBoss.detachAll()` mutated boss-owned state (`namespaces`, attached binding controllers) directly on the caller's thread, unlike `attach()`/`detach()`, which hand off through the boss's own `AgentRunner` task queue and block until that thread has processed them. This left a gap in `EngineBoss`'s own concurrency contract:

- `detachAll()` raced with whatever the boss's own `doWork()` loop was doing concurrently (e.g. an in-flight `Poller` dispatch), instead of serializing with it the way every other namespace attach/detach path does.
- Nothing about `detachAll()` returning guarantees the boss has actually stopped accepting/dispatching by that point.

This was diagnosed from an `hs_err` core dump showing the boss's `AgentRunner` thread still `_thread_in_Java`, mid-dispatch into a worker's inbound ring buffer, at a point where every per-core worker thread had already exited and unmapped its own buffer during shutdown — i.e. `detachAll()` returning did not actually guarantee the boss thread was done producing new dispatches, which is exactly the invariant `Engine.close()` relies on before it starts closing workers.

`detachAll()` now enqueues its work on the same `taskQueue` that `attach()`/`detach()` already use, and blocks until the boss's own thread has processed it, closing the gap between "namespaces detached" and "the boss thread has actually observed and acted on it."

## Testing

- `./mvnw checkstyle:check -pl runtime/engine`
- `./mvnw test -pl runtime/engine -am` — full unit suite passes (0 failures/errors)
- Full `./mvnw clean install` from the repo root — exercises every runtime `*IT` via `EngineRule`, which calls `Engine.close()` (and therefore this exact path) after every scenario

Fixes # — N/A, found via crash/core-dump investigation; no separate tracking issue was filed.

---

https://claude.ai/code/session_01KkSMQndCyKY9gBL3v1Rd7d

---
_Generated by [Claude Code](https://claude.ai/code/session_01KkSMQndCyKY9gBL3v1Rd7d)_


jfallows (commented):
LGTM

PR #2599, opened 2026-09-19
feat(binding-llm): llm server/client/proxy — OpenAI↔Anthropic translation via JsonPipeline

### LINKED ISSUE #2596: binding-llm: replace LlmFlushEx union + event mapper with uniform DATA + LlmDataEx, driven directly by common-json
## Summary

Reworks the wire/event model introduced by #2476 and #2482: every part of the LLM protocol — content or what's currently a structured `LlmFlushEx` union case — becomes a `DATA` frame (fragmented via `init`/`fin` like any oversized value already is), with a minimal `LlmDataEx` (shaped like `binding-sse`'s own `SseDataEx`) carrying a `type` discriminator plus a few cross-cutting fields, and the payload carrying whatever type-specific detail is needed, as JSON. `LlmFlushEx`/`LlmNativeFlushEx` and the separate `LlmEventMapper` class hierarchy (`LlmOpenaiEventMapper`/`LlmAnthropicEventMapper`) go away; per-dialect translation is expressed directly against `common-json`'s `JsonPipeline`/`JsonStream`/`JsonTransform`/`JsonEnvelope`, not the generic engine `ModelHandler`/`ModelTransform` SPI or a catalog-resolved schema.

This directly revisits decisions made in #2476 and #2482 (both closed by open PRs #2505 and #2558, with #2479/PR #2554 and #2586 also building on the current shape). Filing this because implementing the flow-control fix for `llm(client)`'s request-encode path (separately in progress) kept running into the current design's complexity, and working through why surfaced a simpler shape — recorded here before more work lands on the current one.

## Why revisit

The current design's stated rationale (#2476): *"Content deltas map 1-1 across every dialect. All cardinality mismatch is metadata. Splitting the planes confines the mess to FLUSH and leaves DATA as a clean 1-1 stream."* That's a real property, and this proposal gives it up deliberately (see risk 1 below) — the case for doing so is that the current design pays for it in several other ways that compound:

- **A second, bespoke translation mechanism duplicates one that already exists.** `LlmOpenaiResponseTransform`/`LlmAnthropicResponseTransform` (`ModelTransform`-based field renaming, e.g. `$.choices[0].index` ↔ `choiceIndex`, `$.usage.prompt_tokens` ↔ `inputTokens`) already exists and — per its own doc comment — was written with streaming-chunk shapes in mind. But cross-dialect streaming today bypasses it entirely in favor of `LlmOpenaiEventMapper`/`LlmAnthropicEventMapper`, which hand-build JSON via `jakarta.json` from scratch. Same rename logic, two independent implementations.
- **The generic `ModelHandler`/catalog layer is indirection the binding doesn't need.** Each dialect's schema is a known, static resource shipped with the binding, not something registered/discovered at runtime across arbitrary formats — the one thing the catalog-backed `Model` SPI adds. Referencing schemas directly and driving `common-json`'s own schema-validation stage removes a layer that exists only to satisfy an abstraction the binding doesn't need.
- **`pendingContent`-style buffering on the request-encode side exists to bridge an ordering mismatch created by the FLUSH/DATA split**, not by anything inherent to the protocol: content becomes known before its event name (arrives later via FLUSH), but SSE's wire order needs `event:` before `data:`. Carrying the event's `type` alongside its own DATA frame removes the mismatch instead of buffering around it.
- **Same-dialect routes pay full parse/re-render cost for every canonical event today**, even though only `LlmNativeFlushEx` (the `raw` case) is designed as a passthrough escape hatch. Under this proposal, same-dialect routes are pure DATA relay for *any* event, parse-free — matching `binding-sse`'s own same-dialect simplicity.

## Proposed design

**Frame model.** `LlmDataEx` stays minimal, matching `SseDataEx`'s shape (`runtime/binding-sse`'s `sse.idl`):

```
struct SseDataEx extends core::stream::Extension
{
    int64 timestamp = 0;
    string8 id;
    string8 type = null;
}
```

`LlmDataEx` becomes: a `type` discriminator (mirroring `SseDataEx.type`/today's `LlmNativeFlushEx.type`), the existing `logProbability`, and token-usage fields (`inputTokens`/`outputTokens`, extended per #2586's absent-vs-zero requirement) — usage lives on the extension directly, not in the payload, so a metrics consumer reads it uniformly regardless of `type` and never parses payload JSON. Every other type-specific detail (`blockId`, `toolName`, finish `reason`, etc.) lives in the payload as JSON, keyed by `type`. A payload can be absent (`null`), matching `SseServerFactory.onAppData`'s existing `extension.sizeof() > 0`/nullable-payload handling for a comment/keepalive-shaped SSE event with no `data:` line — so a purely structural event (`finish`, `usage`, `keepalive`) is just a DATA frame with no payload, not a new frame type.

`LlmFlushEx`/`LlmNativeFlushEx` (the whole union from #2476) is removed.

**Decode and encode, directly against `common-json`.** Per dialect, one stateful `JsonTransform` (a fresh instance per stream, same as today's `LlmOpenaiResponseTransform`) both renames fields *and* makes the cardinality decisions #2482 scoped to the event mapper — held tokens, block-index maps, one-event lookahead for synthesizing a close event. That state doesn't disappear; it moves into the same transform that already does field renaming, rather than a separate `LlmEventMapper` class. Mechanism, in both directions:

- *Decode* (native chunk → canonical DATA+`LlmDataEx`): the transform writes the outbound event's `type` to `JsonEnvelope` right after emitting `START_DOCUMENT` (synthesized, for a fan-out case) and before the first structural event; a small LLM-owned terminal `JsonSink` reads it there, accumulates the value's bytes, and on `END_DOCUMENT` emits the corresponding `doAppData(...)`/`doAppEnd(...)`-equivalent call directly, as a side effect — no boundary information needs to travel back through `JsonPipelineResult`'s aggregate `consumed`/`produced`. One native chunk producing N canonical events (e.g. OpenAI's flat tool-call delta → Anthropic's separate `content_block_start`/`delta`/`stop`) is N `START_DOCUMENT`/`END_DOCUMENT` cycles within one parse pass, not N re-parses of the same bytes (calling `reset()` to "rewind" already-consumed, possibly-fragmented input would be a misuse of its contract).
- *Encode* (canonical → native SSE bytes): an SSE-aware terminal sink does the same `START_DOCUMENT`(read envelope)/`END_DOCUMENT` dance, but writes `event:`/`id:`/`data:`/`\n\n` directly. Since SSE is self-framing on the wire, the *caller* never needs event boundaries here at all — aggregate bytes can be sliced into DATA frames by window size alone, like plain byte relay.

**Framing layer (JSON vs SSE vs future binary) is unaffected** — `LlmContentDecoder`/`LlmContentEncoder` stays exactly the abstraction it is; a future AWS Bedrock `application/vnd.amazon.eventstream` implementation is a third variant of it, since that's a transport/framing concern, not a payload one (the payload inside is still JSON). Non-JSON literals (OpenAI's `[DONE]`) stay special-cased ahead of the JSON pipeline, as today. This also travels cleanly to a later move of SSE framing into the HTTP binding: whoever unwraps SSE and hands over `HttpDataEx`/`SseDataEx`-tagged DATA frames, the fan-out logic downstream only ever needs "one parsed value + its discriminator," never how that discriminator was produced.

## What disappears

- `LlmFlushEx`, `LlmNativeFlushEx`, and the other 5 union case structs (`llm.idl`)
- `LlmEventMapper`, `LlmOpenaiEventMapper`, `LlmAnthropicEventMapper` — logic absorbed into the per-dialect `JsonTransform`
- `ModelHandler`/`ModelPipeline`/`ModelTransform`/`ModelEnvelope` usage and catalog-based schema resolution (`CatalogedConfig`, `SYSTEM_CATALOG_NAME`) for LLM dialects
- `pendingContent`-style request-encode buffering that exists only to bridge the FLUSH-arrives-after-DATA ordering mismatch

## What stays

- `LlmBeginEx`, `LlmContentDecoder`/`LlmContentEncoder` (JSON/SSE framing split), the dialect SPI's `supplyDecoder`/`supplyEncoder` shape, the `[DONE]`-style non-JSON special-casing

## Risks / open items

1. **Gives up DATA's "clean 1-1, discrimination-free" plane** that #2476 explicitly valued — every DATA frame consumer now inspects `LlmDataEx.type`. Accepted trade for the complexity removed elsewhere, not a free win.
2. **`common-json` needs to actually support multiple document boundaries within one pipeline pass** for the fan-out case — I traced this as likely already supported (`JsonSinkImpl` reports `Status.COMPLETED` purely on `depth == 0`, with no guard against a subsequent `START_DOCUMENT`, and `JsonPipelineImpl.completeDocument()` already pumps further events after a first `COMPLETED`), but that's from reading the driver, not from running it. This should be validated with a real spike — hardest case first (OpenAI→Anthropic tool-call block synthesis with lookahead). If gaps turn up, they're `common-json`'s own requirement independent of this binding: multi-document YAML (`---`-separated) needs the same thing from `JsonPipeline`, so any fix is justified on `common-json`'s own contract, not by naming this binding as the motivating consumer.
3. **Flow-control cost of a (possibly zero-payload) DATA frame per structural event, where FLUSH was previously exempt from window accounting.** Intentional, not a regression: exposes each event's real processing/framing cost end-to-end instead of hiding it.
4. **Superseding PRs #2505, #2554, #2558.** Real cost, taken deliberately — better to settle this now than ship the more complex design and rework it later.
5. **#2586's richer usage shape** (absent-vs-zero, incremental accumulation, partial-on-abort) should fit naturally on `LlmDataEx` directly (see frame model above) but wants explicit confirmation once implementation starts.

## Relationship to open work

Supersedes #2476 and #2482 as written; #2479 (SSE content decoder) and its PR #2554 mostly survive as the framing layer described above, modulo whatever event/data separation assumptions need adjusting. PRs #2505 and #2558 would need reworking or closing in favor of this shape.

jfallows:
Fixes #2596

## Description

Introduces the `llm` binding (`kind: server`/`client`/`proxy`) for bridging OpenAI Chat Completions and Anthropic Messages traffic, plus `examples/llm.proxy` demonstrating it.

**Wire model.** Every part of the app-facing protocol — content and structural signals alike (event name, finish reason, usage) — is a single `DATA` frame carrying a minimal `LlmDataEx` extension (a `type` discriminator plus `logProbability`/`inputTokens`/`outputTokens`), fragmented via ordinary `init`/`fin` flags like any oversized value. There is no separate `LlmFlushEx` union frame for structural signals.

**Translation, directly against `common-json`.** Both directions are expressed against `common-json`'s `JsonPipeline`/`JsonStream`/`JsonTransform`/`JsonSink`/`JsonSchema`, not the generic engine `ModelHandler`/`ModelPipeline`/`ModelTransform` SPI or a catalog-resolved schema:
- Request-side field renaming (`max_tokens`↔`maxOutputTokens`, `top_p`↔`topP`, etc.) is a `JsonTransform` that captures/defers a top-level key on `KEY_NAME`, decides the rename on the paired scalar value, and writes a resumable KEY-then-VALUE pair to its sink.
- Response-side streaming translation between dialects (OpenAI SSE chunks ↔ Anthropic SSE events, including the flat-tool-call-delta ↔ separate-content-block fan-out) is a per-dialect decode `JsonTransform` + encode `JsonSink` pair, sharing one long-lived `JsonPipeline` per response stream: the decode stage emits genuine `START_DOCUMENT`/`END_DOCUMENT` cycles (one per canonical action) to its sink rather than an ad hoc convention, and `JsonPipeline.nextDocument()` (added to `common-json` in this PR) advances the pipeline between native chunks without cascading a state-wiping `reset()` to either stage — `reset()` is reserved for `REJECTED` abandonment and stream teardown.
- Each dialect compiles its own request/response `JsonSchema` directly from a bundled classpath resource at construction; there is no `sys:` namespace catalog or `ModelHandler`-backed schema validation anywhere in the binding.

**Dialects.** `openai` and `anthropic`, each with detection (method/path/content-type or headers), credential-header conventions (`Authorization: Bearer`, `x-api-key`), and a native JSON schema for both directions.

**Routing.** `llm(proxy)` routes purely on `LlmBeginEx.dialect`/`model` — both already resolved by the `llm(server)` that produced the stream — never on request content, mirroring how `mcp(proxy)` routes on tool name. Supports both cross-dialect translation (the default per dialect) and intra-dialect, model-based routing to a second same-dialect deployment (no translation on that leg).

**Credential pass-through.** `options.authorization` templates the caller's own presented credential (`"Bearer {credentials}"` / `"{credentials}"`) rather than configuring a secret of Zilla's own, so a caller's real API key forwards upstream unchanged in either direction.

**`examples/llm.proxy`** demonstrates both frontends (`north_llm_server_openai`/`_anthropic`), the shared `north_llm_proxy`, and four `llm(client)` legs (two dialects × primary/secondary). `etc/zilla.yaml` targets the real OpenAI/Anthropic APIs over TLS by default; `.github/compose.mock.yaml` (merged via `.github/.env.test`'s `COMPOSE_FILE`) swaps in four mock backends and a mock-hostname `zilla.yaml` for CI/local testing without real API keys, mirroring the compose-overlay pattern used for analogous mock-vs-real cases elsewhere.

## Testing

- k3po ITs: `LlmServerIT`, `LlmClientIT`, `LlmProxyIT` (`incubator/binding-llm`), `ApplicationIT`/`NetworkIT` (`incubator/binding-llm.spec`), including openai/anthropic streaming, non-streaming, tool-call round trips, guarded-authorization, cross-dialect, and flow-control (10k/100k) scenarios.
- Unit tests for dialect detection/resolution, request field-rename transforms, and the per-dialect canonical decode/encode transforms (driven through a real `JsonPipeline`, not mocked collaborators), including a test proving cross-chunk state survives `nextDocument()`.
- `examples/llm.proxy`'s own `./.github/test.sh` (8 assertions: both cross-dialect translation directions, both tool-call round trips, both credential pass-through directions, both model-based secondary-deployment routes) run against a locally built image.
- `grep -rn "engine\.model\." incubator/binding-llm/src` returns nothing — confirms no engine Model SPI usage remains anywhere in the binding.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AKZMN6ckSpx1QpHvaE4DwM)_


jfallows:
`Build (25)` failed on commit `4737831e` (superseded) with `common-protobuf` test failures — unrelated to this PR's own module, but caused by it: this PR's `JsonTokenizer` fix (fixing `binding-llm`'s streaming-100k hang) made the tokenizer correctly fragment a scalar value that doesn't start at byte 0 of its input window, which `ProtobufJsonParserImpl` wasn't prepared for — its `messageStep()`/`mapStep()` already decline an incomplete `KEY_NAME` fragment (leaving it unacted-on and re-pulling), but `valueStep()`/`arrayStep()`/the map-entry value step never applied that same treatment to a `VALUE_STRING`/`VALUE_NUMBER` token.

Fixed and pushed in `41f14db6`: give scalar values the same decline as keys already get. `ProtobufJsonTest`/`ProtobufJsonChunkingTest` (the two suites that caught this) are green again, and the full `common-protobuf` and `binding-llm` module verifies pass together at the current head.

---

jfallows:
Third distinct CI failure (unrelated to the previous `common-protobuf` fix): `binding-llm.conf`'s JaCoCo check failed — `instructions covered ratio is 0.99, but expected minimum is 1.00`.

Root cause: two branches in `LlmOptionsConfigAdapter` had no test coverage:

- `defaultPort(String)`'s ternary only had its HTTPS branch exercised (`shouldReadServerOptionWithDefaultHttpsPort`); every other server-option test either supplies an explicit port or uses `https`, so the HTTP-default-port (`80`) branch was never taken.
- `adaptServer()`'s `catch (URISyntaxException ex) {}` block was never reached — the existing `shouldReadMalformedServerOptionAsAbsent` test's input (`"not-a-host-and-port"`) parses successfully as a relative URI and only fails the subsequent host/scheme check, so it never throws.

Fixed in bd99d815 by adding two unit tests: `shouldReadServerOptionWithDefaultHttpPort` (`"http://example.com"`, asserts port `80`) and `shouldReadInvalidServerUriAsAbsent` (a malformed IPv6-literal host that genuinely throws `URISyntaxException`). Verified locally: `binding-llm.conf` now passes `clean verify` with 100% instruction coverage, and the full `binding-llm` `clean verify` (including all k3po ITs) still passes against the updated dependency.

---

jfallows:
Fourth CI failure, on commit bd99d815: `testing (inspect.schema)` — the `examples/inspect.schema` example's golden `schema.expected.json` fixture had drifted from the actual merged JSON schema.

Root cause: this PR's earlier work aligning `llm(client)`'s `server` option with the standard SDK base-URL convention changed its schema `description`/`pattern` (from a `host:port` form to a full base-URL form), but `examples/inspect.schema/.github/schema.expected.json` — which asserts the complete merged schema byte-for-byte via `zilla inspect schema` — was never updated to match.

Fixed in b9836c22 by updating the fixture's `server` property block to the actual current schema (description, `format: uri`, `pattern: ^https?://`). Confirmed the diff matches exactly what CI reported, and that no other `host:port`-pattern occurrences in the fixture are related (the other two are unrelated Kafka `servers` array options).

---

jfallows:
Fifth CI failure, and the significant one: `testing (llm.proxy)` — the example's own end-to-end assertions failed at the very first one (`call_openai_text`): curl got error 18, "transfer closed with outstanding read data remaining", for a plain non-streaming request.

## Root cause

`LlmServerFactory` (`kind: server`) and `LlmClientFactory` (`kind: client`) both sit directly upstream of a real `http` binding. Neither declares a `content-length` for what it writes (the response body, the outbound request body), so `http` always uses chunked `Transfer-Encoding` for these — which means `http` grants a non-zero `padding` in its `WINDOW` frames, to cover its own chunk-framing overhead (chunk-size hex line + CRLFs). Per the established convention elsewhere in the codebase (`HttpServerFactory`'s own `replyPad` usage — see `reserved = length + replyPad` throughout that class — and `runtime/AGENTS.md`'s per-stream field table), a sender must add the receiver's declared padding to `reserved` on every outbound `DATA` frame, and reserve room for it before sizing the frame's payload.

Both `LlmServerFactory` and `LlmClientFactory` ignored `window.padding()` entirely — no field even tracked it — and sent `reserved` equal to the raw payload length. Every existing `binding-llm` k3po test simulates the `http` peer as a synthetic script that (like most k3po fixtures) grants `padding: 0` by default, so this was invisible to the entire existing IT suite; a *real* `http` binding's non-zero padding causes it to reject/mis-account for the frame, producing exactly the observed truncated-response symptom on the reply leg, and an analogous hang on the request leg once padding is large enough to matter there too.

## Fix (0302f15a)

Added `replyPad` (`LlmServerFactory`) / `initialPad` (`LlmClientFactory`) fields, captured from `window.padding()` on the real `WINDOW`, and included them in both the available-window check and the `reserved` value of each outbound `DATA` frame. With `padding=0` (every pre-existing test fixture), the arithmetic is unchanged, so no existing behavior is affected — confirmed by the full `binding-llm`/`binding-llm.spec` test suites passing unmodified.

Added regression coverage that simulates a real `http` peer's non-zero padding via k3po's `option zilla:padding` on the accept/connect side (confirmed each new test fails with the pre-fix code and passes with it):
- `LlmServerIT#shouldForwardOpenaiNonstreamingWithReplyPadding` (response leg)
- `LlmClientIT#shouldForwardOpenaiRequestWithReplyPadding` (request leg)
- plus their `NetworkIT`/`ApplicationIT` peer self-consistency counterparts

Also added `LlmProxyIT#shouldRouteOpenaiToAppZero`, a small (non-10k) single-frame reply-relay case through the proxy — the existing proxy relay tests only ever exercised multi-frame (10k+) replies. This was ruled out as the source of the bug (it passes both before and after the fix) but closes a real coverage gap, so it's kept.

Note: this sandbox has no working Docker daemon, so I could not re-run `examples/llm.proxy`'s own docker-compose end-to-end test directly. Verification here relies on the k3po regression tests above, which reproduce the exact WINDOW-padding mechanics a real `http` binding uses. I'd appreciate the next CI run (which does have Docker) confirming `testing (llm.proxy)` now passes end-to-end.

---

jfallows:
`Build (25)` failed on commit 0302f15a, but not from anything in this PR's diff: the actual failure is `io.aklivity.zilla.runtime.binding.kafka.internal.stream.CacheProduceIT.shouldReconnectPartitionNotLeader` in `binding-kafka` — a module this PR never touches. This looks like a flaky Kafka-reconnection timing test. Re-running the failed jobs once to confirm.

---

jfallows:
## `testing (llm.proxy)` — actual root cause found

The WINDOW-padding fix in `096eb25a`'s predecessor (`0302f15a`) was real and necessary but, as the last update noted, did **not** fix the `call_openai_text` e2e failure. Confirmed via direct local reproduction (not just CI-log archaeology): with a build containing that fix, `curl` against the real `examples/llm.proxy` stack got `200 OK` + `Transfer-Encoding: chunked` headers but zero body bytes, and `mock-anthropic`'s Express/`raw-body` layer aborted the connection before reading a complete request — meaning zilla itself never sent a complete, validly-framed request body to the backend.

Root cause: `LlmClientFactory.java`'s `LlmHttpClient.encodeNet` (the code path that forwards a translated request body from `llm:client` to a real `http:client` binding) computed its outbound DATA frame flags as `initialStarted ? 0 : FLAG_INIT` — i.e. it **never set `FLAG_FIN`** on any request frame, ever. `HttpClientFactory.doEncodeHttp1Body` (`runtime/binding-http`) gates writing the HTTP/1.1 chunk-size-hex prefix on `FLAG_FIN` and the trailing CRLF on `FLAG_INIT`. Since `FLAG_FIN` was never set, every outbound request chunk was missing its size-hex line entirely — the backend received raw JSON bytes with no valid chunked-encoding framing at all, which explains the abort exactly.

Fix (`096eb25a`): always set both `FLAG_INIT | FLAG_FIN` on every physical outbound write in `encodeNet`. Each physical write already becomes its own self-contained wire-level chunk (chunk size scoped to just that write), request headers are written on a separate path unaffected by these flags, and HTTP/2 ignores them entirely — so this is safe regardless of how `encodeNet`'s own backpressure buffering may split a single JSON encode-output slice across multiple physical writes. The now-unused `initialStarted` field was removed.

Notably, no existing k3po test in this module could have caught this: every `LlmClientIT` scenario stubs the network side as a synthetic k3po script that reads decoded HTTP begin/data content directly, never exercising the real `HttpClientFactory` chunk-encoding logic. I'm adding a dedicated regression test that chains `llm:client -> http:client -> tcp:client` for real and asserts the literal wire bytes, to close that coverage gap — will push separately. Full existing `binding-llm` unit + IT suite (136 unit tests, 57 ITs) passes unchanged with this fix.

Re-running local docker-compose e2e now to confirm directly before considering this closed.

https://claude.ai/code/session_01AKZMN6ckSpx1QpHvaE4DwM

---

jfallows:
## `testing (llm.proxy)` — confirmed fixed end-to-end (two bugs, both real)

Root-caused this properly by decoding the actual wire-level frame trace out of the engine's own ring-buffer event log (extracted from a CI artifact and read with a small standalone driver against `runtime/command-logs`' `LogsReader`/`incubator/command-dump`'s `RingBufferSpy` — ground truth beats re-deriving behavior from source alone). That surfaced **two** distinct, sequential bugs blocking this test, not one:

**Bug 1 (fixed in `096eb25a`)** — `LlmHttpClient.encodeNet` never set `FLAG_FIN` on outbound request DATA frames to the real `http:client` binding, so `HttpClientFactory`'s chunk encoder never wrote the HTTP/1.1 chunk-size-hex prefix. CI re-ran against that commit and the symptom changed (confirmed via the frame trace): the request body now arrives at mock-anthropic correctly framed —
```
40\r\n{"model":"gpt-4","messages":[{"role":"user","content":"Hello"}]}\r\n
0\r\n\r\n
```
— but the backend replied `404 Not Found: Cannot POST /v1/messages`, a *different* failure than before (previously the malformed chunking made Express abort mid-parse before routing even ran).

**Bug 2 (fixed in `3bb08ded`)** — `examples/llm.proxy`'s four mock backends (`mock-openai(-secondary)`, `mock-anthropic(-secondary)`) all registered their route on `POST /` under a comment claiming `south_llm_client_*` "always issues its outbound request to '/', regardless of dialect." That was stale: `LlmClientFactory` actually sends each dialect's real canonical path (`/v1/messages` for Anthropic, `/v1/chat/completions` for OpenAI — matching the real upstream APIs), confirmed directly from the same frame trace. Bug 1's malformed body previously masked this, since Express's body-parsing middleware aborted before route matching ever got a chance to 404. Fixed by pointing each mock's route at the dialect's actual path.

**Verified for real, not just in theory**: downloaded the exact `zilla-image` artifact CI built from `096eb25a` (no local rebuild needed) and ran the full local `docker-compose` stack with mock fix `3bb08ded` applied. All 8 assertions in `etc/test/verify.sh` pass end-to-end: cross-dialect text + tool-call translation both directions, credential pass-through both directions (confirmed via each mock's own logs), and model-based secondary routing both directions. Also caught and worked around an unrelated local-only footgun: `compose.yaml`'s `image:` defaults to `:latest` unless `ZILLA_VERSION` is set — CI always sets it, so this never bit CI, only my local repro.

Expect `testing (llm.proxy)` green on the next CI run against `3bb08ded`.

https://claude.ai/code/session_01AKZMN6ckSpx1QpHvaE4DwM

---

